PR #18, opened 2022-03-16
Make an abstraction for 'taps', replace most of the weird include goo using taps with new system.

jbruestle:


risc0/zkp/verify/taps.h
```diff
@@ -0,0 +1,259 @@
...
+enum class RegisterGroup : uint8_t {
+  ACCUM = 0,
+  CODE = 1,
+  DATA = 2,
+  INVALID = 3,
+};
+
+static constexpr size_t kNumRegisterGroups = 3;
+
+namespace impl {
```
flaub, line 40: nit: Idiomatic? c++ code usually uses `detail`. Also thinking that I might want to use `impl` as var name more often then I'd use `detail` as a var name.

risc0/zkp/verify/taps.h
```diff
@@ -0,0 +1,259 @@
...
+  const impl::TapSetData* data_;
+
+public:
+  TapSetRef(const impl::TapSetData* data) : data_(data) {}
+  size_t tapsSize() { return data_->groupBegin[kNumRegisterGroups]; }
+  IteratorRange<TapIterator> taps() {
+    return IteratorRange<TapIterator>(data_->taps,
+                                      data_->taps + data_->groupBegin[kNumRegisterGroups]);
+  }
+  IteratorRange<RegIterator> regs() {
```
flaub, line 206: nit: Would prefer blank lines between methods

risc0/zkp/verify/taps.h
```diff
@@ -0,0 +1,259 @@
...
+
+class TapSet {
+private:
+  void finalize();
+
+  // Stuff used only while building
+  bool finalized_;
+  using reg_t = std::set<size_t>;
+  using group_t = std::map<size_t, reg_t>;
+  using all_t = std::map<RegisterGroup, group_t>;
```
flaub, line 241: nit: Use `ProperCamelCase` for all types (including using/typedef)?

risc0/zkp/verify/taps.h
```diff
@@ -0,0 +1,259 @@
...
+  all_t all_;
+
+  // Stuff that exists post finalization
+  std::vector<impl::TapData> taps_;
+  std::vector<uint16_t> comboTaps_;
+  std::vector<uint16_t> comboBegin_;
+  impl::TapSetData data_;
```
flaub, line 248: nit: Would prefer to drop trailing `_`, but I could be convinced otherwise. I just don't think we've been consistent about this so far.

flaub (commented):
LGTM, other than minor code-style nits.

flaub (approved):

PR #19, opened 2022-03-17
Replace unsafe casts with serde (C++)

flaub:


rlukata (approved):
Able to build without issues.

jbruestle (approved):

PR #21, opened 2022-03-21
Fix Windows Build (include stdexcept where it is required)

ChimeraScorn:
This is required for Windows to build in the x86_x64 Cross Tools dev environment for VS2019 Build Tools 
Notes: 
   This is on a fresh install of the all components, including windows itself
   If you have run bazelisk previously, be sure to run "bazelisk clean --expunge" after updating your build tool chain to remove crufty commands

### DIFF
--- a/risc0/zkp/verify/taps.h
+++ b/risc0/zkp/verify/taps.h
@@ -18,6 +18,7 @@
 #include <map>
 #include <set>
 #include <vector>
+#include <stdexcept>
 
 namespace risc0 {
 


flaub:
@ChimeraScorn Thanks for this! Could you try the latest on `main`? I believe this was already fixed in: https://github.com/risc0/risc0/pull/17/files#diff-681d3575a730af91a02421dd4959550cc65950c19346afe59942c7d3d7c9cdf3R21

PR #26, opened 2022-03-25
NFC: extract RISC-V specific code from core ZKP system

jbruestle:


jbruestle:
NFC - Move files to refactor RISCV support out of core proof system.

flaub (approved):

PR #34, opened 2022-03-31
Step-by-step VM guest debugging and GDB Server

mitchellharper12:
Currently a WIP for first refactoring the VM runtime to support step-by-step debgging, to be followed by implementing the hooks to drive incremental state transition updates and reporting to consumers, followed by implementation of a GDB server for debugging VM guest code as a GDB target.

Creating this draft pull request so that progress can be tracked on the shape of the refactoring changes and API to ensure we are in agreement.

### DIFF
--- a/risc0/r0vm/prove/BUILD.bazel
+++ b/risc0/r0vm/prove/BUILD.bazel
@@ -20,6 +20,7 @@ cc_library(
         "step.cpp",
         "step_context.cpp",
         ":step_inc_gen",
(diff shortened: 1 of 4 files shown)


jbruestle (commented):
I like the overall goal and direction.  Maybe split out 'step' in ExecState as well, and just make run call 'init' and then 'step' in a loop?  Perhaps that was already your plan.  Also some renaming changes incoming, so you may want to rebase before you do too much more.

PR #35, opened 2022-04-01
NFC: Rename/reorg

flaub:
* r0vm -> zkvm
* risc0/r0vm/cpp -> risc0/zkvm/sdk/cpp
* risc0/r0vm/rust -> risc0/zkvm/sdk/rust


flaub:
Feedback welcome!

choongng:
Is the thinking here then we'd refer to it as "Risc Zero ZKVM"?

jbruestle:
> Is the thinking here then we'd refer to it as "Risc Zero ZKVM"?

I think it's more to avoid the 'Smurf Naming Convention' where everything is prefixed with Risc Zero.  Given that the whole repo and the base dir is risc0, I think it's reasonable to just call it 'The ZKVM' and presume the r0 follows from context.  Of course if we were talking about our ZKVM vs another one (say Miden), then we would prefix.

jbruestle (approved):

intoverflow (approved):
Looking good!

In the `zkvm` path, I want to confirm that I know what each of the subdirectories are for:
* "prove" implements the runner, "verify" implements the checker
* "circuit" and "platform" are used by both "prove" and "verify
* sdk contains libs for writing host and guest apps

Is that right?

image

choongng:
> > Is the thinking here then we'd refer to it as "Risc Zero ZKVM"?
> 
> I think it's more to avoid the 'Smurf Naming Convention' where everything is prefixed with Risc Zero. Given that the whole repo and the base dir is risc0, I think it's reasonable to just call it 'The ZKVM' and presume the r0 follows from context. Of course if we were talking about our ZKVM vs another one (say Miden), then we would prefix.

We also know from context it's ZK, could we go all the way down to "vm"? Just as clear but less awkward both in docs and code.

choongng (approved):
vm or zkvm, either way

PR #39, opened 2022-04-02
Modify macro system to allow making table based polynomial verification.

jbruestle:


risc0/zkvm/circuit/poly_context.cpp
```diff
@@ -55,17 +57,17 @@ template <typename Derived, int TypeID> struct PolyOpBase : PolyOpInterface {
 
 using PolyOp = std::shared_ptr<PolyOpInterface>;
 
-struct OpConstant : public PolyOpBase<OpConstant, 0> {
+struct OpConstant : public PolyOpBase<OpConstant, 0, 0> {
```
flaub, line 60: Can we use `false` here instead of `0`? Might be nicer to have a tagged type.

flaub (approved):

PR #41, opened 2022-04-03
Changes to VM executer to prep for debugging support

jbruestle:
Modified the ExecState class to actually understand the RISC-V emulation it is doing, at least a little bit.  This unfortunately makes a bit more coupling between the circuit generation code and the emulation code, but a reasonable trade off for opening the possibility of debugging user code.


mitchellharper12 (approved):
There are parts of the calculation of register values I don't yet fully understand, but this interface will be fine to work with as a stopgap.

risc0/zkvm/prove/exec.h
```diff
@@ -22,14 +22,32 @@
 namespace risc0 {
 
 struct ExecState {
+public:
```
flaub, line 25: Not needed if this is a `struct`

flaub (approved):
Reasonable as a stop-gap

PR #43, opened 2022-04-06
WIP: Rust code-gen prototype

jbruestle:
Prototype of code to autogenerate rust code for circuit from existing C++ macro system, not integrated with build.  Also, not a good long term method for code generation, probably shouldn't land this one, but wanted to make sure to record somewhere.


rust-gen/rust-poly-ops.h
```diff
@@ -0,0 +1,28 @@
+use crate::fp::{Fp};
```
flaub, line 1: Maybe name this `rust-poly-ops.tmpl.rs`

flaub:
@jbruestle What would it take to get this in a land-able state?

PR #45, opened 2022-04-07
Clean up compiler warnings on gcc and clang

shkoo:
This adds -Werror to gcc and clang compilations, and fixes warnings (mostly signed vs unsigned comparisons).

### DIFF
--- a/bazel/rules/cc/defs.bzl
+++ b/bazel/rules/cc/defs.bzl
@@ -9,10 +9,7 @@ def _copts(std):
         "//conditions:default": [
             "-std=" + std,
             "-Werror",
-            # TODO(nils): Clean up type punning violations;
-            # this is easy to do once we get to c++20 because then we
-            # can use std::bit_cast
-            "-Wno-strict-aliasing",
+            "-fno-strict-aliasing",
         ],
     })
 
--- a/bazel/rules/cc/defs.bzl
+++ b/bazel/rules/cc/defs.bzl
@@ -9,7 +9,6 @@ def _copts(std):
         "//conditions:default": [
             "-std=" + std,
             "-Werror",
-            "-fno-strict-aliasing",
         ],
     })
 
--- a/bazel/rules/cc/defs.bzl
+++ b/bazel/rules/cc/defs.bzl
@@ -6,7 +6,14 @@ def _copts(std):
             "/std:" + std,
             "/Zc:preprocessor",
         ],
-        "//conditions:default": ["-std=" + std],
+        "//conditions:default": [
+            "-std=" + std,
+            "-Werror",
+            # TODO(nils): Clean up type punning violations;
+            # this is easy to do once we get to c++20 because then we
+            # can use std::bit_cast
+            "-Wno-strict-aliasing",
+        ],
     })
 
 def cc_binary(name, std = DEFAULT_CXX_STD, copts = [], **kwargs):
--- a/examples/cpp/battleship/core.cpp
+++ b/examples/cpp/battleship/core.cpp
@@ -147,7 +147,7 @@ std::ostream& operator<<(std::ostream& os, const Position& pos) {
 
 std::ostream& operator<<(std::ostream& os, const GameState& state) {
   os << "GameState{";
-  for (unsigned i = 0; i < NUM_SHIPS; i++) {
+  for (uint32_t i = 0; i < NUM_SHIPS; i++) {
     const Ship& ship = state.ships[i];
     if (i > 0) {
       os << ", ";
--- a/examples/cpp/battleship/core.cpp
+++ b/examples/cpp/battleship/core.cpp
@@ -147,13 +147,13 @@ std::ostream& operator<<(std::ostream& os, const Position& pos) {
 
 std::ostream& operator<<(std::ostream& os, const GameState& state) {
   os << "GameState{";
-  for (int i = 0; i < NUM_SHIPS; i++) {
+  for (unsigned i = 0; i < NUM_SHIPS; i++) {
     const Ship& ship = state.ships[i];
     if (i > 0) {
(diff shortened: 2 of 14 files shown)


jbruestle (approved):
Looks good to me, always nice to have -Werror on.

examples/cpp/battleship/core.cpp
```diff
@@ -147,13 +147,13 @@ std::ostream& operator<<(std::ostream& os, const Position& pos) {
 
 std::ostream& operator<<(std::ostream& os, const GameState& state) {
   os << "GameState{";
-  for (int i = 0; i < NUM_SHIPS; i++) {
+  for (unsigned i = 0; i < NUM_SHIPS; i++) {
```
intoverflow, line 150: Should i be unsigned, or should it be size_t?

risc0/zkp/accel/backend/cpu/impl.cpp
```diff
@@ -130,7 +130,7 @@ void batchEvaluateAny(AccelConstSlice<Fp> coeffs,
     size_t id = reinterpret_cast<const uint32_t*>(which.buf()->buf)[i];
     Fp4 x = reinterpret_cast<const Fp4*>(xs.buf()->buf)[i];
     const Fp* coeffLocal = coeffsPtr + (1 << po2) * id;
-    for (size_t i = 0; i < (1 << po2); i++) {
+    for (size_t i = 0; i < (1U << po2); i++) {
```
intoverflow, line 133: Should this be 1U, or should it be (size_t)1?

intoverflow (approved):
Thank you for fixing these warnings!

Only one question (open for discussion): when do we want to use `unsigned` versus `size_t`?

bazel/rules/cc/defs.bzl
```diff
@@ -6,7 +6,14 @@ def _copts(std):
             "/std:" + std,
             "/Zc:preprocessor",
         ],
-        "//conditions:default": ["-std=" + std],
+        "//conditions:default": [
+            "-std=" + std,
+            "-Werror",
+            # TODO(nils): Clean up type punning violations;
+            # this is easy to do once we get to c++20 because then we
+            # can use std::bit_cast
```
flaub, line 14: Yeah, basically I'm not sure we can get away with not doing type punning, since we're writing such low-level stuff. The Linux kernel also requires type punning, so I'm OK with dropping this warning. Although this really only applies to the guest-side code. I've actually already configured this for the RISC-V toolchain here:
https://github.com/risc0/risc0/blob/main/bazel/toolchain/risc0/gcc_toolchain.bzl#L34
https://github.com/risc0/risc0/blob/main/bazel/toolchain/risc0/gcc_toolchain.bzl#L44

So for the host-side, I could see an argument for removing all type punning since I don't know that it's strictly necessary.

Anyways, this looks good for now.

risc0/zkvm/prove/step.cpp
```diff
@@ -28,6 +28,14 @@
 
 namespace risc0 {
 
+// Ignore values in generated code that are computed unnecessarily;
+// the optimizer should remove these.
+#if defined(__clang__)
+#pragma clang diagnostic ignored "-Wunused-variable"
+#elif defined(__GNUC__)
+#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
+#endif
```
flaub, line 37: Yay! love this

flaub (approved):
LGTM, thanks a lot for this!

examples/cpp/battleship/core.cpp
shkoo: NUM_SHIPS is of type uint32_t and it looks like that's what's used elsewhere, so I'll update that.

risc0/zkp/accel/backend/cpu/impl.cpp
shkoo: In c++23 we can use "1z" to get a literal of type size_t, but I guess for now we can cast it manually.

bazel/rules/cc/defs.bzl
shkoo: How about we use -fno-strict-aliasing then instead of just disabling the warning?  It seems less error prone.

flaub: Well on the host-side there are probably important optimizations that can be enabled with strict aliasing. I suppose if we had perf data to support it then that'd be OK.

bazel/rules/cc/defs.bzl
```diff
@@ -6,7 +6,11 @@ def _copts(std):
             "/std:" + std,
             "/Zc:preprocessor",
         ],
-        "//conditions:default": ["-std=" + std],
+        "//conditions:default": [
+            "-std=" + std,
+            "-Werror",
+            "-fno-strict-aliasing",
```
flaub, line 12: Where are we running into type punning on the host side? I'd prefer to avoid this on the host-side.

risc0/zkp/accel/backend/cpu/impl.cpp
```diff
@@ -130,7 +130,7 @@ void batchEvaluateAny(AccelConstSlice<Fp> coeffs,
     size_t id = reinterpret_cast<const uint32_t*>(which.buf()->buf)[i];
     Fp4 x = reinterpret_cast<const Fp4*>(xs.buf()->buf)[i];
     const Fp* coeffLocal = coeffsPtr + (1 << po2) * id;
-    for (size_t i = 0; i < (1 << po2); i++) {
+    for (size_t i = 0; i < ((size_t)1 << po2); i++) {
```
flaub, line 133: I generally try to avoid C-style casts because it's harder to know which kind of cast the compiler is using.

flaub (approved):
LGTM!

PR #49, opened 2022-04-11
WIP: Initial Battleship markdown.

choongng:
More to do but opening for comment.


flaub:
This might be better suited in the risc0/internal-docs repo. @brianretford is it possible to setup netlify for this repository too?

zicklag:
I read through this and it looks like a great start!

I'd like to add that, from the outside, as one who knew very little zk-SNARKs and zk-STARKS, but knew I needed zero knowledge technology to create a privacy-respecting blockchain, risc0 made sense of what you could accomplish. It was almost like realizing what a mutex was. Now I know what zero-knowledge proofs can accomplish in a very concrete way, and how I might use it to develop private protocols.

The Battleship example takes it a step further making it even more clear what can be accomplished with STARKs and how they could be helpful in a simple, relatable, real-life use-case.

That's just a roundabout way to say that a good exposition of the Battleship example could really help this "click" for people.

choongng:
Unless someone objects let's land it and then do an update pass once the changes for the UI code settle down.

jbruestle (approved):
Awesome work!

flaub:
Let's reserve the `WIP:` prefix for draft PRs only (so when it's ready to review, be sure to drop the prefix).

PR #52, opened 2022-04-11
WIP: battleship cli

rlukata:
As discussed, here's a PR for a _working_ cli version of battleship. There's little validation of input, but _happy path_ should yield desired results.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,6 +1,5 @@
 common --experimental_enable_bzlmod
 
-#startup --output_user_root=C:/tmp
 startup --windows_enable_symlinks
 
 build -c opt
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,5 +1,6 @@
 common --experimental_enable_bzlmod
 
+#startup --output_user_root=C:/tmp
 startup --windows_enable_symlinks
 
 build -c opt
--- a/examples/rust/battleship/src/main.rs
+++ b/examples/rust/battleship/src/main.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use battleship_core::*;
-use lib::*;
-use std::{borrow::Borrow, io::*};
+use battleship_core::{GameState, HitType, Ship, ShipDirection};
+use lib::Battleship;
+use std::io::stdin;
 
 struct Players {
     name: String,
@@ -112,15 +112,11 @@ fn main() {
     } else if count_players == 0 {
         run_auto_bot();
     } else {
-        println!("number of players, {}", count_players);
-        let mut player1 = setup_player_1();
-        let mut player2 = setup_player_2();
-
+        let mut player1 = setup_player(1);
+        let mut player2 = setup_player(2);
         let player1_init = player1.battleship.init().unwrap();
         let player2_init = player2.battleship.init().unwrap();
-
         player1.battleship.on_init_msg(&player2_init).unwrap();
-
         player2.battleship.on_init_msg(&player1_init).unwrap();
 
         for n in 1..10 {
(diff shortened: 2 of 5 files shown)


.bazelrc
```diff
@@ -1,5 +1,6 @@
 common --experimental_enable_bzlmod
 
+#startup --output_user_root=C:/tmp
```
flaub, line 3: You can put this in a `.bazelrc.local` file and since it's in `.gitignore` it won't get checked in.

rlukata:
Sample run:

Welcome to R0 Battleship's example in Rust...
How many players today? (0|1|2)
2
number of players, 2
What's player's 1 name?
rami
Where do you want your carrier's coordinates? (i.e.: 0,0)
0,0
What do you want your carrier's direction to be? (vV|hH)
h
Where do you want your battleship's coordinates? (i.e.: 0,0)
3,1
What do you want your battleship's direction to be? (vV|hH)
v
Where do you want your cruiser's coordinates? (i.e.: 0,0)
7,1
What do you want your cruiser's direction to be? (vV|hH)
v
Where do you want your submarine's coordinates? (i.e.: 0,0)
4,3
What do you want your submarine's direction to be? (vV|hH)
v
Where do you want your destroyer's coordinates? (i.e.: 0,0)
6,6
What do you want your destroyer's direction to be? (vV|hH)
h
What's player's 2 name?
Joe
Where do you want your carrier's coordinates? (i.e.: 0,0)
0,0
What do you want your carrier's direction to be? (vV|hH)
h
Where do you want your battleship's coordinates? (i.e.: 0,0)
1,3
What do you want your battleship's direction to be? (vV|hH)
v
Where do you want your cruiser's coordinates? (i.e.: 0,0)
7,1
What do you want your cruiser's direction to be? (vV|hH)
v
Where do you want your submarine's coordinates? (i.e.: 0,0)
4,3
What do you want your submarine's direction to be? (vV|hH)
h
Where do you want your destroyer's coordinates? (i.e.: 0,0)
6,6
What do you want your destroyer's direction to be? (vV|hH)
h

 0.000 (0.000): Opening CPU

Joe turn, what are your attack coordiantes? (i.e.: 0,0)
0,0
It's a Hit

rami turn, what are your attack coordiantes? (i.e.: 0,0)
0,0
It's a Hit

Joe turn, what are your attack coordiantes? (i.e.: 0,0)
1,1
It's a Miss

rami turn, what are your attack coordiantes? (i.e.: 0,0)
0,1
It's a Hit

Joe turn, what are your attack coordiantes? (i.e.: 0,0)
0,0
It's a Hit

rami turn, what are your attack coordiantes? (i.e.: 0,0)
1,1
It's a Miss

Joe turn, what are your attack coordiantes? (i.e.: 0,0)
5,5
It's a Miss

rami turn, what are your attack coordiantes? (i.e.: 0,0)
6,6
It's a Hit

Joe turn, what are your attack coordiantes? (i.e.: 0,0)
7,1
It's a Hit

examples/rust/battleship/src/main.rs
```diff
@@ -12,6 +12,348 @@
...
+        salt: 0xCAFECAFE,
+    };
+
+    return Players {
+        name: player1_string,
+        battleship: Battleship::new(player1_state),
+    };
+}
+
+fn setup_player_2() -> Players {
```
flaub, line 192: Is there any difference between `setup_player_1` and `setup_player_2`? If there is anything, use parameters so that you can have single `setup_player` function that can be called twice.

examples/rust/battleship/src/main.rs
```diff
@@ -12,6 +12,348 @@
...
+        salt: 0xDEADBEEF,
+    };
+
+    return Players {
+        name: player2_string,
+        battleship: Battleship::new(player2_state),
+    };
+}
+
+fn setup_carrier() -> Ship {
```
flaub, line 216: How about having a single `setup_ship` function that can be called with different parameters?

examples/rust/battleship/src/main.rs
```diff
@@ -12,6 +12,348 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use battleship_core::*;
+use lib::*;
+use std::{borrow::Borrow, io::*};
```
flaub, line 17: Looks like you can drop `Borrow`?
Would prefer that we avoid using `::*`, since it makes it less obvious what was pulled in.

examples/rust/battleship/src/main.rs
rlukata: not sure what i was thinking, but totally agree, last commit should address this

examples/rust/battleship/src/main.rs
```diff
@@ -12,6 +12,210 @@
...
+        salt: 0xCAFECAFE,
+    };
+
+    return Players {
+        name: player_string,
+        battleship: Battleship::new(player_state),
+    };
+}
+
+fn setup_ship(ship_name: String) -> Ship {
```
flaub, line 188: Does this need to be a `String`, I'm thinking this could be a `&str` (since we only need to borrow a reference to the string?).

rlukata: definitely need to stop thinking in terms of managed code 😅

rlukata:
Closing PR since codebase has changed enough to deem this approach deprecated.

PR #54, opened 2022-04-12
Make receipt serializable, build binary to run a proof

jbruestle:


risc0/zkvm/r0vm/r0vm.cpp
```diff
@@ -0,0 +1,66 @@
...
+  }
+  if (!feof(stdin)) {
+    fprintf(stderr, "Error reading from stdin\n");
+    return 1;
+  }
+  try {
+    Prover prover(argv[1]);
+    prover.writeInput(data.data(), count);
+    // Run prover
+    Receipt r = prover.run();
```
flaub, line 44: Prefer to avoid single letter variable names except with indexes or math-heavy code.

risc0/zkvm/r0vm/r0vm.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "risc0/zkvm/sdk/cpp/host/receipt.h"
+
+using namespace risc0;
+
+constexpr size_t kMaxStdin = 64 * 1024;
+
+int main(int argc, char* argv[]) {
+  // Check args
+  if (argc != 2) {
+    fprintf(stderr, "Usage: r0vm <elf_file>\n");
```
flaub, line 26: Use C++ std::cerr instead of fprintf

risc0/zkvm/r0vm/r0vm.cpp
```diff
@@ -0,0 +1,66 @@
...
+    fprintf(stderr, "Error reading from stdin\n");
+    return 1;
+  }
+  try {
+    Prover prover(argv[1]);
+    prover.writeInput(data.data(), count);
+    // Run prover
+    Receipt r = prover.run();
+    // Write private output
+    const auto& out = prover.getOutput();
```
flaub, line 46: Prefer to avoid auto unless it improves readability (e.g. iterator types or templates)

flaub (approved):

shkoo (approved):
lgtm

risc0/zkvm/r0vm/r0vm.cpp
jbruestle: Due to a lot of reading/writing that is more annoying to handle in iostream form, I'm using old school C APIs elsewhere, so I'm just trying to be consistent.

PR #55, opened 2022-04-13
Clean up and comment "Buffer" type and stream buffers

shkoo:
* Rename 2 of the 3 different types named "risc0::Buffer" to have different names
* Move the third type named "risc0::Buffer" along with all the other circuit code generation stuff into the new "risc0::circuit" namespace
* Comment stream buffer classes
* Enforce stream buffer classes conform to "stream reader" and "stream writer" models
* Change behavior of CheckedStreamReader to match the other stream readers and writers with respect to read_buffer where len is not a multiple of sizeof(uint32_t)

### DIFF
--- a/risc0/core/archive.h
+++ b/risc0/core/archive.h
@@ -234,7 +234,7 @@ static_assert(is_stream_reader<BufferStreamReader>(),
               "BufferStreamReader must conform to the stream reader model");
 
 // VectorStreamWriter is a stream writer which accumulates written data into a
-// std::fector<uint32_t>.
+// std::vector<uint32_t>.
 struct VectorStreamWriter {
   void write_word(uint32_t word) { vec.push_back(word); }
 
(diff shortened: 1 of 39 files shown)


risc0/core/archive.h
```diff
@@ -179,7 +230,11 @@ struct BufferStreamReader {
 
   uint32_t* ptr;
 };
+static_assert(is_stream_reader<BufferStreamReader>(),
+              "BufferStreamReader must conform to the stream reader model");
 
+// VectorStreamWriter is a stream writer which accumulates written data into a
+// std::fector<uint32_t>.
```
jbruestle, line 237: typo.

risc0/zkvm/circuit/accum_regs.cpp
```diff
@@ -17,6 +17,7 @@
 #include "risc0/zkvm/circuit/step_state.h"
 
 namespace risc0 {
+namespace circuit {
```
jbruestle, line 20: Maybe just 'namespace risc0::circuit {'?

jbruestle (commented):
Looks good overall, see inline comments.

jbruestle (approved):

flaub (approved):
LGTM

PR #56, opened 2022-04-14
Add a benchmark for running a simple loop in a guest

shkoo:
Benchmark runnable with:

bazel run -c opt //risc0/zkvm/prove/bench:benchmark

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -31,6 +31,13 @@ http_archive(
 )
 
 http_archive(
+    name = "com_google_benchmark",
+    sha256 = "3b156bac7800f67858afe6ec2c280e291da70c9fc55377cb47ac11cc83f3128e",
+    strip_prefix = "benchmark-0d98dba29d66e93259db7daa53a9327df767a415",
+    urls = ["https://github.com/google/benchmark/archive/0d98dba29d66e93259db7daa53a9327df767a415.zip"],
+)
+
+http_archive(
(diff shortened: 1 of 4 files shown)


intoverflow:
Thank you!! This will be super useful 👍

I'm on mobile so I can't easily check, so please forgive me for asking a basic question: does the benchmark compare performance for exponentially-increasing number of iterations? That's the graph I'm most excited to see 🤞

shkoo:
> Thank you!! This will be super useful +1
> 
> I'm on mobile so I can't easily check, so please forgive me for asking a basic question: does the benchmark compare performance (run time & output size) for exponentially-increasing number of iterations? That's the graph I'm most excited to see crossed_fingers


No, I totally spaced that the loop runtime isn't linear in the guest.  I updated the benchmark so it will display the number of loops separately and to show the receipt size so you should be able to see that trend.

jbruestle (approved):
Awesome!

Some interesting observations from running it:
1) You can clearly see the PO2 steppings in the perf.  Basically no matter how many cycles the guest *actually* runs, the proof needs to be done on a perfect power of two number of cycles.  If you are just over the line things run about half as fast as if you are just under the line.
2) On my machine, for the better case, I get 1.5k loops/sec.  Looking at the asm from the guest binary via objdump, the inner loop is 4 instructions, so that about 6k insts/sec.

flaub (approved):

PR #60, opened 2022-04-15
Add Lines of Code Badge

zicklag:
Just in case you're interested. :smile:


flaub:
Thanks for the suggestion, but I'm not sure what signal (if any) we get from knowing the number of lines of code.

PR #62, opened 2022-04-16
Battleship demo integration

flaub:
Initial integration of yew/rust/wasm web client with an axum web server.

Use:
```
bazelisk run //examples/rust/battleship/web/server
```

This should launch a web browser pointing to a locally running web server which hosts the web client and provides a service for generating proof receipts (i.e. the prover).


rlukata (approved):

Macil:
I'm really interested in what kinds of guarantees you get in this scenario from using risc0. Is it equivalent to getting commitments from other players about their positions and an assurance that their moves are legal?

jbruestle:
Yes, basically each side commit to their board state (plus a salt) and then when say 'Alice' says B5, 'Bob' is required to make a zero knowledge proof about the result of the shot (Miss, Hit, Sunk my <boattype>), as well as providing the updated state (which track hits to decide when a ship is finished).  So basically, it is full cryptographic 2 player battleship, with the no third party required.  We are basically putting it up on NEAR right now, since NEAR can run our verifier on-chain (since we have a rust version, currently unreleased).  Still not quite complete on that, but should be done and full open sourced soon.

PR #68, opened 2022-04-17
More battleship demo work

flaub:
Switch to using `trunk` for building the client (it's much faster than bazel's rules_rust, especially when adding new deps).

These tools are required:
```
cargo install --locked trunk
cargo install wasm-bindgen-cli
```

In one terminal, run the prover service:
```
bazelisk run //examples/rust/battleship/web/server
```

In another terminal, run the client:
```
trunk serve --open -- examples/rust/battleship/web/client/index.html
```


jbruestle (approved):

jbruestle (commented):
Looks good.

PR #69, opened 2022-04-18
Fix zkvm_serde's Cargo.toml so "cargo test" works and so it's findable via the workspace Cargo.toml

shkoo:


flaub:
Currently this file is only used for the battleship demo, which is under heavy development. My plan is to eventually fix all of the rust bits to use cargo and make it cargo testable. But right now, I think this change will break the wasm build. That's because I do currently depend on `std` for that target.

I would say for now, `cargo test` is not yet supported. We need to be able to set features on a per-target basis for this all to be happy.

PR #78, opened 2022-04-26
Fix rand determinism and uniformity

shkoo:
Also add test for rust verifier so we can test that it reads proofs generated by C++.


risc0/zkvm/verify/src/lib.rs
```diff
@@ -60,28 +60,37 @@ impl Receipt {
...
+    use std::io;
+    use test_log::test;
 
-//     use arrayref::array_ref;
-//     use log::info;
-//     use test_log::test;
+    #[test]
+    fn test_receipt() -> io::Result<()> {
+        log::set_max_level(log::LevelFilter::Info);
+        let data: Vec<u8> = fs::read("src/simple_receipt.receipt")?;
```
jbruestle, line 75: Can we do the annoying bazel stuff to actually generate this from the c++ rather than checking in a binary?  This probably involves making a specific new bin that generates a valid receipt and calling it as a bazel build step, which is annoying but 1) I don't like 100k binaries in git 2) It will make sure that the c++ and rust stay compatible moving forward.

risc0/zkvm/verify/Cargo.toml
```diff
@@ -13,6 +13,7 @@ lazy_static = "1.4"
 risc0-zkp-core = { version = "0.1", path = "../../zkp/core"}
 risc0-zkp-verify = { version = "0.1", path = "../../zkp/verify"}
 serde = { version = "1.0", features = ["derive"] }
+risc0-zkvm-serde = { version = "0.1", path = "../../zkvm/sdk/rust/serde" }
```
jbruestle, line 16: nit: alphabetical order?   Maybe we can find a formatting tool for cargo that does that for us?

jbruestle (changes requested):

risc0/zkvm/verify/src/lib.rs
shkoo: This stuff is all built and tested from cargo, so it's difficult to access bazel.  Maybe after @flaub's reworks it will be easier?

jbruestle: Forgot about that... sigh, I guess binary it is.

jbruestle (approved):

intoverflow:
Late to the party but I want to shout-out this savvy fix 👍

https://github.com/risc0/risc0/pull/78/files#diff-a02ff5f89c003c82c8cdd536d139c0adab35e43abb90ba5a9373d9a0c6bcbd0fR65-R72

PR #79, opened 2022-04-27
Add NEAR smart contract for battleship demo

flaub:
@kandizzy @adam-yucef


flaub:
@jbruestle Unfortunately I couldn't get the `near-sdk` to play nice with all our other dependencies, so it's not part of the overall workspace. This is a bummer because it means we can't rely on a top-level `cargo test` to build and run everything (e.g. for CI).

PR #85, opened 2022-05-05
Optimize rust guest SHA and avoid undefined behavior by serializing objects before computing digest

shkoo:
* Objects are now serialized when written to commit, avoiding
  undefined behavior triggered by -Clto relating to binary layout of
  in-meomry structures.

* BumpPointerAlloc now supports allocating from the heap region, the
  commit region, the output region, and the sha descriptor region.

* Do a better job of guessing at capacity when allocating
  serialization buffers in many cases.

* Fake "Mutex" class that wraps RefCell to make it easy, efficient, and safe to
  use global mutable state in the guest singlethreaded environment

* Significant refactoring of env.rs, sha.rs, and alloc.rs, including
  some somewhat sketchy undefined behavior cases.  We now do
  minimal-copy serialization to the commit region and the output
  region.

* Wrap GPIO ports in a GpioPort structure to encapsulate the unsafe
  logic and make it more convenient to write to them.


risc0/zkvm/sdk/rust/guest/src/sha.rs
```diff
@@ -12,132 +12,178 @@
...
     }
+}
+
+// Calculates the number of words of capacity needed, including end
+// marker and trailer, to take the SHA hash of len_bytes bytes.
+const fn compute_capacity_needed(len_bytes: usize) -> usize {
+    // Add one for end marker
+    let mut len_words = align_up(len_bytes + 1, WORD_SIZE) / WORD_SIZE;
+    // Add two words for 64-bit size
+    len_words += 2;
```
flaub, line 106: Would prefer to avoid `mut` by just doing `let len_words = (some_stuff) + 2;`

risc0/zkvm/sdk/rust/guest/src/sha.rs
```diff
@@ -12,132 +12,178 @@
...
-        let mut digest: Box<mem::MaybeUninit<Digest>> = Box::new_uninit();
-        finalize_into(
-            len,
-            total,
-            self.storage.as_mut_ptr(),
-            digest.as_mut_ptr().cast(),
-        );
-        unsafe { digest.assume_init() }
+// Slower version of add_trailer for use on write-only memory which
+// writes each word exactly once.  LEngth must already be word aligned.
```
flaub, line 112: Typo: `Length`

risc0/zkvm/sdk/rust/guest/src/sha.rs
```diff
@@ -12,132 +12,178 @@
...
+    unsafe {
+        (data_u32.as_mut_ptr() as *mut u8).copy_from_nonoverlapping(data.as_ptr(), len_bytes);
+        data_u32.set_len(align_up(len_bytes, WORD_SIZE) / WORD_SIZE);
+    }
+    add_trailer(&mut data_u32, len_bytes);
+    raw_digest(data_u32.as_slice())
 }
 
+// Do some magic to finialize something relating to the guest environment.
+// TODO(nils): Document this better.
```
flaub, line 182: What this is intending to do is set a marker so that the VM knows when the last SHA descriptor is reached. We need to write to this memory location at least once so that it's not uninitialized, and the value of 0 is the marker that the VM looks for.

risc0/zkvm/sdk/rust/guest/src/sha.rs
```diff
@@ -12,132 +12,178 @@
...
-        self.storage.reserve(len_bytes);
-        unsafe {
-            let end = self.storage.as_mut_ptr().add(self.storage.len());
-            end.copy_from_nonoverlapping(ptr.cast(), len_bytes);
-            self.storage.set_len(self.storage.len() + len_bytes);
-        }
+    fn as_mut_slice(&mut self) -> &mut [u32] {
+        self.as_mut_slice()
+    }
+    fn extend_from_slice(&mut self, data: &[u32]) {
```
flaub, line 92: Please separate each `fn` with a newline
```suggestion
    }
    
    fn extend_from_slice(&mut self, data: &[u32]) {
```

risc0/zkvm/sdk/rust/guest/src/sha.rs
```diff
@@ -12,132 +12,178 @@
...
+// we can hash using a dynamically allocated buffer, or using one of
+// the fixed memory regions.
+pub trait ShaBuf: Extend<u32> {
+    fn len(&self) -> usize;
+    fn push(&mut self, val: u32);
+    fn as_slice(&self) -> &[u32];
+    fn as_mut_slice(&mut self) -> &mut [u32];
+    fn extend_from_slice(&mut self, data: &[u32]);
+    fn resize(&mut self, new_size: usize);
 }
```
flaub, line 77: What's the benefit of a trait here? I only see one impl of this trait.

risc0/zkvm/sdk/rust/guest/src/alloc.rs
```diff
@@ -14,43 +14,259 @@
...
+    cmp::max,
+    mem::size_of,
+    ops::DerefMut,
 };
 
-use crate::{_fault, align_up, REGION_HEAP_END, REGION_HEAP_START, WORD_SIZE};
+use crate::mem_layout::{REGION_HEAP_END, REGION_HEAP_START};
+use crate::mutex::{Mutex, MutexGuard};
+use crate::sha::ShaBuf;
+use crate::{_fault, align_up, WORD_SIZE};
```
flaub, line 25: For consistency, bundle uses:
```suggestion
use crate::{
  mem_layout::{REGION_HEAP_END, REGION_HEAP_START},
  mutex::{Mutex, MutexGuard},
  sha::ShaBuf,
  _fault, align_up, WORD_SIZE
};
```

risc0/zkvm/sdk/rust/guest/src/alloc.rs
```diff
@@ -14,43 +14,259 @@
...
+    fn lock_len<'a>(&'a self) -> MutexGuard<'a, usize> {
+        self.len.lock().unwrap()
+    }
+
+    pub unsafe fn alloc(&self, nitems: usize) -> *mut T {
+        self.alloc_unlocked(&mut *self.lock_len(), nitems)
+    }
+    pub unsafe fn alloc_unlocked(&self, len: &mut usize, nitems: usize) -> *mut T {
+        let ptr = self.start.add(*len);
+        let new_len = *len + nitems;
```
flaub, line 67: Seems like we need a `checked_add` here to deal with overflow

risc0/zkvm/sdk/rust/guest/src/alloc.rs
```diff
@@ -14,43 +14,259 @@
...
+    pub fn as_slice(&self) -> &[T] {
+        self.as_slice_unlocked(*&mut *self.lock_len())
+    }
+    pub fn as_slice_unlocked(&self, len: usize) -> &[T] {
+        unsafe { &*core::slice::from_raw_parts(self.start, len) }
+    }
+
+    // pub fn as_mut_slice(&self) -> &mut [T] {
+    //     self.as_mut_slice_unlocked(*&mut *self.lock_len())
+    // }
```
flaub, line 123: Is this still needed?

risc0/zkvm/sdk/rust/guest/src/alloc.rs
```diff
@@ -14,43 +14,259 @@
...
+    pub fn as_mut_slice_unlocked(&self, len: usize) -> &mut [T] {
+        unsafe { &mut *core::slice::from_raw_parts_mut(self.start, len) }
+    }
+
+    // Returns the slice of new values to be initialized, if any.
+    // pub fn resize(&self, new_size: usize) -> &mut [T] {
+    //     self.resize_unlocked(&mut &mut *self.lock_len(), new_size)
+    // }
```
flaub, line 131: Still needed?

flaub (commented):
I haven't made it completely through this yet. I have some questions about why we need a new bump allocator. I also wonder about the difference between `Once` and using a fake `Mutex`, they seem basically equivalent to me (essentially my reading of `UnsafeCell` is that it's fine in single-core environments). Let's setup a chat so I can understand better some of these changes.

shkoo:
I reverted a bunch of the more dubious of these changes.  It still looks like we have about a 25% speedup, and it now passes tests with -Clto turned on.

flaub (approved):
LGTM!

flaub:
Should we turn `-Clto` on by default now?

PR #91, opened 2022-05-07
Add method id verification to the rust verifier

### LINKED ISSUE #89: Fix rust verifier to correctly verify the code being run
Currently https://github.com/risc0/risc0/blob/main/risc0/zkvm/verify/src/circuit.rs is missing the code required to verify which method is being run (which the C++ has).

shkoo:
Fixes #89


jbruestle (approved):
This looks great.  Generally, the rest of the verifier ends up doing a panic on any error.  Clearly propagating errors is better, but at the time I was porting things, it seemed like a heavy lift (I think I still prefer exceptions), so this is a step in the proper direction.  This is an API breaking change for the NEAR smart contract (https://github.com/risc0/risc0/tree/main/examples/rust/battleship/contract), which we really need to add some tests for, but I don't think we should block this commit on that.

PR #93, opened 2022-05-07
Trim rust verifier dependencies

shkoo:
* Remove dependencies on arrayref, lazy_static, std crates

* Remove direct dependencies on typenum, generic_array crates (we still have indirect dependencies)

* Remove zkvm/verify/src/eval.rs, which looks mostly like a copy of circuit.rs that nothing uses


flaub (approved):
Nice!

PR #95, opened 2022-05-10
Support cargo installable make-id

flaub:
This is a pretty large PR which is a stepping stone for supporting a purely cargo-based build system.

* Use a rust build script (`build.rs`) in order to allow each C++ library to be built via cargo.
* Add a rust-based `make-id` which uses `cxx` to generate FFI so that rust can call into C++ code.
* Conditionally compile `pure`/`cxx` versions of crates. The `pure` feature represents the pure rust implementation. The `cxx` feature represents the FFI wrapped C++ library.
* The NEAR smart contract uses the `pure` feature, while `make-id` uses the `cxx` feature.
* TODO: add support for building TBB from a rust build script.


examples/rust/battleship/methods/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
+target = "../../../../bazel/rules/rust/riscv32im-unknown-none-elf.json"
```
shkoo, line 2: Neat, so that's how you do that without invoking bazel.  TIL! :)

risc0/zkp/core/Cargo.toml
```diff
@@ -1,13 +1,25 @@
...
+version = "0.2.0"
 edition = "2021"
 description = "RISC Zero zero-knowledge proof system core crate"
 license = "Apache-2.0"
 homepage = "https://risczero.com/"
 repository = "https://github.com/risc0/risc0/"
+links = "risc0-zkp-core"
+
+[features]
+default = ["pure"]
```
shkoo, line 12: Can we document what this "pure" feature does somewhere?

risc0/zkp/verify/Cargo.toml
```diff
@@ -1,12 +1,23 @@
...
 description = "RISC Zero zero-knowledge proof system verify crate"
 license = "Apache-2.0"
 homepage = "https://risczero.com/"
 repository = "https://github.com/risc0/risc0/"
+links = "risc0-zkp-verify"
+
+[features]
+default = ["pure"]
+cxx = ["dep:cxx", "risc0-zkp-core/cxx"]
+pure = ["dep:rand"]
```
shkoo, line 14: Can we document what this "pure" feature does?

risc0/zkvm/prove/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "risc0-zkvm-prove"
+version = "0.1.0"
+edition = "2021"
+links = "risc0-zkvm-prove"
+
+[dependencies]
+anyhow = "1.0"
```
shkoo, line 8: 👍 to "anyhow" which looks super cool to make error handling less cumbersome.  I was looking at "quick_error", but this looks somewhat easier to use.

risc0/zkvm/prove/method_id.cpp
```diff
@@ -66,4 +66,18 @@ void writeMethodID(const std::string& filename, const MethodID& id) {
...
 
+namespace rust {
+
+MethodID::MethodID(const std::string& elf_path) : id(makeMethodID(elf_path)) {}
+
+std::unique_ptr<MethodID> new_method_id(const std::string& elf_path) {
+  return std::make_unique<MethodID>(elf_path);
+}
+
+void MethodID::write(const std::string& filename) const {
```
shkoo, line 77: I'm not really excited about the API requiring all this stuff to be in OS files, but I guess that's how it is currently, and it's relatively straightforward to change later.

risc0/zkvm/verify/Cargo.toml
```diff
@@ -1,23 +1,34 @@
...
 description = "RISC Zero zero-knowledge VM verify crate"
 license = "Apache-2.0"
 homepage = "https://risczero.com/"
 repository = "https://github.com/risc0/risc0/"
+links = "risc0-zkvm-verify"
+
+[features]
+default = ["pure", "std"]
+cxx = ["dep:cxx", "dep:risc0-core", "dep:risc0-zkvm-circuit", "risc0-zkp-core/cxx"]
+pure = ["dep:risc0-zkp-verify", "dep:risc0-zkvm-serde", "dep:serde"]
```
shkoo, line 14: Do we need to put e.g. "risc0-zkp-verify/pure" in these features lines in order for the "pure" feature to propagate to the dependencies?  I'm still a bit fuzzy on cargo dependency specifications.

shkoo (changes requested):
Neat!

risc0/zkvm/prove/method_id.cpp
flaub: This PR was getting pretty big so decided to avoid doing functional changes for now, but I agree.

.bazelignore
```diff
@@ -1 +1,2 @@
+examples/rust/battleship/contract
```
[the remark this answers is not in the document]
flaub: Bazel was getting into a recursive loop. I'm puzzled why the `target/` didn't ignore *all* dirs named target, so seems redundant.

I'll try with `.../contract/target`

risc0/zkp/core/Cargo.toml
flaub: Maybe I should call it the `rust` feature? Where should I document this?

risc0/zkvm/verify/Cargo.toml
flaub: It's new for me as well. Maybe? I haven't run into any issues yet. But my understanding is that because the `pure` feature is in the `default` feature of `risc0-zkp-verify`, then it should be OK to elide it here. But maybe it's better to be explicit anyways.

I'm actually puzzled why I didn't need to put `risc0-zkp-verify/cxx` in the `cxx` feature.

flaub: OK, I see what's going on. The `cxx` feature doesn't depend on `risc0-zkp-verify` apparently.

examples/rust/battleship/contract/build.rs
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+use std::env;
+use std::fs;
+use std::path::Path;
+
+fn make_id(name: &str) {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
+    let dest_path = Path::new(&out_dir).join(name);
+    fs::write(&dest_path, "foobar").unwrap();
```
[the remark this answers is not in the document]
flaub: Added a TODO

risc0/zkp/core/Cargo.toml
flaub: I added some comments to this file.

risc0/zkp/verify/Cargo.toml
flaub: Added some comments

shkoo (approved):
Nice!

PR #100, opened 2022-05-11
Use cargo to build zkvm and allow out of tree builds

flaub:


jbruestle (commented):
So this looks good, and I tried a clean checkout and tried both the cargo and bazel builds.  That said, it seems like there are still some issues:  Specifically, it looks like you made make-circuit work in c++ and rust, but that step.cpp.inc is now being checked in?  This seems potentially problematic.  I tried to run make-circuit and at least verify the committed file matched, but the paths used in the build get put into step.cpp.inc all include the source directory, so they don't match.  This is a potential security issue long term if it's not verified by any tests, and moreover I'd like to get rid of checking in generated files (which I totally realize we also do for the rust verifier and the seal tests, but I'd like to fix those ASAP).  Overall, I guess if it's only another day or two, I'd fix that first before we merge it into main to not accumulate technical debt.  If not, let's talk about what the plan would look like.

flaub:
My plan with this PR is to follow up with a few more commits to address remaining issues. Options:
1. deterministic codegen for `make-circuit`
2. figure out a way to make the generated code be its own crate that downstream users can pull in without having to install another tool

flaub:
@jbruestle This version uses some tricks to avoid having to check in the `step.cpp.inc` file.

jbruestle (approved):
Super psyched.  Like the approach with method ids too, which I imagine we can improve by wrapping the build.rs with some utils.  Excited to get some crates soon.

PR #102, opened 2022-05-11
Randomized ship placement

adam-yucef:
This PR replaced fixed ship placement with randomized ones.


examples/rust/battleship/web/client/src/game.rs
```diff
@@ -73,10 +76,120 @@ pub struct GameSession {
...
+    if dir == 0 {
+        bound = x + span <= BOARD_SIZE - 1
+    } else {
+        bound = y + span <= BOARD_SIZE - 1
+    }
+
+    // log::info!("{} {}", "check_bounds", bound);
+
+    bound
+}
```
flaub, line 92: Since the `if` in rust is an expression, you can just do:

```suggestion
fn check_bounds(x: usize, y: usize, span:usize, dir:usize) -> bool {
    if dir == 0 {
        x + span <= BOARD_SIZE - 1
    } else {
        y + span <= BOARD_SIZE - 1
    }
}
```

examples/rust/battleship/core/src/lib.rs
```diff
@@ -33,19 +33,19 @@ pub const SHIP_SPANS: [usize; NUM_SHIPS] = [
 
 pub const SUNK_MASKS: [u8; NUM_SHIPS] = [0x1f, 0x0f, 0x07, 0x07, 0x03];
 
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Hash)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, Hash, Copy)]
```
flaub, line 36: I'd prefer copies to be explicit for structs generally. There's about 3 options I can think of when rust complains about something relating to borrow checking:

1. Implement `Copy` (implicit clone).
2. Explicitly call `.clone()`.
3. Pass by ref with a borrow than by value with a clone/copy.

My assumption is that you wanted to have `Copy` because the compiler complained, but I'd suggest trying to use `.clone()` where it was complaining.

examples/rust/battleship/web/client/src/game.rs
```diff
@@ -103,7 +103,7 @@ fn check_taken(x: usize, y: usize, span:usize, dir:usize, taken:Taken) -> bool {
 fn create_random_ships() -> [Ship; 5] {
     // randomly place 5 ships on the board
     let mut rng = thread_rng();
-    let mut ships = [Ship::default(); 5];
+    let mut ships : [Ship; 5] = array_init::array_init(|_| Ship::default());
```
kandizzy, line 106: @flaub I made an update to remove `Copy`. This was the line that was complaining. I saw a few ways to get around this, but opted for a safe wrapper called `array_init`. Let me know what you think of this approach.

flaub: LGTM

flaub (approved):

PR #105, opened 2022-05-13
Move rust SHA hashing behind a Sha trait.

shkoo:
This way in the future we could have differing implementations for e.g. hashing in the guest, or using a GPU.


jbruestle (commented):
Interesting.  I presume the thinking is that we would then add an implementation that uses the accelerator inside the ZKVM?  Two high level thoughts:  1) I wonder if we could make platform specific part be just the 'engine', i.e. something that takes maybe N blocks (each of 16 words), and does 1 compress per block and outputs 8 words.  Then we could make the padding/encoding/etc shared.  2) For the prover, there is an additional bit of functionality we need: Hashing non-stride 1 lists of things (in particular Fp's).  Basically, I want to make word 'i' of the sha blocks be A[offset + stride * i] for some non-1 stride.  For the GPU case, we do all of the gathering required on the GPU, otherwise we get bottlenecked on GPU<->CPU traffic.

shkoo:
Yup!  The problem with reducing the number of API calls is that different implementations will be able to do different things efficiently.  For instance, copying memory in the guest is very expensive, so we want to have APIs where we copy it as little as possible, since the guest will have to copy it anyways because it needs to add the end marker, padding, and length to the end.

I like having a trait for the API because:

1) It makes it real obvious what the fix should be if there's a ShaImpl that doesn't implement one of the required methods.

2)  It's easy to automate testing, since we can expose e.g. test_sha_impl that will make sure that a new ShaImpl implementation conforms.

risc0/zkp/core/src/sha.rs
```diff
@@ -12,101 +12,64 @@
...
+        &mut self.0
     }
-}
 
-impl Digest {
-    /// Constructs a new `Digest` from a byte array.
-    pub fn new(data: [u32; DIGEST_WORDS]) -> Self {
-        Self(data)
+    pub fn to_hex(&self) -> String {
+        format!("{}", self)
```
flaub, line 66: `format!` is actually pretty expensive in the guest, it causes a large amount of bloat. I'm OK with this for now, but perhaps a TODO would be good.

risc0/zkp/core/src/sha.rs
```diff
@@ -127,3 +90,89 @@ impl Debug for Digest {
...
 }
+
+impl Clone for Digest {
+    fn clone(&self) -> Digest {
+        Digest(self.0)
+    }
+}
+
+// An implementation that provides SHA-256 hashing services.
+pub trait ShaImpl: Clone + Debug {
```
flaub, line 101: I'm a little confused when adding an `Impl` suffix to a `Trait`. I think a more idiomatic rust spelling would just be `pub trait Sha`.

shkoo: Done

risc0/zkp/core/src/sha.rs
shkoo: Done

flaub (approved):

PR #109, opened 2022-05-15
Add support for building out of tree with cargo

flaub:
* Adjust prover/verify API so that the ELF path and the MethodID path can be decoupled.
* Use a convention for cross-compiling methods.
* Add risc0-build crate, which helps get methods cross-compiled and properly linked.
* Add cargo integration tests for the battleship example and for the `risc0-zkvm-host` crate.
* Add `release.toml` files to be used with the `cargo release` tool. This tool is useful for bumping version numbers and publishing crates en masse.
* Make taps builder safe in the face of multiple threads.
* Allow `risc0-zkvm-guest` to compile on the host platform by making it be a no-op. This allows `cargo release` to work with this crate.
* Add missing copyright headers.
* Use `rust-toolchain` to ensure a nightly toolchain is used. Currently, only a nightly toolchain is supported.


flaub:
@jbruestle this is ready to go :)

jbruestle (approved):

PR #115, opened 2022-05-20
Use a 2 phase build system to ensure correctness

flaub:


jbruestle (changes requested):
Can we rename the binary to risc0-build-guest?  Or something similar.

flaub:
I named it `risc0-build-methods`. Note that this is an arbitrary binary name for this repository only. Out of tree users will have to name their binary something but this is a fine convention.

jbruestle (approved):

PR #127, opened 2022-05-24
Change "merke" to "merkle"

BoltonBailey:
I noticed a few files have "Merke" instead of "Merkle" so I changed these. Let me know if there are any conventions I should follow when making PRs.


jbruestle (approved):
Thanks, I’m bad at spelling…

PR #128, opened 2022-05-24
Changes to support builds on docs.rs

flaub:


jbruestle (approved):
Super glad for this.  I'll actually file in docs now.

PR #132, opened 2022-05-26
Minor Prover API Improvements

spaugh:


risc0/build/src/lib.rs
```diff
@@ -159,16 +159,17 @@ pub fn embed_methods() {
                     eprintln!("Try: `cargo run --bin risc0-build-methods`");
                     std::process::exit(-1);
                 }
-                let method_id = MethodID::new(&elf_path.to_str().unwrap()).unwrap();
-                method_id.write(&id_path.to_str().unwrap()).unwrap();
+                let method_id: &[u8];
+                unsafe {
```
intoverflow, line 163: Why is this unsafe?

risc0/zkvm/lib.rs
```diff
@@ -22,24 +24,27 @@ mod ffi {
...
         let_cxx_string!(elf_path = elf_path);
-        let raw = ffi::new_method_id(&elf_path)?;
+        let raw = ffi::make_method_id(&elf_path)?;
         Ok(MethodID { raw })
     }
 
-    pub fn write(&self, filename: &str) -> Result<()> {
-        let_cxx_string!(filename = filename);
-        Ok(self.raw.write(&filename)?)
+    pub unsafe fn to_bytes(&self) -> Result<&[u8]> {
```
intoverflow, line 43: Why is this unsafe?

risc0/build/src/lib.rs
```diff
@@ -159,15 +159,17 @@ pub fn embed_methods() {
                     eprintln!("Try: `cargo run --bin risc0-build-methods`");
                     std::process::exit(-1);
                 }
-                let method_id = make_method_id(&elf_path.to_str().unwrap());
-                let method_id_buf = from_raw_parts(method_id as *const u8, size_of::<MethodID>());
+                let method_id: &[u8];
+                unsafe {
+                    method_id = MethodID::new(&elf_path.to_str().unwrap()).unwrap().to_bytes().unwrap();
+                }
```
flaub, line 165: BTW, in rust you can say:
```
let method_id = unsafe { ... };
```

intoverflow (commented):
Thank you for helping improve the API! I left some naive questions about the use of unsafe.

risc0/zkvm/lib.rs
flaub: `slice::from_raw_parts` is unsafe. But if we can make some good safety assumptions then perhaps we can hide the `unsafe` from the `to_bytes` function signature and instead use an internal `unsafe {}` block.

risc0/zkvm/lib.rs
```diff
@@ -22,24 +24,27 @@ mod ffi {
 
         type MethodID;
 
-        fn new_method_id(elf_path: &CxxString) -> Result<UniquePtr<MethodID>>;
-        fn write(&self, filename: &CxxString) -> Result<()>;
+        fn make_method_id(elf_path: &CxxString) -> Result<UniquePtr<MethodID>>;
     }
 }
 
+// Can we remove this?
```
flaub, line 31: I don't think so?

risc0/build/src/lib.rs
flaub: Good call :)

risc0/zkvm/lib.rs
flaub: After more thought, I don't think we need to convert to a buffer here, since we already have a `MethodID` which has a representation in C++, I think we can just pass a pointer of the raw `ffi::MethodID` down.

risc0/build/src/lib.rs
spaugh: I think it's just leftover from whatever was there before. As a rust noob Im doing a lot of experimentation...This stuff will be cleaned up before I remove the draft status.

risc0/zkvm/lib.rs
spaugh: This is here so we can write the bytes to the methods file though...I think...I'm on mobile and can't really see the whole file.

risc0/zkvm/lib.rs
spaugh: I suppose I should've phrased it "can we reduce the surface area of this api". I don't think it's particularly important but I find those sorts of small challenges useful for building my understanding.

risc0/zkvm/lib.rs
```diff
@@ -12,18 +13,23 @@
...
 
         type MethodID;
 
-        fn new_method_id(elf_path: &CxxString) -> Result<UniquePtr<MethodID>>;
-        fn write(&self, filename: &CxxString) -> Result<()>;
+        // How do I get this constexpr from C++?
+        //static DIGEST_BYTES: usize;
+
+        fn method_id_from_elf(path: &CxxString) -> Result<UniquePtr<MethodID>>;
+        fn method_id_from_bytes(bytes: &[u8; 384]) -> Result<UniquePtr<MethodID>>;
```
spaugh, line 30: This array size exists as a constexpr in C++ (sizeof on the underlying data structure, std::array<ShaDigest, kDigestCount>) but I can't figure out how to get that value in the rust world. Unless someone cleverer than I am figures out how to do that, the options are:

1. redefine the constant on the rust side with a magic number
2. redefine the constant on the rust side by copying all of the underlying constants over and calculating it
3. just use a slice instead of an array, losing some type safety and exposing a bit more of the ffi internals (because slice doesn't have an exact cpp equivalent, so you have to deal with rust::Slice)

thoughts, @flaub? I'm leaning towards #1 as a short term fix and then moving the MethodID definition fully over to rust (with an interface exposed to cpp) as a long term fix, as I don't believe anywhere in c++ world actually cares about the serialization of that thing.

spaugh:
> I'm on my phone on a plane at the moment and will look deeper at this when I get the chance, but this use of memcpy where the bounds checks are implicit and not local to the function here and on line 67 seems brittle on first reading.

Yeah that overload is not great. I need to think a bit about how to make it less brittle.

risc0/zkvm/lib.rs
flaub: We could just use a slice instead of an array, and then just do runtime checking that the len is correct.

risc0/zkvm/lib.rs
```diff
@@ -14,37 +15,46 @@
...
         type MethodID;
 
-        fn make_method_id(elf_path: &CxxString) -> Result<UniquePtr<MethodID>>;
+        // How do I get this constexpr from C++?
+        //static DIGEST_BYTES: usize;
+
+        fn method_id_from_elf(path: &CxxString) -> Result<UniquePtr<MethodID>>;
+        fn method_id_from_bytes(bytes: &[u8; 384]) -> Result<UniquePtr<MethodID>>;
+
+        fn toBytes(self: &MethodID) -> Result<[u8; 384]>;
```
flaub, line 32: In rust, the convention is `snake_case`

spaugh: Yeah, just missed one.

risc0/build/src/lib.rs
```diff
@@ -45,13 +45,11 @@ impl Risc0Metadata {
 struct Risc0Method {
     name: String,
     elf_path: PathBuf,
+    cache_dir: PathBuf, // It's a bit questionable if this belongs here
 }
 
 impl Risc0Method {
```
spaugh, line 51: @shkoo See here. Instead of writing the hash to `method_id.elf_sha`, I wrote the method id to `{hash}.cache`

risc0/build/src/lib.rs
```diff
@@ -45,13 +45,11 @@ impl Risc0Metadata {
...
-    fn make_method_id<P>(&self, method_id_path: P)
-    where
-        P: AsRef<Path>,
-    {
+    fn make_method_id(&self) -> [u8; 384] {
```
spaugh, line 52: @flaub see here (and in a couple other places). I mentioned this in a previous version of the PR but i want to call it out again. There is currently no way to get a constexpr over the cxx bridge, so it either has to be redefined on the rust side or the rust side has to treat it as a [unsized] slice. Here I've got the actual number scattered all over but, if kept, I would collapse it into a constant somewhere. I've noticed a couple other constants duplicated in both rust and c++, so I don't think it'd be setting a bad precedent or anything. It's also unlikely to stick around for very long.

flaub: Yeah it's probably fine for small stuff like this.

spaugh:
@flaub this is a breaking change so it might make sense to get it in before 0.8.0

flaub:
I'll just bump to 0.9.0 once this lands

flaub:
Is this ready for review?

risc0/build/Cargo.toml
```diff
@@ -13,4 +13,5 @@ risc0-zkvm-platform-sys = { version = "0.7", path = "../zkvm/platform" }
 risc0-zkvm-sys = { version = "0.7", path = "../zkvm" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-sha2 = { version = "0.10" }
+anyhow = { version = "1.0", default-features = false }
+sha2 = { version = "0.10" }
```
flaub, line 17: Needs sorting, also needs newline on last line.

risc0/build/src/lib.rs
```diff
@@ -225,29 +216,20 @@ where
 pub fn embed_methods() {
     let pkg = current_package();
 
-    let out_dir_env = env::var_os("OUT_DIR").unwrap();
-    let out_dir = Path::new(&out_dir_env);
-    let guest_target_dir = out_dir.join("riscv-guest");
+    let out_path = env::var_os("OUT_DIR").unwrap();
```
flaub, line 219: I think `out_dir_env` is slightly better name since the thing returned from `env::var_os` isn't actually a path but a string.

risc0/zkvm/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(extern_types)]
```
flaub, line 1: I think we should avoid using unstable features. Especially on the host side; a stable compiler is supported currently.

risc0/zkvm/lib.rs
```diff
@@ -12,34 +13,25 @@
...
+pub mod ffi {
     unsafe extern "C++" {
         include!("risc0/zkvm/prove/method_id.h");
 
-        type MethodID;
-
-        fn new_method_id(elf_path: &CxxString) -> Result<UniquePtr<MethodID>>;
-        fn write(&self, filename: &CxxString) -> Result<()>;
+        #[cxx_name = "makeMethodId"]
+        fn make_method_id_from_elf(path: &CxxString) -> Result<[u8; 384]>;
```
flaub, line 26: Use `METHOD_ID_LEN`

risc0/zkvm/lib.rs
```diff
@@ -12,34 +13,25 @@
...
-    raw: UniquePtr<ffi::MethodID>,
-}
-
-impl MethodID {
-    pub fn new(elf_path: &str) -> Result<Self> {
-        let_cxx_string!(elf_path = elf_path);
-        let raw = ffi::new_method_id(&elf_path)?;
-        Ok(MethodID { raw })
-    }
+pub const METHOD_ID_BYTES: usize = 384; // https://github.com/dtolnay/cxx/issues/1051
```
flaub, line 30: I'd rename this to `METHOD_ID_LEN`

flaub: Also, couldn't we just expose a function that returns the the c++ constant?

spaugh:
yeah - bumped the status into ready

risc0/zkvm/lib.rs
```diff
@@ -12,34 +13,25 @@
...
+pub type MethodId = [u8; METHOD_ID_BYTES];
 
-    pub fn write(&self, filename: &str) -> Result<()> {
-        let_cxx_string!(filename = filename);
-        Ok(self.raw.write(&filename)?)
-    }
+pub fn make_method_id_from_elf(path: &str) -> Result<MethodId> {
+    let_cxx_string!(cxx_path = path);
+    ffi::make_method_id_from_elf(&cxx_path)
+        .map_err(|err| Error::msg(format!("{}", err.what())))
```
flaub, line 36: It's not clear why we need this. Can we say:
```
Ok(ffi::make_method_id_from_elf(&cxx_path)?)
```

risc0/zkvm/prove/method_id.cpp
```diff
@@ -23,20 +24,38 @@
...
-  uint32_t startAddr = loadElf(elfFile, kMemSize, image);
+  uint32_t startAddr = loadElf(elfPath, kMemSize, image);
 
   // Start with an empty return value
-  MethodID ret;
+  MethodDigest digest;
 
   // Make the digest for each level
-  for (size_t i = 0; i < kCodeDigestCount; i++) {
+  for (size_t i = 0; i < numMethodDigests; i++) {
```
flaub, line 54: In C++, use `kSomething` for constants (ala Google coding style).

I think `kCodeDigestCount` is a reasonable name here.

risc0/zkvm/sdk/cpp/host/c_api.cpp
```diff
@@ -77,9 +78,13 @@ void risc0_string_free(risc0_string* str) {
   ffi_wrap_void(&err, [&] { delete str; });
 }
 
-risc0_prover* risc0_prover_new(risc0_error* err, const char* elf_path, const char* method_id_path) {
+risc0_prover* risc0_prover_new(risc0_error* err,
+                               const char* elf_path,
+                               const uint8_t* method_id_buf,
+                               const size_t method_id_len) {
```
flaub, line 84: I don't think we need the length here because it's a constant size.

flaub (commented):
Some minor comments but mostly LGTM.

I think later we might want to have the FFI return an opaque type and the rust code would only ever see the opaque type. At the moment I don't think any rust code needs to look inside. But I think we can do this in a follow up (it's starting to get hard to review as is).

spaugh:
> Some minor comments but mostly LGTM.
> 
> I think later we might want to have the FFI return an opaque type and the rust code would only ever see the opaque type. At the moment I don't think any rust code needs to look inside. But I think we can do this in a follow up (it's starting to get hard to review as is).

Agreed on both fronts. This is sort of a weird diff because I was feeling my way around in the dark for quite a while. And I actually had to back some changes out as it was getting substantially more complex than this and it was making me feel uncomfortable.

risc0/zkvm/sdk/cpp/host/c_api.cpp
spaugh: That's how I had it originally but felt a little uneasy about it. Happy to remove it though - you've definitely got more expertise here.

risc0/zkvm/prove/method_id.cpp
spaugh: Ah. I don't even think that "rename" was intentional. Will change back.

risc0/build/src/lib.rs
```diff
@@ -62,44 +60,36 @@ impl Risc0Method {
...
-        let elf_contents = std::fs::read(&self.elf_path).unwrap();
-        let elf_sha = Sha256::new().chain_update(elf_contents).finalize();
-        let elf_sha_hex: String = elf_sha
-            .as_slice()
-            .iter()
-            .map(|x| format!("{:02x}", x))
-            .collect();
+        let method_id_cache_path = std::fs::read(&self.elf_path)
+            .map(|content| Sha256::new().chain_update(content).finalize())
+            .map(|digest| format!("{:x}", digest))
```
flaub, line 65: This will collide without ensuring the hex is always 2 digits.

risc0/zkvm/lib.rs
spaugh: Oh I missed the question mark in yours. Yeah I think that would work?

risc0/zkvm/lib.rs
spaugh: then we wouldn't be able to use it at compile time, which would defeat the purpose, mostly

risc0/build/src/lib.rs
flaub: I'm actually not sure why this was changed actually. The thing that we compare as to whether we need to re-compute the MethodID is the SHA of the ELF. It seems like the only change needed in this file is just the call to `make_method_id_from_elf` instead of `MethodID::new`. Would be even nicer if we kept the same `MethodID::new` interface.

flaub: To be clear, we need to generate 3 files:
* The ELF file itself
* The SHA of the ELF (for later detection of the need to rebuild MethodID).
* The cached MethodID (so we can reuse it if the SHA hasn't changed)

risc0/zkvm/lib.rs
flaub: Ah yes, good point!

risc0/build/src/lib.rs
flaub: I think I can see why this logic will work now. However, the one issue I see with this new approach is that the target dir will fill up with stale `{hash}.cache` paths that don't go away until someone cleans their target dir.

risc0/zkvm/lib.rs
spaugh: CXX won't allow it...but at least this will break at compile time if the two constants get out of whack

risc0/build/src/lib.rs
spaugh: Yeah the multiple files solution is definitely better because of the lack of pollution. But re: your first comment I had to get the method_id from somewhere. I could've made a different file, put it in the same file and parsed it out, or did what i did. I probably chose the wrong solution and we should fix it. I'm not sure I see the reason why the MethodID::new interface would've been better. I found having the additional type to be a bit confusing because we still had to deal with the byte array no matter what. :shrug:

flaub (approved):

Issue #116, opened 2022-05-20
Make ELF binaries build deterministically 


jbruestle:
Currently when Cargo builds the guest methods, the resulting ELF files are not built deterministically (don't know why) resulting in incompatible method IDs.  Technically only the 'loadable' part of the ELF needs to be deterministic.

shkoo:
rust-lang/cargo#10746

shkoo:
Reopening for now, since std support (#148) requires a host spec file again.

If that cargo pull request gets merged that should fix this issue (but we'll want to add a cargo version check and warn if it's too old).  The other option is to build our own toolchain including cargo, and that's a bigger project.

flaub:
Found some other related issues:
https://github.com/rust-lang/rust/issues/75362
https://github.com/rust-lang/rust/issues/102299

Looks like we could try to use:
```
RUSTFLAGS="--remap-path-prefix=$HOME=/remap-home --remap-path-prefix=$PWD=/remap-pwd"
```

flaub:
https://github.com/rust-lang/rust/issues/98185
https://github.com/rust-lang/rust/labels/A-reproducibility

flaub:
https://github.com/rust-lang/rust/issues/34902

NEAR also uses a docker image: https://github.com/near/near-sdk-rs/issues/151

Pernick:
This also affects builds where multiple members of the workspace have guest methods as a dependency and using its image id for verifying https://github.com/RBFLabs/risc0-id-mismatch

SchmErik:
Here are several code changes that we've added in our tests which are likely due to nondeterministic binaries:

https://github.com/risc0/risc0/pull/553/files#diff-46a860afd43fe55f5e7442917718a4a5357fb4a76b2169d98c8c885158af8e2dR467

The `% 1094` was added because the different instructions could be laid out on page boundaries in certain situations and so when you execute the next instruction, it could cause a page-in. If the build were to be deterministic the difference between cycles would equate to a single value rather than `% 1094` which indicates multiples of the page-in cycle count.

We also increased the segment's power of 2 limit for the continuations test:
https://github.com/risc0/risc0/commit/a0f8ce3aa88df84d7a5613b2fa58c16a6a61d224#diff-86c5437677f80d40396200163171134d15dc3e6eb257f74c37f382c39da52697R207

[...]

PR #152, opened 2022-06-08
Build guest method crates automatically

shkoo:
* Deprecate risc0-build-methods as it's no longer necessary
* Only recalculate method IDs when necessary since it's slow
* Out-of-tree builds with --target-dir now work


risc0/build/src/lib.rs
```diff
@@ -82,99 +197,67 @@ pub fn build(manifest_path: &Path, target_dir: &Path) {
...
-        let manifest_path = Path::new(&pkg.manifest_path)
-            .parent()
-            .unwrap()
-            .join(rel_path)
-            .join("Cargo.toml");
+        // Try to rebuild guest packages if anything changes in the guest package directory.
+        println!(
+            "cargo:rerun-if-changed={}",
+            guest_pkg.manifest_path.parent().unwrap()
+        );
```
flaub, line 250: Ah yes! This is the secret sauce! Nice work.

flaub: Is this transitive?

spaugh:
holy shit. 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍 😍

risc0/build/src/lib.rs
```diff
@@ -82,99 +197,67 @@ pub fn build(manifest_path: &Path, target_dir: &Path) {
...
         std::process::exit(status.code().unwrap());
     }
 }
 
 /// Embeds methods built for RISC-V for use by host-side dependencies.
 ///
+/// This method should be called from a package with a
+/// [package.metadata.risc0] section including a "methods" property
+/// listing the subdirectories that contain riscv guest method
+/// packages.
```
flaub, line 216:
```suggestion
/// This method should be called from a package with a
/// [package.metadata.risc0] section including a "methods" property
/// listing the relative paths that contain riscv guest method
/// packages.
```

risc0/build/src/lib.rs
```diff
@@ -12,171 +12,252 @@
...
         }
     }
+
+    // HACK: It's not particularly practical to figure out all the
+    // files that all the guest crates transtively depend on.  So, we
+    // want to run the guest "cargo build" command each time we build.
+    //
+    // Since we generate methods.rs each time we run, it will always
+    // be changed.
+    println!("cargo:rerun-if-changed={}", methods_path.display());
```
flaub, line 260: Love it, this is a super clever hack!

risc0/build/src/lib.rs
```diff
@@ -12,171 +12,252 @@
...
     process::Command,
 };
 
-use cargo_metadata::MetadataCommand;
+use cargo_metadata::{MetadataCommand, Package};
 use risc0_zkvm_platform_sys::LINKER_SCRIPT;
 use risc0_zkvm_sys::MethodID;
 use serde::Deserialize;
-
-const TARGET_JSON: &[u8] = include_bytes!("../riscv32im-unknown-none-elf.json");
```
flaub, line 30: Let's remove this file from the repo now that it's not needed

risc0/build/src/lib.rs
```diff
@@ -12,171 +12,252 @@
...
+{
+    fs::create_dir_all(target_dir.as_ref()).unwrap();
+
+    let cargo = env::var("CARGO").unwrap();
     let args = vec![
         "build",
         "--release",
         "--target",
-        target_path.to_str().unwrap(),
+	"riscv32im-unknown-none-elf",
```
flaub, line 192: This seems like a rustfmt issue?

flaub (approved):

flaub:
@shkoo Does this PR now fix #116?

PR #155, opened 2022-06-09
Avoid accumulation of cached MethodIDs

flaub:


.github/workflows/main.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [Linux, macOS, Windows]
```
jbruestle, line 36: I this because this breaks windows, or because we want to stop supporting windows for a bit?  I haven't been following CI stuff very much.

jbruestle (approved):
Looks good.  Approving as is, but I'm curious about the windows change, figure we can always roll it back if it wasn't intentional.

.github/workflows/main.yml
flaub: The cargo windows build has race conditions due to our use of `cxx-build`. I tried to do some workarounds but none of them seem to work.

PR #161, opened 2022-06-15
Support rust standard library in guest

### LINKED ISSUE #148: Add std support for guest

shkoo:
Automatically download standard library from risc0's rust repository

Bump current rust nightly channel up to 2022-06-13

Fixes #148

### DIFF
--- a/risc0/zkvm/sdk/rust/serde/src/err.rs
+++ b/risc0/zkvm/sdk/rust/serde/src/err.rs
@@ -59,5 +59,5 @@ impl serde::de::Error for Error {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for Error {}
+// This is an alias for either std::Error, or serde's no_std error replacement.
+impl serde::ser::StdError for Error {}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,24 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
+name = "aes"
+version = "0.7.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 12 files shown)


shkoo:
Unfortunately this also breaks #116 again

risc0/zkvm/sdk/rust/serde/src/err.rs
```diff
@@ -59,5 +59,5 @@ impl serde::de::Error for Error {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for Error {}
+// This is an alias for either std::Error, or serde's no_std error replacement.
+impl serde::ser::StdError for Error {}
```
flaub, line 63: ❤️

flaub:
My inclination would be to wait until #116 is fixed upstream since build reproducibility is a larger need than `std` support at the moment. 

I'm also curious what it would take to get a standard `main` entrypoint. I suspect we need to adjust the zkVM circuit since we rely on a specific ABI for the entry (both on input and the 'result' lands in registers currently).

flaub:
@jbruestle WDYT?

jbruestle:
I am so psyched about std support (and might grab this branch and try out a few things for fun!), but I do also agree that annoyingly for our actual use cases, non-reproducibility is a critical bug, whereas support for std is a hugely valuable but not required feature.  I'm a bit concerned that there hasn't been any review on our cargo PR, but it's only been a few day, so I think it's still impolite to poke.  It seems likely that it could be OK'd at which point this issue goes away, but I guess overall I'm also of the mindset that we shouldn't land this yet.

flaub (changes requested):
Going to block this for now, but very interested to see if we can workaround #116 or wait for rust-lang/cargo#10746 to land

shkoo:
From https://github.com/rust-lang/cargo/blob/master/CONTRIBUTING.md:

> NOTICE: Due to limited review capacity, the Cargo team is not accepting new features or major changes at this time. Please consult with the team before opening a new PR. Only issues that have been explicitly marked as accepted will be reviewed.

So, it could take some additional work to get this into cargo. :(

I'll see if I can figure out another workaround.

jbruestle:
Can we also hit up the discussion forum they mention on the page and talk to the devs about what we are doing?  Even if we can hack around it, I think this is a case where it's a useful feature for them to put in upstream, and developing relations with the Rust devs is probably a good call.

flaub (approved):

PR #165, opened 2022-06-17
Increase max guest execution to 2^24 steps

preston-evans98:
## Notes
The guest previously had a soft cap of 2^20 execution steps and 2^20 words of memory. This change increases both limits to 2^24 steps, which is the hard cap for the current circuit. 

Per conversations with @jbruestle, increasing the cap will slow down proving by about 5% for all programs. 

## Changes
- Increase kMaxCyclesPo2 from 20 to 24
- Increase kMemBits from 20 to 24
- Increase kDataSize from 160 to 162
- Modify memory layout in memory.h and risc0.ld
- Replace hardcoded `20` with kMemBits in mem_check.cpp

### DIFF
--- a/risc0/zkp/core/constants.h
+++ b/risc0/zkp/core/constants.h
@@ -18,7 +18,7 @@
 
 namespace risc0 {
 
-CONSTSCALAR size_t kMaxCyclesPo2 = 20;
+CONSTSCALAR size_t kMaxCyclesPo2 = 24;
 CONSTSCALAR size_t kMaxCycles = size_t(1) << kMaxCyclesPo2;
 
 CONSTSCALAR size_t kQueries = 50; // ~100 bits of conjectured security
--- a/risc0/zkvm/circuit/constants.h
+++ b/risc0/zkvm/circuit/constants.h
@@ -59,7 +59,7 @@ enum {
 } // namespace ShaCycleType
 
 CONSTSCALAR size_t kCodeSize = 16;
-CONSTSCALAR size_t kDataSize = 160;
+CONSTSCALAR size_t kDataSize = 162;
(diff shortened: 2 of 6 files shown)


flaub:
Unfortunately this also causes creation of MethodIds to become 16x slower. To address this, I've pushed #167 which provides a way to specify the expected max code size. Because of the build perf impact, I've set the default to remain at `12` (which is what it would be prior to this PR, e.g. `kMaxCyclesPo2 = 20`). Currently #167 only supports specifying the limit on targets using bazel. It should be easy to add support for rust targets using cargo.


@preston-evans98 Would this be sufficient for your use case?

flaub:
Note that before this can land, the following memory locations need to be updated in rust code as well:
* [risc0/zkvm/sdk/rust/guest/src/lib.rs#L37](https://github.com/risc0/risc0/blob/main/risc0/zkvm/sdk/rust/guest/src/lib.rs#L37)
* [risc0/zkvm/sdk/rust/guest/src/gpio.rs#L15](https://github.com/risc0/risc0/blob/main/risc0/zkvm/sdk/rust/guest/src/gpio.rs#L15)

preston-evans98:
> Unfortunately this also causes creation of MethodIds to become 16x slower. To address this, I've pushed #167 which provides a way to specify the expected max code size. Because of the build perf impact, I've set the default to remain at `12` (which is what it would be prior to this PR, e.g. `kMaxCyclesPo2 = 20`). Currently #167 only supports specifying the limit on targets using bazel. It should be easy to add support for rust targets using cargo.
> 
> @preston-evans98 Would this be sufficient for your use case?

Yes, this is fine for our use case - please feel free to close this PR.

PR #167, opened 2022-06-19
Support configurable MethodID code size limit

### LINKED ISSUE #166: Support configurable MethodID code size limit
Currently, the `make-id` tool generates a table of hashes that have `kCodeDigestCount` entries. In #165, this count goes up by 4 which causes `make-id` to take 16x longer to run.

This feature would allow users to specify the maximum code size limit instead of always using `kCodeDigestCount`.

flaub:
Fixes #166


.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [Linux, macOS, Windows]
+        os: [Linux, macOS]
```
shkoo, line 18: Do we not care about windows anymore?

risc0/build/src/lib.rs
```diff
@@ -51,7 +51,7 @@ struct Risc0Method {
 }
 
 impl Risc0Method {
-    fn make_method_id(&self) -> MethodId {
+    fn make_method_id(&self) -> Vec<u8> {
```
shkoo, line 54: Why don't we want to keep this as a MethodId type?

risc0/zkvm/circuit/constants.h
```diff
@@ -74,4 +75,6 @@ CONSTSCALAR size_t kGlobalSize = kAccumMixGlobalOffset + kAccumMixGlobalSize;
 
 CONSTSCALAR size_t kComboCount = 5;
 
+CONSTSCALAR size_t kMaxCodeDigestCount = log2Ceil(kMaxCycles / kMinCycles) + 1;
```
shkoo, line 78: Can we document what this constant means?  Is it related to the "limit" referenced other places in this PR?

risc0/zkvm/prove/method_id.h
```diff
@@ -14,28 +14,19 @@
...
 
-#include "risc0/core/util.h"
-#include "risc0/zkp/core/sha256.h"
-#include "risc0/zkvm/circuit/constants.h"
+#include "risc0/zkvm/verify/riscv.h"
 
 namespace risc0 {
 
-static constexpr size_t kCodeDigestCount = log2Ceil(kMaxCycles / kMinCycles) + 1;
+constexpr size_t kDefaultCodeDigestCount = 12;
```
shkoo, line 25: Can we document what this does?  Would it make sense to have it in the same place as the kMaxCodeDigestCount?

risc0/zkvm/sdk/rust/host/src/lib.rs
```diff
@@ -25,6 +25,9 @@ pub use exception::Exception;
 #[cxx::bridge]
 mod bridge {}
 
+/// The default digest count when generating a MethodId.
```
shkoo, line 28: This constant seems to pop up a lot; is there a way to share it between locations?

risc0/build/src/lib.rs
flaub: I think because then I'd need to have another ffi to convert raw bytes into a `MethodId`.

I think all of this is basically temporary; once the pure rust prover lands, then we can unify a bunch of things.

risc0/zkvm/sdk/rust/host/src/lib.rs
flaub: This was just a turd, thanks for catching it.

flaub: I don't know how to share constants between C++ and Rust (unfortunately).

Once we finish the Rust port of the prover, a lot of this can be cleaned up.

.github/workflows/main.yml
flaub: The windows agent was out to lunch so I temporarily disabled it. I'll add it back to see if it completes.

Longer term, we need to get more robust windows machines.

shkoo (approved):

PR #168, opened 2022-06-19
Drop Cargo.lock files from git

flaub:
These files aren't needed anymore since the examples have been moved out of this repository. They are also the cause of frequent conflicts.


intoverflow (approved):
lgtm. Are we planning to drop all Cargo.lock from git, or just these?

flaub:
I think it still makes sense to have them for end-user binaries/apps. So I'm thinking we keep them for the example repositories.

PR #170, opened 2022-06-19
Change elf-handling APIs to take an in-memory buffer instead of a file

shkoo:


examples/cpp/battleship/protocol.cpp
```diff
@@ -30,7 +30,7 @@ RoundMessage::Content RoundMessage::decode() const {
 }
 
 InitMessage Battleship::init() {
-  risc0::MethodId methodId = risc0::makeMethodId("examples/cpp/battleship/init_method");
+  risc0::MethodId methodId = risc0::MethodId::fromElfFile("examples/cpp/battleship/init_method");
```
flaub, line 33: These actually should be a `loadMethodId` since bazel already spent the time to compute `.id` files and left them in runfiles.

See: #167

risc0/core/elf.cpp
```diff
@@ -57,15 +59,36 @@ struct ProgHeader {
...
-uint32_t loadElf(const std::string& name, uint32_t maxMem, std::map<uint32_t, uint32_t>& memOut) {
+std::vector<uint8_t> loadFile(const std::string& path) {
+  std::ifstream is(path, std::ios::binary);
+  auto check_err = [&](const char* msg) {
+    if (is.fail() || is.bad()) {
+      std::stringstream ss;
+      ss << "Could not " << msg << " ELF: " << path;
+      throw std::runtime_error(ss.str());
+    }
+  };
```
flaub, line 70: Would it be sufficient to use `is.exceptions(ios_base::failbit | ios_base::badbit)`?

https://en.cppreference.com/w/cpp/io/basic_ios/exceptions

risc0/zkvm/lib.rs
```diff
@@ -23,12 +22,16 @@ pub mod ffi {
     unsafe extern "C++" {
         include!("risc0/zkvm/prove/method_id.h");
 
-        #[cxx_name = "makeMethodId"]
-        fn make_method_id_from_elf(path: &CxxString) -> Result<[u8; 384]>;
+	// Unfortunately cxx::bridge doesn't seem to let us reference METHOD_ID_LEN here. :(
+        unsafe fn methodIdBytesFromElf(elf_contents: *const u8, len: usize) -> Result<[u8; 384]>;
```
flaub, line 26: In #167 I'm using a `Vec<u8>` (to support dynamically sized MethodIDs) and using a cross-language FFI.

In the future, once the rust prover port lands, I suspect most of this FFI stuff will go away.

risc0/zkvm/prove/riscv.h
```diff
@@ -21,6 +21,7 @@
 
 namespace risc0 {
 
-std::unique_ptr<ProveCircuit> getRiscVProveCircuit(const std::string& elfFile, MemoryHandler& io);
+std::unique_ptr<ProveCircuit>
+getRiscVProveCircuit(const uint8_t* elfBytes, size_t elfLen, MemoryHandler& io);
```
flaub, line 25: I'd slightly prefer to use a `const std::vector<uint8_t>& elfContents` here.

risc0/zkvm/sdk/cpp/host/c_api.h
```diff
@@ -60,7 +60,8 @@ void risc0_init();
 // MethodId
 //
 
-risc0_method_id* risc0_method_id_new(risc0_error* err, const char* elf_path, uint32_t limit);
+risc0_method_id*
+risc0_method_id_new(risc0_error* err, const uint8_t* bytes, size_t len, uint32_t limit);
```
flaub, line 64: There seems to be a mismatch, I see `risc0_method_id_compute` in the .cpp

shkoo: Fixed that one and a couple others, added build flag to enforce matching prototypes.

flaub (approved):
I think the big win for this would be to allow embedding the ELF itself in a `methods` crate to make it more portable.

PR #171, opened 2022-06-21
Add constructor to Receipt

intoverflow:
# Summary

The fields of the [Receipt](https://docs.rs/risc0-zkvm-verify/0.9.0/risc0_zkvm_verify/zkvm/struct.Receipt.html) structure are currently private. I do not actually know why this is the case - from reading the source, I would think that Rust would believe they are public. But it does not, and I need a way to build Receipts, so I wrote this PR.

### DIFF
--- a/risc0/zkvm/sdk/rust/verify/src/zkvm/receipt.rs
+++ b/risc0/zkvm/sdk/rust/verify/src/zkvm/receipt.rs
@@ -35,6 +35,7 @@ impl Receipt {
             seal: seal,
         }
     }
+
     pub fn verify(&self, method_id: &MethodID) -> Result<bool, VerificationError> {
         let mut circuit = RV32Circuit::new(method_id);
         let sha = risc0_zkp_core::sha::default_implementation();
--- a/risc0/zkvm/sdk/rust/verify/src/zkvm/receipt.rs
+++ b/risc0/zkvm/sdk/rust/verify/src/zkvm/receipt.rs
@@ -29,6 +29,12 @@ pub struct Receipt {
 }
 
 impl Receipt {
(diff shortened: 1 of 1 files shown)


jbruestle (approved):
Ah, I see, the issue is the current design doesn't allow one to separate the journal + seal for certain use cases such as when the journal is known for other reasons.  Approving.

risc0/zkvm/sdk/rust/verify/src/zkvm/receipt.rs
```diff
@@ -29,6 +29,13 @@ pub struct Receipt {
 }
 
 impl Receipt {
+    pub fn new(journal: Vec<u8>, seal: Vec<u32>) -> Self {
+        Receipt {
+            journal: journal,
+            seal: seal,
```
flaub, line 35:
```suggestion
            journal,
            seal,
```

flaub:
Note that this is the pure rust implementation of the verifier. This is currently only used by contracts that get compiled to WASM. I think it's probably better to stick with the C++ wrapped version until the entire pure rust prover implementation is ported.

flaub:
Odd, this seems like a problem with the tool. Can you try:

```
cargo fmt --all
```

intoverflow:
> Odd, this seems like a problem with the tool. Can you try:
> 
> ```
> cargo fmt --all
> ```

No difference :( (I had actually tried cargo fmt last night before I submitted the PR and was surprised when it made no changes).

I will investigate further

```console
tcarstens@tcarstens-dev1:~/risc0/risc0$ git status
On branch tcarstens/receipt-constructor
nothing to commit, working tree clean
tcarstens@tcarstens-dev1:~/risc0/risc0$
```

PR #174, opened 2022-06-23
Reimplement r0vm in rust

shkoo:
* New shiny command line options!  Verbose mode!  Caches method ids if requested!
* Make "non-pure" host rust implementation of Receipt serializable


risc0/zkvm/r0vm/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::Parser;
+use std::fs;
+use std::io::Write;
```
flaub, line 17: Could you combine these uses?

risc0/zkvm/r0vm/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+            }
+            let computed = MethodId::compute(&elf_contents, args.limit).unwrap();
+            if let Some(method_id_file) = args.method_id {
+                std::fs::write(&method_id_file, computed.as_slice().unwrap()).unwrap();
+                if args.verbose > 0 {
+                    eprintln!("Saved method id to {}", method_id_file);
+                }
+            }
+            computed
+        });
```
flaub, line 105: I'm not sure how I feel about computing the MethodID on the fly. This seems like a security issue?

@jbruestle

risc0/zkvm/r0vm/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[clap(long)]
+    elf: String,
+
+    /// MethodID file; created if it doesn't exist.
+    #[clap(long)]
+    method_id: Option<String>,
+
+    /// Proof output file.
+    #[clap(long)]
+    proof: Option<String>,
```
flaub, line 35: I think we should use the term 'receipt' here.

risc0/zkvm/r0vm/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    /// File to read initial input from.
+    #[clap(long)]
+    initial_input: Option<String>,
+
+    /// Display verbose output.
+    #[clap(short, long, action = clap::ArgAction::Count)]
+    verbose: u8,
+
+    /// Limit the number of hash table entries to compute.
```
flaub, line 45: To clarify, it's the number of hash entries in the MethodID table (might be confusing to people if they associate hash table a standard CS hashtable).

risc0/zkvm/sdk/rust/host/src/lib.rs
```diff
@@ -322,7 +359,21 @@ mod test {
...
+        // TODO(nils): Move this constant into something both the guest and host can
+        // depend on
+        const HEAP_START: u32 = 0x0008_0000;
+
+        let receipt: Receipt = run_memio(&[(HEAP_START, 0)]).unwrap();
+        let ser: Vec<u32> = risc0_zkvm_serde::to_vec(&receipt).unwrap();
+        let de: Receipt = risc0_zkvm_serde::from_slice(&ser).unwrap();
+        assert_eq!(de.get_journal().unwrap(), receipt.get_journal().unwrap());
+        assert_eq!(de.get_seal().unwrap(), receipt.get_seal().unwrap());
+        de.verify(IO_ID).unwrap();
```
flaub, line 376: I think you also want to assert that the verify succeeded?

risc0/zkvm/sdk/rust/host/src/lib.rs
```diff
@@ -152,6 +161,32 @@ impl Receipt {
...
+    /// Deserialize a receipt.
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct ReceiptData {
+            journal: Vec<u8>,
+            seal: Vec<u32>,
+        }
```
flaub, line 184: Maybe we could use this for both serialize and deserialize to ensure both have the same exact encoding, along with the ability to use the serde derive attributes.

risc0/zkvm/r0vm/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[clap(long)]
+    proof: Option<String>,
+
+    /// File to read initial input from.
+    #[clap(long)]
+    initial_input: Option<String>,
+
+    /// Display verbose output.
+    #[clap(short, long, action = clap::ArgAction::Count)]
+    verbose: u8,
```
flaub, line 43: If we attach stdin to the input then the guest could interact with the host and r0vm wouldn't need to know about it. Perhaps we can set the 'default' to be stdin but still allow a file to also be used as input. Of course, the host can't be interactive if it's a file.

risc0/zkvm/sdk/rust/host/src/lib.rs
shkoo: The C++ API throws an error when the verify fails and does not return a bool.  When we lift the rust verifier into zkvm/sdk/rust/host we will need to check this, though.

risc0/zkvm/r0vm/src/main.rs
shkoo: Yup, future plans!

risc0/zkvm/sdk/rust/host/src/lib.rs
```diff
@@ -17,6 +17,8 @@
 
 use std::mem;
 
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
 mod exception;
 mod ffi;
```
flaub, line 23: I think we should order these like `mod` and then `use`.

flaub (approved):
Nice!

risc0/zkvm/r0vm/src/main.rs
flaub: We can always drop this in the future if it's an issue.

PR #175, opened 2022-06-23
add cargo-sort

rlukata:
Add cargo-sort to alphabetically sort all cargo.toml files


.github/workflows/main.yml
```diff
@@ -21,6 +21,8 @@ jobs:
             config: --config=ci
 
     steps:
+      - name: install and run cargo-sort
+        run:  cargo install cargo-sort && cargo-sort -w
```
flaub, line 25: Move this down to the bottom, in the `cargo` job, after the `cargo fmt --all -- --check`

flaub:
We might need to use something like: https://github.com/marketplace/actions/cargo-install

flaub (approved):

PR #177, opened 2022-06-24
Add option to run guest without generating a seal.

shkoo:
riscv guest now runs about 200x faster!*

*: Cryptographic proof of correctness not supported, but very useful for testing.

### DIFF
--- a/Cargo-host.lock
+++ b/Cargo-host.lock
@@ -21,6 +21,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "0.7.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


jbruestle (commented):
Does this return the output/journal?  If not, that might be a good featue.  In fact, it might be reasonable to return a receipt but just with no seal.

risc0/zkvm/sdk/rust/host/src/ffi.rs
```diff
@@ -107,6 +107,8 @@ extern "C" {
     pub(crate) fn risc0_prover_run(err: *mut RawError, prover: *mut RawProver)
         -> *const RawReceipt;
 
+    pub(crate) fn risc0_prover_run_without_proof(err: *mut RawError, prover: *mut RawProver);
```
flaub, line 110: I'd say 'without_seal`

flaub (approved):

PR #185, opened 2022-06-28
Pure rust impl of prove: part 2

flaub:
Some assertions are still failing but it now at least builds and runs with just a few more HAL functions left to implement.


jbruestle (approved):
Rubber stamping to allow downstream development.

PR #193, opened 2022-07-01
Implement CC wrapper

cemozerr:


flaub:
@cemozerr Thanks for your contribution!

Can you provide a brief synopsis for this pull request and what problems you're solving with it?

Thanks!

cemozerr:
Hey @flaub! This PR is for providing a wrapper to the `cc` crate that compiles to `riscv32im` with no boilerplate and reasonable defaults.

I assumed that you wouldn't waste your time with it as its currently a draft, and we probably need to clean up further. 

I'll close this for now, work on our fork and open up a PR when it's ready.

flaub:
I think this is very useful and would love to see an end-to-end example/test of using this to build a c-based or c/rust-hybrid ELF binary that runs within R0VM.

cemozerr:
Totally. 

We locally test end-to-end with our own fork of the [rust-secp256k1](https://github.com/Sovereign-Labs/rust-secp256k1) for example, but testing with that requires making the changes regarding the memory limit.

Maybe I can find another example that does not hit the default memory limit.

PR #199, opened 2022-07-07
Use cargo as source of truth for deps

flaub:


WORKSPACE
```diff
@@ -108,50 +108,20 @@ load("@rules_rust//crate_universe:defs.bzl", "crate", "crates_repository")
 
 crates_repository(
     name = "crates_host",
+    # cargo_config = "//:.cargo/config.toml",
```
shkoo, line 111: Is this for users to use if they have a .cargo/config.toml?  Can we document?

shkoo (approved):
Sweet!  Less stuff we have to edit in both bazel and cargo. :)

PR #200, opened 2022-07-07
rust: Add support for parsing ELF files

flaub:


mitchellharper12:
This is a PR that's reminding me we need at least some place to write down itemized threats for our libraries. I don't see a need to block this PR right now, but anything doing parsing of as complex a format as ELF we will want written review and documentation of how we are using the library and potential points for things parser confusion if there's drift between our parser and the VM loader, weird header tricks that could lead to unexpected behavior, etc.

shkoo (approved):
Ok!

PR #202, opened 2022-07-11
rust: Add ability to embed methods with features

cemozerr:
There is currently no way to pass features to build the guest packages with. 

Having an `embed_methods_with_features` allows user to build guest packages with specified features.


flaub:
Need to run `bazelisk run rustfmt` or `cargo fmt --all`

cemozerr:
@flaub done. Also combined `code_limit` and `method_features` into one `MethodOptions` object to simplify `embed_methods` variants.

PR #204, opened 2022-07-14
fix: naming of MethodOptions to GuestOptions

cemozerr:
Apologies, noticed that the `MethodOptions` naming was wrong as the options object defines options per guest package not method, so here's a quick fix for that.


flaub:
@cemozerr Could you run `bazelisk run rustfmt` and push the changes?

cemozerr:
@flaub done.

flaub (approved):

PR #205, opened 2022-07-15
rust: Implement cc wrapper

cemozerr:
This PR is for providing a wrapper for the `cc` crate that compiles to `riscv32im` if the `CARGO_CFG_TARGET_ARCH` is set to `riscv32` with no boilerplate and reasonable defaults. It's useful for libraries that depend on `c/cpp`.

e.g.
```
fn main() {
    risc0_zkvm::build::rcc::Build::new()
        .file("src/foo.c")
        .compile("foo");
}
```

is equivalent to...

```
fn main() {
     cc::Build::new() // The normal `cc` crate
         .target("riscv32im-unknown-none-elf")
         .no_default_flags(true)
         .flag("-O3")
         .flag("--target=riscv32-unknown-none-elf")
         .flag("-mabi=ilp32")
         .flag("-mcmodel=medany")
         .flag("-fdata-sections")
         .flag("-ffunction-sections")
         .flag("-dead_strip")
         .flag("-flto")
         .flag("-march=rv32im")
         .file("src/foo.c")
         .flag("-static")
         .flag("--sysroot=/opt/riscv/riscv32-unknown-elf")
         .flag("--gcc-toolchain=/opt/riscv")
         .compile("foo");
}

### DIFF
--- a/risc0/zkvm/sdk/rust/methods/test-resources/simple-c-pkg/build.rs
+++ b/risc0/zkvm/sdk/rust/methods/test-resources/simple-c-pkg/build.rs
@@ -1,0 +1,16 @@
+use risc0_zkvm::build::rcc::Build;
+
+fn main() {
+    Build::new()
+        .compiler("/usr/bin/clang")
+        .include("depend/foo_dev.h")
+        .file("src/foo_dev.c")
+        .compile("foo_dev");
+
+    Build::new()
+        .compiler("/usr/bin/clang")
+        .include("depend/foo_rel.h")
+        .file("src/foo_rel.c")
+        .release(true)
+        .compile("foo_rel");
+}
--- a/risc0/zkvm/sdk/rust/methods/test-resources/simple-c-pkg/depend/foo_rel.h
+++ b/risc0/zkvm/sdk/rust/methods/test-resources/simple-c-pkg/depend/foo_rel.h
@@ -1,0 +1,6 @@
+#ifndef RISC0_FOO_REL_H
+#define RISC0_FOO_REL_H
+
+unsigned int c_add_rel(unsigned int, unsigned int);
+
+#endif // RISC0_FOO_REL_H
--- a/risc0/zkvm/sdk/rust/src/build/rcc.rs
+++ b/risc0/zkvm/sdk/rust/src/build/rcc.rs
@@ -1,0 +1,599 @@
+#![cfg(feature = "risc_cc")]
+
+use std::{env, ffi::OsStr, path::Path};
+
+use cc;
+use cc::Error;
+
+#[allow(missing_docs)]
+pub const DEFAULT_RV_CC_INSTALL_PATH: &str = "/opt/riscv";
+
+#[derive(Debug)]
+/// A convenient wrapper around `cc::Build` with sensible defaults for building
+/// to `riscv32im` targets. See `./mod.rs` for further explanation.
+pub struct Build {
+    inner: cc::Build,
+    rv_cc_install_path: String,
+    no_risc0_default_flags: bool,
+    compiler_default_flags: bool,
+    release: bool,
+}
+
+impl Build {
+    /// Construct a new instance with the default riscv32 configuration.
+    ///
+    /// This builder is finished with the [`compile`] function.
+    ///
+    /// [`compile`]: struct.Build.html#method.compile
+    pub fn new() -> Self {
+        let is_riscv_env = env::var("CARGO_CFG_TARGET_ARCH").unwrap() == "riscv32";
+
+        Self {
+            inner: cc::Build::new(),
+            rv_cc_install_path: DEFAULT_RV_CC_INSTALL_PATH.to_string(),
+            no_risc0_default_flags: !is_riscv_env,
+            compiler_default_flags: false,
+            release: false,
+        }
+    }
+
+    /// Add a directory to the `-I` or include path for headers
+    pub fn include<P: AsRef<Path>>(&mut self, dir: P) -> &mut Self {
+        self.inner.include(dir);
+        self
+    }
+
+    /// Add multiple directories to the `-I` include path.
+    pub fn includes<P>(&mut self, dirs: P) -> &mut Self
+    where
+        P: IntoIterator,
+        P::Item: AsRef<Path>,
+    {
+        self.inner.includes(dirs);
+        self
+    }
+
+    /// Specify a `-D` variable with an optional value.
+    pub fn define<'a, V: Into<Option<&'a str>>>(&mut self, var: &str, val: V) -> &mut Self {
+        self.inner.define(var, val);
+        self
+    }
+
+    /// Add an arbitrary object file to link in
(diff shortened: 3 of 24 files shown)

### CODE REVIEW

**Comment by flaub** [comment]
Whenever dependencies are changed, the bazel/cargo lock files need to be updated. To do this, you can run:
```
REPIN=1 bazelisk test //...
```

And then commit and push the changes.

**Comment by flaub** [comment]
Also we now require `Cargo.toml` dependencies to be sorted. There's a handy tool for doing this:
```
cargo install cargo-sort
cargo sort -w
```

**Comment by cemozerr** [comment]
@flaub done.

**Comment by flaub** [comment]
@cemozerr Could you run `bazelisk run rustfmt`?

**Comment by cemozerr** [comment]
@flaub done.

**Comment by cemozerr** [comment]
Not sure why that test has failed. I've re-executed all tests locally and they all seem to pass

File: risc0/zkvm/sdk/rust/tests/cc_unit_test.rs [hunk 1]
```diff
@@ -0,0 +1,217 @@
+#![cfg(feature = "risc_cc")]
+
+use crate::support::{Execution, Test};
+use risc0_zkvm::build::rcc;
+use serial_test::serial;
+
+mod support;
+
+#[test]
+#[serial]
```
flaub, line 10: Why is this necessary?

risc0/zkvm/sdk/rust/tests/support/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+#[cfg(not(target_os = "macos"))]
+fn link_or_copy<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> io::Result<()> {
+    let from = from.as_ref();
+    let to = to.as_ref();
+    fs::hard_link(from, to).or_else(|_| fs::copy(from, to).map(|_| ()))
+}
+
+/// Copy an executable.
+///
+/// On macOS, hard linking the executable leads to strange failures (issue #419), so we just copy.
```
flaub, line 137: Which issue is this?

risc0/zkvm/sdk/rust/tests/cc_end_to_end_test.rs
```diff
@@ -0,0 +1,38 @@
+#![cfg(feature = "risc_cc")]
```
flaub, line 1: I think this is preventing CI from testing all this new code. Is there any harm in dropping the feature and just always including it?

risc0/zkvm/sdk/rust/methods/test-resources/simple-c-pkg/depend/foo_rel.h
```diff
@@ -0,0 +1,6 @@
+#ifndef RISC0_FOO_REL_H
```
flaub, line 1: This file looks identical to the dev version. Are we trying to test that debug vs release builds work? Wouldn't we want to ensure this feature works on the same code?

risc0/zkvm/sdk/rust/methods/test-resources/simple-c-pkg/build.rs
```diff
@@ -0,0 +1,16 @@
+use risc0_zkvm::build::rcc::Build;
+
+fn main() {
+    Build::new()
+        .compiler("/usr/bin/clang")
```
flaub, line 5: I'm surprised this absolute path works. Does it work on Windows?

risc0/zkvm/sdk/rust/tests/cc_unit_test.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    check_risc0_default_flags_do_not_exist(&e);
+    check_release_flags_do_not_exist(&e);
+}
+
+#[test]
+#[serial]
+fn no_risc0_default_flags_via_flag() {
+    reset_env();
+    std::env::set_var("CRATE_NO_RISC0_DEFAULTS", "TRUE");
```
flaub, line 47: Instead of setting env vars globally on the current process, I think we can set the env vars on a per child process basis. So couldn't we say:
```
test.gcc().env("CRATE_NO_RISC0_DEFAULTS", "TRUE").file("foo.c").compile("foo");
```

flaub: I think this approach might drop the need for `#[serial]`

risc0/zkvm/sdk/rust/src/build/rcc.rs
```diff
@@ -0,0 +1,614 @@
...
+
+    #[doc(hidden)]
+    pub fn __set_env<A, B>(&mut self, a: A, b: B) -> &mut Self
+    where
+        A: AsRef<OsStr>,
+        B: AsRef<OsStr>,
+    {
+        self.inner.__set_env(a, b);
+        self
+    }
```
flaub, line 429: I think this is a good public feature to expose (and needed for being able to drop the `#[serial]`.

risc0/zkvm/sdk/rust/src/build/rcc.rs
```diff
@@ -0,0 +1,614 @@
...
+            compiler_default_flags: false,
+            release: false,
+        }
+    }
+
+    /// Add a directory to the `-I` or include path for headers
+    pub fn include<P: AsRef<Path>>(&mut self, dir: P) -> &mut Self {
+        self.inner.include(dir);
+        self
+    }
```
flaub, line 44: Instead of passing thru everything (and thus adding lots of boilerplate to this crate, and the need to document it all and maintain it), I think we could just expose the `inner: cc::Build` directly.

risc0/zkvm/sdk/rust/src/build/rcc.rs
```diff
@@ -0,0 +1,614 @@
+#![cfg(feature = "risc_cc")]
+
+use std::{env, ffi::OsStr, path::Path};
+
+use cc;
+use cc::Error;
+
+#[allow(missing_docs)]
```
flaub, line 8: We need to have justification for missing docs

risc0/zkvm/sdk/rust/tests/cc_end_to_end_test.rs
cemozerr: Its not ideal as the tests will be flaky for people who who don't have the toolchain installed, but its not a huge deal.

risc0/zkvm/sdk/rust/src/build/rcc.rs
cemozerr: I'm a bit unclear how we could do that without extending `inner::cc::Build` (as Rust doesn't seem to support extending structs)

Any suggestions?

I guess, if we're okay with giving up the environment variables specific for this object such as `RCC_COMPILER` and `RV_CC_INSTALL_PATH` we could have a constructor that applies the default risc0 flags and returns the inner Build directly

flaub:
This appears to be stale, closing for now, but feel free to post a new PR on top of the latest head.

PR #209, opened 2022-07-17
Fix older C++ compat

flaub:


flaub:
This apparently fixes an issue that was brought up in discord.

rlukata (approved):
LGTM

jbruestle (approved):
So I think that __LINE__ get's evaluated at the use point, i.e. all tracebacks would show everything originating in source_loc.h (see https://community.intel.com/t5/Intel-C-Compiler/builtin-LINE-builtin-FUNCTION-builtin-FILE-GCC-builtins/td-p/1030060).  However, at least things compile, and I think if there is no __builtin_FILE there is no way to do source locations anyway, so it's a moot point.  All this is a long way of saying 'approved'.

flaub:
Yeah, I figured it was better to get things to compile then worry whether an error message had the right line of code. Also considering that longevity of this code.

PR #210, opened 2022-07-20
Extract risc0-build and risc0-zkvm-guest from risc0-zkvm

flaub:
Trying to integrate the monolithic `risc0-zkvm` crate in downstream projects revealed some build issues. Extracting the `risc0-build` and `risc0-zkvm-guest` crates back out of `risc0-zkvm` seems to resolve those issues.

Another note is that vscode/rust-analyzer wasn't able to provide analysis for the guest module of the monolithic crate.


intoverflow (approved):
LGMT!

PR #211, opened 2022-07-21
Fix warnings in documentation compilation

tzerrell:


tzerrell:
Following GitHub's suggestions on reviewers, sorry if you're not the right people for this.

I'm having a little trouble interpreting the build failure -- it looks like maybe a rustfmt issue, but if I run `cargo fmt` locally everything seems fine...

flaub:
Use `bazelisk run rustfmt`, which will also format comments.

tzerrell:
Poke ... does this look good?

flaub (approved):

PR #214, opened 2022-07-22
Write intro rust docs for zkVM, guest/build

tzerrell:
Add introductions for the risc0-zkvm, risc0-zkvm-guest, risc0-build, and risc0-make-id crates. The risc0-build introduction is the most complex of these, as that's where I explain how to correctly embed/build/link guest methods and access them from the host.

### DIFF
--- a/risc0/zkvm/prove/make-id/src/main.rs
+++ b/risc0/zkvm/prove/make-id/src/main.rs
@@ -12,12 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Creates RISC Zero MethodIDs from ELF files
+//!
+//! A simple command line tool for creating
+//! [MethodID](risc0_zkvm::host::MethodId)s from RISC-V ELF files. It is
+//! intended for basic, low-level use cases. In particular, users looking to
+//! create MethodIDs for Rust code are advised to instead use our zkVM: visit
+//! our [Risc Zero Rust Starter repository](https://github.com/risc0/risc0-rust-starter)
+//! for help getting started.
+
 use clap::Parser;
 use std::fs;
 
 use risc0_zkvm::host::{MethodId, DEFAULT_METHOD_ID_LIMIT};
 
-/// Generates a MethodID for a given RISC-V ELF binary.
+/// The command line arguments necessary to run [main].
 #[derive(Parser)]
 #[clap(about, version, author)]
 struct Args {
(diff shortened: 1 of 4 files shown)


risc0/zkvm/prove/make-id/src/main.rs
```diff
@@ -12,12 +12,21 @@
...
+//! our [Risc Zero Rust Starter repository](https://github.com/risc0/risc0-rust-starter)
+//! for help getting started.
+
 use clap::Parser;
 use std::fs;
 
 use risc0_zkvm::host::{MethodId, DEFAULT_METHOD_ID_LIMIT};
 
-/// Generates a MethodID for a given RISC-V ELF binary.
+/// The command line arguments necessary to run [main].
```
flaub, line 29: Note that this particular doc string gets used by `clap` to generate a nice `--help` message. So in this case, I think we should keep the previous text.

risc0/zkvm/sdk/rust/build/README.md
```diff
@@ -1,6 +1,50 @@
 # risc0-build
 
-This crate contains helper functions to assist with building guest side code,
-and using it in host side code.
+Build RISC Zero zkVM guest code and provide handles to the host side
 
-[TODO: Examples]
+In order for the host to execute guest code in the [RISC Zero zkVM](risc0_zkvm), the host must be provided an ELF file with RISC-V source code and the corresponding [MethodID](risc0_zkvm::host::MethodId). This crate contains the functions needed to take zkVM guest code written in rust, build a corresponding ELF file and MethodID, and make the MethodID and a path to the ELF file available for the host to use.
```
flaub, line 5: The ELF doesn't contain source code, but was compiled from it.

It's also not always the case that all the source code will be based on rust. It's possible to mix with other languages.

risc0/zkvm/prove/make-id/src/main.rs
```diff
@@ -12,12 +12,21 @@
...
 // limitations under the License.
 
+//! Creates RISC Zero MethodIDs from ELF files
+//!
+//! A simple command line tool for creating
+//! [MethodID](risc0_zkvm::host::MethodId)s from RISC-V ELF files. It is
+//! intended for basic, low-level use cases. In particular, users looking to
+//! create MethodIDs for Rust code are advised to instead use our zkVM: visit
+//! our [Risc Zero Rust Starter repository](https://github.com/risc0/risc0-rust-starter)
+//! for help getting started.
```
flaub, line 22: I don't think that binary targets will ever show up in `docs.rs` or rustdocs in general. But it's fine to have comments here describing the binary, perhaps just as normal non-rustdoc comments.

tzerrell: The only reason I even know about this file is it showed up in the crates list on the left when I built locally. So it shows up in at least some circumstances! (Perhaps not on `docs.rs` though)

flaub: TIL :)

flaub (approved):
LGTM!

PR #216, opened 2022-07-26
Add support for println! and eprintln! to rust guest

shkoo:
* Upgrade to new standard library on the rust/from-upstream-nightly-2022-06-13 branch


flaub (approved):
LGTM, but looks like cargo builds are failing

PR #218, opened 2022-08-01
Rust prover/cirgen integration

flaub:


jbruestle (approved):
Checked the c++ stuff, rubber stamping the rust since it’s not really production yet.

flaub:
@shkoo Feel free to chime in, but none of these changes should yet have an effect on anything "in production". The big TODO is to get the I/O (SENDRECV) stuff hooked up. I was thinking I could hand that off to you or I could spend another day or so getting that done.

shkoo (commented):
It looks like step_exec.rs didn't get successfully formatted, but rustfmt doesn't seem to be able to work with it, so... :)

I don't think that the "debug = false" profile setting is working quite right.  I tried plain old "cargo test" in the workspace root and it gave me   process didn't exit successfully: `rustc --crate-name risc0_zkvm_circuit --edition=2021 risc0/zkvm/sdk/rust/circuit/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 -C metadata=3be7fd48f3bbf99e -C extra-filename=-3be7fd48f3bbf99e --out-dir /home/nils/flaub/risc0/target/debug/deps -C incremental=/home/nils/flaub/risc0/target/debug/incremental -L dependency=/home/nils/flaub/risc0/target/debug/deps --extern log=/home/nils/flaub/risc0/target/debug/deps/liblog-2aea261907451eb5.rmeta --extern risc0_zkp=/home/nils/flaub/risc0/target/debug/deps/librisc0_zkp-30f1755382cfe915.rmeta` (signal: 11, SIGSEGV: invalid memory reference)

flaub:
`step_exec.rs` seems to be too large for most tools to deal with. I'm considering generating C code to replace this for the following reasons:
1. likely better performance (but I want to benchmark this first).
2. rust tools (rust-analyzer, rustfmt, etc) don't like large files like this
3. avoid `rustc` segfaults

shkoo (approved):
Looks good!  Yeah, I can work on getting the sendrecv stuff hooked up.

PR #221, opened 2022-08-02
Make the two rust prover APIs (ffi-based and pure rust) more similar …

shkoo:
…than they were before.

zkvm now builds with --features=pure-prove.

(However, it does not pass tests yet)

### DIFF
--- a/risc0/zkvm/sdk/rust/src/host/mod.rs
+++ b/risc0/zkvm/sdk/rust/src/host/mod.rs
@@ -15,12 +15,14 @@
 #![deny(missing_docs)]
 #![doc = include_str!("README.md")]
 
-mod exception;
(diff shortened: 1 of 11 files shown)


risc0/zkvm/sdk/rust/src/host/mod.rs
```diff
@@ -15,12 +15,14 @@
 #![deny(missing_docs)]
 #![doc = include_str!("README.md")]
 
-mod exception;
+#[cfg(not(feature = "pure-prove"))]
```
flaub, line 18: I think the way I was envisioning dealing with this was using the newly added `host` feature to prevent this entire `host` module from being compiled if you wanted the pure-rust version. But I suppose this way works in the interim just as well or perhaps better.

flaub (approved):

PR #223, opened 2022-08-03
Rust prover runs to completion

flaub:


shkoo:
This doesn't pass for me with "cargo test":

---- integration::memory_io stdout ----
thread 'integration::memory_io' panicked at 'attempt to shift left with overflow', risc0/zkvm/sdk/rust/src/prove/exec.rs:116:63

shkoo (approved):

PR #230, opened 2022-08-11
Add METHOD_ELF static const

Vesnica:
Currently, prover have a hard link to guest's elf files, which is inconvenient for deployment. This PR makes it possible that prover can embedded the guest's elf content and running alone.


flaub:
@shkoo WDYT?

I'd actually like to have just contents and not any path.

Vesnica:
Usage example:
https://github.com/Vesnica/risc0-rust-starter/blob/main/cli/src/prover_mul.rs

shkoo:
This sounds like a good idea to me!  Historically the prover could only read an elf file from a path, not from an in-memory buffer, so that's why we have _PATH.

If you feel like being proactive, it would be super neat to get rid of the _PATH constants entirely! But otherwise we at least need a test that uses the _CONTENTS constant.

Thanks!

Vesnica:
I merged upstream accidentally. So I create another PR(https://github.com/risc0/risc0/pull/244), sorry for the inconvenient.

PR #232, opened 2022-08-11
Bump rust-toolchain to nightly-2022-06-20

flaub:
This is necessary to deal with `camino = 1.0.10` that just landed, which relies on some features that appear in rust `1.63`, but weren't added until at least `2022-06-15`


tzerrell (approved):

flaub:
@tzerrell Note that we'll need to make this similar change to all the repositories that depend on risc0.

PR #236, opened 2022-08-15
Fix kGPIO_GetKey constant at zkvm/platform/io.h

denisadapt:
The constant looks suspicious and its value doesn't match to correspondent constant in io.rs


flaub:
@jbruestle Is it OK if we drop this GPIO entirely? I don't think any of the new examples use this feature anymore.

flaub:
The new zkvm implementation uses ecalls rather than GPIO. The GetKey functionality has also been dropped.

PR #238, opened 2022-08-16
Progress towards making zkp field-agnostic

shkoo:
* Define "field::Elem" and "field::ExtElem" types
* Add test_roots_of_unity and test_field_ops test utils to run basic tests on fields (It sounds like kalen and @tzerell will be working more on tests and documentation)
* Rename baby bear field (15*2^27) from {Fp, Fp4} to field::baby_bear::{Elem, ExtElem}; Fp and Fp4 are now aliases
* Change a bunch of Fp and Fp4 calls to use trait items.
* Remove Fp::invalid, so we don't have to keep this complication for all our fields. (and we weren't using it in rust anyways)


tzerrell (approved):
Nice, looks good to me!

@3lkn this is the field genericization API/infrastructure I mentioned.

PR #240, opened 2022-08-17
Clean up Sha implementations in Zkp

shkoo:
* No more Sha trait to pass around everywhere; when we need to choose an accelerator we can use the more general "Hal".  zkp's verify does not use hal.
* sha implementations no longer care about "Fp" or "Fp4" types.
* sha implementations now return "Cow<'static, Digest>' which is a lot more efficient to copy in the guest.

Miscellaneous fixes:

* Trim zkvm features in risc0-build to speed compilation time
* Add "rustfmt.toml" in root of tree so that "cargo fmt" can find it.

### DIFF
--- a/Cargo-guest.lock
+++ b/Cargo-guest.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "anyhow"
-version = "1.0.61"
+version = "1.0.62"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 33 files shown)


flaub:
I'm not sure what's wrong with using a trait for Sha, because we have an issue where zkp verify wants to be able to use something defined in the guest, so there's some natural indirection.

shkoo:
Sounds like we're doing a different approach

PR #250, opened 2022-08-22
Goldilocks implementation in Rust

3lkn:
This is the Rust implementation of Goldilocks; for analogous changes in C++, see [@jbruestle's working branch](https://github.com/risc0/risc0/blob/jeremy-goldilocks/risc0/zkp/core/fp.h). 

**Special review request: Comments overlap heavily with baby_bear.rs, looking for comment review from a documentation perspective so we don't have competing explanation variants.**

Notes:
* Removed options to create field elements from u32 values; may restore this for test purposes somewhere down the (generic) line.

Testing notes:
* Added granular tests for arithmetic operations. 
* In the tests, to generate expected values for field operations, most operations are performed in 128-bit space and then reduced mod P. 
* Because we don't have the 'baby bear' promise not to overflow addition, subtraction is a little case-dependent, and I'd welcome a simplification from any low-level math wizards.

### DIFF
--- a/risc0/zkp/rust/src/field/goldilocks.rs
+++ b/risc0/zkp/rust/src/field/goldilocks.rs
@@ -1,0 +1,735 @@
+// Copyright 2022 Risc0, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// ! Goldilocks field.
+/// ! Support for the base finite field modulo 2^64 - 2^32 + 1
+use crate::field::{self, Elem as FieldElem};
+
+use core::ops;
+
+use bytemuck::{Pod, Zeroable};
+
+/// The Goldilocks class is an element of the finite field F_p, where P is the
+/// prime number 2^64 - 2^32 + 1. Here we implement integer
+/// arithmetic modulo P for both Goldilocks and for a field extension of Goldilocks.
+///
+/// The `Fp` datatype is the core type of all of the operations done within the
+/// zero knowledge proofs, and is the smallest 'addressable' datatype, and the
+/// base type of which all composite types are built. In many ways, one can
+/// imagine it as the word size of a strange architecture,
+/// and its operations as wrapping operations which respect word size P.
+///
+/// The Fp class wraps all standard arithmetic operations to make finite
+/// field elements appear like ordinary numbers (which, for the most part, they are).
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Pod, Zeroable)]
+#[repr(transparent)]
+pub struct Elem(u64);
+
+impl Default for Elem {
+    /// As a default, return the zero [Elem].
+    fn default() -> Self {
+        Self::ZERO
+    }
+}
+
+/// The modulus of our Goldilocks field: 2^64 - 2^32 + 1
+const P: u64 = (0u64.wrapping_sub(1) << 32) + 1;
+// TODO: Vestigial, should be improved on generic implementation
+const P_U64: u64 = P;
+
+impl field::Elem for Elem {
+    const ZERO: Self = Elem::new(0u64);
+    const ONE: Self = Elem::new(1u64);
+
+    /// Compute the multiplicative inverse of `x`, or `1 / x` in finite field
+    /// terms. Since we know by Fermat's Little Theorem that
+    /// `x ^ (P - 1) == 1 % P` for any `x != 0`,
+    /// it follows that `x * x ^ (P - 2) == 1 % P` for `x != 0`.
+    /// That is, `x ^ (P - 2)` is the multiplicative inverse of `x`.
+    /// Note that if computed this way, the *inverse* of zero comes out as zero,
+    /// which we allow because it is convenient in many cases.
+    fn inv(self) -> Self {
+        self.pow((P - 2) as usize)
+    }
+
+    /// Generate a random value within our Goldilocks field
+    fn random(rng: &mut impl rand::Rng) -> Self {
+        // The range of possible RNG-generated u64 integers includes an uneven region modulo P.
+        // We want to reject u64 values from this region because, if mapped to finite field elements
+        // (wrapped), it leads to over-selection of the wrapped values. Here, P happens to
+        // fit only once into a 64-bit space, so we accept only RNG-generated u64 values less than P.
+        let mut val: u64 = rng.gen();
+        while val >= P {
+            val = rng.gen();
+        }
+        Elem::from(val)
+    }
+}
+
+macro_rules! rou_array {
+    [$($x:literal),* $(,)?] => {
+        [$(Elem::new($x)),* ]
+    }
+}
+
+/// Maximum power of two for which we have a root of unity using Goldilocks field
+impl field::RootsOfUnity for Elem {
+    const MAX_ROU_PO2: usize = 32;
+
+    const ROU_FWD: &'static [Elem] = &rou_array![
+        1,
+        18446744069414584320,
+        281474976710656,
+        18446744069397807105,
+        17293822564807737345,
+        70368744161280,
+        549755813888,
+        17870292113338400769,
+        13797081185216407910,
+        1803076106186727246,
+        11353340290879379826,
+        455906449640507599,
+        17492915097719143606,
+        1532612707718625687,
+        16207902636198568418,
+        17776499369601055404,
+        6115771955107415310,
+        12380578893860276750,
+        9306717745644682924,
+        18146160046829613826,
+        3511170319078647661,
+        17654865857378133588,
+        5416168637041100469,
+        16905767614792059275,
+        9713644485405565297,
+        5456943929260765144,
+        17096174751763063430,
+        1213594585890690845,
+        6414415596519834757,
+        16116352524544190054,
+        9123114210336311365,
+        4614640910117430873,
+        1753635133440165772,
+    ];
+
+    const ROU_REV: &'static [Elem] = &rou_array![
+        1,
+        18446744069414584320,
+        18446462594437873665,
+        1099511627520,
+        68719476736,
+        18446744069414322177,
+        18302628881338728449,
+        18442240469787213841,
+        2117504431143841456,
+        4459017075746761332,
+        4295002282146690441,
+        8548973421900915981,
+        11164456749895610016,
+        3968367389790187850,
+        4654242210262998966,
+        1553425662128427817,
+        7868944258580147481,
+        14744321562856667967,
+        2513567076326282710,
+        5089696809409609209,
+        17260140776825220475,
+        11898519751787946856,
+        15307271466853436433,
+        5456584715443070302,
+        1219213613525454263,
+        13843946492009319323,
+        16884827967813875098,
+        10516896061424301529,
+        4514835231089717636,
+        16488041148801377373,
+        16303955383020744715,
+        10790884855407511297,
+        8554224884056360729,
+    ];
+}
+
+impl Elem {
+    /// Create a new [Goldilocks] from a raw integer.
+    pub const fn new(x: u64) -> Self {
+        Self(x % P)
+    }
+}
+
+impl ops::Add for Elem {
+    type Output = Self;
+    /// Addition for Goldilocks field elements [Elem]
+    fn add(self, rhs: Self) -> Self {
+        Elem(add(self.0, rhs.0))
+    }
+}
+
+impl ops::AddAssign for Elem {
+    /// Simple addition case for Goldilocks field elements [Elem]
+    fn add_assign(&mut self, rhs: Self) {
+        self.0 = add(self.0, rhs.0)
+    }
+}
+
+impl ops::Sub for Elem {
+    type Output = Self;
+    /// Subtraction for Goldilocks field elements [Elem]
+    fn sub(self, rhs: Self) -> Self {
+        Elem(sub(self.0, rhs.0))
+    }
+}
+
+impl ops::SubAssign for Elem {
+    /// Simple subtraction case for Goldilocks [Elem]
+    fn sub_assign(&mut self, rhs: Self) {
+        self.0 = sub(self.0, rhs.0)
+    }
+}
+
+impl ops::Mul for Elem {
+    type Output = Self;
+    /// Multiplication for Goldilocks [Elem]
+    fn mul(self, rhs: Self) -> Self {
+        Elem(mul(self.0, rhs.0))
+    }
+}
+
+impl ops::MulAssign for Elem {
+    /// Simple multiplication case for Goldilocks [Elem]
+    fn mul_assign(&mut self, rhs: Self) {
+        self.0 = mul(self.0, rhs.0)
+    }
+}
+
+impl ops::Neg for Elem {
+    type Output = Self;
+    /// Negation for Goldilocks [Elem]
+    fn neg(self) -> Self {
+        Elem(0) - self
+    }
+}
+
+impl From<&Elem> for u64 {
+    fn from(x: &Elem) -> Self {
+        x.0
+    }
+}
+
+impl From<Elem> for u64 {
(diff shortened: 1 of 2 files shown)


risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// ! Goldilocks field.
+/// ! Support for the base finite field modulo 2^64 - 2^32 + 1
+use crate::field::{self, Elem as FieldElem};
+
+use core::ops;
+
+use bytemuck::{Pod, Zeroable};
```
flaub, line 21: The sorting of groups should go:

* std/core/alloc (stuff from rust)
* third party (i.e. bytemuck)
* crate::*

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+}
+
+macro_rules! rou_array {
+    [$($x:literal),* $(,)?] => {
+        [$(Elem::new($x)),* ]
+    }
+}
+
+/// Maximum power of two for which we have a root of unity using Goldilocks
+/// field
```
tzerrell, line 90: This comment applies specifically to `MAX_ROU_PO2`. IIUC, `ROU_FWD` and `ROU_REV` are powers of our canonical 32nd RoU.

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+        }
+        Elem::from(val)
+    }
+}
+
+macro_rules! rou_array {
+    [$($x:literal),* $(,)?] => {
+        [$(Elem::new($x)),* ]
+    }
+}
```
flaub, line 87: What's the purpose of this? Why can't we just create these directly?

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+    // Get two high words
+    let med: u32 = (prod >> 64) as u32;
+    let high: u32 = (prod >> 96) as u32;
+    // Subtract out high bits, add in P if underflow
+    if ret >= (high as u64) {
+        ret = ret.wrapping_sub(high as u64);
+    } else {
+        ret = ret.wrapping_sub(high as u64);
+        ret = ret.wrapping_add(P);
+    }
```
flaub, line 278:
```suggestion
    let ret = 
    if ret >= (high as u64) {
        ret.wrapping_sub(high as u64)
    } else {
        ret.wrapping_sub(high as u64).wrapping_add(P)
    }
```

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+    return if x > lhs { x.wrapping_add(P) } else { x };
+}
+
+/// Wrapping multiplication of [Elem] using Goldilocks prime
+fn mul(lhs: u64, rhs: u64) -> u64 {
+    // To prevent u64 overflow, we first perform with u128
+    let prod: u128 = (lhs as u128).wrapping_mul(rhs as u128);
+
+    // Initialize result to low 64 bits (by converting)
+    let mut ret: u64 = prod as u64;
```
flaub, line 268: In rust, you can 'shadow' variables without ambiguity. So instead of using a `mut`, you could just create a new lvalue for each.

For example, you could say:
```
let mut x = 1;
x = x + 1;
```

Or:
```
let x = 1;
let x = x + 1;
```

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+        ret = ret.wrapping_add(P);
+    }
+    // Compute shifted effect of medium
+    let med_shift = ((med as u64) << 32).wrapping_sub(med as u64);
+    // Add in, if overflow, subtract a P
+    ret = ret.wrapping_add(med_shift);
+    if ret < med_shift || ret >= P {
+        ret = ret.wrapping_sub(P)
+    }
+    ret as u64
```
flaub, line 286: It's not clear why we need to cast at the end. `ret` should be a `u64` at this point.

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+const EXT_SIZE: usize = 2;
+
+/// Instances of `ExtElem` are elements of a finite field `F_p^2`. They are
+/// represented as elements of `F_p[X] / (X^2 - 11)`. This large
+/// finite field (about `2^128` elements) is used when the security of
+/// operations depends on the size of the field. The field extension `ExtElem`
+/// has `Elem` as a subfield, so operations on elements of each are compatible.
+/// The irreducible polynomial `x^2 - 11` was chosen because `11` is
+/// the simplest choice of `B` for `x^2 - B` that makes this polynomial
+/// irreducible.
```
tzerrell, line 299: The adjoined root squared (i.e. 11) is referred to as `BETA` elsewhere in the code, so let's do that in this comment as well.

```suggestion
/// Instances of `ExtElem` are elements of a finite field `F_p^2`. They are
/// represented as elements of `F_p[X] / (X^2 - 11)`. This large
/// finite field (about `2^128` elements) is used when the security of
/// operations depends on the size of the field. The field extension `ExtElem`
/// has `Elem` as a subfield, so operations on elements of each are compatible.
/// The irreducible polynomial `x^2 - 11` was chosen because `11` is
/// the simplest choice of `BETA` for `x^2 - BETA` that makes this polynomial
/// irreducible.
```

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+        let a: u64 = fa.into();
+        let b: u64 = fb.into();
+        assert_eq!(fa + fb, Elem::from(a.wrapping_add(b)));
+    }
+
+    #[test]
+    fn compare_native_subtraction_easy() {
+        // 0 < a - b < P  (when a > b)
+        let fa: Elem = Elem::from(P - 2u64);
+        let fb: Elem = Elem::from(3u64);
```
flaub, line 626:
```suggestion
        let fa = Elem::from(P - 2u64);
        let fb = Elem::from(3u64);
```

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+                fa + fb,
+                Elem::from(((a as u128 + b as u128) % P as u128) as u64),
+                "Error adding `{} + {}`",
+                a,
+                b
+            );
+            // This is a workaround that doesn't need to exist for baby bear
+            // because it doesn't overflow u64 under addition. Here, we could have P - b + a
+            // either overflow or wrap under u64.
+            let diff: u64 = if a < b {
```
flaub, line 724:
```suggestion
            let diff = if a < b {
```

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+        for i in 0..self.0.len() {
+            self.0[i] += rhs.0[i];
+        }
+    }
+}
+
+impl ops::Sub for ExtElem {
+    type Output = Self;
+    /// Subtraction for [ExtElem]
+    fn sub(self, rhs: Self) -> Self {
```
flaub, line 429:
```suggestion
    type Output = Self;
    
    /// Subtraction for [ExtElem]
    fn sub(self, rhs: Self) -> Self {
```

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+    const fn zero() -> Self {
+        Self::from_u64(0)
+    }
+
+    /// Returns the value one.
+    const fn one() -> Self {
+        Self::from_u64(1)
+    }
+
+    /// Returns the constant portion of a [Elem].
```
tzerrell, line 396: IMO "constant portion" is unclear. The algorithm here is that given an element of the form `a + b * sqrt(BETA)`, return `a`. Maybe say "base field portion"?... open to other suggestions...

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+    pub fn const_part(self) -> Elem {
+        self.0[0]
+    }
+
+    /// Returns the elements of an [Elem].
```
tzerrell, line 401: Similar wording confusion IMO. Maybe "Returns [Elem] as a vector of base field values"?... again open to suggestions.

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+        // Test for addition when P < a + b < 2^64
+        let fa = Elem::from(P - 2u64);
+        let fb = Elem::from(5u64);
+        let a: u64 = fa.into();
+        let b: u64 = fb.into();
+        assert_eq!(fa + fb, Elem::from(a.wrapping_add(b)));
+    }
+
+    #[test]
+    fn compare_native_add_u64_wrap() {
```
tzerrell, line 613: Am I missing something or is this the same test as `compare_native_add_subfield_wrap`?

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+            "Error multiplying `{} * {}`",
+            a,
+            b
+        );
+    }
+
+    #[test]
+    fn compare_native_multiplication_subfield_wrap() {
+        // P < a * b < 2^64
+        // We expect the result mod P, but there's no u64 overflow
```
tzerrell, line 691: I think this does overflow u64, as (3/2) P > 2^64 - 1. Which is a reasonable test to have, it's just not what the comment says.

tzerrell (commented):
Great! Looks good overall, I have some inline comments.

Several of my comments are from a documentation perspective, but I think the documentation needs a more thorough overhaul than just what's in my comments -- in particular aligning with & producing unified docs for the baby bear & the field element trait interface generally (i.e. the docs concerns you flagged in your initial comment on this PR). There is duplicated documentation and documentation that's worded as if the multiple fields and the trait interface don't exist as separate entities.

That said, I think that documentation cleanup should be a separate PR. We need it, we should do it promptly while this is fresh in your (and everyone's) mind, but IMO this PR doesn't need to depend on it.

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+
+impl Default for Elem {
+    /// As a default, return the zero [Elem].
+    fn default() -> Self {
+        Self::ZERO
+    }
+}
+
+/// The modulus of our Goldilocks field: 2^64 - 2^32 + 1
+const P: u64 = (0u64.wrapping_sub(1) << 32) + 1;
```
jbruestle, line 49: The way this is computed maybe warrants a mild explaination?  Basically, the obvious thing of computing (1 << 64) - (1 << 32) + 1 would overflow U64, so we need to be trixy.

jbruestle (commented):
Nice work, all the math seems right.  The 'compare_native' tests maybe should be renamed as they are more like 'check various edge cases by hand tests', or maybe like test vectors, but not sure what's a better name.

risc0/zkp/rust/src/field/goldilocks.rs
3lkn: This one I'm not sure on, it's also in `baby_bear.rs` and I did not want to meddle in the affairs of macro wizards. @shkoo ?

3lkn: I am also not sure, it is in lines 97-100 of `baby_bear.rs` and I did not want to meddle in the affairs of macro wizards, @shkoo do we want this left in?

risc0/zkp/rust/src/field/goldilocks.rs
3lkn: Minor note, the `let` in `let ret` on 278 causes new variable to be instantiated, switched it to `ret =` and kept the rest of this suggestion

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,745 @@
...
+macro_rules! rou_array {
+    [$($x:literal),* $(,)?] => {
+        [$(Elem::new($x)),* ]
+    }
+}
+
+/// Maximum power of two for which we have a root of unity using Goldilocks
+/// field
+impl field::RootsOfUnity for Elem {
+    const MAX_ROU_PO2: usize = 32;
```
tzerrell, line 97: To follow up on my earlier comment -- IMO pull this comment to just before the definition of `MAX_ROU_PO2`.
```suggestion
impl field::RootsOfUnity for Elem {
    /// Maximum power of two for which we have a root of unity using Goldilocks
    /// field
    const MAX_ROU_PO2: usize = 32;
```

tzerrell (commented):
Looks good! IMO this just needs a resolution of the macro question and a few doc wording details.

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,755 @@
...
+    }
+}
+
+/// The modulus of our Goldilocks field: 2^64 - 2^32 + 1
+/// Calculation steps chosen to avoid overflowing u64 with 2^64:
+/// 1. Wrapping subtract from 0u64, which leaves all ones
+/// 2. Left-shift ones over by 32, leaving 32 ones and 32 zeros (2^64-2^32)
+/// 3. Add one to get 2^64 - 2^32 + 1
+
+const P: u64 = (0u64.wrapping_sub(1) << 32) + 1;
```
flaub, line 54:
```suggestion
const P: u64 = (0xffffffff_ffffffff << 32) + 1;
```

risc0/zkp/rust/src/field/baby_bear.rs
```diff
@@ -12,15 +12,15 @@
...
-/// ! Support for the base finite field modulo 15*2^27 + 1
-use crate::field::{self, Elem as FieldElem};
-
 use alloc::fmt;
 use core::ops;
 
 use bytemuck::{Pod, Zeroable};
 
+/// ! Baby bear field.
+/// ! Support for the base finite field modulo 15*2^27 + 1
```
flaub, line 21: These module comments should probably appear before all the uses.

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,755 @@
...
+    // Compute shifted effect of medium
+    let med_shift = ((med as u64) << 32).wrapping_sub(med as u64);
+    // Add in, if overflow, subtract a P
+    let ret = ret.wrapping_add(med_shift);
+    let ret = if ret < med_shift || ret >= P {
+        ret.wrapping_sub(P)
+    } else {
+        ret
+    };
+    ret
```
flaub, line 292:
```suggestion
    if ret < med_shift || ret >= P {
        ret.wrapping_sub(P)
    } else {
        ret
    }
```

risc0/zkp/rust/src/field/goldilocks.rs
shkoo: This was the best way I could figure out to create a [Elem] from an array of uints in a const context.  I'd love a better way!

risc0/zkp/rust/src/field/goldilocks.rs
```diff
@@ -0,0 +1,756 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use core::ops;
+
+use bytemuck::{Pod, Zeroable};
+
+/// ! Goldilocks field.
+/// ! Support for the base finite field modulo 2^64 - 2^32 + 1
```
flaub, line 20: Module comments (they should be in the form of `//! Some text`), should go just below the copyright notice.

risc0/zkp/rust/src/field/baby_bear.rs
flaub: Also the format is: `//!`

flaub:
LGTM

3lkn:
> Looks good! IMO this just needs a resolution of the macro question and a few doc wording details.

@tzerrell To my eyes the macro question is for-now-resolved, with this latest round of wording overhaul does this look ready to merge?

tzerrell (approved):

tzerrell:
Let's do a followup PR for the outstanding comment cleanups to get the PRs landed in best merge conflict order

PR #251, opened 2022-08-23
Change Merkle verify asserts to Result Errs

tzerrell:
Replace assertion-based error handling in our Merkle verification code with Result-based error handling.

The intent is for `VerificationError::InvalidProof` to be used when we attempt to verify an alleged proof that is not actually a true proof, and to use other error types for issues not directly related to the validity of the proof.

Split off from #243 for clarity of purpose so that reviewers can more quickly give feedback on whether this is a good approach to handling verification errors.


shkoo (approved):
Neat!

flaub (approved):
Yay!

PR #256, opened 2022-08-26
Add raw byte interface to enable usage of frameworks like rkyv

starkoracles:
This is probably not the cleanest way to do it but it works. It mostly relies on existing plumbing in place for passing in serde in/out. Main changes:
* Separate between the input channel for serde serialized input to raw bytes. It didn't seem to make sense to have zero copy tape combined with serde as the reader knows how many bytes to consume for each serde input. This seems difficult to do in zero-copy. The new channel name is kSendRecvChannel_InitialInputAux / SENDRECV_CHANNEL_INITIAL_AUX_INPUT
* Add interface to write auxiliary input (zero-copy): add_input_u8_slice_aux / add_aux_input. It's more natural to use add_input_u8_slice_aux as serialization is done outside of risc0.
* Currently only allows for a single struct to be serialized using rkyv. I wasn't sure if we can have a single stream that supports multiple structs or if we will have to separate by index to make it easy to leverage using zero-copy.

Would love to change the approach if this seems going in the wrong direction


shkoo (commented):
Hi,

I don't think we need a dedicated "aux" channel.  You can use sendrecv with your own numbering scheme.

See this test for an example of how:
https://github.com/risc0/risc0/blob/c7f2f0ff4b0f4c3a4a5d50b6d5bbf1fe5f899ecb/risc0/zkvm/sdk/rust/src/host/mod.rs#L218-L244  (https://github.com/risc0/risc0/blob/c7f2f0ff4b0f4c3a4a5d50b6d5bbf1fe5f899ecb/risc0/zkvm/sdk/rust/methods/inner/src/bin/sendrecv.rs is the guest side of this test)

PR #257, opened 2022-08-27
Move FRI to generic field

tzerrell:
Allow FRI to work with generic field elems. The bulk of this is methods for representing field elements as sequences of `u32`s and providing conversions -- these are used to read and write field elements from the IoP. 

I also adjusted the Merkle tree tests to work with moderately large elements (of the baby bear field) rather than start from 0, as there were possible errors with casting to a smaller type that wouldn't be caught because all field elements in the merkle tree were small.


risc0/zkp/rust/src/verify/read_iop.rs
```diff
@@ -49,22 +50,28 @@ impl<'a, S: Sha> ReadIOP<'a, S> {
...
+            x[i] = <Fp as Elem>::from_u32_words(&self.proof[words * i..words * (i + 1)]);
+            // Ignore the odd-numbered values, which will be 0 for baby-bear
         }
-        self.proof = &self.proof[x.len()..];
+        self.proof = &self.proof[<Fp as Elem>::WORDS * x.len()..];
     }
 
     pub fn read_fp4s(&mut self, x: &mut [Fp4]) {
+        let elem_words = <Fp4 as ExtElem>::SubElem::WORDS;
+        let ext_size = <Fp4 as ExtElem>::EXT_SIZE;
```
flaub, line 63: These look unused?

flaub (approved):

tzerrell (commented):
Thanks for cleaning up and shepherding through to merging while I was away!

PR #258, opened 2022-08-27
Implement check_code for pure-rust verify

flaub:
* Workaround SHA circuit bug under pure-prove
* All pure-prove tests are 'passing'
* Add pure-prove tests to CI
* Fix guest build warnings
* Ensure risc0-build treats MethodIDs as being distinct between C++ and pure-prove
* Drop integration tests in favor of pure-prove unit tests


risc0/zkvm/sdk/rust/platform/src/memory.rs
```diff
@@ -61,7 +61,15 @@ pub const HEAP: Region = Region::new(0x00A0_0000, mb(20));
...
 pub const GPIO: Region = Region::new(0x01F0_0000, mb(1));
 pub const PROG: Region = Region::new(0x0200_0000, mb(10));
+#[cfg(not(feature = "pure-prove"))]
 pub const SHA: Region = Region::new(0x02A0_0000, mb(1));
+#[cfg(feature = "pure-prove")]
+// This HACK will cause the SHA accelerator to be skipped.
+// This is intended as a temporary workaround for bugs in the SHA circuit.
+// The hash will still be computed on the host via a GPIO signal,
+// but the polynomial constraints will not be applied to the ZKP.
+// *DO NOT USE IN PRODUCTION*
```
shkoo, line 71: Ugh, but yeah, I guess when we make pure-prove the default we'll notice this when we go to remote the feature.

risc0/zkvm/sdk/rust/src/lib.rs
```diff
@@ -47,8 +47,17 @@ pub mod prove;
...
 pub mod receipt;
 pub mod serde;
-#[cfg(feature = "verify")]
-pub mod verify;
 
-// TODO: get this from the cirgen circuit
-const CODE_SIZE: usize = 16;
+#[cfg(feature = "circuit")]
+lazy_static::lazy_static! {
+    pub static ref CIRCUIT: risc0_zkvm_circuit::CircuitImpl =
```
shkoo, line 53: Would it work to have CircuitImpl have a constant constructor so we don't have to use lazy_static?  Or maybe just "CIRCUIT: risc0_zkvm_circuit::CircuitImpl = CircuitImpl;" so we don't have to modify cirgen?

shkoo (approved):

risc0/zkvm/sdk/rust/src/lib.rs
flaub: I don't know if a constant constructor will ensure that only a single copy is resident in memory. That's the intent with the `lazy_static`

PR #259, opened 2022-08-28
Add a test for recursion

flaub:
The test only works with `pure-prove` and currently fails because it doesn't fit within MAX_CYCLES.

* Adjust verify to work with `no_std`
* Reduce verify dependencies
* Add `std` to `GuestOptions`


risc0/zkvm/sdk/rust/methods/verify/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
shkoo, line 1: Can we just use the risc0-zkvm-methods-inner package?  This seems like a large amount of boilerplate just to add a single method.

flaub: Yes, initially it had to be separate because I was building the receipt within the `build.rs`.

flaub: I'm having trouble figuring out how to make `verify.rs` avoid `std` while also being in the same dir as the `Cargo.toml` that has `default = ["std"]`

flaub: OK, I moved the verify package to be a sibling of inner, but I think they need to be separate due to differences in features and dependencies.

shkoo (approved):

PR #262, opened 2022-08-30
added the ability to test in guest environment

dymchenkko:
Added support to run tests in a guest environment.

These changes allow testing the guest environment with printing results to the console.

When the test_mode field is set to true in GuestOptions, in addition to build_guest_package() function, embed_methods_with_options() also runs the test_guest_package() function.

test_guest_package() - installs r0vm in target, runs tests, sets r0vm as a runner for tests and prints outputs.

Compromises made:

- A custom test runner is needed as libtest currently requires fmax/sqrt/fmin functions
- r0vm will always be the latest crates.io upstream, should ideally be parsing guest method Cargo.toml to find the right reference to it

See corresponding pull request in risc0-rust-starter: [https://github.com/risc0/risc0-rust-starter/pull/16](https://github.com/risc0/risc0-rust-starter/pull/16)

NOTE: this is my first open source contribution pull request, be gentle but firm, will happily correct issues!

### DIFF
--- a/risc0/zkvm/sdk/rust/build/src/lib.rs
+++ b/risc0/zkvm/sdk/rust/build/src/lib.rs
@@ -295,7 +295,7 @@ fn build_guest_package<P>(
     pkg: &Package,
     target_dir: P,
     guest_build_env: &GuestBuildEnv,
-    features: Vec<String>,
+    features: &Vec<String>,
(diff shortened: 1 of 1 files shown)


flaub:
This functionality should now be available in an upcoming release.

PR #263, opened 2022-08-30
Eliminate copies in ReadIOP, merkle verifier

shkoo:
* ReadIOP is now field agnostic, and returns slices instead of copying.

* Merkle verifier now splits the merkle tree into "top", which is just referenced by slice, and "rest", which is hashed based on "top".
  It still passes the expanded merkle tests; thanks @tzerrell!

### DIFF
--- a/risc0/zkp/rust/src/verify/merkle.rs
+++ b/risc0/zkp/rust/src/verify/merkle.rs
@@ -37,7 +37,7 @@ pub struct MerkleTreeVerifier<'a, S: Sha> {
     // contains the virtual indexes [top_size..top_size*2).
     top: &'a [Digest],
 
-    // These are the rest of the tree.  These have the virutal indexes [1, top_Size).
+    // These are the rest of the tree.  These have the virtual indexes [1, top_size).
     rest: Vec<S::DigestPtr>,
 }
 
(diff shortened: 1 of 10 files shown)


risc0/zkp/rust/src/verify/merkle.rs
```diff
@@ -28,79 +25,152 @@ use crate::{
...
+
+    // "top" is a reference, top_size long, to the top row.  This
+    // contains the virtual indexes [top_size..top_size*2).
+    top: &'a [Digest],
+
+    // These are the rest of the tree.  These have the virutal indexes [1, top_Size).
+    rest: Vec<S::DigestPtr>,
+}
+
+// Transltaes from virtual indexes to indexes in the "top" and "rest" arrays.
```
flaub, line 44:
```suggestion
// Translates from virtual indexes to indexes in the "top" and "rest" arrays.
```

risc0/zkp/rust/src/verify/merkle.rs
```diff
@@ -28,79 +25,152 @@ use crate::{
...
+    // Conceptually, the merkle tree here is twice as long as the
+    // "top" row (params.top_size), minus element #0.  The children of
+    // the entry at virtual index i are stored at 2*i and 2*i+1.  The
+    // root of the tree is at virtual element #1.
+
+    // "top" is a reference, top_size long, to the top row.  This
+    // contains the virtual indexes [top_size..top_size*2).
+    top: &'a [Digest],
+
+    // These are the rest of the tree.  These have the virutal indexes [1, top_Size).
```
flaub, line 40:
```suggestion
    // These are the rest of the tree.  These have the virutal indexes [1, top_size).
```

flaub (approved):

PR #264, opened 2022-08-30
Group imports by Std External Crate 

rlukata:
Use [group_imports = "StdExternalCrate"](https://rust-lang.github.io/rustfmt/?search=&version=v1.5.1#StdExternalCrate%5C%3A) to format and group imports by:

[std, core, malloc, etc]

[extern]

[crate, super]

### DIFF
--- a/bazel/rules/rust/rustfmt.toml
+++ b/bazel/rules/rust/rustfmt.toml
@@ -1,2 +1,3 @@
 wrap_comments = true
 normalize_comments = true
+group_imports = "StdExternalCrate"
--- a/risc0/zkp/rust/src/core/ntt.rs
+++ b/risc0/zkp/rust/src/core/ntt.rs
@@ -23,7 +23,6 @@ use super::{
     log2_ceil,
     rou::{ROU_FWD, ROU_REV},
 };
-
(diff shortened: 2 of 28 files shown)


flaub:
Nice!

For some reason we have two different `rustfmt.toml` files. We could try to consolidate later, but could you also update the one in the root of the repository? I think this one is used by cargo and maybe tools like vscode.

rlukata:
Looks like the root `rustfmt.toml` is a symbolic link to the file `bazel/rules/rust/rustfmt.toml`.

I did a `find` on the repo and couldn't find additional files:
```
rlukata@Ramis-MBP risc0 % find . -iname "rustfmt.toml"
./rustfmt.toml
./bazel/rules/rust/rustfmt.toml

shkoo:
Yes, I think you have to also format the stuff in methods/inner since it's a different workspace.

rlukata:
@shkoo Looks like they're already formatted and getting formatted. I've manually tried modifying some files in `methods/inner` and they all auto-format even when I view it from the `risc0` workspace.

flaub (approved):

PR #267, opened 2022-09-03
Add backtraces to generated circuits

shkoo:


flaub (approved):

risc0/core/source_loc.cpp
```diff
@@ -0,0 +1,152 @@
+#include "risc0/core/source_loc.h"
```
flaub, line 1: Missing license header

flaub:
This all seems great but I'd try to avoid using C string routines since they're so easy to make a mistake with. (C++ string routines are somewhat safer).

PR #268, opened 2022-09-03
Platform cleanup, std and pure-prove

shkoo:
* Fix rust prover sha bug, sha tests now pass

* Trim dependencies, less stuff to build in guest

* Add some Result propagation

* Guest methods can now decide whether they want to be no_std or not, and get the proper panic and alloc error implementations

* Merge separate "verify" method crate into regular "inner" method crate


risc0/zkvm/sdk/rust/guest/src/lib.rs
```diff
@@ -35,34 +34,63 @@ pub mod sha;
...
+}
+
+/// This defines standalone (no_std) panic and alloc error handlers in
+/// a guest method, for guest methods that do not want to use the
+/// standard library.
+///
+/// Guest methods will also need to specify these at the beginning of the file:
+///
+/// #![no_std]
+/// #![feature(alloc_error_handler)]
```
flaub, line 69: Won't this require guests to always use a nightly compiler even after we get the new target triple added to upstream?

risc0/zkvm/sdk/rust/src/prove/exec.rs
```diff
@@ -751,7 +752,7 @@ where
                 self.code[CodeIndex::Data1High] = round.1;
             } else {
                 self.code[CodeIndex::ShaCtrl] = ONE;
-                self.code[CodeIndex::P1] = Fp::new((sha_phase - 64 + 4) as u32);
+                self.code[CodeIndex::P1] = Fp::new((sha_phase - 68 + 4) as u32);
```
flaub, line 755: Nice find!

risc0/zkvm/sdk/rust/src/prove/mod.rs
```diff
@@ -120,34 +120,35 @@ impl<'a> ProverImpl<'a> {
 }
 
 impl<'a> exec::IoHandler for ProverImpl<'a> {
-    fn on_txrx(&mut self, channel: u32, buf: &[u8]) -> Vec<u8> {
+    fn on_txrx(&mut self, channel: u32, buf: &[u8]) -> Result<Vec<u8>> {
```
flaub, line 123: 👍

flaub (approved):

risc0/zkvm/sdk/rust/guest/src/lib.rs
shkoo: Only for no_std though

PR #272, opened 2022-09-08
Expand generic field usage

tzerrell:
Move forward replacing hardcoded baby bear with generic field (that for now is generally still baby bear). At this point the non-genericized code is largely limited to the automatically generated circuit code and the adapter code that interfaces with it.

I have design questions / uncertainties on a few details, which I'll call attention to with inline comments.


risc0/zkp/rust/src/core/ntt.rs
```diff
@@ -77,31 +73,35 @@ pub fn bit_reverse<T: Copy>(io: &mut [T]) {
...
     // no-op base case
 }
 
 #[inline]
-fn rev_butterfly_0<T>(_: &mut [T]) {
+fn rev_butterfly_0<B, T>(_: &mut [T]) {
     // no-op base case
 }
 
+// TODO: This generates butterfly functions up to $n = 32, but will panic if $n
```
tzerrell, line 85: I am generating butterfly functions up to `$n = 32`, but for some fields (e.g. baby bear) these aren't all legal and some will panic. Would it be better to return `Option`s and handle bad calls?

risc0/zkp/rust/src/verify/mod.rs
```diff
@@ -63,48 +62,154 @@ impl fmt::Display for VerificationError {
...
+    );
 }
 
 #[cfg(feature = "host")]
 mod host {
     use super::*;
     use crate::adapter::{PolyExt, PolyExtContext};
 
-    pub struct CpuVerifyHal<'a, S: Sha, C: PolyExt> {
+    // TODO: Not certain how best to handle the unused Field that's needed to
```
tzerrell, line 192: The `CpuVerifyHal` needs to know which `Field` elements it will be working with, but it doesn't actually directly use the field -- is this the right design for providing that information?

flaub: I think you can use `PhantomData` for this purpose.

https://doc.rust-lang.org/std/marker/struct.PhantomData.html

tzerrell:
The updates in FRI broke performance, so I have reverted them for now; I'm looking into reapplying them without taking a perf hit.

tzerrell:
Thank you to @shkoo for using his new perf tool to track down the performance loss in FRI!

flaub (approved):
LGTM

PR #274, opened 2022-09-12
Add initial support to profile guest runs

shkoo:
This is accessible with the --prof-out option of r0vm, or by using risc0_zkvm::host::profiler::Profiler.

It produces output in a format suitable for use with google's pprof, but does not yet supply full stack traces.


risc0/zkvm/sdk/rust/methods/inner/src/bin/multi_test.rs
```diff
@@ -57,6 +73,21 @@ pub fn main() {
             // our simulation doesn't run faster.
             assert!(total >= 72, "total: {total}");
         }
+        #[cfg(feature = "pure-prove")]
+        3 => unsafe {
```
flaub, line 77: Might be good to use an enum for this rather than a raw 'magic' number.

risc0/zkvm/sdk/rust/src/host/mod.rs
```diff
@@ -291,16 +344,159 @@ mod test {
...
+                .with_trace_callback(prof.make_trace_callback());
+            let mut prover = Prover::new_with_opts(&elf_contents, MULTI_TEST_ID, opts).unwrap();
+            prover.add_input_u32_slice(&[
+                4, // Generate known profiling trace
+                0,
+            ]);
+            prover.run().unwrap();
+        }
+
+        prof.finalize();
```
flaub, line 379: Can we make the profiler use an RAII pattern?

risc0/zkvm/sdk/rust/Cargo.toml
```diff
@@ -41,12 +48,13 @@ test-log = "0.2"
...
 [features]
 circuit = []
-default = ["host", "prove", "std", "verify"]
+default = ["host", "prove", "std", "verify", "profiler"]
 doc = ["std"]
 host = ["risc0-zkp/host"]
 prove = ["circuit", "dep:lazy-regex", "risc0-zkp/prove", "risc0-zkvm-circuit/cpp"]
 std = ["risc0-zkp/std", "serde/std"]
 verify = ["circuit", "risc0-zkp/verify"]
+profiler = ["dep:prost-build", "dep:protobuf-src"]
```
flaub, line 57: Should we add `addr2line`, `gimli` and `prost` to this feature?

flaub (approved):

PR #275, opened 2022-09-13
Starting expansion of ZKVM rustdocs

3lkn:


tzerrell (changes requested):
`cargo doc` is currently broken and this probably shouldn't land without it running. That said, the problem seems to be on `main` and not this PR -- we should add `cargo doc` or a variant to CI, either in this PR or a separate one, to catch this in the future.

risc0/zkvm/sdk/rust/src/receipt.rs
```diff
@@ -88,6 +99,13 @@ impl Receipt {
         self.verify_with_hal(&hal, method_id)
     }
 
+    /// This function is called by [verify](Receipt::verify), which provides the
+    /// CPU HAL generated using the program circuit.
```
tzerrell, line 103: IIUC,  this is the generalized verify, which allows the caller to select whatever HAL they like (whereas `verify` uses the default SHA & CPU HAL). The way this comment is framed feels a bit backwards to me -- it doesn't exist to be called by `verify`, `verify` exists to provide default argument to `verify_with_hal`.

tzerrell (approved):
cool!

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -107,28 +107,96 @@ pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> (&'static [u32], usize) {
     ENV.get().send_recv_as_u32(channel, buf)
 }
 
-/// Read private data from the host.
+/// Reads private data from the host.
+///
+/// # Examples
+/// Values are read in the order in which they are written by the host),
```
flaub, line 113:
```suggestion
/// Values are read in the order in which they are written by the host,
```

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -107,28 +107,96 @@ pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> (&'static [u32], usize) {
...
+/// as in a queue. In the following example, `first_value` and `second_value`
+/// have been shared consecutively by the host via `env::write`.
```
flaub, line 115: On the host side, the call would be `prover.add_input` or some such.

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -107,28 +107,96 @@ pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> (&'static [u32], usize) {
...
+/// ```rust, ignore
+/// let first_value: u64 = env::read();
+/// let second_value: u64 = env::read();
+/// ```
+/// For ease and clarity, we recommend sharing multiple values between guest and
+/// host as a struct. As a practical example, in our [digital signature example](https://github.com/risc0/risc0-rust-examples/tree/main/digital-signature),
+/// we read in a `SigningRequest` that contains both a `Passphrase` and a
```
flaub, line 122: It's be better if we had a self-contained example so that:
1. it requires less context to understand the idea
2. it doesn't depend on another repository where the link might eventually fail

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -107,28 +107,96 @@ pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> (&'static [u32], usize) {
...
 
-/// Write private data to the host.
+/// Writes private data to the host.
+///
+/// # Arguments
+///
+/// * `data` - serialized data to be made available in host-readable memory.
+/// # Example
+/// In this example, the value `42` is written and is then
+/// accessible to the host via `env::read`.
```
flaub, line 138: The host would receive this value thru accessing the journal, `env` is only used on the guest-side.

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -107,28 +107,96 @@ pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> (&'static [u32], usize) {
...
+/// ```rust, ignore
+/// let count = get_cycle_count();
+/// ```
+/// This function can be used to note how many cycles have elapsed during a
+/// guest operation:
+/// ```
+/// let count1 = get_cycle_count();
+/// doSomething();
+/// let count2 = get_cycle_count();
+/// cycles_elapsed = count2 - count1;
```
flaub, line 188:
```suggestion
/// let cycles_elapsed = count2 - count1;
```

risc0/zkvm/sdk/rust/src/receipt.rs
```diff
@@ -88,6 +99,13 @@ impl Receipt {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `hal` - the HAL used to represent the guest program circuit.
```
flaub, line 107: Actually the HAL is independent of the circuit, it's the "Hardware Abstraction Layer" which provides a way to execute portions of the verification in an efficient way based on whether the verify is happening on the host or in the guest (aka recursion).

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -95,8 +95,8 @@ pub(crate) fn finalize(result: *mut usize) {
     ENV.get().finalize(result);
 }
 
-/// Exchanges data with the host, returning the data from the host
-/// as a slice of bytes.
+/// Exchanges data with the host, returning data from the host
+/// as a slice of bytes. In most cases, using env::write() is preferred.
```
flaub, line 99: Maybe word this like:
"See `env::write()` for passing structured data to the host".

Basically it's not really preferred, it just depends on what the needs of the user are. I have a feeling we're going to pretty significantly revamp this part of the API once the zero-copy deserialize lands. There's already some open PRs to provide a better API to allow raw-byte access (which I think will become obsolete when zero-copy lands).

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -34,8 +34,8 @@ use crate::{align_up, memory_barrier, sha};
 
 #[cfg(not(target_os = "zkvm"))]
 // Bazel really wants to compile this file for the host too, so provide a stub.
-/// Stub version of risc0_zkvm_platform::rt::host_sendrecv, re-exported for
-/// easy access through the SDK.
+/// This is a stub version of risc0_zkvm_platform::rt::host_sendrecv,
```
flaub, line 37:
```suggestion
/// This is a stub version of `risc0_zkvm_platform::rt::host_sendrecv`,
```

Or perhaps make an actual rustdoc link here.

risc0/zkvm/sdk/rust/guest/src/sha.rs
```diff
@@ -86,8 +86,8 @@ pub(crate) unsafe fn raw_digest_to(data: &[u32], digest: *mut Digest) {
     GPIO_SHA.as_ptr().write_volatile(desc_ptr);
 }
 
-// Calculates the number of words of capacity needed, including end
-// marker and trailer, to take the SHA hash of len_bytes bytes.
+// Calculates the capacity needed as a number of words, including end
+// marker and trailer, in order to take the SHA hash of len_bytes bytes.
```
flaub, line 90:
```suggestion
// marker and trailer, in order to take the SHA hash of `len_bytes` bytes.
```

risc0/zkvm/sdk/rust/build/src/lib.rs
```diff
@@ -404,12 +404,18 @@ fn build_guest_package<P>(
 /// [`embed_methods_with_options`].
 pub struct GuestOptions {
     /// The number of po2 entries to generate in the MethodID.
+    ///
+    /// By default, this value is set to `DEFAULT_METHOD_ID_LIMIT`.
+    ///
+    /// Successive MethodID entries increase in size by powers of two, so
+    /// increasing the limit by 1 effectively doubles the limit on code
```
pdg744, line 411: perhaps: 

`... increasing code_limit by 1...`

risc0/zkvm/sdk/rust/build/src/lib.rs
```diff
@@ -404,12 +404,18 @@ fn build_guest_package<P>(
 /// [`embed_methods_with_options`].
 pub struct GuestOptions {
     /// The number of po2 entries to generate in the MethodID.
```
pdg744, line 406: Perhaps just:

/// The number of <s>po2</s> entries <s>to generate</s> in the MethodID.

risc0/zkvm/sdk/rust/guest/src/env.rs
```diff
@@ -107,28 +107,96 @@ pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> (&'static [u32], usize) {
...
 
-/// Returns the number of processor cycles that have occured since the guest
+/// Returns the number of processor cycles that have occurred since the guest
 /// began.
+///
+/// # Examples
+/// ```rust, ignore
+/// let count = get_cycle_count();
+/// ```
+/// This function can be used to note how many cycles have elapsed during a
```
pdg744, line 182: rad

risc0/zkvm/sdk/rust/src/lib.rs
```diff
@@ -12,24 +12,23 @@
...
+//! code. No additional information about code execution (such as which inputs
+//! were provided to the guest) is revealed by publishing the
 //! [Receipt](host::Receipt). A high-level overview of how the zkVM is
-//! structured to accomplish this is available in our
-//! [Overview of the zkVM](https://www.risczero.com/docs/explainers/zkvm/zkvm_overview)
-//! explainer.
+//! structured to accomplish this is available in our [Overview of the zkVM](https://www.risczero.com/docs/explainers/zkvm/zkvm_overview).
 //!
-//! Developers new to RISC Zero are encouraged to get started with our
+//! Developers new to RISC Zero are encouraged to begin with our
```
pdg744, line 27: nice

risc0/zkvm/sdk/rust/src/receipt.rs
```diff
@@ -69,6 +69,17 @@ where
 }
 
 impl Receipt {
+    /// Verifies the proof receipt generated when the guest program is run.
```
pdg744, line 72: Perhaps: 

`Verifies a RISC Zero Receipt`

tzerrell (approved):
LGTM once CI is happy

PR #278, opened 2022-09-18
Memset and digest_u8_slice benchmark

stskeeps:
Add a throughput measurement for memset and a benchmark that tests the sha::digest_u8_slice API performance which is quite dramatically worse than raw_sha throughput


flaub:
Superseded by #289

PR #281, opened 2022-11-14
fix: remove default features (std) from rand deps

austinabell:
I didn't verify in-depth, just going off of intuition of what seems like would be a necessary step toward compatibility with more targets to solve issues like https://github.com/risc0/battleship-example/issues/15.


flaub:
I think this is resolved in an upcoming release.

PR #293, opened 2022-11-30
support insecure_skip_seal feature in guest

SchmErik:
This is done by moving insecure_skip_seal() function from prover to receipt. This feature is used to faciliate development of the guest code and not intended for production use.


SchmErik:
Hey @hashcashier could you test this?

flaub (approved):

PR #297, opened 2022-12-05
Correct misleading comment about the choice of polynomial

intoverflow:
Also adds a unit test to pin-down the choice of field extension.


jacobdweightman:
There are also a few (duplicated) comments which mention `X^4 - 11` in zirgen. I'm fine with this as-is, but I would also like to see those fixed.

jacobdweightman (approved):

jbruestle (approved):

risc0/zkp/src/field/baby_bear.rs
```diff
@@ -331,12 +331,12 @@ const fn decode(a: u32) -> u32 {
...
 
 /// Instances of `ExtElem` are elements of a finite field `F_p^4`. They are
-/// represented as elements of `F_p[X] / (X^4 - 11)`. This large
+/// represented as elements of `F_p[X] / (X^4 + 11)`. This large
 /// finite field (about `2^128` elements) is used when the security of
 /// operations depends on the size of the field. The field extension `ExtElem`
 /// has `Elem` as a subfield, so operations on elements of each are compatible.
-/// The irreducible polynomial `x^4 - 11` was chosen because `11` is
-/// the simplest choice of `BETA` for `x^2 - BETA` that makes this polynomial
+/// The irreducible polynomial `x^4 + 11` was chosen because `11` is
```
pdg744, line 338: Looks like line 339 should read x^4 + BETA rather than x^2 + BETA.

intoverflow: Good catch. Thanks!

pdg744:
subtle nit: 

I noticed that in addition to changing - 11 to + 11, this PR also changes `- BETA` to `+ BETA`. It looks like BETA appears in code, and it's not quite clear to me whether we're using:

`f(x) = x - BETA`, where BETA = -11 

`f(x) = x + BETA` where BETA = 11`. 

@jacobdweightman @jbruestle maybe one of you can clarify?

pdg744:
i.e. see BETA vs. NBETA here 

https://github.com/risc0/risc0/blob/c15712debd1788b5691a553c5ca270f2f5dd9b2d/risc0/zkp/src/field/baby_bear.rs#L491

PR #302, opened 2022-12-08
Minimize deps for verify-only use cases

flaub:
* Drop explicit `getrandom` dep
* Use `default-features = false` when depending on `risc0-zkvm` to get verify-only behavior.
* Use `rand_core` to avoid `rand` dependency
* Drop `verify-only `feature
* Introduce `prove` feature (enabled by default)


zie1ony:
I checked locally and can confirm it works.

PR #303, opened 2022-12-14
Update SHA impl so that state is encoded as big-endian words

flaub:
* Make downstream uses less confusing


intoverflow:
It looks like the endianness flip is visible from the Digest. Was it not possible to encapsulate the endianness concern within the circuit?

jbruestle (approved):

jbruestle:
> It looks like the endianness flip is visible from the Digest. Was it not possible to encapsulate the endianness concern within the circuit?

So the main reason to make this change is actually to make hashing of hashes works the way one would expect:  According to the SHA spec, the words in the hash are supposed to written to memory in BE order, but we always kept digests in native (LE in our case) order, which meant that if one did a hash of a digest, the results would be different than most other.  So this change does impact any and all users of the hashing (including doubtless the lean verifier, and also unfortunately my verifier), but it brings us into alignment with the spec finally.

PR #308, opened 2022-12-18
Add MemoryImage: to be used for continuations

flaub:


risc0/zkvm/src/prove/image.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+pub struct MemoryImage {
+    pub image: Vec<u8>,
+    pub root: Digest,
+    info: PageTableInfo,
+}
+
+impl MemoryImage {
+    pub fn new(elf_contents: &[u8], page_size: usize) -> Result<Self> {
```
SchmErik, line 60: Why not return `Self` instead of `Result<Self>` on this function?

risc0/zkvm/src/prove/image.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    fn page_table_size(max_mem: usize, page_size: usize) -> usize {
+        PageTableInfo::new(max_mem, page_size).total
+    }
+
+    #[test]
+    fn check_integrity() {
+        const PAGE_SIZE: usize = 1024;
+        let image = MemoryImage::new(MULTI_TEST_ELF, PAGE_SIZE).unwrap();
+        // std::fs::write("/tmp/test.img", &image.image).unwrap();
```
SchmErik, line 152: This comment looks like it's left over from a previous draft. If that's the case, could you remove it?

SchmErik (commented):
Looks good but I left a few small comments

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -72,6 +70,8 @@ pub mod reg_abi {
     pub const REG_T4: usize = 29; // temporary
     pub const REG_T5: usize = 30; // temporary
     pub const REG_T6: usize = 31; // temporary
+    pub const REG_PC: usize = 32; // program counter
```
SchmErik, line 73: Does this imply that we'll remove pc from the `MachineContext` struct and use this instead? I'm asking because it (positively) affects the gdb support that I'm working on

risc0/zkvm/src/prove/image.rs
flaub: It's actually useful in case you want to examine the entire memory image.

risc0/zkvm/platform/src/syscall.rs
flaub: My plan is to have the PC be recorded in system memory just before pausing a session. We need to record it for the next sessions so that:
1. we can verify that the PC was in a particular position
2. we can resume at the last PC position

The `MachineContext` will probably change a lot, but I don't expect to have all that working for awhile yet.

risc0/zkvm/src/prove/image.rs
flaub: Because `Program::load_elf` can return an error, and we need to avoid panics.

SchmErik (approved):
@flaub thanks for answering my comments/questions! Looks good to me

PR #311, opened 2022-12-19
Another modification to RNG generation to be both constant time, secure, and easy to implement as a circuit.

jbruestle:


intoverflow (approved):
Love it. The use of shifting is especially slick 🤘💯

pdg744 (approved):
Beautiful -- thanks for the clear explanation in the comment!

risc0/zkp/src/field/baby_bear.rs
```diff
@@ -109,18 +109,35 @@ impl field::Elem for Elem {
...
-        // due to new ZK padding, and even at 1 proof-per-second, mean time to
-        // failure is > 2 years. In a better world, we might propagate this
-        // failure and retry at the proof level
-        const REJECT_CUTOFF: u64 = (u64::MAX / (P as u64)) * (P as u64);
-        if val >= REJECT_CUTOFF {
-            panic!("Random number generator got very unlucky");
+        // Normally, we would use rejection sampling here, but our specialized
+        // verifier circuit really wants an O(1) solution to sampling.  So instead, we
+        // sample [0, 2^192) % P.  This is very close to uniform, as we have 2^192 / P
+        // full copies of P, with only 2^192%P left over elements in the 'partial' copy
```
flaub, line 115:
```suggestion
        // full copies of P, with only `2 ^ 192 % P` left over elements in the 'partial' copy
```

risc0/zkp/src/field/baby_bear.rs
```diff
@@ -109,18 +109,35 @@ impl field::Elem for Elem {
...
+        // (which we would normally reject with rejection sampling).
+        //
+        // Even if we imagined that this failure to reject totally destroys soundess,
+        // the probablity of it occuring even once during proving is vanishingly low
+        // (for the about 50 samples our current verifier pulls and at a probability of
+        // less than2^-161 per sample, this is less than 2^-155).  Even if we target
```
flaub, line 121:
```suggestion
        // less than `2 ^ -161` per sample, this is less than `2 ^ -155`).  Even if we target
```

flaub (approved):

flaub (approved):

PR #313, opened 2022-12-22
adding comments for clarifying accum

pdg744:


risc0/zkp/src/prove/accum.rs
```diff
@@ -88,7 +95,9 @@ impl<'a, F: Field> CircuitStepHandler<F::Elem> for Handler<'a, F> {
     fn call(
         &mut self,
         cycle: usize,
+        // TODO what is name?
```
flaub, line 98: This is the name of the extern call to perform.
Examples: `getMajor`, `ramRead`, `syscall`

risc0/zkp/src/prove/accum.rs
```diff
@@ -88,7 +95,9 @@ impl<'a, F: Field> CircuitStepHandler<F::Elem> for Handler<'a, F> {
...
         name: &str,
+        // TODO what is extra?
```
flaub, line 100: This is an extra string argument that is only used by the `log` extern call. 

The extern call interface isn't particularly clean or well formed, but it's complicated by having to deal with calling thru C++ to rust and is generated by zirgen.

risc0/zkp/src/prove/accum.rs
```diff
@@ -21,23 +21,29 @@ use crate::{
...
-    // Plonk kinds and their data.
+    // PLOOKUP relies on two grand product accumulation checks;
+    // one for the memory permutation and a second for a lookup table.
+    // We have two `kinds`: memory and bytes.
+    // TODO check with Jeremy if this is accurate. (what are "kinds?")
     kinds: BTreeMap<String, Vec<E>>,
 }
 
 impl<E: Elem> Accum<E> {
+    // Constructs a new Accum
```
flaub, line 37: This doesn't seem like a very useful comment to me

flaub (commented):
Hopefully all TODOs can be resolved before landing, otherwise LGTM

risc0/zkp/src/prove/accum.rs
```diff
@@ -21,23 +21,29 @@ use crate::{
     field::{Elem, ExtElem, Field},
 };
 
-/// Tracks plonk accumulations for an execution.
+/// Tracks grand product accumulations for PLOOKUP.
```
flaub, line 24: My understanding was that we are using the PLONK permutation argument to verify that memory is valid. In this section of the code, I'm not sure PLOOKUP applies. I think PLOOKUP is used in the circuit to avoid having to split every bit of a byte or word into an individual column.

pdg744: I'm not sure either -- @jbruestle can you clarify whether Accum is just used for the memory-based grand product accumulation or whether this struct is also involved in the bytes-based accumulation?

pdg744: The answer to the TODO about `kinds` will be clarifying on this front.

jbruestle: So both are permutation arguments, but and accum is used for both, PLOOKUP is too specific.  Technically, since at least in our case PLOOKUP is a specialization of PLONK, I'd either call it PLONK, or just a permutation argument.

risc0/zkp/src/prove/accum.rs
```diff
@@ -21,12 +21,15 @@ use crate::{
...
 pub struct Accum<E: Elem> {
     /// Total number of cycles in this run.
     cycles: usize,
 
-    // Plonk kinds and their data.
+    // PLOOKUP relies on two grand product accumulation checks;
+    // one for the memory permutation and a second for a lookup table.
+    // We have two `kinds`: memory and bytes.
+    // TODO check with Jeremy if this is accurate. (what are "kinds?")
```
jbruestle, line 32: Yes, two kinds, memory + bytes, also not PLOOKUP for memory.

risc0/zkp/src/prove/accum.rs
```diff
@@ -38,6 +41,8 @@ impl<E: Elem> Accum<E> {
         }
     }
 
+    // Generates "pre-" side of grand product accumulation
+    // TODO check with Jeremy if this is accurate. (what is "prefix?")
```
jbruestle, line 45: Here 'prefix' is related to prefix-product mechanism for doing parallel computations of products (see https://en.wikipedia.org/wiki/Prefix_sum, same idea since the underlying mechanism works for anything commutative and associative)

risc0/zkp/src/prove/accum.rs
```diff
@@ -57,6 +60,7 @@ impl<E: Elem> Accum<E> {
 }
 
 pub struct Handler<'a, F: Field> {
+    // TODO what is p?
```
pdg744, line 63: @jbruestle one TODO remaining -- can you clarify the role of `p` in the Handler?

flaub: @shkoo Do you know what `p` refers to in this context?

shkoo: p is a mutex lock (an implementation detail so it can be used in multiple threads and as a callback) containing a "Accum" structure, that's commented as "Tracks grand product accumulations for PLONK-style permutation arguments." on line 24

pdg744: Thanks for clarifying, @shkoo
I've edited the TODO to read, "p is a mutex lock that contains an Accum structure;" lmk if you'd revise that.

pdg744:
All TODOs have been taken care of -- thanks for the help, everyone! 

Will land soon (Tuesday AM?) unless anyone has any last edits to suggest.

flaub (approved):

PR #314, opened 2022-12-22
clarifying zkp/src/prove/executor

pdg744:
adding comments to clarify meaning

### DIFF
--- a/risc0/zkp/src/prove/executor.rs
+++ b/risc0/zkp/src/prove/executor.rs
@@ -46,7 +46,7 @@ where
     pub output: Vec<F::Elem>,
     // Power of 2
     pub po2: usize,
-    // steps = 2^po2 is the total number of steps in the zkVM execution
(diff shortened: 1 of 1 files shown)


flaub (approved):

risc0/zkp/src/prove/executor.rs
```diff
@@ -32,16 +32,27 @@ where
...
     code_size: usize,
+    // Execution Trace Data
     pub data: Vec<F::Elem>,
+    // Number of columns used for execution trace data
     data_size: usize,
+    // Circuit output
     pub output: Vec<F::Elem>,
+    // Power of 2
     pub po2: usize,
+    // Counter for zkvm execution
```
jbruestle, line 49: steps is the number of cycles (I probably should have called it num_cycles or something), i.e. it's 2^po2.  cycle is the counter.

PR #326, opened 2023-01-10
Replace MethodIDs with ImageIDs 

flaub:


tzerrell (approved):
🎉

LGTM (of course, I'm not really suited for reviewing the more circuit-integration-y parts of this, but the bits I could review well look good)

risc0/build/risc0.ld
```diff
@@ -21,10 +21,9 @@ EXTERN(__start)
...
 /* Must match risc0/zkvm/platform/src/memory.rs */
 MEMORY {
-  stack        : ORIGIN = 0x00000000, LENGTH =  16M
-  data    (RW) : ORIGIN = 0x01000000, LENGTH =  16M
-  heap         : ORIGIN = 0x02000000, LENGTH =  32M
-  input        : ORIGIN = 0x04000000, LENGTH =  32M
+  stack        : ORIGIN = 0x02000000, LENGTH =  16M
+  data    (RW) : ORIGIN = 0x03000000, LENGTH =  16M
+  heap         : ORIGIN = 0x04000000, LENGTH =  32M
   prog    (X)  : ORIGIN = 0x06000000, LENGTH =  32M
```
jbruestle, line 27: Is there a reason not to put prog @ 0x08... - 0x0c...?  I was of the impression that system memory starts only after 0x0c and that nondet was now below @0x02.  It seems like 64M for heap and 64M for code would be better than 32M for heap and 96M for program.  Or maybe even more room for heap.

risc0/zkp/src/verify/mod.rs
```diff
@@ -258,13 +250,12 @@ pub fn verify<'a, H, C, F>(
     hal: &'a H,
     circuit: &C,
     seal: &'a [u32],
-    journal: &'a [u32],
-    check_code: F,
+    check_globals: F,
```
jbruestle, line 253: We really need to support both checking globals and checking the code root.  I'm OK with the ZKVM temporarily ignoring the code root until we can add the build support to precompute the valid roots, but the recursion support will also need to check the code root, so I'd like to keep the API in place.

risc0/zkvm/platform/src/memory.rs
```diff
@@ -54,17 +54,14 @@ impl Region {
...
-pub const DATA: Region = Region::new(0x0100_0000, mb(16));
-pub const HEAP: Region = Region::new(0x0200_0000, mb(32));
-pub const INPUT: Region = Region::new(0x0400_0000, mb(32));
-pub const PROG: Region = Region::new(0x0600_0000, mb(32));
+pub const INPUT: Region = Region::new(0x0000_0400, mb(32) - kb(1));
+pub const STACK: Region = Region::new(0x0200_0000, mb(16));
+pub const DATA: Region = Region::new(0x0300_0000, mb(16));
+pub const HEAP: Region = Region::new(0x0400_0000, mb(32));
+pub const TEXT: Region = Region::new(0x0600_0000, mb(32));
 pub const OUTPUT: Region = Region::new(0x0800_0000, mb(32));
```
jbruestle, line 62: Is there any reason to have a specific output + commit region anymore?  As far as I can tell there is no reason output can't just send heap pointers to the host.  Also for the commits, cant we just have a single SHA256 state and a single 16 word staging buffer (both in the heap), and just write words to the staging buffer and kick off a compress whenever it's full?

flaub: Ah yes, that's a good point on both counts. Now that we have `sys_sha_buffer`, we can probably drop the COMMIT region. I also don't think we need an OUTPUT region anymore!

risc0/build/risc0.ld
flaub: That all makes great sense. I was also thinking maybe `prog` should live above the system region? Like maybe reading from the PC could use a special memory op.

jbruestle (approved):

SchmErik (approved):

risc0/build/src/lib.rs
```diff
@@ -121,57 +121,21 @@ impl Risc0Method {
             std::process::exit(-1);
         }
 
-        let method_id_path = self.elf_path.with_extension("id");
-        let elf_sha_path = self.elf_path.with_extension("sha");
-        let elf_contents = std::fs::read(&self.elf_path).unwrap();
-
-        let elf_sha = Sha256::new()
-            // Method ID calculation is slow, so only recalculate it if we
-            // actually get a different ELF file.
```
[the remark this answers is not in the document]
flaub: It seems to be pretty fast without having to cache it. Caching comes with all sorts of complexity and potential for bugs. If we find out it's still too slow we can bring caching back.

PR #327, opened 2023-01-10
Move zip download tmpdir to $HOME/.risc0

mothran:
This fixes build issues with fs::rename() across filesystems when /tmp/ is mounted separately from the location of the repo. Hit error message `Invalid cross-device link` while trying to build `main` on Arch Linux.


flaub (approved):
Thanks!!

PR #330, opened 2023-01-14
Improvements to SHA-256 API

nategraf:
Work in progress PR to make improvements to the SHA-256 API and it's usages.

- [x] Refactor the `Sha` trait and `Digest` type to clean up the API and use Rust idioms.
- [x] Enable users to incrementally hash data instead of only being able to do so all at once.
- [x] Create a forked version of the `sha2` crate with support for the zkVM guest.
    - Option 1: Get it working by calling into the risc0 crate for the compress function.
    - Option 2: Use in-line assembly alongside the ASM versions of SHA for x86.
- ~~Change `env::commit` to incrementally hash objects in-place instead of needing to copy them into the commit region.~~
    - ~~If this works, consider downsizing or removing the commit memory region.~~ 
    - (Will look into this while this PR is being finalized and put into a followup PR)
    
Fixes: https://github.com/risc0/risc0/issues/356


nategraf:
Opened this PR as a draft to get the CI checks to run tests across platforms.

risc0/zkp/src/core/sha_cpu.rs
```diff
@@ -44,22 +44,32 @@ impl Impl {
         size: usize,
         stride: usize,
     ) -> Box<Digest> {
-        let mut state = *SHA256_INIT.get();
+        // Because bytes in our digests are in bid-endian order, regardless of host
```
flaub, line 47:
```suggestion
        // Because bytes in our digests are in big-endian order, regardless of host
```

risc0/zkp/src/core/sha.rs
```diff
@@ -26,20 +26,28 @@ use hex::{FromHex, FromHexError};
 use risc0_zeroio::{Deserialize as ZeroioDeserialize, Serialize as ZeroioSerialize};
 use serde::{Deserialize, Serialize};
 
+/// The size of a word used with the SHA-256 implementation and within
+/// a [Digest] and [Block] representations (32-bits = 4 bytes).
+pub const WORD_SIZE: usize = mem::size_of::<u32>();
```
flaub, line 31: BTW, I think this is already defined in the platform crate, I've been meaning to go thru and reduce duplication of constants.

nategraf: Ah yes, I see that. I'll go ahead and de-duplicate that.

nategraf:
This PR now ready for review. I'll still do one more polishing pass tomorrow afternoon, but I don't expect any more major changes.

I have some remaining TODOs left for looking into usages of the non-standard SHA-256 functions and determining whether they A) might result in any issues or B) can be replaced with a standard hashing call. But any results from this will go into a follow-up PR.

risc0/zkp/src/core/sha_cpu.rs
```diff
@@ -35,31 +34,40 @@ fn set_word(buf: &mut [u8], idx: usize, word: u32) {
...
             .iter()
             .skip(offset)
             .step_by(stride)
             .take(size)
             .flat_map(|pod| bytemuck::cast_slice(slice::from_ref(pod)) as &[u8])
-            .cloned()
+            .copied()
             .fuse();
 
+        // TODO(victor): It looks like this function _does not_ implement canonical
```
jbruestle, line 66: This call should only occur inside Seal verification of data structures whose sizes are determined by the protocol (i.e. collisions of different sized data should never be relevant).  That said, the cost to doing standard padding in this case is minimal, so perhaps it's a good idea.

jbruestle, line 66: This call should only occur inside Seal verification of data structures whose sizes are determined by the protocol (i.e. collisions of different sized data should never be relevant). That said, the cost to doing standard padding in this case is minimal, so perhaps it's a good idea.

risc0/zkvm/methods/build.rs
```diff
@@ -11,11 +11,12 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
 use std::{collections::HashMap, env};
 
 use risc0_build::{embed_methods_with_options, GuestOptions};
 
+#[allow(unreachable_code)]
```
jbruestle, line 18: Why?

risc0/r0vm/Cargo.toml
```diff
@@ -13,6 +13,7 @@ bytemuck = "1.12"
 clap = { version = "4.0", features = ["derive"] }
 env_logger = "0.10.0"
 risc0-zkvm = { workspace = true, features = ["default", "profiler", "binfmt"] }
+risc0-zkvm-platform = { workspace = true }
```
flaub, line 16: This looks like a dev-dependency

jbruestle (approved):

risc0/zkp/src/core/sha_cpu.rs
nategraf: I'm going to look into this a bit more, but not in this PR.

risc0/zkvm/methods/build.rs
nategraf: I removed this. Can't remember why I added that. I'll do a pass to make sure nothing else if left from development work.

PR #331, opened 2023-01-14
Update copyright year and automatically check modified files

nategraf:
This PR updates the copyright year to 2023 in `.licenserc.json`, and in all files that have been modified since the start of the year.

Additionally, this PR includes a new copyright check command that ensures a PR will require an updated copyright on any files modified in the PR.

Opened this PR after running into the copyright check in https://github.com/risc0/risc0/pull/330

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
       - run: |
           # Check that no source files updated this year have a missing or out-of-date copyright.
           set -e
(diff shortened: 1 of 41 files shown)


nategraf:
This PR is currently trying to use a `git` command to figure out which files have been touched since the beginning of the year, but there is something about the way GH Actions checks out repos that is preventing that from working. I'll spend a few more minutes looking into this on Monday and, if I can't get it to work, opt for a simpler solution.

flaub (approved):
Thanks!!

PR #344, opened 2023-01-20
Revise Rust docs

tzerrell:
Revises the Rust documentation in preparation for a 0.12 release.

### DIFF
--- a/risc0/zkp/src/prove/mod.rs
+++ b/risc0/zkp/src/prove/mod.rs
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Cryptographic algorithms for producing a ZK proof of compute
+//!
+//! This module is not typically used directly. Instead, we recommend the
+//! higher-level tools offered in [`risc0_zkvm::prove`].
+//!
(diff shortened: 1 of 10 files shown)


risc0/zkp/src/prove/mod.rs
```diff
@@ -12,6 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Cryptographic algorithms for producing a ZK proof of compute
+//!
+//! This module is not typically used directly. Instead, we recommend the
+//! higher-level tools offered in [`risc0_zkvm::prove`].
+//!
+//! [`risc0_zkvm::prove`]: ../../risc0_zkvm/prove/index.html
```
flaub, line 20: I'm actually not sure if relative cross-crate links will work. We could try publishing a `0.12.0-rc.2` to see if this works on `docs.rs`

flaub (approved):

pdg744:
Rad! 
Thanks for pushing these improvements!

risc0/zkp/src/prove/mod.rs
tzerrell: SGTM, it works locally but I agree that doesn't necessarily mean it will work on `docs.rs`.

PR #346, opened 2023-01-24
Fix tap caching

shkoo:
This makes it so verify can run more than one verify per VerifyHal instance on the same circuit in the case where the different proofs have different mix values.


jbruestle (approved):

risc0/zkp/src/verify/mod.rs
```diff
@@ -132,22 +131,24 @@ mod host {
     };
 
     struct TapCache<F: Field> {
+        taps: *const TapSet<'static>,
```
flaub, line 134: Is there a way to use a borrow rather than a pointer?

PR #348, opened 2023-01-25
risc0-sys crate / HAL refactor

mothran:
Move all the native (.cpp/.cu/.h/.metal) code for different HAL layers into its own crate [ `risc0-sys` ]. This breaks some of the dependency links that trigger cargo rebuilds while editing different core crates in the tree and hopefully will increase development iteration speed.

This also paves the way for using `cargo clippy` as part of our development / CI pipeline.


risc0/bootstrap/Cargo.toml
```diff
@@ -5,3 +5,6 @@ edition = "2021"
 
 [dependencies]
 risc0-zkvm = { version = "1.0.0-rc.2", path = "../zkvm" }
+
+[features]
+cuda = ["risc0-zkvm/cuda"]
```
flaub, line 10: I'm not sure why this is necessary

mothran: It was not, removed in merging main into this branch.

mothran:
Moving to ready-for-review, I would appreciate some IRL testing with people's different IDE / editing experience to see if they run into re-builds.

flaub (approved):
Nice work!

PR #350, opened 2023-01-25
Minor build system fixes upstreamed from the risc0-sys crate work

mothran:
* Enabled rebuilds if the `RISC0_SKIP_BUILD` env var changes
* Moves metal builds behind a feature flag vs defaulting on `macos` platforms.


flaub:
I don't see an update to the `risc0-circuit-rv32im`

flaub (approved):

mothran:
@nategraf I think this resolves https://github.com/risc0/risc0/pull/323

Issue #356, opened 2023-01-28
Broken method risc0_zkvm::guest::sha::digest


nategraf:
Based on the test results from this PR https://github.com/risc0/risc0-rust-examples/pull/45 (CI run: https://github.com/risc0/risc0-rust-examples/actions/runs/4028293123/jobs/6925018354), the risc0_zkvm::guest::sha::digest appears to be broken. I confirmed this and was able to work around it by using a different method, but this one should be fixed (or removed).

flaub:
Can you provide details on what symptoms you are seeing and how reproduce the issue?

nategraf:
Ah, yes. The hash result from the method was incorrect. I did not, today, investigate further into the issue, but I plan to on Monday

pdg744:
Someone sent me a message on discord that I think is related to this issue: 

<img width="543" alt="Screen Shot 2023-01-30 at 8 41 52 AM" src="https://user-images.githubusercontent.com/15272444/215508043-479f1a3d-a1e7-45ec-b772-8e0fddb98098.png">

![image](https://user-images.githubusercontent.com/15272444/215507852-0a9a5e76-bc59-465b-a34f-6c51ed3977d5.png)

nategraf:
So I figured out that there _is_ a bug in this method, with the bit counting for the trailer, although this is not why the only reason why the example was failing. Opened https://github.com/risc0/risc0/pull/357 to address this issue.

nategraf:
> Someone sent me a message on discord that I think is related to this issue:

This issue looks unrelated. If you'd like to direct them to me I'd be happy to help though.

flaub:
> > Someone sent me a message on discord that I think is related to this issue:
> 
> This issue looks unrelated. If you'd like to direct them to me I'd be happy to help though.

We could also have them (or us) file a new issue for this one, but my guess is that this is related to an improper mixing of v0.11 and v0.12 together.

PR #357, opened 2023-01-30
Fix bit counting in risc0_zkvm::guest::zkvm::sha::digest

nategraf:
As discussed in https://github.com/risc0/risc0/issues/356, the SHA-256 example started to fail it's
test during the update to risc0 v0.12. There are two causes for this.

1. There is an issue with the bit count where the message is serialized to a `buf` of u32 words, but
   the bit count is calculated as `buf.len() * 8`.
2. The semantics of of the `digest`, which in https://github.com/risc0/risc0-rust-examples/pull/45
   was originally used to replace a usage of `digest_u8_slice` method, actually returns a hash of
   the serialization of a given struct rather than directly hashing it's bytes. This is an issue in
   usage and not the `digest` method itself, so I added a comment to clarify this and hopefully
   avoid other folks making the same mistake.

Fixes: https://github.com/risc0/risc0/issues/356


nategraf:
Note, I marked this as a draft until I ensure there is a test that will cover this function. Will do that today.

flaub (approved):
LGTM, but would like @shkoo to take a pass as well

shkoo (commented):
This seems like a whole lot of copies in the guest to me.  I kind of think we should just get rid of the fn digest<T: Serialize>(...).  It doesn't seem very useful to serialize something, and then only use that serialization for generating a digest.

nategraf:
> This seems like a whole lot of copies in the guest to me.  I kind of think we should just get rid of the fn digest<T: Serialize>(...).  It doesn't seem very useful to serialize something, and then only use that serialization for generating a digest.

Sounds good to me. It's not clear to me that this is a API folks need, and it is somewhat easy to misuse (e.g. feeding a string in to it and not getting the expected result). I'd be happy to remove it.

flaub:
This seems fine to me as well.

nategraf:
Removed this method in 2af13f11e365abb9381a1be3bb6e65eea96f49fd to be included in https://github.com/risc0/risc0/pull/330

PR #359, opened 2023-02-01
Add riscv-tests and fix divide extern

flaub:
https://github.com/riscv-software-src/riscv-tests


jbruestle (approved):
Of course at some point it would be nice to actually build the tests rather than include them by binary, but better to have tests than not!

Issue #328, opened 2023-01-11
Support for C in guests?


matthiasgoergens:
I tried adding some C code to my guest, but when I do `cargo build` I get:

```
methods-guest:   running: "cc" "-O3" "-ffunction-sections" "-fdata-sections" "-fPIC" "-march=rv32im" "-mabi=ilp32" "-mcmodel=medany" "-Wall" "-Wextra" "-o" "/Users/matthias/zk/risc0-rust-starter/target/debug/build/methods-dd8a3f9b912de31d/out/riscv-guest/riscv32im-risc0-zkvm-elf/release/build/methods-guest-6bec23e2e3313bc4/out/src/bin/fib.o" "-c" "src/bin/fib.cpp"   
methods-guest:   cargo:warning=clang: error: the clang compiler does not support '-march=rv32im'   
```

Do we have any guide for how to add C or C++ to guests?

flaub:
There was an earlier attempt to add support here: https://github.com/risc0/risc0/pull/205

matthiasgoergens:
Thanks!  I'm trying to resurrect that attempt.  But it's not trivial.

PR #363, opened 2023-02-02
Remove mention of non-existent C support (328)

matthiasgoergens:
https://github.com/risc0/risc0/issues/328 says that C support doesn't work at the moment.


matthiasgoergens:
Completed?

flaub:
The prover supports running any code that is compiled to `rv32im`. I think it's valid to say that we support these languages but there might be more or less batteries included.

PR #365, opened 2023-02-03
Update the CUDA components with the new buffers

mothran:
Updated CUDA features to build and pass unit tests with `cargo test -F cuda -- --test-threads=1`.  Also addressed an issue with the cpp builds failing to link while using the HAL cache env var.


SchmErik:
> Updated CUDA features to build and pass unit tests with `cargo test -F cuda -- --test-threads=1`. Also addressed an issue with the cpp builds failing to link while using the HAL cache env var.

Is it possible to add `cargo test -F cuda` to CI? Not sure exactly what our current set up is...

flaub (approved):

flaub:
Superseded by #378

PR #366, opened 2023-02-03
[NFC] Start of support for multiple hashes.

jbruestle:
Note: this also fixes build issues for Metal, which no longer runs anyway due to running out of registers.


jbruestle:
Abandoned in favor of another approach.

PR #373, opened 2023-02-09
Migrate SHA and factors examples

tzerrell:
Add the `sha` and `factors` examples from https://github.com/risc0/risc0-rust-examples to this repository, as a proof of concept for merging all our examples into the main repository.

As currently drafted, this PR uses a single workspace for both examples. This will help avoid a profusion of workspaces. However, it leads to an awkward name collision with the `methods` crate -- this PR uses `factors-methods` and `sha-methods`. Probably this is acceptable but I wanted to flag it as a potential point of confusion for people looking at our examples. See the [`8d6ba26`](https://github.com/risc0/risc0/commit/8d6ba269a384ec663f71f368c43be3f1974cb4c1) commit for details on this merge.


examples/factors/rust-toolchain
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
flaub, line 1: I think this file need to live wherever the workspace lives

examples/factors/Cargo.lock
```diff
@@ -0,0 +1,1856 @@
+# This file is automatically @generated by Cargo.
```
flaub, line 1: I'm surprised to see this file here, I would think it would live next to the workspace Cargo.toml

examples/factors/.gitignore
```diff
@@ -0,0 +1,4 @@
+.DS_Store
```
flaub, line 1: I don't think this is necessary, we have a top-level `.gitignore` which should do that right thing

tzerrell: replaced with a `.gitignore` that expects the `example` lock files

examples/factors/Cargo.lock
tzerrell: Yep, this file isn't used in this location; removed.

examples/factors/rust-toolchain
tzerrell: It seems to be happy with this file removed altogether (presumably it's using the top-level toolchain file)

flaub: I think it needs to be alongside the workspace. One way to test is to blow away your `target` dir and perhaps tell `rustup` to remove the target triple also.

examples/.gitignore
```diff
@@ -0,0 +1,4 @@
+!factors/Cargo.lock
```
flaub, line 1: This entry doesn't make sense to me, because with a workspace, we only have a single `Cargo.lock` file for the whole workspace.

examples/.gitignore
```diff
@@ -0,0 +1,4 @@
...
+!factors/methods/guest/Cargo.lock
```
flaub, line 2: This is fine, because it's actually an isolated workspace (see the `[workspace]` section, that's what makes it think its a workspace).

examples/factors/rust-toolchain
tzerrell: Gotcha. Blowing away the `target` dir didn't cause problems, but I didn't have `rustup` remove the target triple.

examples/.gitignore
flaub: Actually, I think the contents of this file could maybe be:
```
!Cargo.lock
```

Which should recursively allow any `Cargo.lock` file

tzerrell: Thanks for the catch! I did fixes in the wrong order and confused myself about which lock files would actually exist if I was cleaning up correctly :)

tzerrell:
Now that I've drafted up #374, I think this approach is better. Although disambiguating `methods` as `factors-methods` and `sha-methods` isn't ideal, I think splitting out the guest methods from the individual examples and putting them in a shared methods directory is more confusing. This PR's version also makes the examples further from the starter template code, creating more changes we need to explain in every example tutorial.

flaub:
If we had a tool for generating a starter project instead of a template repository, would your thinking on this change?

tzerrell:
> If we had a tool for generating a starter project instead of a template repository, would your thinking on this change?

I don't think so. I think these are orthogonal concerns, in that whatever code was generated by our tool would still be structured differently from the examples as written in #374. The key difference is that right now and in #373, if someone writes code starting from the starter template and following our tutorial in https://www.risczero.com/docs/examples/hello_multiply, they can compare to the contents of https://github.com/risc0/risc0-rust-examples/tree/main/factors to see if they did thing right and how they differ from the intended approach. In #374, there is no folder to point them at for this without also saying "and ignore these other examples files and these `members`/`methods` entries in some Cargo.toml files. Whether they start coding from a starter template or from tool-generated starter code, their end result still won't look like our published examples if we go with #374.

My other concern is the one I mentioned in #374 -- that people going to the examples directory will be confused about which folders are examples. Someone coming to look at examples will see e.g.
```
chess
factors
methods
sha
voting-machine
```
and one of those things isn't an example, but that's not obvious. We can alleviate this some with a README, but of course GitHub puts the directory contents above the README, so this is imperfect.

flaub:
Would it make sense to adjust the tutorial and/or the starter so that it matches this?

tzerrell:
> Would it make sense to adjust the tutorial and/or the starter so that it matches this?

I don't know how we could make the tutorial "match this" because I'm worried about the final state of the tutorial code, not its initial state. Fundamentally, unless the tutorial literally involves creating multiple examples, someone following the tutorial will not be able to compare what they create to our examples without understanding our code well enough to know what to ignore, which probably involves telling them "when you go to compare to the example, here is what _is_ the example, and here is what to ignore".

(I did previously have concerns about initial state mismatch too, but thanks to your feedback those concerns are now addressed.)

As for my concern about people going to the example directory not knowing which things are examples: This doesn't involve tutorial/starter code, so how well they do/don't match has no effect on it.

examples/.gitignore
```diff
@@ -0,0 +1 @@
+!Cargo.lock
```
matthiasgoergens, line 1: You might want to make sure this one ends in a newline?

tzerrell:
Using the approach in #374 instead.

PR #374, opened 2023-02-09
Migrate factors & sha examples with shared methods

tzerrell:
A variation on #373 with a shared methods folder and crate. Now edited with updates based on discussion with @flaub on how to unify the methods crate, not just directory.

Relative to #373, this we no longer need the `factors-methods` vs `sha-methods` disambiguation -- now both are part of `methods` and look similar to how they did as standalone examples.

The crate and directory structure is more different from standalone examples than it was in #373, however. Instead of a structure matching https://github.com/risc0/risc0-rust-starter, the host code is pulled up to the root directory, and the guest code now lives in a shared methods folder. This also means that if someone goes straight to this examples directory from our getting started page, they will be looking at
```
examples
-- factors
-- methods
-- sha
```
which IMO is not very clear about what examples we have (or even how many examples we have). Perhaps we could wrap the `methods` directory in a `common` directory to make it clearer what's happening. Fundamentally though, I'm finding this split structure hard to explain and likely to confuse newcomers. I do not relish adding explanation to our first tutorial clarifying which parts of the examples directory are and are not part of the factors example. This isn't unworkably confusing, but it seems substantially more confusing than deconflicting the `methods` name by calling it `factors-methods` or `sha-methods`.

I think we should land #373 instead.


examples/methods/sha/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "methods-guest"
```
flaub, line 2: I think you can use `src/main.rs` if also adjust the name of this crate so that it matches the binary name you expect.

examples/factors/factors/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
flaub, line 1: Could also merge this directory up to its parent, so it goes:
```
examples/factors/Cargo.toml
examples/factors/src/main.rs
```

.github/workflows/main.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           echo "$HOME/.local/bin" >> $GITHUB_PATH
       - run: cargo test -F $FEATURE
       - run: cargo test -F $FEATURE --tests -- --ignored
-      - run: cargo test -F $FEATURE --manifest-path examples/Cargo.toml
+      - run: cargo test --manifest-path examples/Cargo.toml
```
flaub, line 74: I think we should continue to test GPU support on all examples.

flaub (approved):

flaub:
Is this ready to land? I'd like to get some test coverage in for the upcoming release

tzerrell:
> Is this ready to land? I'd like to get some test coverage in for the upcoming release

Do we want all examples in and tested? If so then no, if it's ok to do a partial land and then finish later then yes it can land.

PR #375, opened 2023-02-10
Remove the COMMIT region and hash in-place when possible

nategraf:
Currently, data committed by the guest (i.e. passed to `env::commit`) is serialized into the COMMIT
region, a 32 MB region of memory set aside for this purpose. During guest finalization, this region
is hashed and the digest is written to the output global.

This PR introduces a `CommitHasher` which serializes data into a fixed-width buffer, and compresses
the data into an ongoing hash each time the buffer fills. When given a large byte stream, it is also
able to hash the data inplace without copying.

## WIP

In the current implementation, this PR changes the nature of the commit digest. It replaces the
variant of the SHA-256 hashing that was used before with the standard hash and always hashes the
committed data instead of only when it exceeds the length of a digest. This is backwards
incompatible and it needs to be decided whether this what we want to do or if this should be
changed.

@flaub, if you could provide your thoughts here, that would be helpful.


risc0/zkvm/src/lib.rs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![feature(slice_as_chunks)]
```
flaub, line 15: I think this is only possible on unstable/nightly versions of rustc. I'd like to avoid using since eventually it'd be good to work on all stable features.

risc0/zkvm/src/guest/env.rs
```diff
@@ -158,51 +163,24 @@ impl Env {
     }
 
     fn commit<T: Serialize>(&mut self, data: &T) {
+        // NOTE TO REVIEWERS: This removes the send to the STDOUT channel. It will break
+        // anybody who what relying on that. It's possible send the data to
+        // STDOUT in chunks from the commit hasher if this is important.
```
flaub, line 168: I do feel like there are some examples that use this feature, at least the battleship example was using it (IIRC).

risc0/zkvm/src/lib.rs
nategraf: Sounds good. I'll refactor to avoid this method.

risc0/zkvm/src/guest/env.rs
nategraf: So a challenge with this is going to be that in order to use a fixed buffer, we can only provide a "chunked" view of the output, sending at most 512-bits at a times to the STDOUT channel. I imagine this would be surprising to users who are expecting a single call to `env::commit` to yield a single message.

What do you think about having it be the case that `env:write` works as it does now, serialized the message into one contiguous buffer and sending it in one message to `STDOUT`, but have `env::commit` no longer send to `STDOUT`? In this case, folks that need to receive data over the `STDOUT` channel will use `env::write` possibly sending the data both to commit and output.

nategraf: (Oh course, if this is a key feature, I can work on having the chunks reassembled by the host as to given the client a contiguous buffer for each message)

nategraf:
Changes in this PR were rolled into https://github.com/risc0/risc0/pull/391

Issue #376, opened 2023-02-10
Panic and oom handlers cause issues


austinabell:
Currently, the two handlers setup for no_std environments can cause issues.

https://github.com/risc0/risc0/blob/e6e647e444882c8ec5d8ca97bbe4d5a85f9d168b/risc0/zkvm/src/guest/mod.rs#L87-L101

The reason for this is the following:
- Using formatting by default can cause unnecessary bloat for someone wanting to minimize their bytecode and doesn't want to include core::fmt machinery (it's bulky) https://github.com/risc0/risc0/blob/e6e647e444882c8ec5d8ca97bbe4d5a85f9d168b/risc0/zkvm/src/guest/mod.rs#L93
  - Unsure the type of errors for this architecture, but should be able to downcast the error as per the docs (https://doc.rust-lang.org/core/panic/struct.PanicInfo.html#examples) to avoid using, or at very least allow devs to override to what they want
- The alloc error handler is enabled by default, but if someone wanted to write a no alloc program, this would be unnecessary (yes, right now you can't do that anyway since it's being pulled in by the lib, but that should ideally be optional) https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/guest/mod.rs#L97
  - This forces any user to also use the nightly toolchain, even if they don't alloc and use this feature https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/lib.rs#L17
  - Also, someone very picky might not want to include that static string and just abort https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/guest/mod.rs#L99
- You cannot customize the implementations of either of these

My intuition is that the prover and guest sides should not be merged into the same crate, to allow the guest crate to be as minimal as possible and not bring in unnecessary components, these should be two distinct binaries/programs. I assume this would also mean splitting any shared primitives into a shared crate. Also, with this, there should be features gating these two individually from being included, and can be enabled by default.

I will open a PR with the latter without the refactor/splitting of crates, but I don't think it should be pulled in without the former to avoid churn without solving the core problem.

Let me know if you'd like me to open a separate issue for splitting up the zkvm crate.

flaub:
[...]

PR #377, opened 2023-02-10
put panic and error handlers behind feature flag

austinabell:
Context for this in #376 

In summary, this would be helpful, but my intuition is this is not worth bringing in before a refactor, which is why I'm opening as a draft


austinabell:
Forgot about this. I don't have a plan to update this unless agreed that this should be under a feature flag, so I'll just close the PR for now

PR #379, opened 2023-02-11
Add support for the Poseidon hash and its use by the CPU + Metal Hals for proving and verifying

jbruestle:


jbruestle:
@nategraf At some point (let's see when it fits into your schedule) it would be good to get a very careful audit on the Poseidon code, since I don't have any test vectors for Baby Bear Poseidon, since we are the only ones who use it.

risc0/zkp/Cargo.toml
```diff
@@ -17,6 +17,7 @@ bytemuck = { version = "1.12", features = ["derive"] }
 digest = "0.10"
 fil-rustacuda = { version = "0.1", optional = true }
 hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
+lazy_static = { version = "1.4.0", default-features = false, features = ["spin_no_std"] }
```
flaub, line 20: Needed?

risc0/bootstrap/src/main.rs
```diff
@@ -12,11 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use risc0_zkp::hal::cpu::{BabyBearPoseidonCpuHal, BabyBearSha256CpuHal};
 use risc0_zkvm::Loader;
+// use risc0_zkp::hal::Hal;
```
flaub, line 17: Drop this

risc0/zkp/src/hal/metal.rs
```diff
@@ -45,16 +50,136 @@ const KERNEL_NAMES: &[&str] = &[
...
+    }
+}
+
 #[derive(Debug)]
-pub struct MetalHal {
+pub struct MetalHal<Hash: MetalHash + ?Sized> {
     pub device: Device,
     pub cmd_queue: CommandQueue,
     kernels: HashMap<String, ComputePipelineDescriptor>,
+    pub library: Library,
```
flaub, line 176: Is this still needed? I'm not immediately seeing who uses this externally?

risc0/zkp/Cargo.toml
jbruestle: Yes, for the Poseidon constants since some of them are calculated.

risc0/zkp/src/hal/metal.rs
jbruestle: So I already know third parties are interested in building their own hashes, so I was considering that use case.  Although, I suppose they don't probably need access to the metal library itself, so I'll drop.

PR #386, opened 2023-02-15
Add example tests & GPU support

tzerrell:
Add tests for all current examples and add GPU feature flags for all current examples.


flaub (approved):
Nice!

PR #389, opened 2023-02-17
zkvm: add sys_rand to generate random numbers for the guest

SchmErik:
sys_rand fills a u8 slice with random numbers. This helps enable rust's getrandom crate. While this adds non-determinism to the zkvm, it does not violate any security guarantees of a stark-based system.


flaub:
Unfortunately this breaks the WASM build. We might need to tweak the feature flags or add `#[cfg]` to make this work.

flaub:
I think we need a test that uses this new syscall

mothran:
Notes for adding a custom impl: https://docs.rs/getrandom/0.2.8/getrandom/macro.register_custom_getrandom.html

SchmErik:
@flaub I wrote this code for the test:
```
    let buf = [0u8; 32];
    unsafe { sys_rand(buf.as_ptr(), buf.len()) }
```

And it results in an eqz error like this:
```
---- tests::do_random stdout ----
thread 'tests::do_random' panicked at 'called `Result::unwrap()` on an `Err` value: eqz failed at: cirgen/components/ram.cpp:94', /Users/erik/risc0/risc0/zkp/src/prove/executor.rs:194:22
```
However, when I do things like wrap it in a function or do an `env::commit(...)` after this, the error disappears. I'm going to look at the assembly I wrote more closely but have you run into anything like this before?

SchmErik:
@flaub could you take another look?

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -263,3 +268,25 @@ pub unsafe fn sys_sha_buffer(
     #[cfg(not(target_os = "zkvm"))]
     unimplemented!()
 }
+
+#[inline(always)]
+pub unsafe fn sys_rand(recv_bytes: usize) -> &'static [u8] {
```
flaub, line 273: I think we should pass a pointer to be filled in here at this level (the syscall level). This would allow the use of custom allocators and doesn't assume any allocations in `risc0-zkvm-platform`.

flaub: Also, using a pointer sets us up for potential use from other languages.

risc0/zkvm/methods/guest/src/bin/random_number.rs
```diff
@@ -0,0 +1,29 @@
...
+use risc0_zkvm::guest::env;
+use risc0_zkvm_platform::syscall::sys_rand;
+
+risc0_zkvm::entry!(main);
+
+pub fn main() {
+    unsafe {
+        let rbuf = sys_rand(5);
+        env::commit_slice(&rbuf);
+        // assert_eq!(rbuf, [0u8; 10])
```
flaub, line 27: I think we should add this to the existing multi_test binary.

SchmErik: ah, I forgot about that, thanks!

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -29,6 +29,7 @@ pub mod nr {
     pub const SYS_LOG: u32 = 1;
     pub const SYS_IO: u32 = 2;
     pub const SYS_CYCLE_COUNT: u32 = 3;
+    pub const SYS_RAND: u32 = 4;
```
flaub, line 32: Do we need a separate `SYS_RAND`? I was thinking we could build `getrandom` on top of `SYS_IO` with the `SENDRECV_CHANNEL_RANDOM` channel.

SchmErik: Thanks for catching this! What you're suggesting is what I intended to do and this was left over from previous drafts. I'll remove it.

SchmErik: Thanks! I made this adjustment

flaub (approved):

PR #390, opened 2023-02-17
Add "copy-in" to rv32 circuit

shkoo:
* New major cycle type copies in data from the host at 4 words per cycle
* SOFTWARE ecall type now takes: a0: destination buffer pointer a1: destination size in chunks (words / 4, bytes / 16) and can modify a0 and a1 to return data in registers. It uses the COPY_IN cycle, and calls the following externs: syscall-init: Tells the host to start a syscall syscall-body: Gather the next 4 words to transfer from the host to the guest syscall-fini: Gather the 2 register outputs from the host
* SYS_IO takes the following additional args: a2: source buffer pointer a3: source buffer size in bytes a4: channel id SYS_IO returns arbitrary data from the host in a0 and a1 for data that's small enough to not need a buffer allocated.  (This matches return values in the riscv calling conventions to minimize overhead)
* Removed SYS_COMPUTE_POLY; functionality replaced by SENDRECV_CHANNEL_COMPUTE_POLY via SYS_IO  (Hopefully we can eventually get rid of this)
* Removed SYS_COMMIT; functionality replaced by SENDRECV_CHANNEL_COMMIT via SYS_IO
* Updated guest::env for new SYS_IO API
* Added test to make sure lines-of-code in eval_check/poly_fp doesn't significantly grow unexpectedly


flaub (changes requested):
Let's hold off on this until #391 lands, and then revisit this after.

flaub:
I'd like to get 0.13 out before we make any more big changes. I'm attempting to get all the little PRs landed for the release.

flaub:
I think I'd like to get some benchmarking into CI before landing anymore major changes to circuits.

shkoo:
As we discussed, it sounds like a lines-of-code check for generated eval check code is sufficient in lieu of full benchmarking CI, so that's included with this PR.

shkoo:
Fib benchmark compared to main showing an increase in performance, albeit a pretty small one:

```
$ critcmp main new
group            main                                    new
-----            ----                                    ---
fib/100/proof    1.01  1594.2±176.07ms 20.0 KElem/sec    1.00  1581.9±225.90ms 20.2 KElem/sec
fib/100/run      1.02   367.1±13.10ms 87.0 KElem/sec     1.00    361.3±8.05ms 88.4 KElem/sec
fib/200/proof    1.02  1515.6±97.79ms 21.1 KElem/sec     1.00  1487.8±178.48ms 21.5 KElem/sec
fib/200/run      1.01   373.1±14.83ms 85.6 KElem/sec     1.00   368.5±38.47ms 86.7 KElem/sec

```

risc0/zkvm/src/prove/exec.rs
```diff
@@ -367,11 +333,18 @@ impl<'a, H: HostHandler> CircuitStepHandler<BabyBearElem> for MachineContext<'a,
                 self.log(extra, args);
                 Ok(())
             }
-            "syscall" => {
-                (
-                    (outs[0], outs[1], outs[2], outs[3]),
-                    (outs[4], outs[5], outs[6], outs[7]),
-                ) = self.syscall(cycle)?;
+            "syscall-init" => {
```
flaub, line 336: To be consistent, we could use `camelCase` here

risc0/zkvm/src/guest/env.rs
```diff
@@ -96,16 +97,41 @@ pub(crate) fn finalize() {
...
-pub fn send_recv_as_u32(channel: u32, buf: &[u8]) -> &'static [u32] {
-    bytemuck::cast_slice(send_recv(channel, buf))
+/// Exchange data with the host.
+pub fn send_recv_sized(
+    channel: u32,
+    to_host_buf: &[u8],
+    from_host_bytes: usize,
+) -> (u32, u32, &'static [u8]) {
+    unsafe {
+        let nwords = (from_host_bytes + WORD_SIZE - 1) / WORD_SIZE;
```
flaub, line 112: I think we have an `align_up` function for doing integer rounding up

risc0/zkvm/src/guest/env.rs
```diff
@@ -96,16 +97,41 @@ pub(crate) fn finalize() {
...
+            slice::from_raw_parts(from_host_buf.cast(), from_host_bytes),
+        )
+    }
+}
+
+/// Exchange data with host.  The channel should return the length of
+/// the data to read when presented with an empty output buffer.
+pub fn send_recv(channel: u32, to_host_buf: &[u8]) -> &'static [u8] {
+    let (nbytes, _) = send_host(channel, to_host_buf);
+    let (_, _, bytes) = send_recv_sized(channel, to_host_buf, nbytes as usize);
```
flaub, line 133: It seems like if we're going to have this usage pattern, would it make more sense to have 2 different SYS_TX and SYS_RX syscalls rather than combining into SYS_IO?

risc0/zkvm/src/guest/env.rs
```diff
@@ -96,16 +97,41 @@ pub(crate) fn finalize() {
...
-    bytemuck::cast_slice(send_recv(channel, buf))
+/// Exchange data with the host.
+pub fn send_recv_sized(
+    channel: u32,
+    to_host_buf: &[u8],
+    from_host_bytes: usize,
+) -> (u32, u32, &'static [u8]) {
+    unsafe {
+        let nwords = (from_host_bytes + WORD_SIZE - 1) / WORD_SIZE;
+        let from_host_buf = zkvm_abi_alloc_words(nwords);
```
flaub, line 113: I wonder if we should let the user decide where the buffer to be filled comes from? Maybe they want to use their own custom allocator, or for it live in a local/stack variable?

flaub (approved):

risc0/zkvm/src/guest/env.rs
shkoo: Yeah, pondering this for a bit I think this is kind of confusing and limiting also.

Anyways, I redid the public facing API parts of this PR and separated out "send_recv_raw" from "send_recv_slice".  I think this new way is cleaner and easier to understand and use; can you please take another look?

Issue #381, opened 2023-02-14
Serialization and Deserialization of &[u8] Not Working as Expected


jjtny1:
Given a simple guest that does the following:
1. reads a `&[u8]` from the environment
2. commits the same `&[u8]` back to the environment

The following issues occur:
1. Calling `prover.add_input_u8_slice(...)` with the following bytes: `b922f47c00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001` causes a `DeserializeUnexpectedEnd` when `prover.run()` is called
2. Using `prover.add_input_u32_slice(...)` and using `risc0_zkvm::serde::to_vec` to serialize the `&[u8]` first, stopped the error above, but attempting to deserialize the journal into a `&[u8]` gave the wrong bytes. I made a minimal reproducible test [here](https://github.com/risc0/risc0/commit/08c3da1df3c40e997839165ef19a23176fcba38d)

Changing everything to use `Vec<u8>` instead of `&[u8]` worked as expected.

austinabell:
So this is somewhat because, by default, serde serializes `&[u8]` and `Vec<u8>` as sequences of u8 rather than as bytes. You could have these types be handled at runtime, similar to how [borsh handles these types](https://github.com/near/borsh-rs/blob/9e6831628442036fe3183f8244da3973bd99e8f7/borsh/src/ser/mod.rs#L224-L230) but that comes with its tradeoffs. In the repro, it serializes each byte as an independent u32 word.

This doesn't explain the inconsistency of deserializing as `Vec<u8>` and `&[u8]`, however. `Vec<u8>` is handling it as "expected" where each u8 is deserialized from each u32 word 1:1. The reason deserializing as `&[u8]` doesn't work, is that [serde deserializes it as bytes](https://github.com/serde-rs/serde/blob/0e947e6c3b2ff240e15c47fcf76e3bed5c099d0b/serde/src/de/impls.rs#L663) rather than [as a sequence](https://github.com/serde-rs/serde/blob/0e947e6c3b2ff240e15c47fcf76e3bed5c099d0b/serde/src/de/impls.rs#L1027-L1043) as `Vec<u8>` is handled as.

So, TLDR, this is behaving as expected based on `serde`'s API, but you can modify the serializer to handle this case specifically at runtime and serialize bytes more efficiently. (Also, it seems it would be inline with the zeroio serialization stuff that exists in this repo, but I haven't read through that yet)

PR #391, opened 2023-02-18
Fix for #381

flaub:
This continues work from https://github.com/risc0/risc0/pull/375

* Drop the COMMIT & OUTPUT regions
* Increase HEAP & TEXT from 32MB to 64MB each
* Add `env::read_slice` for reading raw data from the host.
* Add `env::write_slice` for writing raw data to the host.
* Add `env::commit_slice` for committing raw data to the journal.
* Replace `SYS_COMMIT` with a `SYS_IO` call using a new `SENDRECV_CHANNEL_JOURNAL` channel.
* Allow empty journals to compare to either all zeros or the hash of an empty string.


jbruestle (approved):
I tried to read most of it, definitely some good clean up, didn’t see any issues.

PR #393, opened 2023-02-21
Expand examples in documentation

tzerrell:
Expand the Rust documentation with more examples for Prover, Receipt, ans SHA.


risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,51 @@
...
+//!     .expect("Prover should be constructed from an ELF file and matching Image ID");
+//! let receipt = prover
+//!     .run()
+//!     .expect("Prover should run guest code that doesn't crash to generate a receipt");
+//! ```
+//! Once you have a Receipt, you can use the verify function to
+//! cryptographically verify that it was generated faithfully and is associated
+//! with the expected image ID. You can also read the contents of the journal.
+//! ```
+//! # #![feature(slice_flatten)]
```
flaub, line 34: I'm not sure this is going to work on `docs.rs`

risc0/zkvm/src/prove/mod.rs
```diff
@@ -13,6 +13,24 @@
...
+//! [Prover]. See the [Prover] documentation for more detailed usage
+//! information.
+//!
+//! ```ignore
+//! // In real code, the ELF & Image ID would be generated by risc0 build scripts from guest code
+//! use methods::{EXAMPLE_ELF, EXAMPLE_ID};
+//! use risc0_zkvm::Prover;
+//!
+//! let mut prover = Prover::new(&EXAMPLE_ELF, EXAMPLE_ID)
+//!     .expect("Prover should be constructed from an ELF file and matching Image ID");
```
flaub, line 28:
```suggestion
//!     .expect("Specified ELF is invalid");
```

risc0/zkvm/src/prove/mod.rs
```diff
@@ -13,6 +13,24 @@
...
 
 //! Run the zkVM guest and prove its results
+//!
+//! # Usage
+//! The primary use of this module is to provably run a zkVM guest by use of a
+//! [Prover]. See the [Prover] documentation for more detailed usage
+//! information.
+//!
+//! ```ignore
+//! // In real code, the ELF & Image ID would be generated by risc0 build scripts from guest code
```
flaub, line 23:
```suggestion
//! // In real code, the ELF & Image ID would be generated by `risc0-build` scripts for the guest code
```

risc0/zkvm/src/prove/mod.rs
```diff
@@ -127,6 +145,40 @@ impl<'a> Default for ProverOpts<'a> {
...
+/// ```
+/// Provers should essentially always be mutable so that their [Prover::run]
+/// method may be called.
+///
+/// Input data can be passed to the Prover with [Prover::add_input_u32_slice]
+/// (or [Prover::add_input_u8_slice]). After all inputs have been added, call
+/// [Prover::run] to execute the guest code and produce a [Receipt] proving
+/// execution.
+/// ```ignore
+/// prover.add_input_u32_slice(&to_vec(&input)?);
```
flaub, line 170: I don't think this compiles, because `to_vec` isn't qualified and there's no corresponding `use`.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -127,6 +145,40 @@ impl<'a> Default for ProverOpts<'a> {
...
+/// ```
+/// After running the prover, publicly proven results can be accessed from the
+/// [Receipt], while private outputs can be accessed using
+/// [Prover::get_output_u32_vec] (or [Prover::get_output_u8_slice]):
+/// ```ignore
+/// let receipt = prover.run()?;
+/// let slice = prover.get_output_u32_vec()?;
+/// let private_output: OutputType = risc0_zkvm::serde::from_slice(&slice)?;
+/// let proven_result: ResultType = risc0_zkvm::serde::from_slice(&receipt.journal)?;
+/// ```
```
flaub, line 181: It's important to mention the `pub cycles` member that can be used by the host to determine how many cycles the zkvm executed. It's not part of the receipt because it should be private to only the host.

flaub: Also missing is docs for the `pub fn default_hal()`

risc0/zkvm/src/prove/mod.rs
```diff
@@ -185,20 +239,48 @@ impl<'a> Prover<'a> {
...
 
+    /// Provide input data to the guest. This data can be read by the guest
+    /// via [crate::guest::env::read].
+    ///
+    /// It is possible to provide multiple inputs to the guest so long as the
+    /// guest reads them in the same order they are added by the [Prover].
+    /// However, to reduce maintenance burden and the chance of mistakes, we
+    /// recommend instead using a single `struct` to hold all the inputs and
+    /// calling [Prover::add_input_u8_slice] just once (on that input
+    /// `struct`).
```
flaub, line 250:
```suggestion
    /// calling [Prover::add_input_u8_slice] just once (on the serialized representation of the input.
```

flaub: One thing to consider is adding helper functions on the `Prover` which makes it easier to pass serialized inputs. Similar to how the guest has `read` vs `read_slice`. We could have `add_input_u8_slice` for the raw version, and `add_input` which would be generalized over a `Serializer`.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -211,6 +293,14 @@ impl<'a> Prover<'a> {
             .collect())
     }
 
+    /// Run the guest code. If the guest exits successfully, this returns a
+    /// [Receipt] that proves execution. If the guest crashes, this instead
+    /// returns an `Err`.
```
flaub, line 298:
```suggestion
    /// [Receipt] that proves execution. If the guest fails for any reason, this instead
    /// returns an `Err`.
```

risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,55 @@
 // limitations under the License.
 
 //! Manages the output and cryptographic data for a proven computation
+//!
+//! # Usage
+//! Receipts are typically constructed as the output of the prover
```
flaub, line 18: Not sure what this means. I think we should talk about the journal and how that is a commitment or attestation that verifiers can access.

risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,55 @@
...
+//! ```ignore
+//! // In real code, the ELF & Image ID would be generated by risc0 build scripts from guest code
+//! use methods::{EXAMPLE_ELF, EXAMPLE_ID};
+//! use risc0_zkvm::Prover;
+//!
+//! let mut prover = Prover::new(&EXAMPLE_ELF, EXAMPLE_ID)
+//!     .expect("Prover should be constructed from an ELF file and matching Image ID");
```
flaub, line 25: I think example `.expect()` should have messages that pertain to possible ways a function can fail.

risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,55 @@
...
+//! let receipt = prover
+//!     .run()
+//!     .expect("Prover should run guest code that doesn't crash to generate a receipt");
```
flaub, line 28: I don't think "crash" is a good term to use, especially in Rust. There are many reasons this could fail, for instance the embedded verification could fail after constructed a proof. Or there could be a constraint failure in the circuit, or perhaps the guest actually did panic.

risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,55 @@
...
+//! ```
+//! Once you have a Receipt, you can use the verify function to
+//! cryptographically verify that it was generated faithfully and is associated
+//! with the expected image ID. You can also read the contents of the journal.
```
flaub, line 32: Might be better to phrase this without 'you'

risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,55 @@
...
+//! receipt.verify(&IMAGE_ID);
+//! let committed_value: String = risc0_zkvm::serde::from_slice(&receipt.journal)
+//!     .expect("Deserialization should succeed if the types match the committed types");
+//! assert_eq!("test", committed_value);
+//! ```
+//! Directly using the raw bytes of the journal is not recommended. Instead, we
+//! recommend using serialization tools from the zkVM [serde](crate::serde)
+//! module for general use, as shown above. We also have a crate
+//! [zeroio](https://docs.rs/risc0-zeroio/latest/risc0_zeroio/) offering an
+//! alternative serialization approach that is better for some use cases.
```
flaub, line 64: Could phrase this without 'we', like: "An alternative crate exists which ..."

risc0/zkvm/src/receipt.rs
```diff
@@ -13,6 +13,55 @@
...
+//! # };
+//! # use crate::risc0_zkvm::sha::Sha256;
+//! # let IMAGE_ID: [u32; 8] = [0, 0, 0, 0, 0, 0, 0, 0];
+//! // Here `receipt` is a Receipt whose journal contains the String "test"
+//! receipt.verify(&IMAGE_ID);
+//! let committed_value: String = risc0_zkvm::serde::from_slice(&receipt.journal)
+//!     .expect("Deserialization should succeed if the types match the committed types");
+//! assert_eq!("test", committed_value);
+//! ```
+//! Directly using the raw bytes of the journal is not recommended. Instead, we
```
flaub, line 60: I don't think we should make recommendations, rather we should point out the pros/cons of various approaches.

risc0/zkvm/src/sha.rs
```diff
@@ -13,6 +13,37 @@
...
+//! We provide two interfaces for SHA hashing.
+//! The first interface is based on [Rust Crypto](https://github.com/RustCrypto) wrappers and can be found in the [rust_crypto] module, along with documentation and usage notes.
+//!
+//! The other interface is to directly use an implementation of the [Sha256]
+//! trait defined in this module:
+//! ```
+//! use risc0_zkvm::sha::{Impl, Sha256};
+//!
+//! // Hash a u8 array
+//! let data: [u8; 4] = [1, 2, 5, 14];
```
flaub, line 33:
```suggestion
//! let data = [1_u8, 2, 5, 14];
```

risc0/zkvm/src/prove/mod.rs
tzerrell: Hmm, I think I'm missing something because a very similar line works fine in the factors example:
https://github.com/risc0/risc0/blob/861c3959d5a06bb1826a31cf0e5345d97251ec58/examples/factors/src/main.rs#L33
Is it that in factors the `a` is a `u64`, and if `input` were a more complex type there might be ambiguity that doesn't exist with a `u64`?

flaub: The factors example has `use risc0_zkvm::serde::to_vec`. I'm saying the example either needs to have a `use` or it needs to fully qualify `to_vec`.

tzerrell: gotcha, thanks!

risc0/zkvm/src/receipt.rs
tzerrell: I think I am going to move the discussion of what leads to `Err` results into the documentation for the relevant functions (so here `Prover::new`) and replace the `expect`s with `?`s. I think that's a more natural place to find such documentation, and anyone who does look for it in these examples and sees the `?` is likely to guess to check the function documentation, whereas the opposite might not be true.

(And then do as you suggest, document ways possible ways for functions that return `Result`s to fail.)

flaub (approved):
Great!

PR #395, opened 2023-02-22
Use tempdir_in OUT_DIR to prevent cross-fs-unlink

mothran:
By creating a tempdir within the $OUT_DIR we can prevent cross-fs unlink issues if someone has /tmp mounted as tmpfs or on a different filesystem.


flaub (approved):

flaub:
You might need to use the same technique here: https://github.com/risc0/risc0/blob/main/risc0/build_kernel/src/lib.rs#L224

mothran:
> You might need to use the same technique here: https://github.com/risc0/risc0/blob/main/risc0/build_kernel/src/lib.rs#L224

Good catch, I will fix that one and do a once over for other tmpdirs that I might have missed.

PR #398, opened 2023-02-23
Benchmark workflows

mothran:
Adding workflow to support basic benchmark'ing on PR and saving benchmark results to a github page.

## TODO
- [ ] Save criterion diff reports between the PR branch and `main` to the benchmarks repo pages (https://github.com/benchmark-action/github-action-benchmark/pull/145 potential fix here, or we customize the action itself)
- [x] Add report diffs into the comment created on the PR with charts.


.github/workflows/bench.yml
```diff
@@ -0,0 +1,54 @@
...
+      FEATURE: ${{ matrix.feature }}
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: risc0/actions-rs-toolchain@v1
+        with:
+          toolchain: stable
+      - run: cargo bench -F $FEATURE --bench fib -- --output-format=bencher | tee output.txt
+      - name: Store benchmark result
+        uses: risc0/github-action-benchmark@7eeba8924d4ed651010d72d7319b064fc0c10355
```
flaub, line 43: Let's pin to a specific version unless you expect certain behavior on a non-released commit?

mothran: Github recommends commit hash over version for actions generally to prevent non-determinism but since we own the forks I am happy to swap to a version here.

flaub (approved):
Nice!!

PR #399, opened 2023-02-23
Add waldo example

flaub:


tzerrell (approved):
Great!

flaub:
The next thing is to get a test running which actually runs the prove/verify. In my testing, it requires lots of time and memory, so maybe this is our first candidate for a nightly test.

PR #401, opened 2023-02-23
Using Blake2b as a hash function

kubaplas:
Implementation of Blake2b hash function. It uses generics to allow supplying different hashers implementations.

### DIFF
--- a/risc0/zkp/Cargo.toml
+++ b/risc0/zkp/Cargo.toml
@@ -13,6 +13,7 @@ harness = false
 
 [dependencies]
 anyhow = { version = "1.0", default-features = false }
+blake2 = { version = "0.10.6", default-features = false }
 bytemuck = { version = "1.12", features = ["derive"] }
 digest = "0.10"
 fil-rustacuda = { version = "0.1", optional = true }
--- a/risc0/zkp/src/core/blake2b.rs
+++ b/risc0/zkp/src/core/blake2b.rs
@@ -1,0 +1,155 @@
+// Copyright 2023 RISC Zero, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A Blake2b HashSuite.
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use core::marker::PhantomData;
+
+use blake2::{
+    digest::{Update, VariableOutput},
+    Blake2bVar,
+};
+use rand_core::{impls, Error, RngCore};
+use risc0_core::field::baby_bear::{BabyBear, BabyBearElem, BabyBearExtElem};
+use risc0_core::field::Elem;
+use risc0_core::field::ExtElem;
+
+use super::config::HashSuite;
+use super::digest::Digest;
+use crate::core::config::{ConfigHash, ConfigRng};
+
+/// Hash function trait.
+pub trait Blake2b {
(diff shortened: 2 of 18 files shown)


risc0/zkp/src/core/blake2b.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.current = next;
+
+        ((next[0] as u32) << 24) +
+            ((next[1] as u32) << 16) +
+            ((next[2] as u32) << 8) +
+            ((next[3] as u32) << 0)
+    }
+
+    fn random_elem(&mut self) -> BabyBearElem {
+        BabyBearElem::new(self.random_u32())
```
jbruestle, line 128: When the BB element is constructed here, the uniformly distribute value x \in [0, 2^32) is moved into the range [0, P) via x % P, which will result in a significantly non-uniform distribution. as some elements appear 3 times in [0, 2^32) % P, and some appear only 2 times.  This non-uniformity breaks the assumptions used in the soundness arguments for the proof system.  You can either use a technique similar to the random_elem implementation in Sha2 which produces something cryptographically close to uniform, or you can use rejection sampling (https://en.wikipedia.org/wiki/Rejection_sampling).  Rejection sampling is probably  easier and more efficient on average in terms of samples drawn, the reason we use an alternate technique in the Sha2 implementation is to make sure the number of samples is always fixed, which matters for certain recursion use cases.

risc0/zkp/src/core/blake2b.rs
```diff
@@ -0,0 +1,139 @@
...
+use super::config::HashSuite;
+use super::digest::Digest;
+
+/// Hash function trait.
+pub trait Blake2bHasher {
+    /// A function producing a hash from a list of u8.
+    fn blake2b<T: AsRef<[u8]>>(data: T) -> [u8; 32];
+}
+
+/// Implementation of blake2b using CPU.
```
jbruestle, line 36: The reason we use a trait for the underlying hasher in the Sha example is to allow the guest to use a special Sha256 acceleration circuit.  You are welcome to similarly abstract Blake2, but you can also probably skip it if that's simpler, since it will probably be a while before we add a Blake2 accelerator.

risc0/zkp/Cargo.toml
```diff
@@ -29,6 +29,7 @@ rustacuda_core = { version = "0.1", optional = true }
 rustacuda_derive = { version = "0.1", optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 tracing = { version = "0.1", default-features = false, features = ["attributes"] }
+blake2 = { version = "0.9.0", default-features = false }
```
jbruestle, line 32: Is there any specific reason you are using an older version of the package?

jbruestle (commented):
Nicely done overall, especially given the lack of documentation on how to add a new hash.  Some specific concerns inline, at least one of which is a blocker (the non-uniform selection of field elements).

risc0/zkp/src/core/blake2b.rs
kubaplas: Thanks for pointing me in the direction I should go, it's really helpful!

risc0/zkp/src/core/blake2b.rs
kubaplas: Isn't it already abstracted in a similar way? We are using it to replace the blake2b implementation with a Casper host one.
I've changed the trait's name to `Blake2b` to reflect this better.

risc0/zkp/Cargo.toml
kubaplas: No, I just happened to have a code snippet written for an earlier version.
I have updated it to the latest and updated the relevant code.

risc0/zkp/src/core/blake2b.rs
jbruestle: Yup, looks good.

risc0/zkp/Cargo.toml
jbruestle: Great.  Fixing that also fixed the additional requirements that got added to cargo locks files

jbruestle (approved):
Looks good!  The CI needs you to sort your deps alphabetially (see https://crates.io/crates/cargo-sort), otherwise should be good.

PR #403, opened 2023-02-25
Add starter template for Bonsai applications

nategraf:
This PR includes a starter tenplate for building applications on [Bonsai]. This template is intended to serve as a starting point for building applications, as well as an example of usage for both the zkVM and Bonsai. It is integrated into the CI for the `risc0` monorepo such that it will continously be tested with each me release of the RISC Zero platform.

See the included README.md for more information about the template.

[Bonsai]: https://dev.bonsai.xyz


nategraf:
It looks like we will need `npm` on the self-hosted machines. Not sure how we will want to manage that (e.g. installed as part of CI or pre-installed om the system).

templates/bonsai/contracts/tests/contract_tests/utils.rs
```diff
@@ -0,0 +1,141 @@
...
+
+pub struct BonsaiMock {
+    pub contract_address: Address,
+    pub task_handle: JoinHandle<()>,
+}
+
+impl BonsaiMock {
+    pub async fn spawn(
+        client: Client,
+        registry: HashMap<[u8; 32], &'static [u8]>,
```
flaub, line 66: Might be good to use a type or at least a way to remove the magic length number.

```suggestion
        registry: HashMap<Digest, &'static [u8]>,
```

or

```suggestion
        registry: HashMap<ImageId, &'static [u8]>,
```

I'm also not sure it's a good idea to use the `'static` lifetime.

flaub (approved):

PR #405, opened 2023-02-27
Drop vulnerable remove_dir_all dep

flaub:
See: https://github.com/risc0/risc0/security/dependabot


mothran (approved):
nice! Did dependabot find this?

PR #407, opened 2023-02-27
One comment per benchmark PR check.

mothran:
Using https://github.com/risc0/criterion-compare-action/pull/5 and artifact upload / downloads we can collapse all the markdown generated by the criterion-compare action into a single comment created per workflow run vs 4.


flaub (approved):

flaub:
Yay!!

PR #408, opened 2023-02-27
Have Bonsai template use path dep and address review comment from #403

nategraf:
Follow-up to #403, this PR switches to using a path dependency for `risc0` such that it reflects the
state of the repo at the current commit.

Additionally, it addresses https://github.com/risc0/risc0/pull/403#discussion_r1118427060 by adding
the `Hash` trait to the `sha::Digest` type, and by using generics to allow for owned vectors to be
used for the to be used for the values in the local ELF registry. (I tested this to make sure
`HashMap::<Digest, Vec<u8>>` workds as well as `HashMap::<Digest, &'static [u8]>`.


templates/bonsai/Cargo.toml
```diff
@@ -8,9 +8,9 @@ members = [
 [workspace.dependencies]
 hello-bonsai-methods = { path = "methods" }
 hello-bonsai-contracts = { path = "contracts" }
-risc0-zkp = { version = "0.13" }
-risc0-zkvm = { version = "0.13" }
-risc0-build = { version = "0.13" }
+risc0-zkp = { path = "../../risc0/zkp" }
+risc0-zkvm = { path = "../../risc0/zkvm" }
+risc0-build = { path = "../../risc0/build" }
```
flaub, line 13: I think we might want both, which will:
1. When used locally, it uses the `path`.
2. When the crate is published, it uses the `version`.

Or perhaps we can arrange for the `cargo risc0` tool to tweak this when it deploys the template.

templates/bonsai/contracts/tests/contract_tests/utils.rs
```diff
@@ -61,9 +62,9 @@ pub struct BonsaiMock {
 }
 
 impl BonsaiMock {
-    pub async fn spawn(
+    pub async fn spawn<Bytes: AsRef<[u8]> + Send + 'static>(
```
flaub, line 65: Is `'static` what we want here or do we want to generalize `spawn` so that it takes a `'a`. I think this is probably fine for tests, but I can imagine scenarios where the registry might not last for the entire lifetime of the program.

templates/bonsai/Cargo.toml
nategraf: Are you saying to have `risc0-zkp = { path = "../../risc0/zkp", version = "0.13" }`? Would that work?

Another option I thought about would be to have this handled by a patch statement down below, making it easy to copy the template out and simply drop the patch statements, but that feels a bit error prone, and we can figure out the right way to do this when we write the `cargo risc0` template generator.

flaub: Yes, you can specify both `path` and `version`, but I'm not entirely sure if that's actually what we want here.

Agreed that maybe best place to solve this is with the generator tool.

templates/bonsai/contracts/tests/contract_tests/utils.rs
nategraf: So a `'static` lifetime bound is required when passing data into a block for `tokio::spawn` (Since the thread is allowed to live for the lifetime of the program). This was new to me, so I learned more about it from https://tokio.rs/tokio/tutorial/spawning#static-bound

I think given the context, the `'static` lifetime bound is correct. This trait bound in particular can be satisfied by `&'static [u8]` or `Vec<u8>`. If the user has a buffer they want to reference that may not last until the end of the program, they can put it into an `Arc` and pass it in IIUC. At least with `tokio`, I think this is the the way of things.

flaub: Thank you! This clears up some misconceptions I had in my head and now I see that we're just specifying a bound, not an annotation to the reference itself.

flaub (approved):

PR #409, opened 2023-02-28
[Review] Use .gitignore in license checker script

nategraf:
Suggested changes to #406 to use `git`, and therefore `.gitignore` to decide which files to look at.

Wrote up this PR when I noticed that this was failing on my branch that included the new `template`
dir and the `node_modules` within it, which had C headers.

### DIFF
--- a/license-check.py
+++ b/license-check.py
@@ -5,8 +5,6 @@ import os
 from pathlib import Path
 import subprocess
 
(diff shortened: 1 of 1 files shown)


flaub:
Very cool! That's a neat idea. Not a big fan of globals, so I'll fix that, but thanks for this!

Issue #410, opened 2023-02-28
Replace or determine fix for action-rs being unmaintained


mothran:
As per:

https://github.com/actions-rs/toolchain/issues/219

Our upstream action dep on https://github.com/actions-rs/toolchain/ appears to be unmaintained and is starting to throw warnings about nodejs12 being deprecated:

```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: risc0/actions-rs-toolchain@v1.
```

Looks like a possible alternative is to drop the dependency all together with:

```yml
    - name: Setup Rust
      run: |
        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
        rustup default stable
```

Removing the action also would remove the need to maintain a fork of it.

PR #413, opened 2023-02-28
Add actions for rustup and sccache

flaub:
Fixes https://github.com/risc0/risc0/issues/410


mothran (approved):
Nice to have these as action modules, good call.

rlukata:
Didn't know about composite actions, looks great!

PR #416, opened 2023-03-01
[Bonsai Template] Update guest serialization to latest Bonsai commit

nategraf:
Bonsai's proxy contract and bridge dropped the addition of the input length from the guest input in
a change on Februrary 28th. This PR updates the starter template to reflect this change.

Note that this new version of the service is what is deployed for use by Eth Denver hackathon
participants. Participants will need to use the template as updated by this PR, or make the same
changes to their own code.

### DIFF
--- a/templates/bonsai/methods/guest/src/bin/fibonacci.rs
+++ b/templates/bonsai/methods/guest/src/bin/fibonacci.rs
@@ -29,10 +29,12 @@ fn fibonacci(n: U256) -> U256 {
     return curr;
 }
 
+const INPUT_LEN: usize = core::mem::size_of::<U256>();
+
 pub fn main() {
     // NOTE: env::read_slice requires a length argument. Reads must be of known
     // length. https://github.com/risc0/risc0/issues/402
-    let input = ethabi::decode_whole(&[ParamType::Uint(256)], env::read_slice(32)).unwrap();
+    let input = ethabi::decode_whole(&[ParamType::Uint(256)], env::read_slice(INPUT_LEN)).unwrap();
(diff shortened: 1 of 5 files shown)


nategraf:
NOTE: Marked this PR as draft until I am able to run this change against the real service.

flaub (approved):

templates/bonsai/methods/guest/src/bin/fibonacci.rs
```diff
@@ -30,10 +30,9 @@ fn fibonacci(n: U256) -> U256 {
 }
 
 pub fn main() {
-    // NOTE: Currently env::read_slice requires a length argument. Bonsai passes the length at the
-    // start of the input. https://github.com/risc0/risc0/issues/402
-    let input_len = env::read_slice::<u32>(1)[0] as usize;
-    let input = ethabi::decode_whole(&[ParamType::Uint(256)], env::read_slice(input_len)).unwrap();
+    // NOTE: env::read_slice requires a length argument. Reads must be of known
+    // length. https://github.com/risc0/risc0/issues/402
+    let input = ethabi::decode_whole(&[ParamType::Uint(256)], env::read_slice(32)).unwrap();
```
flaub, line 35: Might be good to use a CONST var instead of magic number(s).

nategraf: Done in [2267c8e](https://github.com/risc0/risc0/pull/416/commits/2267c8e563317dc22ec83d98b36815e9287d4c36)

nategraf:
I managed to ensure this would work against the latest Bonsai service

PR #421, opened 2023-03-02
Change RNG semantics to help Poseidon.

jbruestle:
Change RNG semantics to not require 32 bits, only the number actually required, so that Poseidon can make RNGs reasonably

### DIFF
--- a/risc0/zkp/src/core/blake2b.rs
+++ b/risc0/zkp/src/core/blake2b.rs
@@ -117,14 +117,8 @@ impl<T: Blake2b> ConfigRng<BabyBear> for Blake2bRng<T> {
         self.current = T::blake2b(concat);
     }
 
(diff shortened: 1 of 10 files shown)


flaub:
The latest benchmark check shows 10% increase on linux-cuda, not sure what to do with that.

mothran:
> Re the benchmark, is there a way I can just rerun. Currently, the code doesn't change anything related to the normal code path. For all the SHA based hals, it's effectively a no-op, so I strongly suspect it's just noise.

Its probably noise. Currently benchmarks don't trigger PR blocking are just for information. I will look into if we can get that noise floor down though.
If you wanna trigger a re-run of a workflow like the Benchmark Check you can open up the completed workflow in the actions view and then click `rerun all jobs` in the upper left.

flaub (approved):
New benchmark results seem fine

PR #424, opened 2023-03-06
Fixed example code linking to wrong youtube video

KaiGeffen:
Fixed the password-checker README linking to the wrong video (Was chess example video). Switched to the correct video.


tzerrell (approved):
Thanks for fixing this!

PR #425, opened 2023-03-06
Full documentation for `risc0_zkvm`

tzerrell:
Complete at least brief documentation for all of the `risc0_zkvm` crate and turn on `warn(missing_docs)` for this crate.

Major topics with added documentation include:
* Serde
* MemoryImage
* ControlID
* Loader
* I/O
* Poseidon HAL

I learned about some of these as I documented, so there's definitely a possibility of errors -- I would appreciate review for accuracy not just style, and have tagged reviewers I think would know about these topics. If you see something that you think needs review by someone I haven't tagged please add them.

### DIFF
--- a/risc0/zkvm/src/lib.rs
+++ b/risc0/zkvm/src/lib.rs
@@ -16,7 +16,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
 #![deny(rustdoc::broken_intra_doc_links)]
-#![warn(missing_docs)]
+#![deny(missing_docs)]
 
 extern crate alloc;
 
--- a/risc0/zkvm/src/lib.rs
+++ b/risc0/zkvm/src/lib.rs
@@ -16,6 +16,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![warn(missing_docs)]
 
 extern crate alloc;
 
--- a/risc0/zkvm/src/lib.rs
+++ b/risc0/zkvm/src/lib.rs
@@ -51,12 +51,16 @@ pub use crate::receipt::Receipt;
 
 const CIRCUIT: risc0_circuit_rv32im::CircuitImpl = risc0_circuit_rv32im::CircuitImpl::new();
 
+/// A collection of hashes attesting to the circuit architecture
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ControlId {
+    /// The hashes comprising the [ControlId]
     pub table: alloc::vec::Vec<Digest>,
 }
 
+/// A trait for objects with an associated [ControlId]
 pub trait ControlIdLocator {
+    /// Get the [ControlId] associated with this object
     fn get_control_id() -> ControlId;
 }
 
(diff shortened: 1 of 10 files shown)


risc0/zkvm/src/lib.rs
```diff
@@ -16,6 +16,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
 #![deny(rustdoc::broken_intra_doc_links)]
+#![warn(missing_docs)]
```
flaub, line 19:
```suggestion
#![deny(missing_docs)]
```

risc0/zkvm/src/prove/io.rs
```diff
@@ -38,6 +55,11 @@ pub trait SliceIoHandler {
 ///
 /// Users may find SliceIoHandler more friendly to use.
 pub trait RawIoHandler {
+    /// Host side IO handling
```
flaub, line 58:
```suggestion
    /// Host side I/O handling
```

risc0/zkvm/src/prove/io.rs
```diff
@@ -27,8 +37,15 @@ use risc0_zkvm_platform::WORD_SIZE;
 /// elements are to be sent back to the guest, and the second call
 /// actually returns the elements after the guest allocates space.
 pub trait SliceIoHandler {
+    /// Type for data received from guest
     type FromGuest: Pod;
+    /// Type for data sent to guest
     type ToGuest: Pod;
+    /// Host side IO handling
```
flaub, line 44:
```suggestion
    /// Host side I/O handling
```

risc0/zkvm/src/serde/serializer.rs
```diff
@@ -46,30 +51,51 @@ where
...
     }
 
+    /// Write a slice
     fn write_slice<T: Pod>(&mut self, slice: &[T]) -> Result<()>;
 
+    /// Return the written data as a [StreamWriter::Output]
     fn release(&mut self) -> Result<Self::Output>;
 }
 
+/// Interface for serializing
```
flaub, line 85: I'd avoid the term 'interface' since I associate that with the trait.

risc0/zkvm/src/serde/serializer.rs
```diff
@@ -402,13 +428,18 @@ impl<'a, W: StreamWriter> serde::ser::SerializeStructVariant for &'a mut Seriali
     }
 }
 
+/// A vector of bytes that can be written to with a stream interface
```
flaub, line 431: This is specifically a host-side `StreamWriter` impl, it's not meant to be used in the guest.

tzerrell: :+1: Added a remark indicating it's for host use.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -193,6 +193,9 @@ cfg_if::cfg_if! {
         use risc0_circuit_rv32im::cuda::CudaEvalCheck;
         use risc0_zkp::hal::cuda::CudaHal;
 
+        /// Returns the default SHA-256 HAL for the RISC Zero circuit
+        ///
+        /// See the non-cuda version of this documentation for details
```
flaub, line 198: Does this show up in the docs? I don't think it will actually be visible unless these features are set.

tzerrell: It does not, but if it's missing the `cuda` feature can't build due to `deny(missing_docs)` (see [this build](https://github.com/risc0/risc0/actions/runs/4347444670/jobs/7594765734)). Hence my "See the non-cuda version of this documentation for details" instead of a full duplication of the documentation.

Happy to hear other approaches for addressing this failure.

flaub: I'm just thinking we can drop the reference to the non-cuda version

flaub (approved):

PR #426, opened 2023-03-06
Simplify guest I/O

shkoo:
* "syscalls" are now how guest talks with host; no more distinction between "syscall" and "channel"
* (Posix style file descriptors can be implemented on top of syscalls, since posix semantics are all byte-based)
* Syscalls are now named to avoid numbering conflicts.
* It's now just as easy (if not easier) for users to implement their own syscalls as it was for them to implement their own sendrecv channels.
* Added "syscall!" macro to invoke a syscall from the guest side to reduce the number of separate asm! wrappers


risc0/zkvm/src/guest/env.rs
```diff
@@ -185,27 +184,31 @@ pub fn commit_slice<T: Pod>(slice: &[T]) {
 /// Return the number of processor cycles that have occured since the guest
 /// began.
 pub fn get_cycle_count() -> usize {
-    unsafe { sys_cycle_count() }
+    unsafe {
+        let (a0, _) = syscall!(SYS_CYCLE_COUNT, core::ptr::null_mut(), 0);
```
flaub, line 188: It seems like some syscalls (like `sys_cycle_count`) isn't verify similar to I/O. There might be other syscalls we want to add that also don't really fit that mold. It feels like we're trying to shoehorn all things into I/O, but I'm not sure that makes the most conceptual sense.

flaub:
Can these syscalls be used from other languages? One of the goals of the existing syscall interface is to make them callable by other languages that have the notion of a "C" FFI.

examples/waldo/core/src/merkle.rs
```diff
@@ -21,12 +21,16 @@ use merkle_light::{
...
+};
 use serde::{Deserialize, Serialize};
 
-/// RISC0 channel identifier for providing oracle access to a vector to the
-/// guest from the host.
-pub const VECTOR_ORACLE_CHANNEL: u32 = 0x09ac1e00;
+declare_syscall!(
+    /// RISC0 syscall for providing oracle access to a vector to the
+    /// guest from the host.
+    pub SYS_VECTOR_ORACLE);
```
flaub, line 33: It feels like we're exposing a deeper interface to end-users. Channels feel like, these are standard ways to extend notion of I/O, and we're hiding the details of what it takes to do a syscall. I think it should be possible for advanced users to add their own syscalls, but I feel like channels are a nice intermediate level extension.

flaub (approved):
I personally would like to avoid the macros and just write out all the assembly for each `syscall_N`, but otherwise this seems like the right direction to me.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+pub unsafe extern "C" fn sys_panic(msg_ptr: *const u8, len: usize) -> ! {
+    syscall_2(nr::SYS_PANIC, null_mut(), 0, msg_ptr as u32, len as u32);
+    unreachable!()
+}
+
+pub unsafe extern "C" fn sys_log(msg_ptr: *const u8, len: usize) {
+    syscall_2(nr::SYS_LOG, null_mut(), 0, msg_ptr as u32, len as u32);
+}
+
+pub extern "C" fn sys_cycle_count() -> usize {
```
flaub, line 298: My mental model is that all syscalls are unsafe, so it seems odd that this one is special.

examples/waldo/core/src/merkle.rs
nategraf: I definitely like not asking the developer to pick their favorite number

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+        let Return(a0, _) = syscall_0(nr::SYS_CYCLE_COUNT, null_mut(), 0);
+        a0 as usize
     }
-    #[cfg(not(target_os = "zkvm"))]
-    unimplemented!()
+}
+
+/// Returns the number of bytes currently available for reading on the
+/// given file descriptor.
+pub unsafe extern "C" fn sys_read_avail(fd: u32) -> usize {
```
nategraf, line 307: What is the expected behavior this call for streaming file descriptors where a length may not be easily available? And for file descriptors with infinite or indeterminate length (e.g. reading from `/dev/urandom`)?

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+    let Return(navail, _) = syscall_1(nr::SYS_READ_AVAIL, null_mut(), 0, fd);
+    navail as usize
+}
+
+/// Reads the given number of bytes into the given buffer, posix-style.  Returns
+/// the number of bytes actually read.  On end of file, returns 0.
+pub unsafe extern "C" fn sys_read(fd: u32, recv_buf: *mut u8, nrequested: usize) -> usize {
```
nategraf, line 314: So as a safety requirement, the caller needs to be sure that the pointer `recv_buf` can accept `nrequested` bytes, is that correct? E.g. if someone calls `sys_read` with a pointer to a stack buffer, and passes in an `nrequested` that is too big, could that cause an exploitable stack buffer overflow? If so, we should at least call this out in the doc here, although it is already unsafe and so hopefully will only be used with caution.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+    // The SYS_READ system call can do a given number of word-aligned reads
+    // efficiently. The semantics of the system call are:
+    //
+    //   (nread, word) = syscall_2(nr::SYS_READ, outbuf, outwords, fd, nbytes);
```
nategraf, line 318: What is `outwords` here?

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+    //
+    // This reads exactly nbytes from the file descriptor, and fills the words
+    // in outbuf, followed by up to 4 bytes returned in "word", and fills
```
nategraf, line 321: I feel like I can guess from surround context, but it's not immediately clear what 'followed by up to 4 bytes returned in "word"' means.

risc0/zkvm/platform/src/syscall.rs
nategraf: Are the semantics here "available now" or "available until end of file"? E.g. Is it acceptable for a call to `SYS_READ_AVAIL` to return `0` and then when called a second time return `1`?

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+
+/// Returns the number of bytes currently available for reading on the
+/// given file descriptor.
+pub unsafe extern "C" fn sys_read_avail(fd: u32) -> usize {
+    let Return(navail, _) = syscall_1(nr::SYS_READ_AVAIL, null_mut(), 0, fd);
+    navail as usize
+}
+
+/// Reads the given number of bytes into the given buffer, posix-style.  Returns
+/// the number of bytes actually read.  On end of file, returns 0.
```
nategraf, line 313: When the caller asks for `nrequested` bytes, but the number of read bytes is fewer does that mean the fd is closed? (My guess would be yes, but that's not clear in this comment)

risc0/zkvm/src/guest/env.rs
```diff
@@ -97,45 +100,38 @@ pub(crate) fn finalize() {
     }
 }
 
-/// Exchange data with the host.
-pub fn send_recv_raw(channel: u32, to_host: &[u8], from_host: &mut [u32]) -> (u32, u32) {
+/// Exchange data with the host
+pub fn send_recv_raw(
```
nategraf, line 104: It looks like we are changing semantics to syscalls rather than send_recv. Will this function be deprecated/removed/renamed?

risc0/zkvm/src/prove/io.rs
```diff
@@ -123,16 +122,142 @@ impl<H: SliceIoHandler> RawIoHandler for SliceIoWrapper<H> {
...
+        Self {
+            read_fds: RefCell::new(BTreeMap::new()),
+            write_fds: RefCell::new(BTreeMap::new()),
+        }
+    }
+
+    pub fn with_read_fd(self, fd: u32, reader: Rc<RefCell<impl BufRead + 'a>>) -> Self {
+        self.read_fds.borrow_mut().insert(fd, reader);
+        self
+    }
```
nategraf, line 164: I remember you mentioned needing a `RefCell` in the platform sync, but I don't see why it's needed here at least, this is an owned reference. I see in the `Syscall::syscall` method it needs some mutability. Could that be solved by using a `&mut self` receiver. I'm guessing I'm missing something. Since this exhibits observable mutability, I agree that it would be nice to make that clear in the type system, but that `RefCell` is a decent backup option.

risc0/zkvm/platform/src/syscall.rs
shkoo: "currently available", so available now.  I also unexported sys_read_avail as I didn't mean to expose it to users.

sys_read has standard posix read semantics, so updated the comment.

risc0/zkvm/platform/src/syscall.rs
shkoo: Yes, the semantics are the same as posix read.  Users should prefer a higher-level abstraction.

risc0/zkvm/platform/src/syscall.rs
shkoo: Renamed to num_words_in_outbuf

risc0/zkvm/platform/src/syscall.rs
shkoo: Not necessarily; it has standard posix semantics.  Clarified comment.

risc0/zkvm/src/guest/env.rs
flaub: Maybe this is `ioctl`?

flaub: Or maybe just `syscall()`?

risc0/zkvm/platform/src/lib.rs
```diff
@@ -17,9 +17,16 @@
...
 pub mod abi;
-pub mod io;
 pub mod memory;
+#[macro_use]
 pub mod syscall;
 
 pub const WORD_SIZE: usize = core::mem::size_of::<u32>();
 pub const PAGE_SIZE: usize = 1024;
+
+pub mod fileno {
```
flaub, line 27: Need docs on public things

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -86,86 +78,131 @@ const fn round_up(a: u32, b: u32) -> u32 {
...
-        asm!(
-            "ecall",
-            in("t0") ecall::SOFTWARE,
-            out("a0")  cycle,
-            inout("a1") 0 => _,
-            in("a2") nr::SYS_CYCLE_COUNT,
-        );
-        cycle
+#[repr(C)]
+/// Returned registers (a0, a1) from a syscall invocation.
```
flaub, line 140:
```suggestion
/// Returned registers (a0, a1) from a syscall invocation.
#[repr(C)]
```

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -245,12 +282,116 @@ pub unsafe fn sys_sha_buffer(
...
+    syscall_2(nr::SYS_PANIC, null_mut(), 0, msg_ptr as u32, len as u32);
+    unreachable!()
+}
+
+pub unsafe extern "C" fn sys_log(msg_ptr: *const u8, len: usize) {
+    syscall_2(nr::SYS_LOG, null_mut(), 0, msg_ptr as u32, len as u32);
+}
+
+pub unsafe extern "C" fn sys_cycle_count() -> usize {
+    unsafe {
```
flaub, line 299: This is redundant

risc0/zkvm/src/prove/mod.rs
```diff
@@ -415,7 +421,7 @@ impl<'a> Prover<'a> {
 
         // Attach the full version of the output journal & construct receipt object
         let receipt = Receipt {
-            journal: self.inner.journal.clone(),
+            journal: take(&mut self.inner.journal.buf.borrow_mut()),
```
nategraf, line 424: Using `RefCell::take` has the nice property that it will panic if there is an outstanding borrow.
```suggestion
            journal: self.inner.journal.buf.take(),
```

PR #436, opened 2023-03-14
Updates to Wordle text and yellow highlight

3lkn:
Updated the Wordle text to slightly more friendly interaction.
Changed yellow highlight to not over-highlight letters (don't count dupes of "correct letter in correct spot" as "correct letter in incorrect spot").


examples/wordle/core/src/lib.rs
```diff
@@ -49,6 +50,6 @@ impl WordFeedback {
             }
             print!("{:}", guess_word.chars().nth(i).unwrap());
         }
-        println!("\x1b[0m");
+        println!("\x1b[0m\n");
```
flaub, line 53: This is weird, because `println!` includes a `\n`

flaub: If you want a newline, I think it's clearer to say:
```
println!("\x1b[0m");
println!("");
```

flaub (approved):
Thanks for adding a test for the false-positive case.

examples/wordle/src/main.rs
```diff
@@ -105,22 +114,24 @@ fn play_rounds(server: Server, player: Player, rounds: usize) -> bool {
...
+    println!("Welcome to fair Wordle! Enter a five-letter word.");
 
     let server = Server::new(wordlist::pick_word());
     let player = Player {
         hash: server.get_secret_word_hash(),
     };
 
     if play_rounds(server, player, 6) {
-        println!("You won!");
+        println!("You won!\n");
```
flaub, line 125: I think we should do the same for all these cases

tzerrell (approved):
great!

flaub:
Bump, would be good to get this landed sooner than later

PR #438, opened 2023-03-15
Add no-std fix for WASM/Substrate chains

justinFrevert:
This fixes a no-std issue to allow Substrate-based chains to use the zkvm within their WASM runtimes - at least for verification, as I'd tested. It's valuable to maintain support for WASM environments, as it opens Risc0 up to be used by hundreds more projects.

The change simply resolves an error on the Substrate-side which is fairly common for no-std issues. Though this one points at num_traits:
```
  error[E0152]: duplicate lang item in crate `sp_io` (which `sp_runtime` depends on): `panic_impl`.
    |
    = note: the lang item is first defined in crate `std` (which `num_traits` depends on)
    = note: first definition in `std` loaded from /home/justin/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/wasm32-unknown-unknown/lib/libstd-a643c5931c4d12de.rlib
    = note: second definition in `sp_io` loaded from /home/justin/dev/substrate-risc0/target/release/wbuild/node-template-runtime/target/wasm32-unknown-unknown/release/deps/libsp_io-3b94b351090fd32d.rmeta

  For more information about this error, try `rustc --explain E0152`.
  error: could not compile `sp-runtime` due to previous error
```

I tested in a sample Substrate project and found that [the change](https://github.com/justinFrevert/substrate-risc0/blob/9b7d0f3ba23a4eaf3fc553f5456800c121af2355/pallets/template/src/lib.rs#L58-L71) works.


SchmErik:
Hi @justinFrevert thanks for your PR. So we have a build of WASM and I'm surprised that it didn't catch this...

SchmErik:
I see what's going on now. Our build of the wasm verifier doesn't include a lot of other crates so we don't run into this build issue!

justinFrevert:
@SchmErik is that with the `wasm` crate? I actually didn't figure out how that to run with/without the changes.

justinFrevert:
It's possible I may be able to help some with the WASM check, but would need help understanding the aim for this crate a bit. Are we just wanting more dependencies for this `wasm` crate? @SchmErik

mothran:
@justinFrevert heads up, minor issue in our CI. That failure is on us, and I should have a fix in soonish.
Sorry about that.

risc0/zkvm/Cargo.toml
```diff
@@ -43,7 +43,7 @@ gimli = { version = "0.27", optional = true }
 lazy-regex = { version = "2.3", optional = true }
 log = "0.4"
 num-derive = "0.3"
-num-traits = "0.2"
+num-traits = { default-features = false, version = "0.2" }
```
flaub, line 46:
```suggestion
num-traits = { version = "0.2", default-features = false }
```

flaub (approved):
LGTM, would love to add some way to ensure that CI catches this problem in the future

justinFrevert:
> LGTM, would love to add some way to ensure that CI catches this problem in the future

I am not sure if I have thoughts on a solution atm. I added https://github.com/risc0/risc0/issues/441

PR #445, opened 2023-03-16
zkvm: add guest support for getrandom

SchmErik:
This is a "batteries-inculded" solution for programmers to be able to use getrandom without having to depend on the `getrandom` crate


flaub:
Do you think we could register a custom implementation inside the guest crate so that it's already hooked up and ready to go if users select a new `getrandom` feature on the `risc0-zkvm` crate?

flaub (approved):

risc0/zkvm/Cargo.toml
```diff
@@ -23,6 +23,7 @@ protobuf-src = { version = "1.1", optional = true }
 anyhow = { version = "1.0", default-features = false }
 bytemuck = "1.12"
 cfg-if = "1.0"
+getrandom = { version = "0.2", features = ["custom"] }
```
mothran, line 26: I think the `getrandom` crate has a `std` flag, might be worth double checking if we need to optionally pass that feature flag on/off depending on if the zkvm is has it. https://docs.rs/crate/getrandom/0.2.8/features#std

mothran (approved):

risc0/zkvm/Cargo.toml
SchmErik: I've been playing around with various combinations and I'm unable to get a failure at least on the `getrandom` crate. I've run into some things using rand on the `multitests` but i'm able to get `rand` to work on the starter template. I'll merge this but I'll keep the `std` flag in mind.. I feel like we have some clean up to do on that front...

PR #447, opened 2023-03-16
Partial experimental preflight implementation for riscv32im

shkoo:
Select this experimental mode by setting the RISC0_EXPERIMENTAL_PREFLIGHT envornment variable or by using ProverOpts::with_preflight

Preflight does not yet generate all the data needed to generate a full circuit execution trace, so currently only works when skip_seal is specified.

All of the risc0-zkvm tests pass in preflight mode, except for the ones which depend on exact cycle counts or on execution tracing/profiling.


flaub (approved):
Nice that we found a crate for RISC-V itself

PR #452, opened 2023-03-17
Support for stdin and getenv in guest

shkoo:
* Actually export the sys_* ABI symbols without mangling them
* Add sys_getenv to get environment variables
* Add with_env_var to ProverOpts to specify environment variables to include for the guest
* Add --env VARNAME=value option to r0vm to specify environment variables to include
* Plumb through stdin() so that std::io::stdin() works in the guest as expected


risc0/zkvm/methods/std/src/bin/standard_lib.rs
```diff
@@ -12,9 +12,35 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::io::{stdin, stdout, Read, Write};
+
 risc0_zkvm::entry!(main);
 
 pub fn main() {
```
mothran, line 19: Nice way to integrate std::env testing.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -393,3 +404,49 @@ pub unsafe extern "C" fn sys_write(fd: u32, write_buf: *const u8, nbytes: usize)
...
     );
 }
+
+/// Retrieves the value of an environment variable, and stores as much
+/// of it as it can it in the memory at [out_words, out_words +
+/// out_nwords).  Returns the length of the value.
+///
+/// This is normally called twice to read an environment variable:
+/// Once to get the length of the value, and once to fill in allocated
+/// memory.
```
mothran, line 414: Heads up, in the future if we integrate clippy, unsafe functions will warn with missing a `# Safety` block in the docstr, if you happen to have a rough idea of the lifetime assumptions of the args, putting them in a Safety block now would be a nice way to jump start that documentation.

reference: https://rust-lang.github.io/rust-clippy/rust-1.56.0/index.html#missing_safety_doc

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -393,3 +404,49 @@ pub unsafe extern "C" fn sys_write(fd: u32, write_buf: *const u8, nbytes: usize)
...
+        a0 as usize
+    }
+}
+
+// Number of words remaining in the heap that haven't yet been allocated.
+static mut HEAP_WORDS_REMAINING: usize = crate::memory::HEAP.len_words();
+
+#[no_mangle]
+pub unsafe extern "C" fn sys_alloc_words(nwords: usize) -> *mut u32 {
+    // SAFETY: Single threaded, so nothing else can touch this while we're working.
```
mothran, line 441: Maybe move to a `# Safety` section in the docstr.

mothran (approved):
Looks really good.

risc0/zkvm/platform/src/syscall.rs
shkoo: @flaub in https://github.com/risc0/risc0/pull/426#discussion_r1134625722_ indicated a preference for marking all syscalls as "unsafe", even if they don't write to memory or do anything else unsafe.

How do we want to reconcile that?   Should we just add a "# Safety" block saying "This is safe to call!  No worries here."?

mothran: Hmmm good question, if there is not a straight forward answer now we could just defer until we decide to incorporate clippy.

flaub: I think we should probably always have a safety comment that goes with any unsafe calls. The justification could be: "There are no known issues with calling this function" or something to that effect.

risc0/zkvm/platform/src/syscall.rs
flaub: I think we also need to get into the habit of docstrings for all public functions

PR #461, opened 2023-03-21
Continuations, part 1

flaub:
Breaking Changes:
* Drop the `image_id` parameter to the `risc0_zkvm::Prover`. The Prover already computes the `image_id` from the supplied ELF binary.
* Replace `risc0_zkvm::MemoryImage::root` with `risc0_zkvm::MemoryImage::get_root()`, which computes the root merkle tree entry on-demand.
* Drop `VerificationError::SealJournalLengthMismatch`.
* Rename `VerificationError::JournalSealRootMismatch` to `VerificationError::JournalDigestMismatch`

Updates:
* New rv32im circuit which includes continuations support
* Add `env::pause` to the guest environment API
* Add a test for pause/continue

TODO:
* Implement system initiated splits


risc0/zkvm/platform/src/syscall.rs
```diff
@@ -141,85 +146,211 @@ impl SyscallName {
...
+        Return(a0, a1)
+    }
+    #[cfg(not(target_os = "zkvm"))]
+    unimplemented!()
+}
 
 #[inline(always)]
 #[no_mangle]
-pub unsafe extern "C" fn sys_halt() {
+pub unsafe extern "C" fn sys_halt() -> ! {
```
shkoo, line 325: Could we have sys_halt take a halt::* argument instead of having three separate asm blocks?

risc0/zkvm/src/receipt.rs
```diff
@@ -326,3 +311,54 @@ impl Receipt {
...
+
+#[derive(Debug)]
+struct Global {
+    pre: SystemState,
+    _post: SystemState,
+    _check_dirty: u32,
+    output: Digest,
+}
+
+impl SystemState {
```
shkoo, line 329: This is supposed to match the layout in the circuit I'm guessing?  I think this should really come from the exported layout so we don't have to worry if stuff shifts around.  But I guess I can do that later.

shkoo (approved):

risc0/zkvm/platform/src/syscall.rs
flaub: There's only two cases (AFAICT) that are user-visible (HALT & PAUSE). I guess I left it separate because perhaps the interface for pause will change, and also `sys_halt` wants to return `!` whereas `sys_pause` can't do that.

risc0/zkvm/src/receipt.rs
flaub: I'd also love to have serialization (not just deserialization). Generally I think we need to improve the way that the host interacts with the globals, but this is a first tiny step in that direction.

PR #463, opened 2023-03-22
Add concurrency to CI

rlukata:
This PR adds concurrency to the Github workflows to automatically cancel current in-progress workflow runs when new commits are added.

The definition `${{ github.workflow }}` makes it so that only this workflow is cancelled instead of a widespread cancelling of workflows.

The definition group: `${{ github.event.pull_request.number || github.ref }}` targets workflows within the same PR, branch, or tag.

### DIFF
--- a/.github/workflows/bench_pr.yml
+++ b/.github/workflows/bench_pr.yml
@@ -10,6 +10,10 @@ on:
 
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   bench:
     runs-on: [self-hosted, bench, "${{ matrix.os }}", "${{ matrix.device }}"]
--- a/.github/workflows/bench_trends.yml
+++ b/.github/workflows/bench_trends.yml
@@ -12,6 +12,10 @@ on:
   # pull_request:
   #   branches: [ main ]
 
+concurrency:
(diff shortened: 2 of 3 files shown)


.github/workflows/bench_trends.yml
```diff
@@ -12,6 +12,10 @@ on:
   # pull_request:
   #   branches: [ main ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
flaub, line 17: This only runs on main, we probably dont' need it here

.github/workflows/bench_pr.yml
```diff
@@ -10,6 +10,10 @@ on:
 
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
flaub, line 15: Sometimes I want to get benchmarks manually, does this prevent that?

rlukata: No, it just cancels previous pending/in-progress runs. Doesn't prevent from manually triggering flows.

.github/workflows/bench_trends.yml
rlukata: I added it because I saw that this workflow can run if in PR (only when testing) and because I imagine there's a situation where 2 PRs land at similar times and we only care about the last one.

flaub (approved):
TY!

PR #464, opened 2023-03-22
adding warning about memory

pdg744:


tzerrell (commented):
Thanks!

Same comment as in https://github.com/risc0/website/pull/133, I think this runs ok on my 32GB machine but I'll test again when I have a chance to make sure I'm right about that

pdg744:
I made up that recommendation based on Frank's comment here: https://github.com/risc0/risc0-rust-examples/issues/53

tzerrell (approved):

PR #465, opened 2023-03-22
Addressed missing libm symbols in guest

mothran:
Imports the [externc-libm](https://github.com/HaruxOS/externc-libm) and forces the symbols to be present in the guest build. Fixes "rust-lld missing symbol" errors for libm floating point symbols.

This PR pulls in the `externc-libm` crate from github and not crates.io because the published crates is slightly behind the repo.  Mentioned here: https://github.com/HaruxOS/externc-libm/issues/2#issuecomment-1480206267

Questions:
* Should we test this in MultiTest when building `examples/evm` will also validate this fix in CI?
* Should we move this fix behind the `std` flag? externc-libm is designed to work in no_std environments so it should be fine to provide it in no_std?


SchmErik:
>     * Should we test this in MultiTest when building `examples/evm` will also validate this fix in CI?

I think adding this to MultiTest reduces our dependency on revm so I think we should add a small test that calls a function from libm.

SchmErik:
> 
>     * Should we move this fix behind the `std` flag? externc-libm is designed to work in no_std environments so it should be fine to provide it in no_std?

This is the perplexing part of the libm issue for me: the example/evm directory has std enabled... Is libm supposed to be a part of std or something? I'm curious about the root cause of this issue...

SchmErik:
I was looking around and found this: https://crates.io/crates/libm

SchmErik:
Screenshot 2023-03-22 at 2 54 21 PM

mothran:
@SchmErik Yeah the libm crate is re-exported by `externc-libm` which wraps the methods in extern blocks and makes it so any dep in the tree will have the symbols exposed to the linker. The externc-libm is specifically designed for a situation like we are in where we want the symbols to just be provided.

The evm / revm parts are just examples of a crate that triggered the bug, the minimal guest code to replicate the issue that I found was:
```rust
use core::hint::black_box;
let f = black_box(1.0_f32);
black_box(f.min(1.0));
```

As for the origin of this issue, it appears to be something in how rust-lang handles floating point operations, its possible we could fix this in the custom triple for our arch + platform but I don't know the exact location that rust-lang links out to libm implicitly.

SchmErik:
it looks like this https://github.com/rust-lang/compiler-builtins could be helpful too.

SchmErik:
> @SchmErik Yeah the libm crate is re-exported by `externc-libm` which wraps the methods in extern blocks and makes it so any dep in the tree will have the symbols exposed to the linker. The externc-libm is specifically designed for a situation like we are in where we want the symbols to just be provided.
> 

I see! Thanks for the explanation! I didn't realize that `externc-libm` wrapped the `libm` crate. I assumed it wrapped another libm.

> The evm / revm parts are just examples of a crate that triggered the bug, the minimal guest code to replicate the issue that I found was:
> 
> ```rust
> use core::hint::black_box;
> let f = black_box(1.0_f32);
> black_box(f.min(1.0));
> ```
> 
> As for the origin of this issue, it appears to be something in how rust-lang handles floating point operations, its possible we could fix this in the custom triple for our arch + platform but I don't know the exact location that rust-lang links out to libm implicitly.

Given this and seeing that the linker error shows up in the evm example which has std enabled, maybe it's ok to expose it regardless of `std`... @flaub do you have any thoughts?

risc0/zkvm/Cargo.toml
```diff
@@ -23,6 +23,7 @@ protobuf-src = { version = "1.1", optional = true }
 anyhow = { version = "1.0", default-features = false }
 bytemuck = "1.12"
 cfg-if = "1.0"
+externc-libm = { git = "https://github.com/HaruxOS/externc-libm", rev = "873e2ee" }
```
shkoo, line 26: I think specifying a "git" package source dependency isn't allowed when publishing packages, so we should depend on this through crates.io like all our other dependencies.

risc0/zkvm/src/guest/mod.rs
```diff
@@ -63,6 +63,8 @@ pub mod sha;
 
 use core::{arch::asm, mem, ptr};
 
+// Empty import to get the externc_libm symbols linked into the final linker run.
+use externc_libm::math::fminf;
```
shkoo, line 67: Ugh , there's got to be a better way to do this but I don't know what it is. :)

risc0/zkvm/Cargo.toml
mothran: Hmmm that might be a blocker because: https://github.com/HaruxOS/externc-libm/issues/2#issuecomment-1480206267 causes a duplicate symbol error at link time with the existing crate.io published version.

risc0/zkvm/src/guest/mod.rs
mothran: Found it!
```rust
extern crate externc_libm;
```

risc0/zkvm/Cargo.toml
flaub: I'm not aware of crates.io preventing the use of git dependencies, but I also can't find any evidence for that in either direction.

mothran: I think this is the tracking issue on `cargo` https://github.com/rust-lang/cargo/issues/6722

flaub (approved):
I don't understand the root cause of this, but having a workaround in place seems like a good first step. It's not clear to me how this interacts with `std` or `no_std`.

mothran:
> I don't understand the root cause of this, but having a workaround in place seems like a good first step. It's not clear to me how this interacts with `std` or `no_std`.

I think we might be blocked from merging this PR until we resolve: https://github.com/risc0/risc0/pull/465#discussion_r1146709961 

I will hold this PR until I get an update from the upstream or we need to deploy a workaround.

flaub:
I think we should consider maintaining our own fork

PR #468, opened 2023-03-23
Migrate guest's concept of "initial input" to be a more standard "stdin"

shkoo:
* env::read now reads from stdin by default.
* risc0_zkvm::serde now serializes type "char" (i.e. single characters that are serialized by themselves) as its 4-byte codepoint rather than a 4-byte length followed by its serialized bytes.  Strings are unchanged.
* Removed risc0_zkvm::env::Env structure, since its last remaining use is gone.
* Renamed Stream{Reader,Writer} to more rusty names
* New "with_stdin" options for ProverOpts to specify the standard input to supply to the guest; this accepts anything that implements std::io::Read.
* When "std" is enabled, env::{stdin,stdout,stderr} implement std::io::{Read,Write}.
* Deserialized objects are now owned and filled in directly instead of referencing a buffer.  This means that serialized types will need to be owned types, like "String" instead of "&str".


tzerrell (commented):
I am surprised this is passing CI. Shouldn't this line in the Chess example no longer work? https://github.com/risc0/risc0/blob/396356fe659c1894be9c4eee90805b11f0655d47/examples/chess/methods/guest/src/main.rs#L24

And I'd expect other similar example failures.

But looking at the CI logs, the test at https://github.com/risc0/risc0/blob/396356fe659c1894be9c4eee90805b11f0655d47/examples/chess/src/main.rs#L75 did run, and it did pass, and it should be exercising the line of code I'm concerned about. So... I'm confused.

tzerrell:
> Shouldn't this line in the Chess example no longer work?

Ah, maybe I'm wrong and that line _is_ still the intended approach, maybe this is only meant to replace `env::send_recv_slice` and not read? In which case this is just my personal confusion.

shkoo:
> > Shouldn't this line in the Chess example no longer work?
> 
> Ah, maybe I'm wrong and that line _is_ still the intended approach, maybe this is only meant to replace `env::send_recv_slice` and not read? In which case this is just my personal confusion.

This PR replaces env::read with a different implementation, but doesn't touch env::send_recv_slice.  

I'm not sure I understand what you think should no longer work; what behavior from this PR would you expect that you're not seeing?

flaub (approved):

tzerrell:
> I'm not sure I understand what you think should no longer work; what behavior from this PR would you expect that you're not seeing?

The changes in the `methods` directories (e.g. [here](https://github.com/risc0/risc0/pull/468/files#diff-b48d5a0971cb03b3b1444a919cae1e341ed4c923728505beecfbcc523a4765cb)) made me think that the `risc0_zkvm::guest::env` API was changing. Which made me surprised things in the `examples` directory were passing tests without changes. Then I looked at that linked file again, and realized it was just `env::send_recv_slice` changing to `env::stdin().read_words_to_end`, not a change to `env::read`. This made me think the examples may not need to change after all, hence my "just my personal confusion" comment.

However, I'm now confused again by this comment:
> This PR replaces env::read with a different implementation, but doesn't touch env::send_recv_slice.

Why did the linked methods file change from `env::send_recv_slice` to `env::stdin().read_words_to_end` then?

PR #469, opened 2023-03-24
Continuations, part 2: system initiated splits

flaub:


risc0/zkvm/src/prove/mod.rs
```diff
@@ -82,6 +82,8 @@ use crate::{
     ControlIdLocator, MemoryImage, CIRCUIT, PAGE_SIZE,
 };
 
+const DEFAULT_SEGMENT_LIMIT_PO2: usize = 20; // 1M cycles
```
shkoo, line 85: I have a feeling this might be a breaking change, since without everyone changing their code to run the prover more than once, everyone's will now start ending at 1M cycles instead of 16M cycles.

shkoo (approved):

risc0/zkvm/src/prove/mod.rs
flaub: OK, I changed this to be 16M cycles for now

PR #471, opened 2023-03-27
adding thors examples

pdg744:
do we want to add thor's examples to the examples library?

### DIFF
--- a/examples/dotproduct/.gitignore
+++ b/examples/dotproduct/.gitignore
@@ -1,0 +1,4 @@
+.DS_Store
+Cargo.lock
+methods/guest/Cargo.lock
+target/
--- a/examples/dotproduct/Cargo.toml
+++ b/examples/dotproduct/Cargo.toml
@@ -1,0 +1,5 @@
+[workspace]
+members = [
+    "dotproduct",
+    "methods",
+]
--- a/examples/dotproduct/README.md
+++ b/examples/dotproduct/README.md
@@ -1,0 +1,9 @@
+# Risc-0 Dotproduct example
+An example, computing the dotproduct of two 2-dimensional vectors with Risc0. To run the code, you'll need a working
+rust installation, then just `cargo run`.
+
+Main tips:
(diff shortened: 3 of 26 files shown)


flaub:
Could you add a better title and description of this PR?

tzerrell:
A bit of context: These are (approximately?) the examples Thor wrote to explain what RISC Zero is in the context of other ZKP projects for ZK Hack folks. There are 2 examples: One vectorizing our `factors` example and one extracting the Merkle tree component of the `waldo` example.

I haven't done a detailed review, but my high-level thoughts are:
* Vectorized factors doesn't seem likely to have much value beyond the existing scalar factors example. My intuition is that including it would more pollute our examples directory with redundant examples. It would also add to the maintenance load. (There's a contrasting view that many similar examples with slight changes can make clearer what are core shared techniques and what are problem specific. I don't think this is right for this case, but could be persuaded otherwise.)
* Merklization for committing to a large amount of data and then reading only a small subset of it is clearly useful. That said, I'm not sure this Merkle example is the correct next step after the waldo example. I think in the medium term we want a Merklization function or tool. In the interim -- maybe this Merklization example is helpful? Perhaps I'll have a stronger opinion once I can do a real review, but I'm interested if others have thoughts on this.

pdg744:
The intention here was to surface the new examples and invite discussion about what (if anything) we want to do with them.

For the [ZK Hack prep sesh](https://www.youtube.com/watch?v=75M-GeuGke4&list=PLj80z0cJm8QFGB6AsiAG3EB06L7xr5S1c&index=51), Thor made a dot product and merkle tree example in Circom, Noir, and RISC Zero. The question here is whether we want to add one or both of those examples to our examples folder.

pdg744:
More or less agree with Tim's comment that a dotproduct example doesn't add much value; adding a Merkle example feels pretty high value to me.

pdg744:
Worth noting that Thor found it non-trivial to build a merkle tree example -- and also found it non-trivial to make sense of the merkle implementation in `waldo`

flaub:
Can we split this PR into 2?
1. Merkle-tree lib
2. Vectorized factors

At that point we can do a PR on each and decide whether to land both, each, or none.

PR #473, opened 2023-03-28
Add tool for datasheet generation

flaub:


mothran (approved):
Let me know what your thinking for making this automated, I assume doing something like pushing to the `benchmarks` repo would fit and we could do it on a nightly trigger.

flaub:
Yes, that sounds fantastic

tzerrell (commented):
🎉

PR #480, opened 2023-03-29
Fixes for cargo-risczero new

flaub:


risc0/cargo-risczero/src/commands/new.rs
```diff
@@ -66,8 +66,12 @@ pub struct NewCommand {
     ///
     /// Toggles the `#![no_std]` in the guest main() and the `std` feature flag
     /// on the `risc0_zkvm` crate.
-    #[clap(value_parser, long, global = true)]
+    #[clap(value_parser, long, global = false)]
```
mothran, line 69: Does this swap us to defaulting to `std` for the templates?

mothran (approved):
Great fixes!

tzerrell (approved):
Great, looks good to me and runs correctly for me locally.

flaub:
BTW, the way I tested this was:
```
cd /local/risc0/clone
cargo run -p cargo-risczero risczero new --template templates/rust-starter /tmp/rust-starter --path $PWD
cd /tmp/rust-starter
cargo run
```

Which ensures that it's actually using the template from the current checkout *and* sets deps so that it builds against the same checkout.

tzerrell:
Conveniently, I tested via a fairly different path, using the `--branch` and `--use-git-branch` flags (which I did not find especially intuitive, but upgrading their DevEx can be a later problem):
```
cargo install --path risc0/risc0/cargo-risczero
cargo risczero new --branch flaub/fix-r0-new --use-git-branch flaub/fix-r0-new --template risc0/examples/factors temp-factors
cd temp-factors
cargo run -r
```

PR #493, opened 2023-04-01
Hash function cleanups

flaub:
* Move hashing functions into `risc0_zkp::hash`
* Rename `ConfigHash` to `HashFn`
* Renamed `ConfigRng` to `Rng`


tzerrell (commented):
The API adjustments seem reasonable to me.

I can let people more involved in the implementation speak up in their reviews if they have any concerns with how the implementations were moved / adjusted.

jbruestle (approved):

PR #494, opened 2023-04-01
Update to latest ethers & revm

flaub:
Also dropping `risc0-build` from all guest crates. This should resolve most of the dependabot alerts.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -50,10 +50,50 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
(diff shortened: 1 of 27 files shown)


mothran (approved):
Thanks! With the RPC url change in cargo test, we should also update the README:
https://github.com/risc0/risc0/blob/main/examples/evm/README.md?plain=1#L32
As well as mention that a user can use that RPC for initial testing.

mothran (approved):

tzerrell:
Thank you for the `build-dependencies` cleanup

PR #500, opened 2023-04-05
Add "cargo risczero staticlib" command to generate a static library for use with non-rust guests

shkoo:
* Expose "run cargo command in guest environment" functionality from risc0-build
* Add C++ "hello world" example in examples/cpp
* Move rust language features (panic handler, alloc fault) into risc0-zkvm-platform


examples/cpp/Makefile
```diff
@@ -0,0 +1,10 @@
+CXX=riscv32-unknown-elf-c++
```
mothran, line 1: Do I need to update the CI AMI's to install this compiler / toolchain? If so what package on ubuntu 22.04 provides it?

mothran (approved):
LGTM, the changes to risc0-build should make `cargo risczero test` much easier to implement. Nice!

PR #501, opened 2023-04-07
Crates.io validation tooling

mothran:
Adds tooling to generate risc0 projects with different guest crates for testing building, running and (eventually) crate unit tests. Included is a github workflow that triggers the tooling to do a full end to end run each night as well as record its results.

## TODO

- [x] Create landing page index.html for benchmarks static data pages (benchmarks / crates testing)
- [x] Push the outputted JSON to https://github.com/risc0/benchmarks
- [x] Create a small prototype HTML UI for exploring the results data
- [x] Capture build errors and place into ValidationResults object
- [x] Fix any small TODO's within the source code

## Next steps.

The following are things that we can optionally add or add later:

- [ ] Rename risc0/benchmarks repo
- [x] ~~Rough bucketing of build errors - example: detect missing std by searching logs~~
~~This Gives us a nice view of what the highest impact fixes would be.~~
- [x] ALTERNATIVE to bucketing: Do the bucketing in the javascript UI + search box to do str search in error logs.
- [ ] Extend the manual `profiles.rs` with further hand checking the result logs ask @mothran for details here
- [ ] Add `cargo risczero test` to the validation suite
- [ ] Add a configurable list of crates to add to the test suite. Allows us to add in important crates that might not be in the TOP N of downloads.


tools/crates-validator/profiles/test.json
```diff
@@ -0,0 +1,19 @@
...
+        {
+            "name": "lazy_static",
+            "version": "1.4.0",
+            "std": true,
+            "import_str": "use lazy_static::lazy_static;\nlazy_static! {\n\tstatic ref EXAMPLE: u8 = 42;\n}",
+            "should_fail": false,
+            "run_prover": true
+        }
+    ]
+}
```
tzerrell, line 19:
```suggestion
}

```

tools/crates-validator/src/profiles.rs
```diff
@@ -0,0 +1,64 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::CrateProfile;
+
+pub const SKIP_CRATES: &[&str] = &[
+    "miow", // Windows specific crate that depends on `windows-sys`
+    "windows-sys",
+    "redox_syscall", // Tools for redox OS, not supported
```
tzerrell, line 20: Not blocking, but if it's easy to have these comments appear when you click "Skipped" in the same way the error logs appear when you click "BuildFail", that would be a nice addition.

tools/crates-validator/src/lib.rs
```diff
@@ -0,0 +1,562 @@
...
+            .current_dir(working_dir)
+            .output()?;
+
+        let status = output.status;
+        let res = if status.success() || (!status.success() && profile.should_fail) {
+            info!("{} - build - SUCCESS", profile.name);
+            (true, String::new())
+        } else {
+            // Write out the build log to stderr
+            // TODO: Capture and analyze this log for common classes of errors?
```
tzerrell, line 422: Definitely would be cool and helpful! And also until we can resource such a feature this can be done by manually looking at all the failures.

tzerrell (approved):
:tada: Awesome, so glad to have this!

tools/crates-validator/src/lib.rs
mothran: Yeah my current thinking is to actually put that all into the UI code, since we can scan the snipped build logs in javascript. But maybe move it over here later, idk.

tools/crates-validator/src/profiles.rs
mothran: Yeah probably a good call, I can probably make this a tuple of (crate_name, "description str") and then return that into the results if_skipped.

mothran:
Blocked on some issues with the EVM rpc in examples, tmp fix here: https://github.com/risc0/risc0/pull/508

flaub:
This is great! Would love the ability to sort/filter the table. Maybe just the ability to export as a CSV would be sufficient so I could open it in excel. Also these are all nice to haves, not a blocker IMO.

PR #503, opened 2023-04-09
WIP: Implementation of new API

flaub:


risc0/build/src/lib.rs
```diff
@@ -441,14 +447,18 @@ impl Default for GuestOptions {
...
-    let skip_var_name = "RISC0_SKIP_BUILD";
-    println!("cargo:rerun-if-env-changed={}", skip_var_name);
-    if env::var(skip_var_name).is_ok() {
-        return;
-    }
-
     let out_dir_env = env::var_os("OUT_DIR").unwrap();
-    let out_dir = Path::new(&out_dir_env);
+    let out_dir = Path::new(&out_dir_env); // $ROOT/target/$profile/build/$crate/out
+    let guest_dir = out_dir
```
mothran, line 452: Should we perform a sanity check here to ensure the resolved `guest_dir` is a cargo `target/$profile` directory? Just in case someone invokes risc0-build in a non-standard layout.

risc0/zkvm/src/exec/mod.rs
```diff
@@ -0,0 +1,575 @@
...
+use crate::{
+    opcode::{MajorType, OpCode},
+    ExitCode, MemoryImage, Session,
+};
+
+const DEFAULT_SEGMENT_LIMIT_PO2: usize = 20; // 1M cycles
+
+const DEFAULT_SESSION_LIMIT: usize = 64 * 1024 * 1024; // 64M cycles
+
+// The number of cycles required to compress a SHA-256 block.
```
mothran, line 60: I think these should be doc comments (`///`) so `--document-private-items` of rustdoc will pick them up.

risc0/build/src/lib.rs
flaub: I actually don't know what situation would cause it to be a different layout, and I'm not sure what sanity check I should perform, any suggestions?

risc0/zkvm/src/exec/mod.rs
flaub: I'm trying hard to shrink the public interface to reduce the burden of ensuring all public interfaces have docs. Do you use a local rustdoc build with `--document-private-items`?

flaub:
@shkoo I think this is at a good landing point, tests are passing and the remaining TODOs are now related to the proving phase. Basically I think the execution phase is about 90% complete with this PR.

risc0/zkvm/src/lib.rs
```diff
@@ -23,12 +23,18 @@ extern crate alloc;
 #[cfg(feature = "binfmt")]
 pub mod binfmt;
 mod control_id;
+#[cfg(feature = "prove")]
+mod exec;
```
mothran, line 27: If a user wants to implement a syscall handler in the Executor, wouldn't this need to be pub to access the `risc0_zkvm::exec::io::{Syscall, SyscallContext}`? 

I am currently testing using the new Executor api and hit the missing SYS_RANDOM in Execute mode, so tried to implement it as a custom handler and hit this.

mothran: Makes sense to me!

tzerrell (commented):
:tada: great! I mostly was looking at the API structure rather than implementation details.

A few inline questions, plus I'm putting together a documentation PR to merge into this PR since I think we should get rid of docs TODOs wherever practical (which I believe should be everywhere but the receipt metadata).

tzerrell:
See also my suggested docs update in #509

risc0/zkvm/src/session.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub(crate) pre_image: MemoryImage,
+    pub(crate) post_image_id: Digest,
+    syscalls: Vec<SyscallRecord>,
+    pub(crate) exit_code: ExitCode,
+}
+
+impl Session {
+    /// TODO
+    pub fn new(segments: Vec<Segment>) -> Self {
+        Self { segments }
```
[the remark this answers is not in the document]
flaub: Yes, I think asserts will be very useful

risc0/zkvm/src/receipt.rs
```diff
@@ -361,3 +361,53 @@ impl Global {
...
+}
+
+impl SegmentReceipt {
+    /// TODO
+    pub fn get_metadata(&self) -> ReceiptMetadata {
+        todo!()
+    }
+
+    /// TODO
+    pub fn verify(&self) -> Result<()> {
```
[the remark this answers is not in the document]
flaub: The image_id is inside the seal, via the `ReceiptMetadata`

mothran (approved):

PR #507, opened 2023-04-10
Add with_stdin_obj and with_stdout_obj to ProverOpts

shkoo:
These make serialization/deserialization of standard input and output cleaner for user use.

### DIFF
--- a/risc0/zkvm/methods/guest/src/bin/multi_test.rs
+++ b/risc0/zkvm/methods/guest/src/bin/multi_test.rs
@@ -28,10 +28,7 @@ use risc0_zkvm::{
     sha::{Digest, Sha256},
 };
 use risc0_zkvm_methods::multi_test::{MultiTestSpec, SYS_MULTI_TEST};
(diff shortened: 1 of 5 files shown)


tzerrell:
Sorry it took me a bit to get back to this. I did end up using it for #513 -- that was going to be substantially messier without this.

There did seem to be some lifetime restrictions around using the object passed to `with_stdout_obj` that I found cumbersome. I don't know whether they are a design issue or my relative inexperience with Rust.

tzerrell (commented):
Oh, and for my actual review: This seems reasonable -- also am I correct that upcoming continuations work will make much of this obsolete? Is this primarily for getting interactive host-guest communication like in #513 working so we have useful tests in making that transition? Or am I confused about the future direction of our I/O?

If this is not intended to be largely temporary, it might be good to get people beyond me to review.

flaub:
We'll need to re-apply this to the new API if we want to keep it, and it looks like a useful addition to me.

PR #510, opened 2023-04-13
Add from_image constructor for MemoryImage

spaugh:
It's useful for me to be able to build a new MemoryImage from just the Vec<u8>. This constructor allows that.


flaub:
It seems like this won't be as useful once we add the PC, do we still need this?

PR #513, opened 2023-04-14
Restore voting-machine example

tzerrell:
The voting-machine example was accidentally omitted in the migration from https://github.com/risc0/risc0-rust-examples, this restores that example.


tzerrell:
We may choose to revive this someday, but none of the code from this PR will be helpful in doing so.

flaub (approved):

PR #514, opened 2023-04-17
Implementation of new API, part 2

flaub:
This completes the new API so that execution and proving phases work together. 

TODO:
* Revive tracing and profiling support


risc0/zkvm/src/lib.rs
```diff
@@ -49,13 +47,13 @@ pub use risc0_zkvm_platform::{declare_syscall, memory::MEM_SIZE, PAGE_SIZE};
...
 pub use self::binfmt::{elf::Program, image::MemoryImage};
+pub use self::receipt::{SegmentReceipt, SessionReceipt};
 #[cfg(feature = "prove")]
-pub use self::exec::{Executor, ExecutorEnv, ExecutorEnvBuilder};
-#[cfg(feature = "prove")]
-pub use self::prove::{loader::Loader, Prover, ProverOpts};
-pub use self::receipt::{Receipt, SegmentReceipt, SessionReceipt};
-#[cfg(feature = "prove")]
-pub use self::session::{ExitCode, Segment, Session};
+pub use self::{
```
mothran, line 52: Should we be Exposing `Syscall` and `SyscallContext` here? I might have missed if they are exposed to the public API somewhere else.

templates/rust-starter/host/src/main.rs
```diff
@@ -2,29 +2,21 @@
...
-
-    // TODO: Implement communication with the guest here
-
-    // Run prover & generate receipt
-    let receipt = prover.run().expect(
-        "Code should be provable unless it had an error or exceeded the maximum cycle limit",
-    );
+    let env = ExecutorEnv::default();
+    let mut exec = Executor::from_elf(env, METHOD_NAME_ELF).unwrap();
+    let session = exec.run().unwrap();
```
mothran, line 13: I would recommend keeping the .expect() with helper error messages. Provides a better first use experience for new users.

flaub: The reason I went this way is because I think we're going to re-write these all again soon. I think a higher-level API is probably needed to reduce the boilerplate.

risc0/zkvm/src/lib.rs
flaub: We'll definitely follow up with this in a future PR, this one is already getting pretty big :)

mothran: Sounds good to me.

risc0/zkvm/src/session.rs
```diff
@@ -81,12 +84,19 @@ pub struct Segment {
...
+
+    /// The number of cycles in powers of 2.
+    pub po2: usize,
 }
 
 impl Session {
-    /// Create a new Session from its constituent [Segment]s and journal.
-    pub fn new(segments: Vec<Segment>, journal: Vec<u8>) -> Self {
-        Self { segments, journal }
+    /// TODO
```
mothran, line 93:
```suggestion
    /// Construct a Session from Vec of [Segment], journal, and [ExitCode] 
```

risc0/zkvm/src/session.rs
```diff
@@ -63,6 +63,9 @@ pub struct Session {
 
     /// The data publicly committed by the guest program.
     pub journal: Vec<u8>,
+
+    /// TODO
```
mothran, line 67:
```suggestion
    /// [ExitCode] of final [Segment] in this session
```

risc0/zkvm/src/prove/exec.rs
```diff
@@ -802,23 +534,7 @@ impl<'a, H: HostHandler> MachineContext<'a, H> {
...
-        let mut to_guest_buf = vec![0u32; to_guest_words as usize];
-
-        self.memory.cur_cycle = cycle;
-        self.syscall_out_regs = self
-            .handler
-            .on_txrx(&self.memory, &name, &mut to_guest_buf)?;
-        trace!("SYS_IO[{cycle}] (a0, a1): {:?}", self.syscall_out_regs);
-        trace!("SYS_IO[{cycle}] data sent to guest: {to_guest_buf:?}");
-        self.syscall_out_data = to_guest_buf.into();
+    fn syscall_init(&mut self, _cycle: usize) -> Result<()> {
```
SchmErik, line 537: Can this function be removed all together? The only other reference I see is on line 229

risc0/zkvm/src/prove/exec.rs
```diff
@@ -362,245 +257,126 @@ impl<'a, H: HostHandler> CircuitStepHandler<BabyBearElem> for MachineContext<'a,
...
     }
 
-    fn trace(&mut self, cycle: usize, pc: BabyBearElem) -> Result<()> {
-        if self.trace_enabled {
-            let cycle = cycle as u32;
-            let pc = pc.into();
-            self.handler
-                .on_trace(TraceEvent::InstructionStart { cycle, pc })?
-        }
+    fn trace(&mut self, _cycle: usize, _pc: Elem) -> Result<()> {
```
SchmErik, line 366: I think this function can also be removed, similar to https://github.com/risc0/risc0/pull/514/files#r1170563305

flaub: Eventually, yes, but that will require changes to the circuit, which I didn't want to do just yet. I think there's a few minor modifications to the circuit that I plan to make soon.

risc0/zkvm/src/prove/exec.rs
flaub: Yes, sort of. Right now the circuit expects this extern to exist, but we might be able to remove it from the circuit in the future. Leaving it in for now as a note that it's still used by the circuit.

mothran (approved):
Looking great! Makes sense to continue follow up work in additional PRs.

SchmErik (approved):
LGTM!

flaub:
Thanks for all the comments!

PR #517, opened 2023-04-18
Added templ_subdir flag to cargo risczero new

mothran:
Provides the optional ability to select from multiple template locations within a folder|repo at `--template`


risc0/cargo-risczero/src/commands/new.rs
```diff
@@ -34,6 +34,14 @@ pub struct NewCommand {
     #[clap(value_parser, long, short, default_value = RISC0_GH_REPO)]
     pub template: String,
 
+    /// location of the template
```
tzerrell, line 37: Match capitalization (sorry I think you borrowed this from how I formatted my suggestion)

```suggestion
    /// Location of the template
```

tzerrell (approved):

Issue #472, opened 2023-03-27
[DevEx] Document how to enable guest profiler


flaub:

flaub:
@shkoo Can provide additional details if needed

PR #518, opened 2023-04-19
Revive profiler, add docs, cleanups

flaub:


risc0/r0vm/src/bin/r0vm.rs
```diff
@@ -40,6 +40,13 @@ struct Args {
     /// Add environment vairables in the form of NAME=value.
     #[clap(long, action = clap::ArgAction::Append)]
     env: Vec<String>,
+
+    /// Write "pprof" protobuf output of the guest's run to this file.
```
mothran, line 44: It might be nice to show a end-to-end example set of commands in a `r0vm` crate level README ( and/or landing page for crates.io) of how to use profiling / other features of r0vm so its a little more accessible to users.

risc0/zkvm/src/lib.rs
```diff
@@ -47,9 +47,12 @@ pub use risc0_zkvm_platform::{declare_syscall, memory::MEM_SIZE, PAGE_SIZE};
 
 #[cfg(feature = "binfmt")]
 pub use self::binfmt::{elf::Program, image::MemoryImage};
+#[cfg(feature = "profiler")]
+pub use self::exec::profiler::Profiler;
 pub use self::receipt::{SegmentReceipt, SessionReceipt};
 #[cfg(feature = "prove")]
 pub use self::{
+    exec::io::{Syscall, SyscallContext},
```
mothran, line 55: yay!

mothran (approved):
Looks good!

risc0/zkvm/src/lib.rs
flaub: MultiTest doesn't really test the external API, it really just tests the host/guest interaction.

risc0/r0vm/src/bin/r0vm.rs
flaub: Agreed, see: https://github.com/risc0/risc0/issues/472

risc0/zkvm/src/lib.rs
flaub: I'm also not super keen on exposing the `Syscall` interface directly, I think there are usually better ways to extend the host/guest (like using I/O). If we can drop this re-export I'd prefer to.

mothran: I am also happy with that.

tzerrell (commented):
Great!

PR #520, opened 2023-04-20
Replace default_hal with default_prover

flaub:
Now all HALs are registered and then can be selected either programmatically by name or at runtime with an environment variable (`RISC0_HAL`) to override the default. The following names are registered:
* -F default
  * `cpu`
  * `$default`
  * `cpu:poseidon`
  * `$poseidon`
* -F cuda
  * `cuda`
  * `$gpu`
  * `$default` (this overrides the CPU one)
  * `cuda:poseidon`
  * `$poseidon` (this overrides the CPU one)
* -F metal
  * `metal`
  * `$gpu`
  * `$default` (this overrides the CPU one)
  * `metal:poseidon`
  * `$poseidon` (this overrides the CPU one)

The CUDA HAL needed to be updated; `rustacuda` is replaced with `cust`. The latter has better context handling and prevents segfaults and ContextIsDestroyed errors. There's one hack added to CI to get the CUDA toolkit to be detected, this is needed until this crate gets a new release: https://github.com/Rust-GPU/Rust-CUDA/pull/72.


risc0/circuit/rv32im/benches/eval_check.rs
```diff
@@ -38,7 +38,7 @@ pub fn eval_check(c: &mut Criterion) {
     #[cfg(feature = "cuda")]
     for po2 in [2, 8, 16, 20, 21].iter() {
         let params = EvalCheckParams::new(*po2);
-        let hal = std::rc::Rc::new(risc0_zkp::hal::cuda::CudaHalSha256::new());
+        let hal = std::sync::Arc::new(risc0_zkp::hal::cuda::CudaHalSha256::new());
```
mothran, line 41: Did we switch to a Arc because the hal now cross thread boundaries?

risc0/zkp/src/hal/dual.rs
```diff
@@ -148,51 +158,56 @@ where
...
+    fn batch_expand(
+        &self,
+        output: &Self::Buffer<Self::Elem>,
+        input: &Self::Buffer<Self::Elem>,
+        count: usize,
+    ) {
         self.hal1.batch_expand(&output.buf1, &input.buf1, count);
         self.hal2.batch_expand(&output.buf2, &input.buf2, count);
-        output.assert_eq();
+        // output.assert_eq();
```
mothran, line 183: Still needed? Also a few more commented out below this line.

risc0/zkp/src/hal/dual.rs
```diff
@@ -318,46 +355,49 @@ where
...
+    H1: Hal<Field = F>,
     E1: EvalCheck<H1>,
-    H2: Hal,
+    H2: Hal<Field = F>,
     E2: EvalCheck<H2>,
-    H1::BufferElem: Buffer<H2::Elem>,
-    H1::BufferExtElem: Buffer<H2::ExtElem>,
-    H2::BufferElem: Buffer<H1::Elem>,
-    H2::BufferExtElem: Buffer<H1::ExtElem>,
+    // H1::BufferElem: Buffer<H2::Elem>,
```
mothran, line 366: Needed?

risc0/zkvm/src/prove/mod.rs
```diff
@@ -229,16 +281,87 @@ impl Segment {
...
+        table.insert("$poseidon".to_string(), prover);
+    }
+    table
+}
+
+/// Return a default [Prover] based on environment variables, falling back to a
+/// default CPU-based prover.
+pub fn default_prover() -> Arc<dyn Prover> {
+    let provers = provers();
+    if let Ok(requested) = std::env::var("RISC0_PROVER") {
```
mothran, line 337: We should probably document this in a more expanded rustdoc str and potentially the README. I think we are now getting to the point where we should consider adding all the env vars to list with explanations of each.

mothran (commented):
Looking really good, the clarity of the prover/hal interface is quite nice.

SchmErik (approved):
looks good to me. I like how this change simplifies the example code!

risc0/zkvm/src/prove/mod.rs
3lkn: This is a great idea -- creating an issue for it.

jbruestle (approved):
Awesome, so glad we no longer need to recompile to change HALs...

risc0/circuit/rv32im/benches/eval_check.rs
flaub: Initially I tried to make a singleton `OnceCell` registry of all provers, but ran into trouble with getting this to work with rustacuda. Eventually I replaced rustacude with cust, which fixed those issues, but by this point I abandonded the singleton approach. Now there's no reason to use `Arc` so I'll revert back to `Rc`.

risc0/zkp/src/hal/dual.rs
```diff
@@ -36,19 +44,21 @@ where
...
 
-    fn assert_eq(&self) {
-        self.buf1.view(|a| {
-            self.buf2.view(|b| {
-                assert_eq!(a.len(), b.len());
-                assert_eq!(a, b);
-            });
-        })
-    }
+    // fn assert_eq(&self) {
```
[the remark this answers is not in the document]
flaub: Yes, I'll just drop the `DualHal` entirely since it doesn't compile anymore. Maybe someday someone can fix this :)

Issue #380, opened 2023-02-14
Document what cargo features are available


tzerrell:
Add documentation indicating what cargo features our crates have available, what they do, and which ones are enabled by default.

flaub:
I think this is done?

PR #521, opened 2023-04-20
Adding feature flag documentation to crate READMEs

3lkn:
This PR corresponds to https://github.com/risc0/risc0/issues/380 (document what cargo features are available).

Note: two crate feature flags `_dual` and `_binfmt` have been omitted from documentation and are now marked by underscore prefix to hide them from rustdoc:
* `_dual` is currently possibly not functional, used primarily for internal A/B testing of HALs
* `_binfmt` is also an internal tool for build system efficiency (related to rebuilding .cpp files)


risc0/circuit/rv32im/README.md
```diff
@@ -1,2 +1,18 @@
...
+The RISC Zero zkVM circuit
+
+# Crate Feature Flags
+
+The following crate feature flags are available. They are configured in your `Cargo.toml`.
+
+* cuda
+  * Optional (default is CPU only)
+  * Turns on GPU acceleration for guest code with CUDA
+* metal
```
mothran, line 11: For all `metal` flags it might be worth noting that it is only supported on `macos` OS's

risc0/zkvm/README.md
```diff
@@ -27,3 +27,23 @@ channel](https://www.youtube.com/@risczero) has many more videos as well.
...
   when it gets resolved to a permanent solution.
 * And more on [the RISC Zero website](https://www.risczero.com/)!
+
+# Crate Feature Flags
+
+The following crate feature flags are available. They are configured in your `Cargo.toml`.
+
+* cuda
+  * Optional (default is CPU only)
+  * Turns on GPU acceleration for guest code with CUDA
```
flaub, line 37: This isn't acceleration for guest code, it's acceleration specifically for the prover. Basically `prove` must be set when `metal` or `cuda` is selected.

risc0/zkvm/README.md
```diff
@@ -27,3 +27,23 @@ channel](https://www.youtube.com/@risczero) has many more videos as well.
...
+* metal
+  * Optional (default is CPU only)
+  * Turns on GPU acceleration for guest code with Metal
+* profiler
+  * Optional, experimental
+  * Use with `r0vm` binary tool to profile guest code
+  * Counts cycles during guest execution as an aid to code optimization
+* prove
+  * Enabled by default
+  * Disable this feature if you only need to use the verifier
```
flaub, line 47: Or if you want to use the guest module in guest-side code

risc0/zkvm/README.md
```diff
@@ -27,3 +27,23 @@ channel](https://www.youtube.com/@risczero) has many more videos as well.
...
+The following crate feature flags are available. They are configured in your `Cargo.toml`.
+
+* cuda
+  * Optional (default is CPU only)
+  * Turns on GPU acceleration for guest code with CUDA
+* metal
+  * Optional (default is CPU only)
+  * Turns on GPU acceleration for guest code with Metal
+* profiler
+  * Optional, experimental
```
flaub, line 42: I don't think this is experimental anymore

flaub: It's optional mostly because it introduces a lot of dependencies (like protobuf) that are pretty heavy and aren't needed most of the time

risc0/circuit/rv32im/README.md
3lkn: That's a really good point -- folks who don't use Mac might not recognize it as MetalTM. I wonder if we should also note support issues w/ CUDA on Mac or just leave it as an exercise for the reader since it's complicated.

risc0/zkvm/README.md
3lkn: Ok I'll pull this out of the code comments as well then.

risc0/build/README.md
```diff
@@ -53,4 +53,8 @@ be named `methods::MULTIPLY_ELF`. These are included at the beginning
 of the host-side code:
 ```text
 use methods::{MULTIPLY_ELF, MULTIPLY_ID};
-```
\ No newline at end of file
+```
+
+# Crate Feature Flags
+
+There are currently no feature flags available for `build`.
```
tzerrell, line 60: I'm not sure we even need this section for a crate with no feature flags (but probably it doesn't cause any harm, either).

risc0/circuit/rv32im/README.md
```diff
@@ -1,2 +1,20 @@
 
-The RISC Zero zkVM circuit
\ No newline at end of file
+The RISC Zero zkVM circuit
+
+# Crate Feature Flags
+
+The following crate feature flags are available. They are configured in your `Cargo.toml`.
```
tzerrell, line 6: Probably nice to link the cargo docs about feature flags here (and in the analogous parts of other readmes): https://doc.rust-lang.org/cargo/reference/features.html. With that linked, you may not need the  line about "configured in `Cargo.toml`".

risc0/zkvm/README.md
```diff
@@ -27,3 +27,25 @@ channel](https://www.youtube.com/@risczero) has many more videos as well.
...
+  * Optional (default is CPU only)
+  * Turns on Metal GPU acceleration for the prover
+  * Only available for MacOS
+* profiler
+  * Optional
+  * Use with `r0vm` binary tool to profile guest code
+  * Counts cycles during guest execution as an aid to code optimization
+* prove
+  * Enabled by default
+  * Disable this feature if you only need to use the verifier, or if you want to use the `guest` module in guest-side code
```
tzerrell, line 48: I would flip the order on this to highlight the more important "zkvm in guest" case first: Using `guest::env` in the guest shows up in literally all of our examples. Running the verifier w/out the prover is another use, and I do think it's worth mentioning, but it's not even strictly necessary to exclude the prover in that case (you just build unnecessary stuff). Whereas the build fails if you try to make guest code depend on `risc0_zkvm` without disabling the `prove` feature.

I'd make the analogous change in the zkp readme as well.

tzerrell (commented):
Great! I have a few inline comments

risc0/build/README.md
3lkn: Ah, I was thinking it would disambiguate between "I forgot to add this" and "These are not present" but if it's odd enough to stick out I'll pull it

risc0/zkvm/Cargo.toml
```diff
@@ -68,11 +68,11 @@ tar = "0.4"
 test-log = { version = "0.2", features = ["trace"] }
 
 [features]
-binfmt = ["dep:elf", "std"]
+_binfmt = ["dep:elf", "std"]
 cuda = ["risc0-circuit-rv32im/cuda", "risc0-zkp/cuda"]
 metal = ["risc0-circuit-rv32im/metal", "risc0-zkp/metal"]
 default = ["prove"]
-dual = []
+_dual = []
```
flaub, line 75: It should be safe to drop this now

3lkn: Feature flag is dropped.

risc0/zkvm/README.md
3lkn: Tables added to top-level README and to all relevant crate-level READMEs.
Language updated for "zkvm in guest" case and order flipped in all crate notes.

tzerrell (approved):

README.md
```diff
@@ -138,6 +138,18 @@ structured, and other resources useful to developers new to RISC Zero, see our
 | risc0-zkvm               | [![x](https://img.shields.io/badge/crates.io-v0.14-orange)](https://crates.io/crates/risc0-zkvm)               | [![](https://img.shields.io/docsrs/risc0-zkvm)](https://docs.rs/risc0-zkvm)                             |
 | risc0-zkvm-platform      | [![x](https://img.shields.io/badge/crates.io-v0.14-orange)](https://crates.io/crates/risc0-zkvm-platform)      | [![](https://img.shields.io/docsrs/risc0-zkvm-platform)](https://docs.rs/risc0-zkvm-platform)           |
 
+## Feature flags
+
+The following feature flags are present in one or more of the crates listed above:
+
+| Feature | Target(s)  | Implies | Description | Crates |
+| --- | --- | --- | ---- | --- |
+| cuda |windows, linux| prove, std | Turns on CUDA GPU acceleration for the prover | rv32im, zkp, zkvm |
```
flaub, line 147: I think the crate names need to be the full names, not shorthands

risc0/zkp/README.md
```diff
@@ -1,3 +1,30 @@
...
+
+The following [crate feature flags](https://doc.rust-lang.org/cargo/reference/features.html) are available.
+
+
+| Feature | Target(s)  | Implies | Description |
+| --- | --- | --- | ---- |
+| cuda |windows, linux| prove, std | Turns on CUDA GPU acceleration for the prover |
+metal |macos|prove|Turns on Metal GPU acceleration for the prover
+prove|all except rv32im|std| Disable this feature if you want to use the `guest` module in guest-side code
+std|all|--|Support for stdlib in the guest, including `std::io` read and write in the guest
```
flaub, line 15: I don't know why `std::io` is being called out specially here

flaub: This isn't specific to the guest either, this just enables std support, period.

README.md
tzerrell: Oh, good call -- perhaps ideally even docs.rs links? Full name feels more definitely important, though.

flaub (approved):

PR #523, opened 2023-04-21
Add Clone derive to Segment

spaugh:
Needed for storing Segments.


flaub (approved):
I would try to avoid doing clones since it's going to be fairly expensive, but I also don't see any issue with adding this derive.

mothran (approved):

PR #524, opened 2023-04-21
Poseidon 254

jbruestle:
This adds experimental support for a different version of Poseidon which is efficient to use inside SNARKs used on ETH as a step in building a STARK -> SNARK recursion.  It is largely untested, but should have no impact on the normal code path.


risc0/zkp/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 
 extern crate alloc;
+extern crate ff;
```
flaub, line 21: This shouldn't be needed

jbruestle: If I don't put it there I get a ton of errors.

flaub (approved):

PR #525, opened 2023-04-21
Added index to Segment / SegmentReceipt

mothran:
Useful for reconstructing a SessionReceipt from a unordered list of SegmentReceipts.

### DIFF
--- a/risc0/zkvm/src/exec/mod.rs
+++ b/risc0/zkvm/src/exec/mod.rs
@@ -165,6 +165,7 @@ impl<'a> Executor<'a> {
             .with_write_fd(fileno::JOURNAL, journal.clone());
 
         let mut run_loop = || -> Result<ExitCode> {
+            let mut idx = 0;
             loop {
                 if let Some(exit_code) = self.step()? {
                     let total_cycles = self.total_cycles();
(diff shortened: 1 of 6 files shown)


risc0/zkvm/src/exec/mod.rs
```diff
@@ -183,7 +184,9 @@ impl<'a> Executor<'a> {
                         syscalls,
                         exit_code,
                         log2_ceil(total_cycles.next_power_of_two()),
+                        idx,
```
flaub, line 187:
```suggestion
                        segments.len(),
```

risc0/zkvm/src/receipt.rs
```diff
@@ -134,6 +134,9 @@ pub struct SegmentReceipt {
     /// [SegmentReceipt::verify]) to cryptographically prove that this Segment
     /// was faithfully executed.
     pub seal: Vec<u32>,
+
+    /// Segment index within the [SessionReceipt]
+    pub index: u64,
```
flaub, line 139: 4 billion segments seems like a lot :)

flaub (approved):

risc0/zkvm/src/receipt.rs
mothran: I like to be optimistic hahaha, but u32 is more portable. Fixed in: c041649754ba2478d005ad8b77cd01fe1c8102d5

risc0/zkvm/src/exec/mod.rs
mothran: Fixed in: c041649754ba2478d005ad8b77cd01fe1c8102d5 with additional removal of the idx var.

PR #527, opened 2023-04-21
Add mdBook stub

tzerrell:
This PR starts a new mdBook in `docs/mdbook`, and adds it to CI.

Currently this is a stub. Also, CI includes running tests (although there is nothing to test yet), but does not include a deployment, on the principle that we shouldn't deploy something that doesn't have content, and that even as we add content it will be ugly for a while. I did write a short list of links to other content for the stub, so if we think it's important to be deploying this from day 1 we could set that up and not confuse our users too badly. It does deploy successfully locally with `mdbook serve --open`.

CI appears to be working fine, giving the same output as on my local machine, but I'm interested in any feedback about whether organizing this into the `doc` check was the right approach.


.github/workflows/main.yml
```diff
@@ -101,6 +101,13 @@ jobs:
       - uses: ./.github/actions/rustup
       - uses: ./.github/actions/sccache
       - run: RISC0_SKIP_BUILD=1 cargo doc --no-deps --exclude=risc0-zkvm-methods --workspace
+      - name: Install mdbook
+        uses: risc0/cargo-install@2cfb0024938d23011106cbf127b393bc83fddba1
+        with:
+          crate: mdbook
+          version: "0.4"
+      - run: mdbook test
```
mothran, line 109: This looks good, might be good to add a single test to verify its working as expected.

We could look into auto publishing (on main or on-release workflows): https://github.com/peaceiris/actions-mdbook Because we already have a GH token to publish to the `benchmarks` repo which has a public HTML page: https://risc0.github.io/benchmarks/

flaub: Agreed and then we could probably just rename `benchmarks` to something more generic, "assets", "internal-website", ???.

mothran: Agreed, it got overloaded when I started pushing crate-validator results there. I had it on my TODO to rename that repo but required changing a few other things. I can probably do that sooner though if we land on a name.  `generated-content` might work?

tzerrell: The only reason I was leaving out auto-publishing was to "not include a deployment, on the principle that we shouldn't deploy something that doesn't have content" -- but it's not like this has content we don't want anyone to see. So I'll look into doing auto publishing, thanks for the references!

.github/workflows/mdbook.yml
```diff
@@ -0,0 +1,28 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup mdBook
+        uses: peaceiris/actions-mdbook@v1
```
mothran, line 19: TBD: review the code and fork it to `risc0` org.

tzerrell:
Based on in-person discussion my plan is to roll back the deployment portion of this PR and split it off into its own PR. Then this can land before we've fully sorted out the deployment story.

flaub (approved):

PR #528, opened 2023-04-24
Keep CUDA context alive

flaub:


risc0/zkp/src/hal/cuda.rs
```diff
@@ -42,6 +43,15 @@ use crate::{
 
 const KERNELS_FATBIN: &[u8] = include_bytes!(env!("ZKP_CUDA_PATH"));
 
+lazy_static! {
+    static ref CONTEXT: Context = {
+        let device = Device::get_device(0).unwrap();
```
mothran, line 48: Could we have a ENV var with a default of `0` could be useful for future multi-GPU hosts.

mothran (approved):

risc0/zkp/src/hal/cuda.rs
flaub: Yes, I plan to do that in the future

PR #534, opened 2023-04-25
Added pc to MemoryImage

mothran:
I think I got all the places we need to update it in `Executor` but I might be missing something.
I was unsure of the right way to unit test it so happy to hear any suggestions there.


risc0/zkvm/src/exec/mod.rs
```diff
@@ -217,6 +217,7 @@ impl<'a> Executor<'a> {
         self.insn_counter = 0;
         self.segment_cycle = self.init_cycles;
         self.pre_pc = self.pc;
```
SchmErik, line 219: Do we still need this field?

mothran: ooo great catch! I think I can delete even more fields! I was even able to drop a clippy warning suppression in the `Session::new()` method.

Fixed in: 56ae572c776e00d4dfe9d1631d71bfc41a2f883c

PR #535, opened 2023-04-25
zkVM: add the ability to use bonsai to register binaries and run proofs

SchmErik:
This allows running bonsai from some zkVM functions in order to run proofs. This is intended for "dogfooding" so that we're running bonsai as much as possible. For now, we'll run proofs and retrieve the proof receipts from bonsai.


SchmErik:
@mothran do you have any suggestions on how to do implement tests? With the code how it is now, I need to set `BONSAI_DOGFOOD_URL` environment variable when doing `cargo test` which could reveal information that we want to keep private...

mothran:
> @mothran do you have any suggestions on how to do implement tests? With the code how it is now, I need to set `BONSAI_DOGFOOD_URL` environment variable when doing `cargo test` which could reveal information that we want to keep private...

I would look into if there are proc_macro's that can disable certain unit tests if a given env var is not present until the API is accessible from the CI machines.

risc0/zkvm/Cargo.toml
```diff
@@ -49,8 +50,20 @@ num-traits = { version = "0.2", default-features = false }
...
 rand = { version = "0.8", optional = true }
 rayon = { version = "1.5", optional = true }
+reqwest = { version = "0.11.16", features = [ "blocking", "json" ] }
 rrs-lib = { version = "0.1", optional = true }
 sha2 = { version = "0.10", optional = true }
+serde_json = "1.0.96"
+serde_with = { version = "2.0.0", features = ["time_0_3"] }
+time = "0.3.11"
+utoipa = "3.3.0"
+sqlx = { version = "0.6.2", features = [
```
mothran, line 60: Do we need a database driver? The types should only need serde traits. This brings in a lot of code to the tree.

risc0/zkvm/Cargo.toml
```diff
@@ -49,8 +50,20 @@ num-traits = { version = "0.2", default-features = false }
 prost = { version = "0.11", optional = true }
 rand = { version = "0.8", optional = true }
 rayon = { version = "1.5", optional = true }
+reqwest = { version = "0.11.16", features = [ "blocking", "json" ] }
 rrs-lib = { version = "0.1", optional = true }
 sha2 = { version = "0.10", optional = true }
+serde_json = "1.0.96"
+serde_with = { version = "2.0.0", features = ["time_0_3"] }
+time = "0.3.11"
+utoipa = "3.3.0"
```
mothran, line 59: We shouldn't need this for the client side.

risc0/zkvm/Cargo.toml
```diff
@@ -49,8 +50,20 @@ num-traits = { version = "0.2", default-features = false }
 prost = { version = "0.11", optional = true }
 rand = { version = "0.8", optional = true }
 rayon = { version = "1.5", optional = true }
+reqwest = { version = "0.11.16", features = [ "blocking", "json" ] }
 rrs-lib = { version = "0.1", optional = true }
 sha2 = { version = "0.10", optional = true }
+serde_json = "1.0.96"
```
mothran, line 56: Where is this used?

risc0/zkvm/src/exec/mod.rs
```diff
@@ -143,19 +155,55 @@ impl<'a> Executor<'a> {
...
+        ))
     }
 
     /// Run the executor until [ExitCode::Paused] or [ExitCode::Halted] is
     /// reached, producing a [Session] as a result.
     pub fn run(&mut self) -> Result<Session> {
+        // Bonsai only needs the proof ID to retrieve the SessionReceipt. So a "session"
+        // can be represented by a proof ID.
+        if std::env::var("BONSAI_DOGFOOD_URL").is_ok() {
+            println!("created special session");
```
mothran, line 198: Remove println! use log::* macros.

risc0/zkvm/src/session.rs
```diff
@@ -66,6 +66,9 @@ pub struct Session {
 
     /// The [ExitCode] of the session.
     pub exit_code: ExitCode,
+
+    /// The proof ID used for Bonsai execution
+    pub proof_id: Option<i64>,
```
mothran, line 71: I would make this pub(crate) and consider hiding it from serialization with: `#[serde(skip_serializing)]`

risc0/zkvm/src/exec/mod.rs
```diff
@@ -143,19 +155,55 @@ impl<'a> Executor<'a> {
...
     /// Construct a new [Executor] from an ELF binary.
     pub fn from_elf(env: ExecutorEnv<'a>, elf: &[u8]) -> Result<Self> {
         let program = Program::load_elf(&elf, MEM_SIZE as u32)?;
         let image = MemoryImage::new(&program, PAGE_SIZE as u32)?;
-        Ok(Self::new(env, image, program.entry))
+        let mut bonsai_proof_id: Option<i64> = None;
+
+        // TODO: move this to the run() function once we switch the ELF registration to
+        // use MemoryImages.
+        if std::env::var("BONSAI_DOGFOOD_URL").is_ok() {
```
mothran, line 170: Here you can use a `if let Some(...) = ....` pattern: https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html

risc0/zkvm/src/prove/mod.rs
```diff
@@ -227,6 +227,13 @@ where
     }
 
     fn prove_session(&self, session: &Session) -> Result<SessionReceipt> {
+        if std::env::var("BONSAI_DOGFOOD_URL").is_ok() {
```
mothran, line 230: if-let here.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -227,6 +227,13 @@ where
...
+            println!("running bonsai prove");
```
mothran, line 231: println

risc0/zkvm/src/bonsai_api.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Once we are able to use the proof ID to get the receipt, we can get rid of
+    // this call.
+    log::debug!("Getting receipt hash for proof request {}", proof_id);
+    let mut proof_request: ProofRequest;
+    loop {
+        let res = client
+            .get(format!("{bonsai_url}/{REQUEST_PATH}/{proof_id}"))
+            .header(reqwest::header::CONTENT_TYPE, "application/json")
+            .send()
+            .unwrap();
```
mothran, line 219: All these unwraps should probably bubble up via `?` instead.

risc0/zkvm/src/exec/mod.rs
```diff
@@ -125,6 +126,17 @@ impl Write for Journal {
 impl<'a> Executor<'a> {
     /// Construct a new [Executor] from a [MemoryImage] and entry point.
     pub fn new(env: ExecutorEnv<'a>, image: MemoryImage, pc: u32) -> Self {
+        Self::new_with_id(env, image, pc, None)
+    }
+
+    /// Construct a new [Executor] from a [MemoryImage], entry point, along with
+    /// the proof ID generated from bonsai
+    fn new_with_id(
```
mothran, line 134: I know @flaub and I chatted about making a Executor trait like interface, similar to how the Prover pluggable system works. I think this is a good use case for that new API.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -227,6 +227,13 @@ where
     }
 
     fn prove_session(&self, session: &Session) -> Result<SessionReceipt> {
```
mothran, line 229: The `Prover` trait system lets us have pluggable provers, I would recommend having the Bonsai based prover be a implementation of that which we can register / fetch via get_prover(). Then in `default_prover()` check for the env var and swap to the BonsaiProver.

mothran (changes requested):
Lots of little style things. My big items are:

* why do we have so many REST / SQL based imports? I don't think we need nearly as many for the serialization of those structures to REST.
* Should we consider moving this behind a feature flag?
* Moving things to Prover + Executor (TBD) traits I think will be the long term place for these to live, but we currently only have the Prover interface and not the Executor.

risc0/zkvm/Cargo.toml
SchmErik: shoot! I had unintentionally left these in there from when I lifted the struct definitions! Thanks for catching it!

SchmErik:
@mothran I implemented a `RemoteProver`. I saw that we needed access `segments[0].preimage.get_root()` for the image ID to call verify and it kind of seemed a little clunky in my case. I added an `Optional<Digest>` in session representing this instead.. Let me know what you think...

risc0/zkvm/src/prove/mod.rs
```diff
@@ -337,6 +381,12 @@ fn provers() -> HashMap<String, Rc<dyn Prover>> {
 /// default CPU-based prover.
 pub fn default_prover() -> Rc<dyn Prover> {
     let provers = provers();
+    if std::env::var("BONSAI_DOGFOOD_URL").is_ok() {
+        println!("give bonsai prover");
```
mothran, line 385: println

mothran:
> @mothran I implemented a `RemoteProver`. I saw that we needed access `segments[0].preimage.get_root()` in `LocalProver` to get the image ID to call verify and it kind of seemed a little clunky in my case. I added an `Optional<Digest>` in session representing this instead.. Let me know what you think... I remember Frank talking about the pluggable executor thing too. I'll give this a try tomorrow. I think it'll make things cleaner.

Yeah I see, I guess I would have probably opted to access the the pre_image from segment[0] vs add a new attribute. But it makes sense either way to me.

SchmErik:
superseded by #677

PR #539, opened 2023-04-27
Added running WASM within the zkvm example

mothran:
Added a example the runs the supplied WASM bytecode within the zkvm and commits the results. Using a simple `fib()` WAT module as an example function.


examples/wasm/methods/guest/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let wasm: Vec<u8> = env::read();
+    let iters: i32 = env::read();
+
+    // Derived from the wasmi example: https://docs.rs/wasmi/0.29.0/wasmi/#example
+    let module = Module::new(&engine, &mut &wasm[..]).expect("Failed to create module");
+    type HostState = u32;
+
+    let linker = <Linker<HostState>>::new(&engine);
+    let mut store = Store::new(&engine, 42);
```
tzerrell, line 35: Why does this store `42`? This feels like it may be an accidental relic of the source example...

tzerrell (approved):
Awesome!

flaub (approved):

examples/wasm/methods/guest/src/main.rs
mothran: Yeah I think its an artifact of https://docs.rs/wasmi/latest/wasmi/#example and it requires some data to be passed with the correct type: https://docs.rs/wasmi/latest/wasmi/struct.Store.html#method.new

tzerrell: :+1: gotcha, so the value is arbitrary and unused, but having _something_ there is required.

PR #540, opened 2023-04-28
API adjustment: executor segment callback

flaub:


risc0/zkvm/src/exec/mod.rs
```diff
@@ -209,9 +220,11 @@ impl<'a> Executor<'a> {
         };
 
         let exit_code = run_loop()?;
-        let mut segments = Vec::new();
-        std::mem::swap(&mut segments, &mut self.segments);
-        Ok(Session::new(segments, journal.buf.take(), exit_code))
+        Ok(Session::new(
+            take(&mut self.segments),
```
nategraf, line 224: Is there a sensible default to be left in the place of `self.segments`?

risc0/zkvm/src/exec/mod.rs
```diff
@@ -159,6 +159,15 @@ impl<'a> Executor<'a> {
...
     /// reached, producing a [Session] as a result.
     pub fn run(&mut self) -> Result<Session> {
+        self.run_with_callback(|segment| Box::new(SimpleSegmentRef::new(segment)))
+    }
+
+    /// Run the executor until [ExitCode::Paused] or [ExitCode::Halted] is
+    /// reached, producing a [Session] as a result.
+    pub fn run_with_callback<F>(&mut self, mut callback: F) -> Result<Session>
+    where
+        F: FnMut(Segment) -> Box<dyn SegmentRef>,
```
nategraf, line 169: Should like be a result? I'm guessing that that a callback may, for instance, upload the `Segment` to an S3 bucket, or otherwise store it at some network accessible location.
```suggestion
        F: FnMut(Segment) -> Result<Box<dyn SegmentRef>>,
```

nategraf: Another option on this would be to make either `Executor` or `Executor::run` generic over some kind of `SegmentStorage` trait. It would functionally be the same, just using an object instead of a closure. E.g. 

```rust
trait SegmentStore {
    type Ref: SegmentRef;

    fn store(&mut self, segment: Segement) -> Result<Self::Ref>;
}

pub struct SimpleSegmentStore;

impl SegmentRef for Segment {
    fn resolve(&self) -> Result<Segment> { Ok(self.clone()) }
}

impl SegmentStorage for SimpleSegmentStore {
    type Ref = Segment;

    fn store(&mut self, segment: Segement) -> Result<Self::Ref> { Ok(Segment) };
}

impl Executor {
...
    pub fn run(&mut self) -> Result<Session> {
        self.run_with_store(SimpleSegmentStore)
    }

    pub fn run_with_callback(&mut self, callback: &mut impl SegmentStore) -> Result<Session> {
        ...
    }
...
}
```

nategraf (commented):
If having execution happen in a `Iterator<Item = Segment>` object, which it seems unclear if this is better. A solution to extracting the `Session` as the end is to have the object implement a `self` consuming function that returns the session.

```rust
pub struct Exection { ... }

impl Iterator for Execution {
    type Item = Segment
    ...
}

impl Execution {
    fn session(self) -> Option(Session) { ... }
}
```
Unclear if that actually is a reasonable solution or improves over what's here already, but I was thinking about it after reading the Slack thread.

risc0/zkvm/src/exec/mod.rs
flaub: Yes, for sure the callback should return a `Result`. `SegmentStorage` is an interesting idea.

flaub:
> If having execution happen in a `Iterator<Item = Segment>` object, which it seems unclear if this is better. A solution to extracting the `Session` as the end is to have the object implement a `self` consuming function that returns the session.
> 
> ```rust
> pub struct Exection { ... }
> 
> impl Iterator for Execution {
>     type Item = Segment
>     ...
> }
> 
> impl Execution {
>     fn session(self) -> Option<Session> { ... }
> }
> ```
> 
> Unclear if that actually is a reasonable solution or improves over what's here already, but I was thinking about it after reading the Slack thread. I feel like the calling pattern would be a bit weird.

Another use case is to be able to run the Executor again, maybe multiple times. The first time may hit a pause, but the next one may hit a halt.

risc0/zkvm/src/exec/mod.rs
flaub: Yes, it's whatever `Vec::default()` is, which is an empty vector.

risc0/zkvm/src/exec/mod.rs
flaub: After thinking about this, I like the callback a little better because it allows more choice by the caller. If the caller wants to use a trait, they can do that and wire up the callback to a function on the trait. But from the Executor's point of view, only a call is needed. I would go with this trait being part of the `run_with_callback` if there was more than just a call that was needed.

tzerrell (approved):
Looks reasonable from a DevEx perspective.

PR #541, opened 2023-04-30
Changes to the RISC-V circuit semantics for recursions

jbruestle:


jbruestle:
This commit makes a number of changes to make the RISC-V circuit compliant with it's description in the recursion RFC and fixes some potential sources of bugs.
1) Update the circuit to a new version, which 
  a) outputs the the full system exit code, and add a user exit code 
  b) Makes the output process atomic (part of halt) by doing it during 2 additional reset cycles
  c) Adds an 'input' global and provides a new system call to access it (replacing the output  call)
2) Make the requires updates to the rust zkvm code
3) Correctly resets the journal commitment on pause
4) Update the riscv tests (adding one instruction to set output, which is not required)

flaub (approved):

PR #543, opened 2023-05-02
Added libm externs to zkvm guest

mothran:
Fixes #465 without the git based dep.

Added a module into the zkvm::guest module that exports the extern versions of the libm functions. Additional added unit tests to ensure it works correctly and removed the externc-libm from evm / wasm examples.


flaub (approved):
Nice!

SchmErik (approved):

PR #545, opened 2023-05-03
serde: add support to serialize and deserialize 32 bit and 64 bit floating point values

SchmErik:
This pull request enables 32 bit and 64 bit floating point values to be used as inputs to guest programs.

### DIFF
--- a/risc0/zkvm/src/prove/tests.rs
+++ b/risc0/zkvm/src/prove/tests.rs
@@ -44,6 +44,20 @@ fn prove_nothing(name: &str) -> Result<SessionReceipt> {
 }
 
 #[test]
+fn fp_inputs() {
+    let input = to_vec(&MultiTestSpec::Float).unwrap();
+    let env = ExecutorEnv::builder()
+        .add_input(&input)
+        .add_input(&to_vec(&3.14f32).unwrap())
+        .add_input(&to_vec(&2.71f64).unwrap())
+        .build();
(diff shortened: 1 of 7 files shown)


risc0/zkvm/src/prove/tests.rs
```diff
@@ -43,6 +43,20 @@ fn prove_nothing(name: &str) -> Result<SessionReceipt> {
...
+        .add_input(&input)
+        .add_input(&to_vec(&3.14f32).unwrap())
+        .add_input(&to_vec(&2.71f64).unwrap())
+        .build();
+    let mut exec = Executor::from_elf(env, MULTI_TEST_ELF).unwrap();
+    let session = exec.run().unwrap();
+    let prover = get_prover("$default");
+    prover.prove_session(&session).unwrap();
+}
+
```
flaub, line 59: Since running the prover is pretty expensive, and since the prover doesn't really care about whether we use floating point or not, I think this test has marginal utility. I see we have coverage in the serde tests, so I feel like that's probably sufficient.

flaub, line 59: Since running the prover is pretty expensive, and since the prover doesn't really care about whether we use floating point or not, I think this test has marginal utility. I see we have coverage in the serde tests, so I feel like that's probably sufficient.

flaub (approved):

mothran (approved):

PR #546, opened 2023-05-03
Always provide risc0_zkvm::guest module for easier analyzing, even when not on target=zkvm

shkoo:
* This allows rust-analyzer to analyze guest code without having to configure it to cross compile for the zkvm target.
* Fixed flaky integration test by remapping absolute pathnames to relative pathnames when compiling the guest

### DIFF
--- a/risc0/build/src/lib.rs
+++ b/risc0/build/src/lib.rs
@@ -382,7 +382,18 @@ fn build_guest_package<P>(
 
     let mut cmd = Command::new(cargo);
     let mut child = cmd
-        .env("CARGO_ENCODED_RUSTFLAGS", "-C\x1fpasses=loweratomic")
+        .env(
+            "CARGO_ENCODED_RUSTFLAGS",
+            [
+                // Replace atomic ops with nonatomic versions since the guest is single threaded.
+                "-C",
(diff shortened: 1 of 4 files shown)


mothran (approved):

risc0/build/src/lib.rs
```diff
@@ -382,7 +382,18 @@ fn build_guest_package<P>(
 
     let mut cmd = Command::new(cargo);
     let mut child = cmd
-        .env("CARGO_ENCODED_RUSTFLAGS", "-C\x1fpasses=loweratomic")
+        .env(
+            "CARGO_ENCODED_RUSTFLAGS",
+            [
+                // Replace atomic ops with nonatomic versions since the guest is single threaded.
+                "-C",
+                "passes=loweratomic",
```
mothran, line 390: Quick question: would this impact our crate compatibility at all? Or will it just effect the deeper lowering phase?

shkoo: It just affects the lowering phase, and the "passes=loweratomic" was already there; I just reformatted some. :)

flaub:
Is there a way to add CI tests to ensure this doesn't get broken in the future?

PR #547, opened 2023-05-04
Use crypto-bigint instead of num-bigint

nategraf:
As part of working on the interfaces to the bigint accelerator, I am switching usages of
num-bigint to crypto-bigint, which is better suited to the fixed-width computations we intend.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -2234,7 +2234,6 @@ dependencies = [
  "autocfg",
  "num-integer",
  "num-traits",
- "rand",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


SchmErik (approved):
This looks good to me! I have a question but it's more about about the structure of the bigint circuit so I'll ask internally

SchmErik:
@nategraf it looks like your commits are unverified

nategraf:
> @nategraf it looks like your commits are unverified

Ah, ya I need to add my new signing key to my account. I'll do that on Monday before merging this PR.

nategraf:
> @nategraf it looks like your commits are unverified

Ah, ya I need to add my new signing key to my account. I'll do that on Monday and before merging this PR.

PR #550, opened 2023-05-08
Fix spurious build warning about "duplicate package `wordle`".

choongng:
It appears having two packages with the same name can cause this error even when they are (I think) not built for the same target. Fix by renaming the guest code in the wordle example. Tests for wordle pass.


mothran (approved):
Good catch!

flaub (approved):
Nice, thanks!

flaub:
@tzerrell Could you ensure this is documented someplace?

mothran:
> I see the CUDA machine failed CI but it doesn't seem like there's any possible way this is related. What should I do in this situation?

known issue in the CI right now, working up a patchwork fix now.

PR #556, opened 2023-05-10
Fix executor by accounting for extra cycles and patch continuations test for CI

SchmErik:


mothran (commented):
Nice find! This all makes sense to me, I will defer to @flaub review for final approval though.

flaub (approved):

PR #558, opened 2023-05-10
fix trace test to account for possible page-in cycles

SchmErik:


jbruestle (approved):

risc0/zkvm/src/exec/tests.rs
```diff
@@ -458,7 +458,17 @@ fn trace() {
...
             }] = window
             {
-                assert_eq!(cycle1 + 1, cycle2, "li should take 1 cycles: {:#?}", window);
+                // Note: it's possible that these instructions could lie between page
+                // boundaries. If that is the case, it means that the difference between cycle2
+                // and cycle1 could be multiples of page-in, which takes up 1094 cycles. Once we
+                // figure out a way to get reproducible builds, we should restrict the
+                // difference of cycles to a single number rather than taking the mod.
+                assert_eq!(
+                    (cycle2 - cycle1) % 1094,
```
flaub, line 467: I think this magic number should be a constant that is named.

PR #560, opened 2023-05-11
Add `FileSegmentRef`

tzerrell:
Add a variety of `SegmentRef` that saves the segment to a file (at a user-specified location). Also, use this `FileSegmentRef` in the EVM example.

### DIFF
--- a/examples/evm/src/main.rs
+++ b/examples/evm/src/main.rs
@@ -90,7 +90,8 @@ async fn main() {
         .add_input(&to_vec(&zkdb).unwrap())
         .build();
     let mut exec = Executor::from_elf(env, EVM_ELF).unwrap();
-    let session = exec.run_with_callback(|segment| Ok(Box::new(FileSegmentRef::new(&segment, &temp_dir())?))).unwrap();
+    let segment_dir = temp_dir();
+    let session = exec.run_with_callback(|segment| Ok(Box::new(FileSegmentRef::new(&segment, &segment_dir)?))).unwrap();
     let receipt = session.prove().unwrap();
 
     let res: EvmResult = from_slice(&receipt.journal).expect("Failed to deserialize EvmResult");
--- a/examples/evm/src/main.rs
+++ b/examples/evm/src/main.rs
@@ -25,7 +25,7 @@ use evm_methods::EVM_ELF;
 use log::info;
 use risc0_zkvm::{
     serde::{from_slice, to_vec},
-    Executor, ExecutorEnv,
+    Executor, ExecutorEnv, TempFileSegmentRef,
 };
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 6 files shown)


examples/evm/src/main.rs
```diff
@@ -86,7 +90,7 @@ async fn main() {
         .add_input(&to_vec(&zkdb).unwrap())
         .build();
     let mut exec = Executor::from_elf(env, EVM_ELF).unwrap();
-    let session = exec.run().unwrap();
+    let session = exec.run_with_callback(|segment| Ok(Box::new(FileSegmentRef::new(&segment, &temp_dir())?))).unwrap();
```
mothran, line 93: I think this calls temp_dir() each callback so each segment would get its own new directory. I would consider calling it once before run_with_callback and then just passing a reference on each callback.

risc0/zkvm/src/session.rs
```diff
@@ -155,3 +160,36 @@ impl SimpleSegmentRef {
...
+        file.read_to_end(&mut contents)?;
+        let segment: Segment = bincode::deserialize(&contents)?;
+        Ok(segment)
+    }
+}
+
+impl FileSegmentRef {
+    /// Construct a [FileSegmentRef]
+    ///
+    /// This builds a FileSegmentRef that stores `segment` in a file at `path`.
```
mothran, line 187: This is a little bit of a odd suggestion but should we consider linking to the EVM code in the docstr kinda like: "if you wanna see an example of using this check here ...." ?

mothran (changes requested):
Thank you! Only minor tweaks in the comments from me.

risc0/zkvm/src/session.rs
tzerrell: Interesting! Perhaps a bit odd, but probably a discoverability win so I like it :+1:

mothran (approved):
Looks good! Does it have the memory overhead properties you were hoping for? I can run `heaptrack` on it if you need.

tzerrell:
> Looks good! Does it have the memory overhead properties you were hoping for? I can run `heaptrack` on it if you need.

I'm still testing this version, but an earlier version ([da2b55d](https://github.com/risc0/risc0/pull/560/commits/da2b55d88dac58b0716f6a56d22aff5fc98844a9)) was an approximately constant ~12GB for runs between 1.5M and 56M cycles, IIRC. So yes, this seems to be working as intended in that regard.

examples/evm/src/main.rs
```diff
@@ -86,7 +86,10 @@ async fn main() {
         .add_input(&to_vec(&zkdb).unwrap())
         .build();
     let mut exec = Executor::from_elf(env, EVM_ELF).unwrap();
-    let session = exec.run().unwrap();
+    let segment_dir = temp_dir();
```
flaub, line 89: I think we should use the crate `tempfile` so that the tempdir gets removed when it goes out of scope.

PR #561, opened 2023-05-12
Store memory image as a btree of pages

shkoo:
This significantly improves execute performance, both in memory usage and runtime.


flaub:
Perf looks amazing!!

flaub (approved):

PR #564, opened 2023-05-16
Updating instructions in factors README and comments in starter template

3lkn:


tzerrell (commented):
Thank you for catching this! I thought we had already done these updates. Not a great sign for the ease of keeping the tutorial up-to-date...

Your title mentions changes to the starter template, but I only see factors README changes -- did you forget to commit some of your changes?

3lkn:
> Thank you for catching this! I thought we had already done these updates. Not a great sign for the ease of keeping the tutorial up-to-date...
> 
> Your title mentions changes to the starter template, but I only see factors README changes -- did you forget to commit some of your changes?

Why yes I did, added the other update now!

templates/rust-starter/host/src/main.rs
```diff
@@ -2,13 +2,25 @@
...
+    serde::{from_slice, to_vec},
+    Executor, ExecutorEnv,
+};
 
 fn main() {
+    // First, we construct an executor environment
     let env = ExecutorEnv::default();
+
+    // TODO: add guest input to the executor environment using
+    // ExecutorEnv::add_input()
```
flaub, line 15: This isn't correct, `add_input` is on the `ExecutorEnvBuilder`

examples/factors/README.md
```diff
@@ -57,8 +55,8 @@ Don't worry about why these lines are included yet; for now, we're just being di
 Here are what the other two lines with `METHOD_NAME_ELF` and `METHOD_NAME_ID` should look like after updating:
 
 ```rust
-    let mut prover =
-        Prover::new(MULTIPLY_ELF).expect("Prover should be constructed from valid ELF binary");
+    let mut env =
+        ExecutorEnv::new(MULTIPLY_ELF).expect("Prover should be constructed from valid ELF binary");
```
tzerrell, line 59:
```suggestion
        ExecutorEnv::new(MULTIPLY_ELF).expect("ExecutorEnv should be constructed from valid ELF binary");
```

tzerrell: Thanks! I clearly didn't really turn my brain on for this part of the review.

So yeah, suggestion is instead use the code that actually works from the factors example and describe it here.

3lkn: I can make the correct update and check the rest of the factors README for strays as well (MULTIPLY_ELF along with env as params for Executor construction).

examples/factors/README.md
```diff
@@ -83,38 +81,46 @@ fn main() {
...
 
 ```rust
-    // Make the prover.
-    let mut prover =
-        Prover::new(MULTIPLY_ELF).expect("Prover should be constructed from valid ELF binary");
+    // First, we construct an executor environment
+    let env = ExecutorEnv::default();
+
+    // TODO: add guest input to the executor environment using
+    // ExecutorEnv::add_input()
```
tzerrell, line 91: If this is also wrong in the in-code comment, that will need to be updated as well.

```suggestion
    // ExecutorEnvBuilder::add_input()
```

examples/factors/README.md
```diff
@@ -83,38 +81,46 @@ fn main() {
...
-    // Make the prover.
-    let mut prover =
-        Prover::new(MULTIPLY_ELF).expect("Prover should be constructed from valid ELF binary");
+    // First, we construct an executor environment
+      let env = ExecutorEnv::builder().build();
 
-    prover.add_input_u32_slice(&to_vec(&a).expect("should be serializable"));
-    prover.add_input_u32_slice(&to_vec(&b).expect("should be serializable"));
+          // TODO: add guest input to the executor environment using
+    // ExecutorEnv::add_input()
```
tzerrell, line 112:
```suggestion
    // ExecutorEnvBuilder::add_input()
```

examples/factors/README.md
```diff
@@ -57,14 +55,14 @@ Don't worry about why these lines are included yet; for now, we're just being di
 Here are what the other two lines with `METHOD_NAME_ELF` and `METHOD_NAME_ID` should look like after updating:
 
 ```rust
-    let mut prover =
-        Prover::new(MULTIPLY_ELF).expect("Prover should be constructed from valid ELF binary");
-
+fn main() {
+    // Here we make an executor, loading the (renamed) ELF binary.
+    let mut exec = Executor::from_elf(env, METHOD_NAME_ELF).unwrap();
```
flaub, line 60: This takes an `env` parameter, which isn't defined anywhere.

Here's a suggestion, let's figure out how to make this README something that gets run as part of doc builds, which would then give us the ability to 'unit test' this example code.

3lkn: Looking into this and working out errors right now -- before I continue, is this an improvement we should be aiming to include in today's release?

flaub: I think we should try to get as much of our docs under test as possible

tzerrell: Looking at this thread: https://github.com/rust-lang/cargo/issues/383, it seems like built in Rust tooling won't accomplish this. It includes a comment suggesting a fairly straightforward macro for adding README tests: https://github.com/rust-lang/cargo/issues/383#issuecomment-720873790, however I haven't gotten this working -- I'll take a less rushed look tomorrow.

3lkn: That's the one I've been trying out and running into errors, going to keep troubleshooting today.

3lkn: More detailed version: `#![doc = include_str!("../README.md")]` is included in most of our crates under `src/lib.rs`, but I'm getting fingerprint errors that seem to be complaining about files changing out from under the compiler when I add the line to the factors example under `src/main.rs`. I'm going to pull my branch down and rerun in case I've done something funny locally, but it's not yet complaining about the syntax of various README snippets, my marker for at least having the README successfully included.

3lkn: The docs are building now so the branch issue was probably just git flow cruft but it's strange that only the digital signature example README is sensitive to asserting untruths.

3lkn: From convo with @flaub for now will add factors function and tests into a `lib.rs` to "solve" this problem but will come back to this (and see what else is out there with untested README snippets).

3lkn: Working through decisions around how to present code snippets to get this functional but still educational. The one that's most displeasing is aligning with our per-example naming conventions. We change `methods` to `factors-methods`, and thus
`use methods::{MULTIPLY_ID, MULTIPLY_ELF};` fails as an import unless I tell people to first complete this rename and get `use factor_methods::{MULTIPLY_ID, MULTIPLY_ELF};`.

I love that we've cut a clean path in our README to "minimal steps necessary to successfully build" to get people into the edit-compile loop as quickly as possible. Injecting this rename task breaks that a tiny bit, but I'm writing out instructions to feel it out.

3lkn: A second sub-ideal situation is that the final product is now split into `main.rs` and `lib.rs`, so there's not a pleasant 1:1 to follow along with. That said, the cost of having outdated README here is sufficiently high to warrant proceeding. There might be room to tack on a "let's make a library" step. It just won't give folks an instant visual clue. 

Loving @tzerrell 's suggestion to just include the README and nothing else. Tradeoff here is not teaching "let's make a library" so I'll leave it off for now but it's an option to consider if we strongly dislike the complexity of this next revision.

examples/factors/README.md
```diff
@@ -182,19 +221,47 @@ Once committed to the journal, anyone with the [receipt] can read this value.
...
+    };
+
+    let a: u64 = 17;
+    let b: u64 = 23;
+
+    let env = ExecutorEnv::builder()
+    // Send a & b to the guest
+    .add_input(&to_vec(&a).unwrap())
+    .add_input(&to_vec(&b).unwrap())
+    .build();
```
flaub, line 257:
```suggestion
    // Send a & b to the guest
    let env = ExecutorEnv::builder()
        .add_input(&to_vec(&a).unwrap())
        .add_input(&to_vec(&b).unwrap())
        .build();
```

examples/factors/src/main.rs
```diff
@@ -12,15 +12,28 @@
...
+mod test_readme {
+    macro_rules! external_doc_test {
+        ($x:expr) => {
+            #[doc = $x]
+            extern "C" {}
+        };
+    }
+
+    external_doc_test!(include_str!("../README.md"));
+}
```
flaub, line 32: Why do we need this *and* have lib.rs refer to README.md to get test coverage?

3lkn: Unnecessary, will remove. Artifact of re-testing after refactor whether we really needed the lib

templates/rust-starter/host/src/main.rs
```diff
@@ -12,7 +12,12 @@ fn main() {
     let env = ExecutorEnv::default();
 
     // TODO: add guest input to the executor environment using
-    // ExecutorEnv::add_input()
+    // ExecutorEnvBuilder::add_input().
+    // To access this method, you'll need to use the alternate construction
+    // ExecutorEnv::builder(), which creates an ExecutorEnvBuilder. When you're
+    // done adding input, call ExecutorEnvBuilder::build().
+
+    // For example: let env = ExecutorEnv::builder().add_input(&vec).build();
```
tzerrell: Perhaps we should pre-separate this as
```
let env = ExecutorEnv::builder().build();
```
and then we can tell people to add there inputs in the middle.

tzerrell (commented):
I can definitely see some of the tradeoffs you were mentioning in getting this tested, but I do think we're all right that the testing is worth those tradeoffs.

examples/factors/README.md
```diff
@@ -136,16 +173,18 @@ We'll then publicly commit their product to the [journal] portion of the [receip
 
 Here is the complete guest program.
 We'll break this down step by step below:
-```rust
+```no_compile
```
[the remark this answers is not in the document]
3lkn: This hasn't worked for me but I will change if I find a workaround that allows them both.

templates/rust-starter/host/src/main.rs
3lkn: > I can definitely see some of the tradeoffs you were mentioning in getting this tested, but I do think we're all right that the testing is worth those tradeoffs.

Agreed -- other than accepting that a few lines are commented out here and there or have bonus explanatory comments, I feel all right about the final product.

3lkn: > Perhaps we should pre-separate this as
> 
> ```
> let env = ExecutorEnv::builder().build();
> ```
> 
> and then we can tell people to add there inputs in the middle.

Okay great, if there's no reason that `ExecutorEnv::default()` was left in the starter template (and I can't see why there would be, since it's just wrapping `ExecutorEnv::builder().build()`), I'll change both in the README and the starter template.

examples/factors/README.md
```diff
@@ -182,19 +221,47 @@ Once committed to the journal, anyone with the [receipt] can read this value.
...
+    // Run the executor to produce a session.
+    let session = exec.run().unwrap();
+
+    // Prove the session to produce a receipt.
+    let receipt = session.prove().unwrap();
+
     // TODO: Implement code for transmitting or serializing the receipt for
     // other parties to verify here
+
```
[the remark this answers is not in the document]
3lkn: This isn't in the `factors` example but it's present in lines 31-32 of `templates/rust-starter/host/src/main.rs`.I'll grant you though it's an odd step to instruct on now that in the `factors` example the journal content extraction is in `multiply_factors` and the journal verification is in `main.rs`. Still for now I think since it lines up reasonably well with the original template content, it's better to leave it.

examples/factors/README.md
```diff
@@ -26,101 +27,128 @@ cd wherever/you/want
...
 In the main project folder, build and run the project using `cargo run --release`.
 Nothing exciting will happen yet, but it should give you an error if you made one of the above changes incorrectly.
 Use this command any time you'd like to check your progress.
 
-## Step 3 (Host): Share two values with the guest
+## Concept break: How do we run and prove the guest program?
+
+Our next objective is to provide the guest program with input. Before we implement this, let's take a closer look at how we run and prove the guest program in `factors/src/main.rs`.
+
+In the starter template project, our host driver program creates an executor environment before constructing an executor.  When the `Executor::run()` command executes, it will produce a session. The `Session::prove()` command will cause our guest program to execute, producing a receipt:
```
flaub, line 92:
```suggestion
In the starter template project, our host driver program creates an executor environment before constructing an executor.  When `Executor::run()` is called, it will produce a session. Calling `Session::prove()` will cause our guest program to execute, producing a receipt:
```

examples/factors/README.md
```diff
@@ -26,101 +27,128 @@ cd wherever/you/want
...
-Edit `methods/guest/Cargo.toml`, changing the line `name = "method_name"` to instead read
+The methods package contains the program that executes on the guest zkVM as well as a few supporting libraries. Before we proceed, let's change its name in `methods/Cargo.toml` to match the name of our project, renaming `name = methods` to `name = factors-methods`.
+
+Parts of this package are included in the driver program `src/main.rs`, so change the line
+
+```no_compile
+use methods::{METHOD_NAME_ELF, METHOD_NAME_ID};
 ```
-name = "multiply"
+
```
flaub, line 43: feels like we're missing a "to" here, but it's hard to parse in this diff format

tzerrell, line 43: I checked in View File and there does indeed need to be a "to" here.
```suggestion

to

```

tzerrell: I think this may have been accidentally resolved, this is not actually in yet.

tzerrell (commented):
Looks like one change didn't make it in

tzerrell (approved):

PR #566, opened 2023-05-17
Adjust image_id to be the hash of (merkle_root, pc)

flaub:


flaub:
The check for exit_code is failing and I plan to fix this soon. But I wanted to get eyes on this in the meantime.

jbruestle (approved):

spaugh (approved):

risc0/zkvm/src/exec/mod.rs
```diff
@@ -316,7 +318,7 @@ impl<'a> Executor<'a> {
     }
 
     fn advance(&mut self, opcode: OpCode, op_result: OpCodeResult) -> Option<ExitCode> {
-        log::debug!(
+        log::trace!(
```
mothran, line 321: Thank you! I just hit this and thought it should be `trace`

risc0/zkvm/src/receipt.rs
```diff
@@ -341,3 +368,12 @@ impl ReceiptMetadata {
         })
     }
 }
+
+/// Compute and return the ImageID of the given `(merkle_root, pc)` pair.
+pub fn compute_image_id(merkle_root: &Digest, pc: u32) -> Digest {
+    use risc0_zkp::core::{digest::DIGEST_WORDS, hash::sha::Sha256};
```
mothran, line 374: I generally think we should avoid in-function imports because it can be hard to track down import scoping when introducing new feature flags.  But its only a small nit.

mothran (approved):

risc0/zkvm/src/receipt.rs
flaub: Agreed, but in this particular case, I couldn't figure out how to get it to work because the rest of the file needs `Sha256` but is a different trait with the same name.

PR #567, opened 2023-05-17
Make `sys_sha_buffer` available as extern and add RustCrypto sha2 patching example

nategraf:
This PR marks the `sys_sha_buffer` syscall as `extern "C"` so that it can be linked to the patched RustCrypto/sha2 crate.

This PR also includes a variation on the SHA-256 example that uses a patched version of the `sha2` crate. It includes a `#[test]` that excersizes that code path for execution in CI.

Including the test in the example may decrease the clarity of the example. If this is the case, I can move the test e.g. into `risc0/zkvm/methods`.


examples/sha/src/main.rs
```diff
@@ -18,18 +18,27 @@ use risc0_zkvm::{
...
+use sha_methods::{HASH_ELF, HASH_RUST_CRYPTO_ELF};
 
-fn provably_hash(input: &str) -> Digest {
+fn provably_hash(input: &str, use_rust_crypto: bool) -> Digest {
     let env = ExecutorEnv::builder()
         .add_input(&to_vec(input).unwrap())
         .build();
 
-    let mut exec = Executor::from_elf(env, HASH_ELF).unwrap();
+    let elf = if use_rust_crypto {
```
tzerrell, line 28: Maybe add a comment here saying that this enables/disables circuit acceleration patching for the RustCrypto sha2 crate by switching between guest code that does/doesn't use it; and probably also directing people to the corresponding guest files.

tzerrell (approved):
I think the additional complexity is pretty minor and doesn't get in the way of understanding the essentials of what this example is doing. And it may even be helpful to show how one can use patching. I think adding a comment about what is getting switched on is probably helpful -- I put an inline suggestion but feel free to use your own judgment if you think you have a better idea for what to say.

flaub (approved):

nategraf:
@flaub I made a drive-by change in 25045682f2507815fe445686f6fe9e9453b6493a to add `#[must_use]` to the verify methods. I have been seeing the attribute use in the cryptography libraries I have been looking at and it helps sand off the sharp edge that a developer would forget to check and ensure the method did not return an error (e.g. because the assumed it would `assert!`).

PR #568, opened 2023-05-18
[WIP] bonsai-starter template based on foundry

hashcashier:


.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "templates/bonsai-starter/lib/forge-std"]
```
mothran, line 1: I think we generally should avoid submodule in the root project that submodule into the template. Should this .gitmodules be in the template directory?

templates/bonsai-starter/methods/src/main.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2023 RISC Zero, Inc.
```
mothran, line 1: Having the `main.rs` be in the `methods` crate is a little bit of a deviation from our normal starter pattern of the `host` vs `methods` crates.

.github/workflows/main.yml
```diff
@@ -31,14 +31,8 @@ jobs:
...
       - run: cargo fmt --all --check
       - run: cargo fmt --all --check --manifest-path examples/Cargo.toml
-      - run: cargo fmt --all --check --manifest-path templates/bonsai/Cargo.toml
       - run: cargo sort --workspace --check
       - run: cargo sort --workspace --check examples
-      - run: cargo sort --workspace --check templates/bonsai
-      - run: npm install
-        working-directory: templates/bonsai
-      - run: npm run ci
-        working-directory: templates/bonsai
```
flaub, line 41: Hopefully all these changes are temporary?

.gitmodules
flaub: I'd like to avoid submodules entirely, can we discuss why this is necessary and possible alternatives?

.github/workflows/main.yml
hashcashier: These were used for testing the old bonsai template. They won't be needed to test the new one I believe. I need the foundry installation action pulled in so I can use it to create the test.

PR #557, opened 2023-05-10
[WIP] Add cargo-risczero build command

nategraf:
Work in progress command to build a selected crate for the RISC Zero target.

Wrote up this very basic version to help in developing a patch for the Rust Crypto `crypto-bigint`
crate.

## Usage

Example of building a crate using this command

```bash
cargo run -p cargo-risczero -- risczero build --manifest-dir ~/risc0/rust-crypto/bigint/
```

Example of building tests (does not actually generate functional binaries right now)

```bash
cargo run -p cargo-risczero -- risczero build --test --manifest-dir ~/risc0/rust-crypto/bigint/
```

PR #571, opened 2023-05-18
Add error string to opcode decode failure in Executor

nategraf:
Add an error string which includes the bad opcode on decode failure. A small change that helped be
debug a failure I ran into when trying to run tests compiled as guest code.

This is a small PR, just to preserve the one remaining useful change from
https://github.com/risc0/risc0/pull/557, which is now closed.

### DIFF
--- a/risc0/zkvm/src/opcode.rs
+++ b/risc0/zkvm/src/opcode.rs
@@ -89,7 +89,7 @@ impl OpCode {
         // log::debug!("decode: 0x{word:08X}");
 
         let decode_error_str = || {
-            format!("malformed rv32im instruction at {insn_pc:#x}: {insn:032b} ({opcode:07b}, {funct3:03b}, {rs2:05b}, {funct7:07b})")
+            format!("illegal instruction at {insn_pc:#x}: {insn:032b} ({opcode:07b}, {funct3:03b}, {rs2:05b}, {funct7:07b})")
         };
 
         Ok(match opcode {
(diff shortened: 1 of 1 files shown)


mothran (approved):
Nice! Looks very helpful.

SchmErik (approved):

risc0/zkvm/src/opcode.rs
```diff
@@ -88,14 +88,18 @@ impl OpCode {
...
+
         Ok(match opcode {
             0b0000011 => match funct3 {
                 0x0 => OpCode::new(insn, insn_pc, "LB", 24, 1),
                 0x1 => OpCode::new(insn, insn_pc, "LH", 25, 1),
                 0x2 => OpCode::new(insn, insn_pc, "LW", 26, 1),
                 0x4 => OpCode::new(insn, insn_pc, "LBU", 27, 1),
                 0x5 => OpCode::new(insn, insn_pc, "LHU", 28, 1),
-                _ => unreachable!(),
+                _ => unreachable!("{}", decode_error_str()),
```
flaub, line 102:
```suggestion
                _ => bail!("{}", decode_error_str()),
```

We probably should avoid panics here

nategraf: Changed to `bail!` in [d1e8ec4](https://github.com/risc0/risc0/pull/571/commits/d1e8ec4ddfc156ea99b0049ffca0ade0d89f8918)

flaub (approved):

PR #572, opened 2023-05-19
Adding a few Rust doc examples to Executor and ExecutorEnv

3lkn:


risc0/zkvm/src/exec/env.rs
```diff
@@ -151,6 +213,39 @@ impl<'a> ExecutorEnvBuilder<'a> {
     }
 
     /// Add initial input that can be read by the guest from stdin.
+    /// This step must occur before the executor environment is built.
```
flaub, line 216: I don't understand this comment because structurally, it's not possible to `add_input` after the env has been built.

risc0/zkvm/src/exec/env.rs
```diff
@@ -131,18 +155,56 @@ impl<'a> ExecutorEnvBuilder<'a> {
     }
 
     /// Set a session limit, specified in number of cycles.
+    /// # Example
+    /// ```
+    /// # use risc0_zkvm::{
+    /// #   ExecutorEnv,
+    /// #   ExecutorEnvBuilder};
```
flaub, line 162: Why are these commented out? Seems useful to have in the example.

risc0/zkvm/src/exec/env.rs
```diff
@@ -131,18 +155,56 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
+    ///
+    /// const NEW_SESSION_LIMIT: usize = 32 * 1024 * 1024; // 32M cycles
+    ///
+    /// let env = ExecutorEnv::builder()
+    ///     .session_limit(NEW_SESSION_LIMIT)
```
flaub, line 167:
```suggestion
    ///     .session_limit(32 * 1024 * 1024) // 32M cycles
```

risc0/zkvm/src/exec/env.rs
```diff
@@ -131,18 +155,56 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
         self
     }
 
     /// Add an environment variable to the guest environment.
+    /// # Example
+    /// ```
+    /// # use risc0_zkvm::{
+    /// #   ExecutorEnv,
+    /// #   ExecutorEnvBuilder};
+    /// # use std::collections::HashMap;
```
flaub, line 202: I don't think this is needed

risc0/zkvm/src/exec/env.rs
```diff
@@ -131,18 +155,56 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
+    ///
+    /// let mut vars = HashMap::new();
+    /// vars.insert("VAR1".to_string(), "SOME_VALUE".to_string());
+    /// vars.insert("VAR2".to_string(), "SOME_VALUE".to_string());
+    ///
+    /// let env = ExecutorEnv::builder()
+    /// .env_vars(vars)
+    /// .build();
```
flaub, line 189:
```suggestion
    ///     .env_vars(vars)
    ///     .build();
```

risc0/zkvm/src/exec/env.rs
```diff
@@ -63,6 +68,17 @@ pub struct ExecutorEnv<'a> {
 
 impl<'a> ExecutorEnv<'a> {
     /// Construct a [ExecutorEnvBuilder].
+    /// # Example
+    /// ```
+    /// # use risc0_zkvm::{
+    /// #   ExecutorEnv,
+    /// #   ExecutorEnvBuilder,
+    ///     serde::to_vec};
```
flaub, line 76: This isn't needed

risc0/zkvm/src/exec/env.rs
```diff
@@ -63,6 +68,17 @@ pub struct ExecutorEnv<'a> {
...
+    ///
+    /// let a: u64 = 400;
+    ///
+    /// let env = ExecutorEnv::builder().build();
```
flaub, line 80: This example doesn't seem especially useful

risc0/zkvm/src/exec/env.rs
```diff
@@ -50,6 +50,11 @@ pub struct ExecutorEnvBuilder<'a> {
 }
 
 /// The [super::Executor] is configured from this object.
+///
+/// The executor environment holds configuration details that inform how the
+/// guest environment is set up prior to guest program execution. This
+/// includes initial guest program inputs, session size limits, segment size
+/// limits, and environment variables.
```
flaub, line 57: Adding more context is good, but enumerating the fields in this description isn't good for maintenance and it doesn't actually help understand what this thing does any better (IMHO).

risc0/zkvm/src/exec/env.rs
```diff
@@ -151,6 +213,39 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
+    ///
+    /// let env = ExecutorEnv::builder()
+    ///     .add_input(&to_vec(&a).unwrap())
+    ///     .build();
+    /// ```
+    ///
+    /// Most of the time, the best practice is to combine guest program inputs
+    /// into a single struct. For cases where you want to add inputs that
+    /// the guest will access using consecutive reads, you can call
+    /// `ExecutorEnvBuilder::add_input()` iteratively:
```
flaub, line 234: Instead of making a prescription, I'd just say what the behavior of calling `add_input` is, which is to concatenate the inputs together

risc0/zkvm/src/exec/mod.rs
```diff
@@ -129,6 +130,11 @@ impl Write for Journal {
 
 impl<'a> Executor<'a> {
     /// Construct a new [Executor] from a [MemoryImage] and entry point.
+    /// Before a guest program is proven, the [Executor] is responsible for
+    /// deciding where a zkVM program should be split into [Segment]s and what
+    /// work will be done in each segment. This is the execution phase:
+    /// the guest program is executed to determine how its proof should be
+    /// divided into subparts.
```
flaub, line 137:
```suggestion
    /// Construct a new [Executor] from a [MemoryImage] and entry point.
    ///
    /// Before a guest program is proven, the [Executor] is responsible for
    /// deciding where a zkVM program should be split into [Segment]s and what
    /// work will be done in each segment. This is the execution phase:
    /// the guest program is executed to determine how its proof should be
    /// divided into subparts.
```

risc0/zkvm/src/exec/mod.rs
```diff
@@ -129,6 +130,11 @@ impl Write for Journal {
...
     pub fn new(env: ExecutorEnv<'a>, image: MemoryImage, pc: u32) -> Self {
```
flaub, line 138: Now that the PC is part of the MemoryImage, we should drop the `pc` argument here

risc0/zkvm/src/exec/mod.rs
flaub: I think how the executor works should be described on the `Executor` and not the constructor. I also think execution is more than merely deciding where splits occur. It also consumes all the inputs to generate a trace that the prover can use.

risc0/zkvm/src/exec/mod.rs
```diff
@@ -151,7 +157,20 @@ impl<'a> Executor<'a> {
...
+    /// configuration details.
+    /// # Example
+    /// ```
+    /// # use risc0_zkvm::{serde::to_vec, Executor, ExecutorEnv, Session};
+    /// # use risc0_zkvm_methods::{BENCH_ELF, bench::{BenchmarkSpec, SpecWithIters}};
+    /// #
+    /// # let spec = SpecWithIters(BenchmarkSpec::SimpleLoop, 1);
+    /// # let env = ExecutorEnv::builder()
+    /// #   .add_input(&to_vec(&spec).unwrap())
+    /// #   .build();
```
flaub, line 171: I think showing this is useful for this example

risc0/zkvm/src/exec/env.rs
```diff
@@ -151,6 +213,39 @@ impl<'a> ExecutorEnvBuilder<'a> {
     }
 
     /// Add initial input that can be read by the guest from stdin.
+    /// This step must occur before the executor environment is built.
+    /// # Examples
+    /// ```
+    /// # use risc0_zkvm::{
+    /// #   ExecutorEnv,
+    /// #   ExecutorEnvBuilder,
+    ///     serde::to_vec};
```
tzerrell, line 222: This `use` is dangling. Probably just un-hide the entire `use` block, but if you really want to show just `to_vec`, make it a separate statement, e.g.
```
# use risc0_zkvm::{
#   ExecutorEnv,
#   ExecutorEnvBuilder
# };
use risc0_zkvm::serde::to_vec;
```

risc0/zkvm/src/exec/mod.rs
```diff
@@ -160,6 +179,18 @@ impl<'a> Executor<'a> {
...
+    /// # let mut exec = Executor::from_elf(env, BENCH_ELF).unwrap();
```
tzerrell, line 191: I think showing the `env` & `exec` gives useful context. If the `spec` and `BENCH_ELF` names seem confusing (and they probably are), this could be further upgraded by hidden lines like `let EXAMPLE_ELF = BENCH_ELF` and naming `spec` to instead by `example_input` or something similar.

tzerrell (commented):
I'm glad to see examples for these, thanks!

@flaub pointed out specific places where this would be helpful, but in general with code snippets I think showing context is one of the main benefits. More concretely, when documenting a function, the automatically generated signature documentation gives the reader a pretty good sense of what the exact line where you call the function looks like. What isn't generated that we need to manually write is what happens around that line: are there expectations or common techniques for other functions to be called ahead of time? Is there a standard or common way to construct one or more of the inputs? What sorts of things can you do with the result of the function? Is any cleanup needed? What do you need to import to be able to use the function in typical ways?

There's definitely a tradeoff here, for the general reasons that concision is good (readers won't bother to read, or read carefully, snippets that are too long). Nevertheless, in general I would hide less of the supporting code relative to what you're doing here. This is especially true when you only give one snippet: if you have three snippets and they each use the same set up code, you sometimes only need to show it on the first snippet, but if you only have one snippet -- well you still have to show it on the first snippet, but that's the only snippet.

risc0/zkvm/src/exec/env.rs
3lkn: Yeah this is trivially true, pulling it out.

risc0/zkvm/src/exec/env.rs
3lkn: You're right -- going off of TimZ's feedback, I'm going to un-comment a majority of the omitted code, including includes.

risc0/zkvm/src/exec/env.rs
3lkn: Oops! Copied and pasted without pulling that out.

risc0/zkvm/src/exec/env.rs
3lkn: That makes a lot of sense, the fragility of explanatory text is still fragility

risc0/zkvm/src/exec/env.rs
3lkn: Removing the prescriptive first line but making note that guest can access concatenated input with consecutive reads, can take it out if that's still too prescriptive.

3lkn:
> 

Thank you for writing this out, these are excellent guidelines.

risc0/zkvm/src/exec/env.rs
```diff
@@ -131,18 +152,55 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
     /// Add an environment variable to the guest environment.
+    /// # Example
+    /// ```
+    /// # use risc0_zkvm::{
+    /// #   ExecutorEnv,
+    /// #   ExecutorEnvBuilder};
+    ///
+    /// let env = ExecutorEnv::builder()
+    /// .env_var("VAR1", "SOME_VALUE")
+    /// .build();
```
tzerrell, line 202:
```suggestion
    /// let env = ExecutorEnv::builder()
    ///     .env_var("VAR1", "SOME_VALUE")
    ///     .build();
```

risc0/zkvm/src/exec/env.rs
tzerrell: Maybe instead of an example, add a redirection to other parts of the doc, e.g. "Combine with [ExecutorEnvBuilder] functions like [...add_input] to set up the environment and I/O data your guest needs. Finalize with [ExecutorEnvBuilder::build] to get a configured [ExecutorEnv]."

tzerrell (commented):
Looking good! One suggested revision, one typesetting suggestion

tzerrell (approved):

tzerrell (approved):

PR #575, opened 2023-05-19
add test for session limit overflow

SchmErik:
This parameter of the executor env had no test. This PR adds a simple test case that exercises the session limit overflow.


flaub:
This is a great start! I think there are lots of other edge cases to check for too. Especially when there are multiple segments.

PR #581, opened 2023-05-23
Add Graviton / arm64 + linux runners into CI

imdaisyadams:


.github/actions/sccache/action.yml
```diff
@@ -5,22 +5,30 @@ runs:
...
     - if: runner.os == 'macOS' && runner.arch == 'ARM64'
       run: |
         echo "SCCACHE_ARCH=aarch64-apple-darwin" >> $GITHUB_ENV
+        echo "SCCACHE_VERSION=v0.3.3" >> $GITHUB_ENV
+      shell: bash
+
+    - if: runner.os == 'Linux' &&  runner.arch == 'ARM64'
+      run: |
+        echo "SCCACHE_ARCH=aarch64-unknown-linux-musl" >> $GITHUB_ENV
+        echo "SCCACHE_VERSION=v0.4.0-pre.5" >> $GITHUB_ENV
```
mothran, line 27: Is there a published version that is the same between all versions of sccache? I worry that mixed versions on different hosts could cause weird random failures.

mothran (changes requested):

.github/actions/sccache/action.yml
imdaisyadams: just updated sccache to v.0.5.0 across all systems, and builds are completing successfully

.github/workflows/main.yml
```diff
@@ -59,6 +59,9 @@ jobs:
           - os: macOS
             feature: metal
             device: apple_m2_pro
+          - os: Linux
+            feature: cuda
+            device: graviton2
```
flaub, line 64: Wouldn't the device be `nvidia_rtx_????` here?

flaub: AFAIK, graviton represents the CPU, but we still have NVidia gear that we're running accelerated code on

imdaisyadams: yeah you're right, I've updated the tag to be: `nvidia_tesla_t4g`

flaub (approved):

.github/actions/sccache/action.yml
```diff
@@ -5,7 +5,7 @@ runs:
   steps:
     - if: runner.os == 'Linux' && runner.arch == 'X64'
       run: |
-        echo "SCCACHE_ARCH=x86_64-unknown-linux-musl" >> $GITHUB_ENV
+          echo "SCCACHE_ARCH=x86_64-unknown-linux-musl" >> $GITHUB_ENV
```
mothran, line 8: Extra indent?

mothran (approved):
Just one little indent change but looks good!

PR #585, opened 2023-05-24
Enable feature-gated programmatic access to `methods.rs` contents

hashcashier:
# Guest List Implementation

This PR enables `risc0-build` to create an additional `GUEST_LIST` variable in generated `methods.rs` files to enable dynamic lookup of elfs/ids/paths using guest metadata. This is gated behind a newly introduced `guest-list` feature.

## New Guest List Entry Struct
The following structure is added to the `risc0-build` crate:
https://github.com/risc0/risc0/blob/d40687ba803474277158eabca6f801e13324a354/risc0/build/src/lib.rs#L55-L66

## `Risc0Method` Amendment

The existing `Risc0Method` struct is extended with the following function, which translates it into `GuestListEntry` instantiation rust code:
https://github.com/risc0/risc0/blob/d40687ba803474277158eabca6f801e13324a354/risc0/build/src/lib.rs#L112-L123

## `methods.rs` embedding changes

First, a vector is created to keep track of the guest entry strings to be written, and the import of the `GuestListEntry` struct from the `risc0-build` crate is added to the beginning of the generated file:

https://github.com/risc0/risc0/blob/d40687ba803474277158eabca6f801e13324a354/risc0/build/src/lib.rs#L511-L516

For each elf/id/path, a corresponding guest entry string is stored in the vector above:

https://github.com/risc0/risc0/blob/d40687ba803474277158eabca6f801e13324a354/risc0/build/src/lib.rs#L535-L542

Lastly, the list of guest entry strings is appended at the end of the `methods.rs` file:

https://github.com/risc0/risc0/blob/d40687ba803474277158eabca6f801e13324a354/risc0/build/src/lib.rs#L544-L553

## Sample generated output

```rust
use risc0_build::GuestListEntry;

pub const FIBONACCI_ELF: &[u8] = &[...];
pub const FIBONACCI_ID: [u32; 8] = [...];
pub const FIBONACCI_PATH: &str = r#"..."#;

pub const FIBONACHOS_ELF: &[u8] = &[...];
pub const FIBONACHOS_ID: [u32; 8] = [...];
pub const FIBONACHOS_PATH: &str = r#"..."#;

pub const GUEST_LIST: &[GuestListEntry] = &[
    GuestListEntry {
        name: "FIBONACCI",
        elf: FIBONACCI_ELF,
        image_id: FIBONACCI_ID,
        path: FIBONACCI_PATH,
    },
    GuestListEntry {
        name: "FIBONACHOS",
        elf: FIBONACHOS_ELF,
        image_id: FIBONACHOS_ID,
        path: FIBONACHOS_PATH,
    }];
```

### DIFF
--- a/risc0/build/src/lib.rs
+++ b/risc0/build/src/lib.rs
@@ -60,7 +60,7 @@ pub struct GuestListEntry {
     /// The compiled ELF guest binary
     pub elf: &'static [u8],
     /// The image id of the guest
-    pub image_id: [u32; 8],
+    pub image_id: [u32; DIGEST_WORDS],
     /// The path to the ELF binary
     pub path: &'static str,
 }
--- a/risc0/build/src/lib.rs
+++ b/risc0/build/src/lib.rs
@@ -119,8 +119,7 @@ pub const {upper}_PATH: &str = r#"{elf_path}"#;
         elf: {upper}_ELF,
         image_id: {upper}_ID,
         path: {upper}_PATH,
-    }}
-            "
+    }}"
         )
     }
 }
@@ -547,7 +546,7 @@ pub fn embed_methods_with_options(mut guest_pkg_to_options: HashMap<&str, GuestO
             .write_all(
                 format!(
                     "pub const GUEST_LIST: &[GuestListEntry] = &[{}];",
-                    guest_list_entries.join(",\n")
+                    guest_list_entries.join(",")
                 )
                 .as_bytes(),
             )
--- a/risc0/build/src/lib.rs
+++ b/risc0/build/src/lib.rs
@@ -52,6 +52,19 @@ impl Risc0Metadata {
     }
 }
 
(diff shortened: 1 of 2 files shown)


risc0/build/src/lib.rs
```diff
@@ -52,6 +52,19 @@ impl Risc0Metadata {
...
 
+#[cfg(feature = "guest-list")]
+/// Represents an item in the generated list of compiled guest binaries
+pub struct GuestListEntry {
+    /// The name of the guest binary
+    pub name: &'static str,
+    /// The compiled ELF guest binary
+    pub elf: &'static [u8],
+    /// The image id of the guest
+    pub image_id: [u32; 8],
```
flaub, line 63:
```suggestion
    pub image_id: [u32; DIGEST_WORDS],
```

flaub (approved):

PR #589, opened 2023-05-26
evm: adding pointer from readme to blog

pdg744:
Now that the continuations blog has landed, it makes sense to make that findable from the examples repo. 

This PR accomplishes that by adding a pointer from the top of the evm readme to the continuations blog.


tzerrell (approved):
Good call

PR #590, opened 2023-05-26
Add ECDSA verification example using bigint acceleration support

nategraf:
This PR adds an ECDSA verification example, which shows how to use the patched `k256` library to get accelerated signature verification on secp256k1 in the zkVM guest.

As a dependency for the `k256` patch, it also updates the allocator to support alignments that are larger than a word. (In particular, the constant-time lookup table implementation uses a 1024-byte alignment). https://github.com/risc0/rust/pull/11 is related to this PR. Based on work by @shkoo.

As a drive-by change, this PR includes an additional check in the executor that execution did not previously end with `ExitCode::Halted`, which is a non-resumable state. I added this after running into this error by accident and having a hard time debugging it based on the errors I received.


examples/ecdsa/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+risc0-zkvm = { path = "../../../../risc0/zkvm", default-features = false, features = ["std"] }
+hex-literal = "0.4"
+k256 = { version = "=0.13.1", features = ["arithmetic", "serde", "expose-field", "std", "ecdsa"], default_features = false }
+
+[patch.crates-io]
+# Placing these patch statement in the workspace Cargo.toml will add RISC Zero SHA-256 and bigint
+# multiplicaation accelerator support for all downstream usages of the following crates.
+# DO NOT MERGE: Use tagged git references instead of branches
+sha2 = { git = "https://github.com/risc0/RustCrypto-hashes", tag = "sha2-v0.10.6-risc0" }
+k256 = { git = "https://github.com/risc0/RustCrypto-elliptic-curves", branch = "victor/use-risc0-accel"  }
```
flaub, line 18: I don't think we should rely on `username/` branches persisting. So I suggest we rename these branches to something more stable and then updating those references here.

examples/ecdsa/methods/guest/src/bin/benchmark.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+#[inline(never)]
+fn bench_inner(name: &str, func: impl FnOnce(), silent: bool) {
+    black_box({
+        let start = env::get_cycle_count();
+
+        black_box(func());
+
+        let end = env::get_cycle_count();
```
flaub, line 41: Unforunately using cycle count in this way won't be very accurate, due to the difference between the way pages are emulated in the executor and performed in the prover. I'm not sure how much we can trust these numbers in this way, let's chat about this offline at some point so I can provide deeper context.

examples/ecdsa/methods/guest/Cargo.toml
nategraf: Definitely. Strategy is to use tags primarily and have all of our tags live on the `risc0/use-accel` branch in our forks. I won't merge this PR until those tags are available.

flaub:
I have a concern about using standard benchmarking frameworks unless we can turn down the iteration count to 1 since there should be no difference at all of cycle counts behaving differently between runs.

nategraf:
> I have a concern about using standard benchmarking frameworks unless we can turn down the iteration count to 1 since there should be no difference at all of cycle counts behaving differently between runs.

So the bench-marking example currently included doesn't use any standard bench-marking library (e.g. criterion) if that's what you are referring to, it just runs a given closure twice inside of a `core::hint::black_box` declaration. Running it twice is simply to ensure paging-in is not recorded. That file though was mostly for my own optimization work, and I hadn't decided yet what to do with it before marking this PR as ready.

nategraf:
@flaub and @tzerrell , this PR is ready with the exception of the exact patch references. which will be updated once https://github.com/risc0/RustCrypto-crypto-bigint/pull/1 and then https://github.com/risc0/RustCrypto-elliptic-curves/pull/1 are merged.

tzerrell (approved):
LGTM (up to the patch reference issue already discussed)

PR #595, opened 2023-06-01
Add ZK financial calculation example

choongng:
Initial draft. Looking for feedback on structure and overall design before polishing.


flaub:
Missing expected license header (see other examples)

examples/prorata/core/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub csv_hash: Vec<u8>,
+}
+
+pub fn allocate(amount: Decimal, recipients: Vec<Recipient>) -> Vec<Allocation> {
+    /* Completely allocate the total funds `amount` by shares into dollars and
+    cents amounts for each recipient.
+
+    amount - total amount to distribute
+    recipients - list of recipients with their share of the total amount
+    */
```
flaub, line 36: Prefer using `//` rather than multi-line comments

examples/prorata/methods/guest/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+use risc0_zkvm::guest::env;
+use rust_decimal::Decimal;
+
+risc0_zkvm::guest::entry!(main);
+
+pub fn main() {
+    // load the amount, recipients, and target user from the environment
+    let amount: Decimal = env::read();
+    let recipients_csv: Vec<u8> = env::read();
+    let query: String = env::read();
```
flaub, line 15: This would be cleaner to have a struct that is read in, instead of reading in each element separately.

examples/prorata/methods/guest/src/main.rs
```diff
@@ -0,0 +1,22 @@
...
+
+    // use allocate_for() to compute the allocation for the requested recipient
+    let result: QueryResult = allocate_for_csv(amount, recipients_csv, &query);
```
flaub, line 18: This could be a function on the struct that is read in above.

flaub, line 18:
```suggestion
    let result = allocate_for_csv(amount, recipients_csv, &query);
```

examples/prorata/host/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+                from_slice(&receipt.journal).expect("Failed to deserialize result");
+            print!("{}", format_query_result(result));
+        }
+        Err(e) => println!("Receipt is invalid: {}", e),
+    }
+}
+
+// todo should be able to gate this on std availability
+fn format_query_result(result: QueryResult) -> String {
+    let mut s = Vec::<String>::new();
```
flaub, line 106: Avoid single letter variable names (unless it's heavy math formulas or `i` for loops, etc).

examples/prorata/host/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+    let recipients_csv = std::fs::read(&input).expect("Failed to read input file");
+    let query = recipient;
+
+    println!("Query: {}", query);
+
+    let env = ExecutorEnv::builder()
+        // Send a & b to the guest
+        // TODO: this is very ugly but to_vec() seems to be necessary otherwise
+        // we get a panic due to an alignment problem in bytemuck
+        // > thread 'main' panicked at 'cast_slice>TargetAlignmentGreaterAndInputNotAligned'
```
flaub, line 72: I haven't experienced this before, would you mind filing a bug report for this one?

examples/prorata/host/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+        } => allocate(input, output, recipient, amount),
+        Commands::Verify { input } => verify(input),
+    }
+}
+
+fn allocate(input: &str, output: &str, recipient: &str, amount: &Decimal) {
+    // read the file in the path specified as args.input into a vec of u8 called
+    // recipients_csv
+    let recipients_csv = std::fs::read(&input).expect("Failed to read input file");
+    let query = recipient;
```
flaub, line 64: Why the rename/rebinding? How about just rename `recipient` parameter to `query`?

examples/prorata/core/src/lib.rs
flaub: Also this seems like a comment that would go just before the function, and perhaps even use `///`

choongng: I'll read up on how to do this so it fits with the rust doc tooling.

examples/prorata/host/src/main.rs
choongng: Will file if it still repros.

examples/prorata/methods/guest/src/main.rs
choongng: Agree that would be cleaner, this is the pattern we demonstrate elsewhere (at least the factors example). Should we update the examples to be single struct in / struct out?
https://github.com/risc0/risc0/blob/3ce67c81ae3fcdc87e6301c2f9371aa53747a3c1/examples/factors/src/lib.rs#L26

tzerrell: The factors example is one of (maybe literally) the only examples that uses this multi-read; this is to keep it as barebones as possible to make it as simple of a "Hello World" as possible.

It's possible that's the wrong decision (and I'd be happy to discuss), but it was a deliberate choice.

tzerrell (commented):
Nothing to suggest that wasn't already mentioned.

I appreciate that you serialize the receipt and have a separate deserialize-and-verify command -- it's a good pattern and I'm not sure we show it elsewhere in our examples.

examples/prorata/methods/guest/src/main.rs
choongng: That makes some sense but I think by showing the multi-read pattern in our first example without a clear warning we're effectively endorsing it. The simplest fix would be to put a note in the code and tutorials pointing to an example of the better practice.

choongng:
> Nothing to suggest that wasn't already mentioned.
> 
> I appreciate that you serialize the receipt and have a separate deserialize-and-verify command -- it's a good pattern and I'm not sure we show it elsewhere in our examples.

We can save longer discussion for another thread but one thing the example as written doesn't do is build & save a fixed version of the guest code. That means the exact same build must be used for both signing and verification which has some obvious problems. I think we'll want to provide a standard pattern for this but haven't quite worked out my functional wish list for that yet.

flaub:
I think the pattern around how to get multiple parties to rely on the same image_id and thsu the same code for signing vs verfication is probably blocked on: https://github.com/risc0/risc0/issues/116

choongng:
For the narrow case of something like prorata it shouldn't matter if builds are reproducible, we only need to save the ELF as built once and do whatever distribution steps (signing, upload to Bonsai registry, etc). I don't think it should block this example but it is something that will unlock a lot of real uses of zkVM.

examples/Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
     "json/core",
     "password-checker",
     "password-checker/core",
+    "prorata/host",
```
choongng, line 14: This looks different from the other examples but let the example use the same overall layout as our template. Is that a reasonable practice?

choongng:
Ok I think this is fairly close, should I move this on from draft?

examples/prorata/core/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+        write!(f, "CSV hash: {}\n", hex::encode(&self.csv_hash))
+    }
+}
+
+pub fn allocate(amount: Decimal, recipients: Vec<Recipient>) -> Vec<Allocation> {
+    //! Completely allocate the total funds `amount` by shares into dollars and
+    //! cents amounts for each recipient.
+    //!
+    //! amount - total amount to distribute
+    //! recipients - list of recipients with their share of the total amount
```
flaub, line 97:
```suggestion
/// Completely allocate the total funds `amount` by shares into dollars and
/// cents amounts for each recipient.
///
/// amount - total amount to distribute
/// recipients - list of recipients with their share of the total amount
pub fn allocate(amount: Decimal, recipients: Vec<Recipient>) -> Vec<Allocation> {
```

examples/prorata/core/src/lib.rs
```diff
@@ -0,0 +1,285 @@
...
+    amount: Decimal,
+    recipients: Vec<Recipient>,
+    target: &str,
+) -> Option<Allocation> {
+    // Allocate the total funds `amount` by shares into dollars and cents amounts
+    // for the recipient with name `target`.
+    //
+    // amount - total amount to distribute
+    // recipients - list of recipients with their share of the total amount
+    // target - name of the recipient to allocate for
```
flaub, line 143:
```suggestion
/// Allocate the total funds `amount` by shares into dollars and cents amounts
/// for the recipient with name `target`.
///
/// amount - total amount to distribute
/// recipients - list of recipients with their share of the total amount
/// target - name of the recipient to allocate for
pub fn allocate_for(
    amount: Decimal,
    recipients: Vec<Recipient>,
    target: &str,
) -> Option<Allocation> {
```

flaub:
Is there a good one-line description for this PR that we could use?

choongng:
"Add example of ZK financial calculation in examples/prorata/"

I poked around a bit and it's not obvious how to change the message GitHub will use for the auto-merge, do I need to squash it all manually?

flaub:
There's an edit button at the very top of the PR UI. I'll go ahead and update it now.

flaub (approved):

PR #596, opened 2023-06-01
Remove python from /opt/hostedtoolcache

imdaisyadams:
Removing python/3.10 from /hostedtoolcached seems to have fix the issue seen [here](https://github.com/risc0/risc0/actions/runs/5138834761/jobs/9249808060#step:15:9)


flaub:
From what I can tell, these failures of the license check are legit instances where the new code is missing the expected license header.

PR #603, opened 2023-06-05
Publish Recursion Verifier

SchmErik:
This pull request moves the verifier for the recursion circuit from our private repo. In order to do this, I split out the `CircuitDef` trait into a `CircuitCoreDef` and `CircuitProveDef` traits in order to keep the recursion prover code inside the private repo and brought over several files used for verification such as receipt.rs. In addition to this, I brought over the generated rust files info.rs, taps.rs, poly_ext.rs, and control_id.rs to this repo.


SchmErik:
marking as a draft for now - I found some changes that need to be made in bootstrap..

risc0/zkvm/src/recursion/receipt.rs
```diff
@@ -0,0 +1,288 @@
...
+#[cfg(not(target_os = "zkvm"))]
+use crate::receipt::compute_image_id;
+#[cfg(not(target_os = "zkvm"))]
+use crate::recursion::CIRCUIT_CORE;
+use crate::{
+    sha::{self},
+    ControlId,
+};
+
+/// TODO
```
mothran, line 31: We should probably fill in all these TODO docstrs before merging.

risc0/zkvm/src/recursion/receipt.rs
```diff
@@ -0,0 +1,288 @@
...
+        all_ids.push(Digest::from_hex(digest_str).unwrap());
+    }
+    for digest_str in crate::recursion::control_id::RECURSION_CONTROL_IDS {
+        all_ids.push(Digest::from_hex(digest_str).unwrap());
+    }
+    all_ids
+}
+
+/// todo
+pub fn tagged_struct(tag: &str, down: &[Digest], data: &[u32]) -> Digest {
```
mothran, line 61: `pub` ?

risc0/zkvm/src/recursion/receipt.rs
```diff
@@ -0,0 +1,288 @@
...
+        let output_hash = read_sha_halfs(&mut seal_meta);
+        if output_hash != self.meta.digest() {
+            return Err(VerificationError::JournalDigestMismatch);
+        }
+        // Everything passed
+        Ok(())
+    }
+}
+
+/// todo
```
mothran, line 239: This is the datatype bonsai will most likely return, so we should probably get this fully documented.

mothran (changes requested):
Looks good, just some comments about docstr's and pub visibility

mothran (dismissed):
Looks good, just some comments about docstr's and pub visibility

risc0/zkvm/src/recursion/receipt.rs
SchmErik: thanks, this doesn't need pub!

risc0/zkvm/src/recursion/receipt.rs
```diff
@@ -0,0 +1,288 @@
...
+#[cfg(not(target_os = "zkvm"))]
+use crate::recursion::CIRCUIT_CORE;
+use crate::{
+    sha::{self},
+    ControlId,
+};
+
+/// TODO
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RecursionReceipt {
```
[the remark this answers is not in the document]
SchmErik: thanks for pointing this out, this can actually move back to wip!

flaub (commented):
Looking good, but I think we need to get all the docstr TODOs finished before landing. A strategy that can help is to keep modules private and then re-export specific types from those modules to make them public. This reduces the amount of docstrs needed.

risc0/circuit/rv32im/src/taps.rs
```diff
@@ -16,7 +16,9 @@
 
 use risc0_zkp::taps::{TapData, TapSet};
 
-pub(crate) const TAPSET: &'static TapSet = &TapSet::<'static> {
+#[allow(missing_docs)]
+
+pub const TAPSET: &'static TapSet = &TapSet::<'static> {
```
flaub, line 21:
```suggestion
#[allow(missing_docs)]
pub const TAPSET: &'static TapSet = &TapSet::<'static> {
```

flaub (approved):

risc0/circuit/rv32im/src/taps.rs
SchmErik: Unfortunately, this is auto generated so it'll revert back every time we update from private. When I get a little bit of time after this sequence of merges, I can open a PR to remove it from bootstrap

PR #608, opened 2023-06-06
Make session_limit optional

intoverflow:
This PR makes two minor changes to session and segment limits:

* Switch session_limit to Option
* Impose max value for segment_limit_po2

### DIFF
--- a/risc0/zkvm/src/exec/env.rs
+++ b/risc0/zkvm/src/exec/env.rs
@@ -40,9 +40,6 @@ use super::{
 /// to try and fit with 8GB of RAM.
 const DEFAULT_SEGMENT_LIMIT_PO2: usize = 20; // 1M cycles
 
-/// The default session limit specified in cycles.
-const DEFAULT_SESSION_LIMIT: usize = 64 * 1024 * 1024; // 64M cycles
-
 /// A builder pattern used to construct an [ExecutorEnv].
 #[derive(Clone)]
 pub struct ExecutorEnvBuilder<'a> {
@@ -57,7 +54,7 @@ pub struct ExecutorEnvBuilder<'a> {
 pub struct ExecutorEnv<'a> {
     env_vars: HashMap<String, String>,
     pub(crate) segment_limit_po2: usize,
-    session_limit: usize,
+    session_limit: Option<usize>,
     syscalls: SyscallTable<'a>,
     pub(crate) io: Rc<RefCell<PosixIo<'a>>>,
     input: Vec<u8>,
@@ -84,7 +81,7 @@ impl<'a> ExecutorEnv<'a> {
         1 << self.segment_limit_po2
     }
 
-    pub(crate) fn get_session_limit(&self) -> usize {
+    pub(crate) fn get_session_limit(&self) -> Option<usize> {
         self.session_limit
     }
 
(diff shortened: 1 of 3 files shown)


risc0/zkvm/src/exec/env.rs
```diff
@@ -146,8 +143,11 @@ impl<'a> ExecutorEnvBuilder<'a> {
     }
 
     /// Set a segment limit, specified in powers of 2 cycles.
+    ///
+    /// If the given value exceeds [risc0_zkp::MAX_CYCLES_PO2], then the value
+    /// [risc0_zkp::MAX_CYCLES_PO2] is used instead.
     pub fn segment_limit_po2(&mut self, limit: usize) -> &mut Self {
-        self.inner.segment_limit_po2 = limit;
+        self.inner.segment_limit_po2 = usize::min(limit, risc0_zkp::MAX_CYCLES_PO2);
```
flaub, line 150: This might be confusing to users. Maybe we should make the `build` function be failable and so it can do integrity checks on the env and return if its valid or not.

risc0/zkvm/src/exec/mod.rs
```diff
@@ -277,7 +277,7 @@ impl<'a> Executor<'a> {
     ///
     /// This can be directly used by debuggers.
     pub fn step(&mut self) -> Result<Option<ExitCode>> {
-        if self.session_cycle() > self.env.get_session_limit() {
+        if self.session_cycle() > self.env.get_session_limit().unwrap_or(usize::MAX - 1) {
             return Ok(Some(ExitCode::SessionLimit));
         }
```
flaub, line 282:
```suggestion
        if let Some(limit) = self.env.get_session_limit() {
            if self.session_cycle() > limit {
                return Ok(Some(ExitCode::SessionLimit));
            }
        }
```

risc0/zkvm/src/exec/env.rs
intoverflow: Good question. I agree that the behavior as-written might be confusing (I included a note in the docs, but it's still likely to confuse people who haven't read that note).

We could make it do that `build()` returns `Result<>`. We could also make it so that `segment_limit_po2()` returns `Result<>`, depending on where we'd want to surface the error.

I don't know how to judge between these options :/ What do you think would be best?

flaub: I think we break the builder pattern if we mage `segment_limit_po2()` return a Result, I'm actually not sure how that would work.

So I think if we return a `Result` from `build`, then we have the ability to do verification checks for this, and beyond.

intoverflow: `Result<>` from `build()` sounds reasonable to me. I'll update the PR

risc0/zkvm/src/exec/mod.rs
intoverflow: While we're in here, how do you feel about me making the comparison `>=` instead of `>` ?

(That way, if someone sets limit to `usize::MAX`, we still get reasonable behavior.)

flaub: sgtm

flaub (approved):

PR #612, opened 2023-06-06
only run workflow on prs ready for review

imdaisyadams:


flaub:
Hmm, there's many times where I've used CI in order to see if my draft PR that maybe isn't ready for review works on all the different GPU backends and other platforms.

PR #615, opened 2023-06-07
Improve execution performance

flaub:


jbruestle (approved):
As an additional idea, couldn't we also just page in the registers at the beginning of the segment (since they will certainly be used) and then drop the 'load_region_in_page' entirely?  Or even further, couldn't we also just keep a local copy of the registers in the Executor (maybe a 'write-through' cache)?

flaub:
There's many more commits I plan to follow up with. I posted this mainly to see what results I get in CI.

SchmErik (approved):
LGTM

PR #616, opened 2023-06-08
implement SessionReceipt trait

SchmErik:
Previous SessionReceipt has been renamed as FlatSessionReceipt


flaub:
It seems like we could have a single `SessionReceipt` and abstract over the `SegmentReceipt`. Isn't the verify for `SessionFlatReceipt` and `SessionRollupReceipt` basically identical?

SchmErik:
> It seems like we could have a single `SessionReceipt` and abstract over the `SegmentReceipt`. Isn't the verify for `SessionFlatReceipt` and `SessionRollupReceipt` basically identical?

I'm looking at the verify code for these two Receipts and while I think we might be able to make it the same, it'll take some work to match the data structures/modifications to the prover. I had another concern about having a single `SessionReceipt`. It would be good to get your take on it...

So a `SessionFlatReceipt` has a journal and a `Vec<SegmentReceipt>` meaning that it could have zero or more Segment Receipts but `SessionRollupReceipt` should only have single `SegmentRecursionReceipt`. When we talked about abstracting over the SegmentReceipt and combining the two Session receipt types, I thought that we meant something like this (please correct me if I'm misunderstanding):

```
pub struct SessionReceipt {
    pub journal: Vec<u8>;
    pub receipts: Vec<SegReceipt>; // let's call SegReceipt the trait that is implemented by SegmentReceipt and SegmentRecursionReceipt
}
```

This would make sense for a `SessionFlatReceipt` but I'm concerned about the implication of this definition for SessionRecursionReceipt. This struct definition allows instances where the `receipts` field could contain many SegmentRecursionReceipts or a collection of SegmentRecursionReceipt and SegmentReceipts.

risc0/zkvm/src/receipt.rs
```diff
@@ -149,24 +150,64 @@ pub struct ReceiptMetadata {
...
+    fn verify(&self, image_id: Digest) -> Result<(), VerificationError>;
+
+    /// All session receipts have a journal
+    fn get_journal(&self) -> &Vec<u8>;
+
+    /// this provides a way to serialize a receipt
+    fn encode(&self) -> Vec<u8>;
+
+    /// this is used for downcasting, primarily used for testing
+    fn as_any(&self) -> &dyn Any;
```
flaub, line 177: I'd really like to drop this somehow. can we add more methods to the trait so we don't need to downcast?

risc0/zkvm/src/receipt.rs
```diff
@@ -149,24 +150,64 @@ pub struct ReceiptMetadata {
...
+}
+
+/// A free function that verifies the receipt
+///
+/// This function is a wrapper for [SessionReceipt::verify] that allows the
+/// caller to pass parameters that implement type conversion to [Digest] by
+/// calling the [into] function.
+#[cfg(not(target_os = "zkvm"))]
+pub fn verify(
+    receipt: Box<dyn SessionReceipt>,
```
flaub, line 187:
```suggestion
    receipt: &dyn SessionReceipt,
```

risc0/zkvm/src/receipt.rs
SchmErik: How about we do `#[cfg(test)]` so that it's used for testing purposes only? I think it's useful to be able to downcast for tests so that we can more freely dive into testing the `FlatSessionReceipt`

SchmErik: And we could remove it all together once we implement the combined receipt

flaub:
To be consistent with naming, we could have `FlatSessionReceipt` and `RollupSessionReceipt`, or `SessionFlatReceipt` and `SessionRollupReceipt`.

PR #617, opened 2023-06-08
Expose `SystemState` for mocking `ReceiptMeta` instances in tests

hashcashier:
Can't create a `ReceiptMeta` without a `SystemState` instance.


SchmErik (approved):
Looks good to me!

PR #624, opened 2023-06-13
Update Cronjob Time

imdaisyadams:
The nightly cronjobs are triggering around 12:07:00 AM UTC which is around 5:07pm PST, but we actually need to run these around 07:00:00 AM UTC which is our midnight PST.


mothran (approved):
Nice catch!

morgangibbs87 (approved):

Issue #636, opened 2023-06-16
Generating Bonsai template project results in project Cargo.tomls with relative paths


justinFrevert:
Generating a project with the bonsai template specifically, as specified by the `templ-subdir`flag generates a new project with relative paths to risc0 dependencies. The expected behavior is likely that generated Cargo.tomls target risc0 dependencies by version. This is  the case with freshly generated projects based on `rust-starter`

Steps to reproduce:
`cargo risczero new hello-risc0 --templ-subdir templates/bonsai`

PR #637, opened 2023-06-16
Fix deps in bonsai template

justinFrevert:
Fixes https://github.com/risc0/risc0/issues/636 where generating a project with the bonsai template resulted in relative paths to risc0 deps.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,10 +31,8 @@ jobs:
           node-version: 18
       - run: cargo fmt --all --check
       - run: cargo fmt --all --check --manifest-path examples/Cargo.toml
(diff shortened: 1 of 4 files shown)


justinFrevert:
I removed the check on the bonsai template crate because there wasn't a similar check on the `rust-starter` one. My best guess as to the reason why is that the check breaks on the template variable values placed into the `Cargo.toml` files

flaub:
cc: @mothran @nategraf @hashcashier 

I think we plan to drop the bonsai template in this repository in favor of: https://github.com/risc0/bonsai-foundry-template. I don't think the bonsai template here was designed to work with the `risczero` tool.

Sorry for the confusion and if I'm wrong about this please re-open.

justinFrevert:
Ah okay, I didn't catch that template around the time I ran into this. Sounds good

PR #638, opened 2023-06-16
Bonsai Guest SDK

hashcashier:
# bonsai-guest-sdk

This PR introduces a new crate designed to reduce boilerplate for Bonsai guest code.

The first feature set includes two convenience methods for reading ethereum abi-encoded guest input and decoding it using `ethabi`:
1. `fn decode_eth_abi_input(types: &[ParamType]) -> Result<Vec<Token>, ethabi::Error>`, which uses `ethabi::decode`.
2. `fn decode_whole_eth_abi_input(types: &[ParamType]) -> Result<Vec<Token>, ethabi::Error>`, which uses uses `ethabi::decode_whole`.
Both methods use the input abi-encoding `types` to keep reading the necessary number of bytes for the `ethabi` decoding call to succeed, then return its output.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 resolver = "2"
 members = [
+  "bonsai/guest-sdk",
   "bonsai/sdk",
   "risc0/bootstrap",
   "risc0/bootstrap/poseidon",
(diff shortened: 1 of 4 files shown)


nategraf:
Nice work implementing the Solidity ABI! I actually found another solution to this problem yesterday based on POSIX-style I/O we quietly added a month or so ago. Here is how to succinctly read all of the input available to the guest and then decode it, which should work well for Bonsai.

https://github.com/risc0/bonsai-foundry-template/blob/victor/hackathon-suggestions/methods/guest/src/bin/fibonacci.rs

With that we can avoid maintaining our own ABI decoder.

hashcashier:
thanks, I wish that addition was a bit louder though! Let's look back on this after the conference then because I want to get a feel for whether devs will construct their input all at once using one encoding (then we can probably discard this), or whether they will append multiple independently encoded pieces together, in which case this might still be useful.

flaub:
@hashcashier @nategraf Is this still needed?

hashcashier:
I don't think so. No one needed to parse any overly complicated fragments of abi-encoded input at the hackathon, so I would say it's safe to just leave it out to avoid bloat.

PR #641, opened 2023-06-16
Migrate from BONSAI_ENDPOINT -> multi env var

mothran:


flaub (approved):

nategraf:
Note that we will want to port this to the 0.15 release branch so folks can use the changes from there.

PR #646, opened 2023-06-22
Add experimental cargo-risczero test command and support for building rust-runtime.a

nategraf:
This PR adds a new `cargo riszero test` command, which can be used to build the current crate for
the RISC Zero target, and run it's tests inside the Executor.

This is intended to allow developers of libraries intended to support the RISC Zero zkVM a way to
test their code, using all existing tests and any new ones they write for zkVM specific
functionality, without writing a guest wrapper or duplicating effort. As an example, see
https://github.com/risc0/RustCrypto-crypto-bigint/pull/2 which adds support for running all tests
for the crypto-bigint library, which has been modified to use the bigint multiplication accelerator
when built for the RISC Zero target. This may also be used to test crate compatibility with the zkVM
without any modifications e.g. in the crate validator.

As a dependency, this PR includes support for building the zkVM rust runtime as a linkable `.a`
file. This is used within the `cargo risczero test` command to link required functionality, such as
the syscalls, for the test binaries. More generally it can be used as a tool for developing other
runtimes, as the symbols can be used in any language with C FFI.


flaub (changes requested):
Could we meet to discuss this PR? At a minimum I think it probably needs to get split up possibly, but I'd also like to understand some of the design choices better.

nategraf:
> Could we meet to discuss this PR? At a minimum I think it probably needs to get split up possibly, but I'd also like to understand some of the design choices better.

Absolutely. Would love your thoughts on how to make this better, or even whether to include it at all.

flaub (approved):
Thanks for walking thru this with me and answering my questions :)

nategraf:
Great! I am going to give this a final round of polish and fix up some things at some point in the next week.

flaub:
I guess I'm wondering why we need to build a `runtime.a` as opposed to having this functionality live in the target triple fork of rust. I think it makes sense for a user to want to run a test of an arbitrary crate, but if the risc0 toolchain is installed, could we make that be enough to accomplish this?

nategraf:
> I guess I'm wondering why we need to build a runtime.a as opposed to having this functionality live in the target triple fork of rust. I think it makes sense for a user to want to run a test of an arbitrary crate, but if the risc0 toolchain is installed, could we make that be enough to accomplish this?

So for most guests, getting the panic handle, allocator, and entrypoint from the Rust toolchain seems to clearly be the right approach. As of my work on Friday, the `runtime.a` file is not generated for guest binary builds that use our standard structure (i.e. import `risc0_zkvm` and have a `main` marked by the entrypoint macro). Right now the panic handler and allocator are still in this repository, but moving them into the toolchain also seems to me to be the right approach. Whether we do that now, or later, I am not sure of.

One of the motivations, and the immediate goal of this PR, is to enable binaries build with `cargo test` and without `risc0_zkvm` declared as a Cargo dependency to be linked against `risc0` symbols (e.g. `sys_bigint`) and run with our allocator, panic handler, and entrypoint. Moving the guest SDK into the toolchain would solve this problem. It's not totally clear to me that we want our syscall symbols to be shipped with our toolchain, but it also doesn't seem unreasonable. If we don't, using the `.a` file to define those symbols can continue to be the approach when they need to be late-bound. (Example motivation for late binding https://github.com/risc0/RustCrypto-hashes/pull/2)

The less immediate use-case is support of binaries written in C/C++, or any language that can link against C symbols (e.g. Python, Golang, Zig). In those cases, the `runtime.a` file provides the guest SDK. To me providing this `.a` file seems like the right way to support those environments. In this case the `.a` file would need to include definitions for syscall symbols. It seems it would also need to implement the entrypoint so that init and finalization occur properly, but I'm not certain this approach actually works generally.

nategraf:
@jjtny1 I believe all the build system issues are worked out. It have to head out now, but I'll try out the actual `cargo risczero test` command tommorrow

risc0/cargo-risczero/Cargo.toml
```diff
@@ -30,12 +30,21 @@ reqwest = { version = "0.11", default-features = false, features = [
   "gzip",
 ] }
 risc0-binfmt = { workspace = true }
+risc0-build = { workspace = true }
+risc0-zkvm = { workspace = true, features = ["prove"] }
```
flaub, line 34: So this will essentially break the client/server model because it means anyone that wants to install `cargo-risczero` will now have to build the prover, which will cause very long compile times.

Why does `cargo-risczero` now depend on the `prove` feature?

risc0/cargo-risczero/build.rs
```diff
@@ -0,0 +1,56 @@
...
+use std::env;
+
+#[cfg(feature = "experimental")]
+mod runtime {
+    use std::{env, fs, io, path::Path};
+
+    use risc0_build::build_rust_runtime;
+    use zip::{write::FileOptions, CompressionMethod, ZipWriter};
+
+    pub fn build_and_zip_runtime() {
```
flaub, line 24: It seems like we should make this library part of the toolchain, rather than part of `cargo-risczero`

risc0/cargo-risczero/src/commands/new.rs
```diff
@@ -116,16 +116,20 @@ impl NewCommand {
                 format!("git = \"https://github.com/risc0/risc0.git\", branch = \"{branch}\"");
             template_variables.push(format!("risc0_build={spec}"));
             template_variables.push(format!("risc0_zkvm={spec}"));
+            template_variables.push(format!("risc0_zkvm_platform={spec}"));
```
flaub, line 119: Doesn't `risc0-zkvm` depend on `risc0-zkvm-platform` already?

.github/actions/rustup/action.yml
nategraf, line 1: @jjtny1 unless we enable using the upstream version of the `rustup` actions directly, we need to changes made to this file that were removed in your most recent commits. These changes are for use by CI for the crypto forks. https://github.com/risc0/RustCrypto-crypto-bigint/pull/2

nategraf: There was a discussion on this in Slack. If you search for "de-fork actions", you'll find it.

nategraf:
Question: Should we still gate `cargo risczero test` behind a feature flag?

nategraf:
Great progress. I tried to use this to test the updates to `curve25519-dalek` (at commit 42c4faf7dc3f640a7e3f7e4cbf43e7d5d6d46b67). I was not successful. Could you look into this failure as well.

risc0/cargo-risczero/Cargo.toml
nategraf: Addressed in c75f8b613b940896a4ae5be82eedfbad9a999b7f by making it activated by the `experimental` flag. It is required because the test command uses the `Executor`.

risc0/cargo-risczero/src/commands/new.rs
nategraf: This is vestigial. Addressed in c75f8b613b940896a4ae5be82eedfbad9a999b7f.

nategraf:
Sounds good to me. Definitely don't want to lose track of the comments earlier https://github.com/risc0/risc0/pull/646#pullrequestreview-1638790122

> I'm wondering if we can land it so that it doesn't get behind and then incrementally fix bugs/issues that arise, and also add CI.

Sounds good. I pushed a commit to deal with the two comments from earlier that were relevant to the codebase outside of the new command. c75f8b613b940896a4ae5be82eedfbad9a999b7f

PR #649, opened 2023-06-23
updating "getting started" link to new docs site

pdg744:
The [getting started] link currently points to https://risczero.com/docs, which is stale. 

This PR updates the link in the README to https://dev.risczero.com/zkvm/quickstart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@
 [risc-v]: https://en.wikipedia.org/wiki/RISC-V
 [crates.io]: https://crates.io
 [cargo-risczero-readme]: https://github.com/risc0/risc0/blob/main/risc0/cargo-risczero/README.md
-[website-getting-started]: https://www.risczero.com/docs
+[website-getting-started]: https://dev.risczero.com/zkvm/quickstart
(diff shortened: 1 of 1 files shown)


pdg744:
We may want to adjust this "Getting Started" section more substantially; here's what we have on the website for the zkVM quick start page: 

Screen Shot 2023-06-23 at 6 53 50 AM

tzerrell (approved):
Thanks for finding and fixing the link!

As for a more substantial update: It feels like there's room for improvement, with too much indirection in the status quo. This page isn't itself a quickstart but instead refers the reader to other quickstart resources. And I'm not sure we make the tutorial and cargo risczero links prominent enough for the reader to promptly click on them without some friction/disorientation.

That said, I'm not sure switching out the quickstart for a redirect straight to the tutorial in factors (or to cargo risczero, for that matter) would be great either -- it would be a bit disorienting to click on "Quickstart" and be dumped into a github readme saying "here's how this example works".

So I see a problem, but don't immediately see the best solution. Let's land this and iterate.

PR #653, opened 2023-06-24
Improve CUDA performance

flaub:


risc0/zkp/src/hal/cuda.rs
```diff
@@ -564,23 +564,22 @@ impl<CH: CudaHash> Hal for CudaHal<CH> {
...
         let eval_count = which.size();
         assert_eq!(xs.size(), eval_count);
         assert_eq!(out.size(), eval_count);
-        let count = 1 << po2;
 
         let stream = Stream::new(StreamFlags::DEFAULT, None).unwrap();
-        let kernel = self.module.get_function("batch_evaluate_any").unwrap();
-        let params = self.compute_simple_params(eval_count);
+        let kernel = self.module.get_function("multi_poly_eval").unwrap();
+        let params = self.compute_simple_params(out.size() * 256);
```
jbruestle, line 575: I think this is unsafe if default threads per group is not 256.  In general, probably all the 256 constants in the kernel should be blockdim.x as well…

jbruestle (commented):
Looks good minus a the fixed constant and hidden dependence on another constant.

jbruestle (approved):

PR #658, opened 2023-06-28
Unify session receipt types into a single SessionReceipt type

SchmErik:
This done by abstracting over segment receipts via the new `SegmentReceipt`
trait and using the same `verify` function for both receipt types. Now each
`SessionReceipt` contains a vector of `SegmentReceipt` trait objects. This
simplifies the Receipt types to facailitate dev-mode and dogfooding.

This PR switches the default hash function to be poseidon (which is more friendly for recursion).
In the future we need to clean up the verifier so that we can more easily allow usage of different hash functions.

### DIFF
--- a/risc0/zkvm/src/prove/mod.rs
+++ b/risc0/zkvm/src/prove/mod.rs
@@ -62,8 +62,8 @@ use risc0_zkvm_platform::WORD_SIZE;
 
 use self::{exec::MachineContext, loader::Loader};
 use crate::{
-    receipt::SessionReceipt, ControlId, Segment, SegmentReceipt, Session, SessionFlatReceipt,
-    CIRCUIT,
+    receipt::{SegmentReceipt, SessionReceipt},
+    ControlId, Segment, SegmentBaseReceipt, Session, CIRCUIT,
 };
 
 /// HAL creation functions for CUDA.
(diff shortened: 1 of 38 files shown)


risc0/zkvm/src/prove/mod.rs
```diff
@@ -292,52 +292,53 @@ where
...
-        table.insert("cpu".to_string(), prover.clone());
-        table.insert("$default".to_string(), prover);
+        let prover = Rc::new(LocalProver::new("cpu:sha256", cpu::sha256_hal_eval()));
+        table.insert("cpu:sha256".to_string(), prover.clone());
+        table.insert("$sha256".to_string(), prover);
 
         let prover = Rc::new(LocalProver::new("cpu:poseidon", cpu::poseidon_hal_eval()));
         table.insert("cpu:poseidon".to_string(), prover.clone());
-        table.insert("$poseidon".to_string(), prover);
+        table.insert("$poseidon".to_string(), prover.clone());
```
mothran, line 315: warning: we found poseidon to be far slower on CPU than the sha256 hashing so this change will probably slow down the majority of local proving.

SchmErik: Bummer! Thanks for the heads up. I saw the perf numbers and was thinking that it was something else..

SchmErik: I'll make a quick fix and bring this up internally

flaub: The related benchmark shows a significant improvement in performance, we might be reading the results wrong

flaub:
Superseded by #666

PR #662, opened 2023-06-28
Bonsai SDK - image UUID to imageID

mothran:
This PR makes a few changes to the bonsai alpha SDK
* Replaces `anyhow` with `thiserror` to make programmatic access to error types cleaner
* Replaces the sample usage in the README with a updated and more complete example with working imports
* Adds support for the REST API changes on bonsai to migrate from using UUID's for images to ImageId generated on the client side


flaub:
My impression was that you use `thiserror` internally, but it doesn't get used in the public API. Instead, you return an `anyhow::Result` which internally comes from a `thiserror`. But maybe my mental model is wrong here.

capossele (approved):
The SdkErr is going to be very useful! You read my mind!

capossele:
> My impression was that you use `thiserror` internally, but it doesn't get used in the public API. Instead, you return an `anyhow::Result` which internally comes from a `thiserror`. But maybe my mental model is wrong here.

I think that only happens in the example, but the actual SDK returns an SdkErr type

mothran:
> My impression was that you use `thiserror` internally, but it doesn't get used in the public API. Instead, you return an `anyhow::Result` which internally comes from a `thiserror`. But maybe my mental model is wrong here.

So my understanding is actually flipped, and reading through comments from the author of the two crates I think the are intended in this direction.

`thiserror` is the types that are used on your library interface that other users consume.
`anyhow` is for binary/application or "end of the chain" code that is not consumed, generally its recommended to not expose anyhow errors to library users because it adds a additional dependency vs thiserror which only provides derive features so the Error types are all `std`. 

References for this:
https://www.shakacode.com/blog/thiserror-anyhow-or-how-i-handle-errors-in-rust-apps/
https://docs.rs/thiserror/latest/thiserror/#details 
```
Thiserror deliberately does not appear in your public API. You get the same thing as if you had written an implementation of std::error::Error by hand, and switching from handwritten impls to thiserror or vice versa is not a breaking change.
```
https://nick.groenen.me/posts/rust-error-handling/#application-errors

mothran:
> The SdkErr is going to be very useful! You read my mind!

As soon as I went to add handling for the `ImageIdExists` I realized `anyhow` Result objects are very hard to downcast into their respective error types. I am really glad the interface will work for you. I am not sure if I am gonna loose some error message fidelity but let me know if you have any suggestions. I was kinda debating `snafu` vs `thiserror` but I have just used `thiserror` more in the past so I picked the one I know.

Open to suggestions, I really wanna get this interface correct!

flaub:
This was a parsing error in my brain, but I think I now understand what the `thiserror` docs are trying to say. Yes, this looks wonderful, sorry for the fuss.

flaub (approved):

PR #666, opened 2023-07-03
Receipt unification

flaub:
Refactor zkp and zkvm such that receipt types (`SegmentReceipt` and `RollupReceipt`) implements a `Receipt` trait which allows the `SessionReceipt` to hold segments or rollups.

This allows a receipt to optionally have a hash function name which is stamped at creation time. Later, a verifier can use this information to select the appropriate matching hash function.

Key changes:
* Adjust `HashFn` and `Rng` to be object-safe traits (which means removing generics in function singatures, and requiring a self receiver).
* Add `RngFactory` to support creation of `Rng`
* Adjust `HashSuite` to be a struct containing a trait-object of `HashFn` and `Rng`.
* Adjust the `Hal` to replace `HashSuite` generics with the new struct.
* Update `prove` and `verify`


risc0/zkvm/src/receipt.rs
```diff
@@ -149,76 +157,58 @@ pub struct ReceiptMetadata {
...
     /// The public data written by the guest in this Session.
     ///
     /// This data is cryptographically authenticated in
-    /// [SessionFlatReceipt::verify].
+    /// [SessionReceipt::verify].
     pub journal: Vec<u8>,
 }
 
+/// Provide common functionality implemented by all receipt types.
+#[dyn_partial_eq]
```
SchmErik, line 182: I just wanted to point this out: using `dyn_partial_eq` implicitly enables downcasting. You can see it here: https://github.com/risc0/risc0/pull/666/files#diff-86c5437677f80d40396200163171134d15dc3e6eb257f74c37f382c39da52697R205

I initially added this in my draft PR so that we could get `partialeq` on the structs implementing this trait for the serialize/deserialize tests. I don't think it's a problem to expose this but I thought I'd point it out

risc0/zkvm/src/receipt.rs
```diff
@@ -343,56 +312,33 @@ impl SessionFlatReceipt {
...
+        use hex::FromHex;
+        let check_code = |_, control_id: &Digest| -> Result<(), VerificationError> {
+            POSEIDON_CONTROL_ID
+                .into_iter()
+                .chain(SHA256_CONTROL_ID)
+                .chain(BLAKE2B_CONTROL_ID)
+                .find(|x| Digest::from_hex(x).unwrap() == *control_id)
+                .map(|_| ())
+                .ok_or(VerificationError::ControlVerificationError)
+        };
```
SchmErik, line 327: This `check_code` closure will create an iterator with all control ID's from `poseidon`, `sha256`, and `blake2b`. I think it would be more accurate if we could select the correct set of control_id's from the hash suite... I guess another question is this: is the possibility of getting the incorrect control id somehow bounded away so that it's not possible to get an accidental mismatch?

risc0/zkvm/src/receipt.rs
flaub: Ah ya I noticed that after I tried to add `as_any`, apparently that's already supported with `dyn_partial_eq`. Thanks for the heads up.

risc0/zkvm/src/receipt.rs
flaub: In the near future, I'm thinking we could add this to the `VerifierContext`:
```
control_ids: HashMap<(String, String), ControlIds>
```

Where the key is the `(circuit_id, hashfn)`.

But I think this current approach is safe in the meantime, at least from a cryptographic perspective.

SchmErik: Also - I think the logic in this PR means that we can simplify `check_code` to only take `&Digest` as a parameter. I think this can be done in a separate PR though

SchmErik (approved):
LGTM - I don't have a deep understanding of the material in the zkp folder. I know that it works for what you did in this PR but I wonder if anyone else has thoughts/comments on them?

PR #668, opened 2023-07-05
add a hook in Session to support the proof generation progress bar

dimitrysuen:
It would be great for users to view the proof generation progress. We put a hook in `Session` to make it extensible.

Also, I tested it locally. Here is the screenshot:
screenshot


risc0/zkvm/src/session.rs
```diff
@@ -92,13 +97,24 @@ pub struct Segment {
...
 }
 
+/// The Events of [Session]
+pub trait Events {
+    /// Fired before the proving of a segment starts.
+    fn on_pre_prove_segment(&self, _: &Segment) {}
+
+    /// Fired after the proving of a segment ends.
+    fn on_post_prove_segment(&self, _: &Segment) {}
+}
```
flaub, line 107:
```suggestion
pub trait SessionEvents {
    /// Fired before the proving of a segment starts.
    fn on_pre_prove_segment(&self, segment: &Segment) {}

    /// Fired after the proving of a segment ends.
    fn on_post_prove_segment(&self, segment: &Segment) {}
}
```

flaub:
This is pretty nice, thank you!

Could you add a test for this so that we don't accidentally break this feature in the future?

risc0/zkvm/src/session.rs
dimitrysuen: Thanks for the suggestion. We've renamed `Events` and `_`.

dimitrysuen:
> This is pretty nice, thank you!
> 
> Could you add a test for this so that we don't accidentally break this feature in the future?

Yeah. We added the test for the hook.

risc0/zkvm/src/session.rs
```diff
@@ -92,13 +97,25 @@ pub struct Segment {
     pub insn_cycles: usize,
 }
 
+/// The Events of [Session]
+pub trait SessionEvents {
+    /// Fired before the proving of a segment starts.
+    // The awkward let _ is required to suppress the unused variables warning
+    fn on_pre_prove_segment(&self, segment: &Segment) { let _ = segment; }
```
flaub, line 104:
```suggestion
    #[allow(unused)]
    fn on_pre_prove_segment(&self, segment: &Segment) { let _ = segment; }
```

dimitrysuen: Fixed. Thanks.

PR #669, opened 2023-07-05
Make Bonsai SDK Client Asynchronous

jjtny1:
Currently, the Bonsai SDK using `reqwest::blocking::Client` to make requests to Bonsai Alpha. However, we cannot use `reqwest::blocking` with `tokio` (https://github.com/seanmonstar/reqwest/issues/1233). This causes a panic when using the relayer (which uses the SDK to make calls to Bonsai) because the relayer relies heavily on `tokio` for asynchronous processing of proof requests.

This PR refactors the Client in the SDK to be asynchronous


bonsai/sdk/src/alpha.rs
```diff
@@ -110,30 +110,30 @@ impl SessionId {
...
-            let body = res.text()?;
+            let body = res.text().await?;
             return Err(SdkErr::InternalServerErr(body));
         }
-        Ok(res.json::<SessionStatusRes>()?)
+        Ok(res.json::<SessionStatusRes>().await?)
     }
 }
 
 /// Represents a client of the REST api
```
capossele, line 125:
```suggestion
/// Represents a client of the REST api
#[derive(Debug, Clone)]
```

capossele (approved):
Let's wait for @mothran before merging though

bonsai/sdk/src/alpha.rs
capossele: Ah and also we should make adjustments to the README to update the examples

bonsai/sdk/src/sync/mod.rs
```diff
@@ -0,0 +1,485 @@
+// Copyright 2023 RISC Zero, Inc.
```
mothran: This file seems like a duplication of the async impl. I would prefer if we could have a single implementation that is driven by two frontends. The `reqwest` blocking has a reasonable example of unifying the core implementation: https://github.com/seanmonstar/reqwest/blob/master/src/blocking/client.rs

PR #675, opened 2023-07-06
Added required dependencies

criadoperez:
On a clean Linux system the application won't build because not having the required dependencies.

Added a note on installing the required dependencies for a successful build. 

Tested on Debian 12 (bookworm).

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,12 @@ First, [install Rust][install-rust] if you don't already have it, then install t
 cargo install cargo-risczero
 ```
 
+For the above commands to build succesfully you will need to have installed the required dependencies. On a Linux system you can install them with:
+
+```bash
+sudo apt install curl build-essential libssl-dev pkgconf-bin
+```
+
 Then, create a new project (named `my_project` in this example):
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,6 +103,12 @@ First, [install Rust][install-rust] if you don't already have it, then install t
 cargo install cargo-risczero
 ```
 
+For the above commands to build succesfully you will need to have installed the required dependencies. On a Linux system you can install them with:
```
flaub, line 106:
```suggestion
For the above commands to build successfully you will need to have installed the required dependencies. On a Ubuntu system you can install them with:
```

flaub (approved):

criadoperez:
The package `pkgconf-bin` is available on [Debian](https://packages.debian.org/sid/pkg-config). On Ubuntu it's only available on version 23.04 and future version mantic. ([here are the details](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=pkgconf-bin&searchon=names)). In other words this would not work on Ubuntu current LTS, 22.04.

However the package [pkgconf](https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=pkgconf&searchon=names) is available on all versions of Ubuntu and with also work on [Debian](https://packages.debian.org/search?searchon=names&keywords=pkgconf).

So I'm replacing `pkgconf-bin` with `pkgconf`.

Tested on Debian 12 and Ubuntu 22.04LTS

PR #677, opened 2023-07-07
zkVM: add remote proving by using the bonsai sdk

SchmErik:
This change enables generating proofs remotely by using bonsai using the zkVM API. In order to execute, add the following environment variables before running the host code: BONSAI_API_KEY, BONSAI_API_URL. A new function has been added to generate an executor if these environment variables are set.


SchmErik:
This is my initial idea for integrating the bonsai alpha sdk. I haven't changed the examples and I haven't run any workloads but I did change r0vm to see what this looks like. Before I go change all of the examples, I thought this is a good stopping place for initial feedback

risc0/zkvm/src/exec/mod.rs
```diff
@@ -28,7 +28,9 @@ mod tests;
 
 use std::{cell::RefCell, fmt::Debug, io::Write, mem::take, rc::Rc};
 
+use ::bonsai_sdk::alpha::SdkErr;
```
mothran, line 31: You could probably remove this line and replace it with `use bonsai_sdk::SdkErr` after you alias bonsai_sdk::alpha below.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -190,6 +191,70 @@ pub trait Prover {
...
+                let receipt: SessionReceipt = bincode::deserialize(&receipt_buf)?;
+                return Ok(receipt);
+            } else {
+                return Err(anyhow!("Bonsai prover workflow exited: {}", res.status));
+            }
+        }
+    }
+
+    fn prove_segment(&self, _ctx: &VerifierContext, _segment: &Segment) -> Result<SegmentReceipt> {
+        Err(anyhow!(
```
mothran, line 252: Should we consider using https://doc.rust-lang.org/std/macro.unimplemented.html instead of returning a failable error? Im not sure either.

mothran (commented):
It might be helpful for review to use this new higher level API in the rust starter-template first before examples. Plus it would be a good one to convert to test / review you API changes. The SDK integration looks clear to me.

risc0/zkvm/src/prove/mod.rs
SchmErik: yeah... I'm not sure about this. I saw all of the cases that use prove_segment. the unimplemented will panic which I know we're trying to reduce...

SchmErik:
I'm thinking about renaming the `Exec` trait as `Executor` and change the existing Executor to something else so that it's easier to understand from the zkVM API user's perspective...

risc0/zkvm/Cargo.toml
```diff
@@ -40,6 +40,7 @@ tracing = { version = "0.1", default-features = false, features = ["attributes"]
 [target.'cfg(not(target_os = "zkvm"))'.dependencies]
 addr2line = { version = "0.20", optional = true }
 bincode = { version = "1.3", optional = true }
+bonsai-sdk = { path = "../../bonsai/sdk" }
```
flaub, line 43:
```suggestion
bonsai-sdk = { workspace = true }
```

flaub:
> I'm thinking about renaming the `Exec` trait as `Executor` and change the existing Executor to something else so that it's easier to understand from the zkVM API user's perspective...

How about we rename `Executor` to `LocalExecutor` so its in parity with `RemoteExecutor` and that frees up `Executor` to be the trait.

flaub:
Should we drop `SdkErr::ImageIdExists`? I would expect `upload_img` to be idempotent, so the fact that it already exists should just be transparent and never considered an error.

flaub:
I think we should split `LocalExecutor` and RemoteExecutor` into their own modules, since they aren't very related, and leave the `Executor` trait in the `exec` module.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -341,6 +406,12 @@ fn provers() -> HashMap<String, Rc<dyn Prover>> {
 /// default CPU-based prover.
 pub fn default_prover() -> Rc<dyn Prover> {
     let provers = provers();
+
+    if std::env::var("BONSAI_API_URL").is_ok() && std::env::var("BONSAI_API_KEY").is_ok() {
+        if let Some(prover) = provers.get("$bonsai") {
+            return prover.clone();
+        }
+    }
```
flaub, line 414: I think this should come after `RISC0_PROVER` so that users can override

risc0/zkvm/src/prove/mod.rs
SchmErik: @flaub do you have any thoughts on this?

risc0/zkvm/src/exec/mod.rs
```diff
@@ -124,6 +125,72 @@ impl Write for Journal {
...
 }
 
+pub trait Exec {
+    fn run(&mut self) -> Result<Session>;
+}
+
+/// Construct a new [Exec] from the ELF binary of the guest program you
+/// want to run and an [ExecutorEnv] containing relevant environmental
+/// configuration details.
+pub fn make_exec_from_elf<'a>(env: ExecutorEnv<'a>, elf: &[u8]) -> Result<Box<dyn Exec + 'a>> {
```
flaub, line 135:
```suggestion
pub fn default_executor_from_elf<'a>(env: ExecutorEnv<'a>, elf: &[u8]) -> Result<Box<dyn Exec + 'a>> {
```

risc0/zkvm/src/prove/mod.rs
flaub: I think in this case a panic is probably fine since it indicates a programmer mistake. One way to think about it is if this panics, will it bring down a decentralized node? I'd say in this case perhaps, but I'd argue that was a logical mistake by the programmer of the node software itself, and not related to any outside user input.

SchmErik: Thanks for your clarification!

flaub (approved):
I like this

PR #678, opened 2023-07-08
Revive the dual HAL

flaub:


risc0/zkvm/src/prove/mod.rs
```diff
@@ -319,6 +319,28 @@ fn provers() -> HashMap<String, Rc<dyn Prover>> {
         let prover = Rc::new(LocalProver::new("cuda:poseidon", cuda::poseidon_hal_eval()));
         table.insert("cuda:poseidon".to_string(), prover.clone());
         table.insert("$poseidon".to_string(), prover);
+        #[cfg(feature = "dual")]
```
SchmErik, line 322: I think you can move this hunk and the hunk under `feature = "metal"` and place it under a clause like
`#[cfg(all(feature = "dual"), (any(feature = "metal", feature="cuda")]`
to reduce duplication.

flaub: There's a subtle difference. One of them needs `cuda::xxx` and the other `metal::xxx`. While I could try to reduce duplication, I think it might harm readability. In the future, I'd like to try and use the `inventory` crate or some other way to register HALs and provers without it all being in one file and one function.

SchmErik (approved):

PR #680, opened 2023-07-10
Add a new executor benchmark

flaub:


morgangibbs87 (approved):
LGTM

flaub:
I've tried and failed up to this point to get the benchmark comments to be more robust. I was hoping I could just report on the exact output from `critcmp` but I keep running into issues with it. This should definitely not land in its present state.

PR #683, opened 2023-07-10
updating bevy readme

pdg744:
Right now, the readme organization on the examples in `risc0/examples` is very haphazard, and I'd like to standardize (at least partially) the organization of these readmes. 

To that end, I have: 
- made some notes about the status quo readme contents on this [Notion page](https://www.notion.so/risczero/0be3745d465c405d8a0967b824ca9672?v=574615b9a1d64ab8b160e06188dcb338)
- suggested a standard readme organization to use moving forward (see the `[template]` row at the top of the Notion page). 
- re-written the `bevy` readme using that suggested organization, in order to make this suggestion more concrete for review.

Leaving this as a draft for now, in order to give some space for iterating on the section organization. 
If the suggested organization feels right to people, I can re-write the other readmes to match.


tzerrell (approved):
Having some organization is great, and this seems the right level of detail!

What do think about replacing "Project Organization" with "Approach"? Where "Approach" might talk about the organization but might also talk about why the example was coded a certain way rather than some alternative approach. My thinking here is "how the project is structured" is only sometimes helpful, whereas "why we coded it like this" as almost always good information in an example. Which _can_ be organization in some conceptually simple projects, but can also be more motivational.

tzerrell:
Definitely agreed that "Approach" is more ambiguous. To me that was a feature rather than a bug -- I don't think every example is well-served by a "Project Organization" section.

Relatedly, I don't think we should have verbiage along those lines in every README, essentially for verbosity reasons: If we overexplain, we'll lose readers' interest and also make it hard for people to find the information we most want them to have. I worry that this information is redundant with just looking at the code, and we should instead focus most READMEs on motivation and design information that isn't part of the code. (I also think this will vary by example: for instance, the factors example will have our most novice users and should be very explicit about even rudimentary aspects of the code.)

That said, the line between overexplaining and underexplaining is a judgment call. I'm very curious to hear what other reviewers' thoughts are on this.

pdg744:
I guess my thought is that any of these demos might be a new user's first exposure to our tech. If I'm a brand new user perusing the code for this demo, I'm going to need some hand-holding to help me make sense of wtf is going on -- what's a guest, what's a host, what's a receipt, etc. 

imo, adding a dozen lines to the README on each demo in order to offer pointers to orient readers to those terms is worth the extra verbosity

pdg744:
An alternative would be to add READMEs for the `guest/src` and `src` folders... 

One way or another, it feels important that users on GitHub have navigation paths that point to [relevant parts of] the dev docs.

examples/bevy/README.md
```diff
@@ -1,12 +1,42 @@
...
-can verify the receipt and also determine that the primary entity moved `n`
-units. While this is a fairly minimal example, we encourage you to bring your
-creativity to take this example to the next level!
+To link gameplay to a particular player, you may want to pair this demo with the [ECDSA] demo, which would allow a player to sign their moves.
+
+## Project Organization
+zkVM applications are organized into a [host program] and a [guest program].
+The host program can be found in [`src/main.rs`](src/main.rs) and the guest program can be found in [`methods/guest/src/main.rs`](methods/guest/src/main.rs).
+
+The [host] first [executes] the guest program and then [proves the execution] to construct a [receipt].
```
tzerrell, line 21: The `[proves the execution]` link is dead -- either switch it to point to `#prover` not `#prove`, or add a `#prove` section to the terminology page. Personally I lean toward the latter.

tzerrell (approved):
I like the updated `Project Organization` section. Given that our plan is to make its copy a part of the example readme template, it's worth everyone doing a careful pass on, but I really like it.

examples/bevy/README.md
```diff
@@ -1,12 +1,42 @@
...
+## Quick Start
+First, [install Rust] if you don't already have it.
+Then, run the example with:
+```
+  cargo run --release
+```
 
-## Proving movement
+## Use Cases
+By using this demo as part of a [Bonsai application], you could build an app where on-chain payment depends on off-chain gameplay.
```
3lkn, line 12: No notes, just really like the changes to this section.

3lkn (approved):
Looks good!

PR #685, opened 2023-07-10
Updated bonsai sdk README example code

mothran:
Changed to `SessionRollupReceipt` -> `SessionReceipt` in the sample code of the SDK.

It would be nice to get that README sample code running as a doctest but I was unsure of how to do so without creating circular deps in `dev-dependencies`


SchmErik (approved):
LGTM

flaub (approved):

PR #660, opened 2023-06-28
chore: remove unused Once type

austinabell:
was reading code and noticed this. API looked strange and seemed unused, opening PR to see if necessary or if CI runs a test I missed in verifying

Edit: there seems to be a lot of strange things ignored by the fact the guest crate uses `#![allow(unused)]`. Would you like me to take a stab to fix these issues and clean it up?

PR #688, opened 2023-07-11
chore: remove allow(unused) from guest code, cleanup

austinabell:
Had time today to follow up on https://github.com/risc0/risc0/pull/660#issuecomment-1611780339

Opinionated points:
- The illegal instruction for non-compliant hosts was only used in `sys_panic` and wasn't clean to ignore the unreachable code. I decided to inline the logic in that function, but let me know if this isn't ideal
- `read_slice` implementation didn't handle unexpected EoF errors; I added the handling of it. This is changing logic, and I'm happy to switch it to ignoring the values, but I couldn't intuitively see a reason why this would intentionally be the case


SchmErik (dismissed):

austinabell:
> Oop! I should have run this on CI before giving approval. Looks like we have some test failures

Failure is unrelated to the changes here. Seems there was a crate updated recently within `critcmp` to use the `IsTerminal` feature which was stabilized in 1.70 Rust https://github.com/rust-lang/rust/commit/afd45c2e10d9a5dbb277852384bb3d2245239f63 (and https://github.com/BurntSushi/critcmp/commit/9a50980f4dc7dfa0481f1bcb57c4f7f46c9e1817). 

The quick fix might just be to pin the critcmp installation in https://github.com/risc0/criterion-compare-action/tree/risc0. Let me know if that's something you'd like me to PR for.

flaub:
Thanks! I recently added `--locked` but apparently pinning is also needed.

risc0/zkvm/src/guest/mod.rs
```diff
@@ -101,28 +99,17 @@ core::arch::global_asm!(include_str!("memcpy.s"));
...
-    #[cfg(target_os = "zkvm")]
-    unsafe {
-        asm!("sw x0, 1(x0)")
-    };
-    unreachable!();
-}
-
 /// Aborts the guest with the given message.
 pub fn abort(msg: &str) -> ! {
     // A compliant host should fault when it receives this syscall.
```
flaub, line 104: Perhaps replace this comment (that is now not entirely accurate since it doesn't matter if the host is compliant) with a `SAFETY:`

flaub (approved):
I generally like these changes, thank you!

austinabell:
> Thanks! I recently added `--locked` but apparently pinning is also needed.

Yeah, it's because the lockfile was updated before the most recent release `0.1.8` (https://github.com/BurntSushi/critcmp/commit/b5928445f8e07530952661ee4bc45140a9a7675f).

> I generally like these changes, thank you!

So I'm still a bit on the fence about inlining the illegal instruction in `sys_panic`. An alternative would be to switch `sys_panic` to have a null return if you're not guaranteeing an abort from that syscall, and add the illegal instruction back just for `abort` as it was.

Unclear exactly the case where the syscall and the `unreachable!()` panic will not exit the program as to why you need the illegal instruction in the first place, so maybe I'm missing why it could be more ideal to be just in the `abort` call.

SchmErik:
> So I'm still a bit on the fence about inlining the illegal instruction in `sys_panic`. An alternative would be to switch `sys_panic` to have a null return if you're not guaranteeing an abort from that syscall, and add the illegal instruction back just for `abort` as it was.
>
 
We want to guarantee an abort on the panic syscall but we trust the host to do the right thing (which is to return an error that propagates up to the executor). If this doesn't happen, it's a really nice fallback to have the illegal instruction because that will 


> Unclear exactly the case where the syscall and the `unreachable!()` panic will not exit the program as to why you need the illegal instruction in the first place, so maybe I'm missing why it could be more ideal to be just in the `abort` call.

austinabell:
> We want to guarantee an abort on the panic syscall but we trust the host to do the right thing (which is to return an error that propagates up to the executor). If this doesn't happen, it's a really nice fallback to have the illegal instruction because that will

I understand this, but what I'm curious about is when the `unreachable!()` panic that follows after the illegal instruction doesn't do exactly this? Is that also not guaranteed to panic? Looking into the asm the unreachable code compiles down to ends with an `unimp` instruction which is documented [here](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#instruction-aliases).

austinabell:
Also, curious why the [panic handler for no_std environments](https://github.com/risc0/risc0/blob/0d02111b2f84a53cf6198a0064086b281ed0f749/risc0/zkvm/src/guest/mod.rs#L130) calls `abort` (and then if for whatever reason that and the illegal instruction fails to halt, this will infinitely loop because panic is called within the panic handler) but there isn't a [panic hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) setup for std environments to match this behaviour. Is that intentional?

flaub:
> > We want to guarantee an abort on the panic syscall but we trust the host to do the right thing (which is to return an error that propagates up to the executor). If this doesn't happen, it's a really nice fallback to have the illegal instruction because that will
> 
> I understand this, but what I'm curious about is when the `unreachable!()` panic that follows after the illegal instruction doesn't do exactly this? Is that also not guaranteed to panic? Looking into the asm the unreachable code compiles down to ends with an `unimp` instruction which is documented [here](https://github.com/riscv-non-isa/riscv-asm-manual/blob/master/riscv-asm.md#instruction-aliases).

I wasn't aware that `unreachable!` turns into an illegal instruction on rv32im. Do you have a reference to the asm for this? If that's all true then I agree that we can drop the explicit asm to do an illegal instruction.

flaub:
> Also, curious why the [panic handler for no_std environments](https://github.com/risc0/risc0/blob/0d02111b2f84a53cf6198a0064086b281ed0f749/risc0/zkvm/src/guest/mod.rs#L130) calls `abort` (and then if for whatever reason that and the illegal instruction fails to halt, this will infinitely loop because panic is called within the panic handler) but there isn't a [panic hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) setup for std environments to match this behaviour. Is that intentional?

The current version of the circuit for rv32im will guarantee that an illegal instruction causes the zkvm to halt. Ideally the host will implement `sys_panic` in a reasonable way and halt immediately. So there should be no way for this to cause a loop.

For std support, I think we just rely on the existing builtin panic to do an illegal instruction, but I'm actually not certain about this. @shkoo Can you shed some light on this?

austinabell:
> I wasn't aware that `unreachable!` turns into an illegal instruction on rv32im. Do you have a reference to the asm for this? If that's all true then I agree that we can drop the explicit asm to do an illegal instruction.

I just compiled the following code:
```rust
#![no_std]

risc0_zkvm::guest::entry!(fnnc);

#[no_mangle]
#[inline(never)]
pub fn fnnc() {
    unsafe {
        core::arch::asm!("sw x0, 1(x0)")
    };
    unreachable!();
}

```

and checked the disassembled code with `riscv64-unknown-elf-objdump -d -x -S ./target/riscv-guest/riscv32im-risc0-zkvm-elf/release/method_name`. Although now that I look more carefully at it, it seems like the `jalr` instructions seem to go in an infinite loop and the unimp instructions maybe aren't hit. I don't have high confidence on this right now on what is actually executed, but I can look more into this later. My guess is that this is maybe an infinite loop of calls because `unreachable` is called within the panic handler for no_std environments.

> For std support, I think we just rely on the existing builtin panic to do an illegal instruction, but I'm actually not certain about this.

My main curiosity is why this would be inconsistent between `no_std` and `std`, and not have the std panic hook setup to call the panic syscall and issue the illegal instruction the same as `no_std`. Also, it seems strange to have an infinite recursive call of panics instead of just having an infinite loop after the panic syscall and the illegal instruction (if that's actually what's happening, again not high confidence on this).

austinabell:
So I didn't mean to bikeshed my own PR, so I'll open an issue with this question and suggestion -- also not that important of a detail. Sound good?

PR #689, opened 2023-07-11
Align implementations of image ID and system state struct hash

nategraf:
Currently the hash of the memory image root and program counter (PC) used as the "image ID" is distinct from the struct hash used within the recursion predicates for hashing the `ReaceiptMetadata` and the `SystemState` within it. In particular, `compute_image_id(...)` currently uses a non-standard SHA-256 padding when combining the Merkle root and PC, which raises difficulties in Solidity where we do not have an implementation of the SHA-256 compress function.

This PR changes `compute_image_id(...)` to use the `SystemState::digest(...)` function, making the memory commitments used in these two contexts equal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 [![Discord chat][discord-badge]][discord-url]
 [![Twitter][twitter-badge]][twitter-url]
 
-[crates-badge]: https://img.shields.io/badge/crates.io-v0.15-orange
+[crates-badge]: https://img.shields.io/badge/crates.io-v0.16-orange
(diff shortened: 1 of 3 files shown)


SchmErik (approved):

flaub (approved):

nategraf:
@SchmErik @flaub I made the digest method `pub(crate)` at first, but realized now that I actually do need access to it in my application. (I need to be able to get the post-state `SystemState` digest).

I disabled automerge to give y'all a chance to object, but feel free to hit merge.

flaub:
LGTM, thanks!

PR #690, opened 2023-07-11
cargo risczero: update README with reference to remote proving

SchmErik:
The bonsai quick start guide refers to this documentation as one of 4 options for using Bonsai. This README does not contain any mention of remote proving and could be confusing for users who wish to use cargo risczero to generate a template for remote proving. Add a simple reference to remote proving to provide additional clarity.


pdg744 (approved):
Great, thank you!

PR #697, opened 2023-07-14
refactor(ethereum-relay): Make websocket connection more reliable

Cardosaum:
This change prevents the downloader from panicking the relay[^1], as well as implementing reconnection logic for websockets.

[^1]: Assuming the node we connect to is not down for more than 7 days, see comment in the discussion session below for more information.


bonsai/ethereum-relay/src/downloader/proxy_callback_proof_request_stream.rs
```diff
@@ -11,61 +11,127 @@
...
-            let event: CallbackRequestFilter = ethers::contract::parse_log(log)?;
-            self.event_processor.process_event(event).await?
+        loop {
+            self.recreate_client(
+                &mut client,
+                &mut recreate_client,
+                MAX_RETRIES,
+                WAIT_DURATION,
+            )
+            .await?;
```
Cardosaum, line 69: It's important to not that this function will return an error if we exhaust `MAX_RETRIES`, and by extension we could panic the downloader. By current settings, if we are not able to reconnect within 7 days, we'll panic.

capossele (approved):

PR #701, opened 2023-07-16
Add xtask install

flaub:
Use `cargo xtask install` as a way to install local development tools.

See: https://github.com/matklad/cargo-xtask


capossele (approved):
Nice!

PR #703, opened 2023-07-16
Replace Receipt trait with InnerReceipt enum

flaub:
This is a breaking change and so should be deferred until after Paris.

This is one step towards supporting a WASM-based verifier.


SchmErik:
looks good to me. Let's circle back on this after Paris

risc0/zkvm/src/receipt.rs
```diff
@@ -173,45 +170,141 @@ pub struct SessionReceipt {
...
+/// [SuccinctReceipt].
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub enum InnerReceipt {
+    /// The [SegmentReceipts].
+    Flat(SegmentReceipts),
+
+    /// The [SuccinctReceipt].
+    Succinct(SuccinctReceipt),
+
+    /// A fake receipt for testing and development.
```
nategraf, line 183: Adding this suggested comment since it's the first question I had when I saw this enum variant.
```suggestion
    /// A fake receipt for testing and development.
    /// Verification will always return [VerificationError::InvalidProof]
```

risc0/zkvm/src/receipt.rs
```diff
@@ -173,45 +170,141 @@ pub struct SessionReceipt {
...
-pub trait Receipt: Debug {
+/// An inner receipt can take the form of a collection of [SegmentReceipts] or a
+/// [SuccinctReceipt].
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub enum InnerReceipt {
+    /// The [SegmentReceipts].
+    Flat(SegmentReceipts),
+
+    /// The [SuccinctReceipt].
+    Succinct(SuccinctReceipt),
```
nategraf, line 181: If I understand correctly, the primary difference between `Flat` and `Succint` is the circuit that produced the result, and `Succinct` will also be used for the outputs of all other recursion predicates in the future. If this is correct, I'd advocate for calling these the `Rv32im` and `Recursion`.
```suggestion
    /// [Rv32imReceipts], resulting from proving with the RISC-V circuit.
    Rv32im(Rv32imReceipts),

    /// [RecursionReceipt], resulting from proving with the recursion circuit.
    Recursion(RecursionReceipt),
```

risc0/zkvm/src/receipt.rs
```diff
@@ -173,45 +170,141 @@ pub struct SessionReceipt {
...
 }
 
-/// Provide common functionality implemented by all receipt types.
-#[dyn_partial_eq]
-#[typetag::serde(tag = "type")]
-pub trait Receipt: Debug {
+/// An inner receipt can take the form of a collection of [SegmentReceipts] or a
+/// [SuccinctReceipt].
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub enum InnerReceipt {
```
nategraf, line 176: Since we won't have a `Receipt` trait, should we simply call this enum `Receipt`?

nategraf (approved):

risc0/zkvm/src/receipt.rs
flaub: This will actually form the basis for "dev mode", and in that mode it won't return InvalidProof.

risc0/zkvm/src/receipt.rs
flaub: @jbruestle Will we need a different `InnerReceipt` type for union/finalized predicates? My impression was that this would be needed to handle proof sets and resolution.

risc0/zkvm/src/receipt.rs
flaub: If anything, I was thinking we'd want to rename `SessionReceipt` to just `Receipt` since that's basically what users are going to interact with.

SchmErik (approved):
LGTM

tzerrell (approved):
Looks good. Some docs are not fully updated (in particular the module level docs for the `risc0_zkvm::receipt` module inaccurately describe how many receipt types they are and what those are), but I think that can happen in a separate PR as we iterate on the receipt API.

Issue #694, opened 2023-07-14
`thiserror` doesn't allow `risc0_zkp` to compile with `no_std`


SkymanOne:
## Sumary
`risck0_zkp` has introduced `thiserror` (https://github.com/risc0/risc0/blob/0d02111b2f84a53cf6198a0064086b281ed0f749/risc0/zkp/Cargo.toml#L38C9-L38C9) crate since `0.16.0` version which doesn't allow to compile `risc0_zkvm` and `risc0_zkp` in `no_std` environment.

I know this might be my niche case. I am working on the [ink!](https://github.com/paritytech/ink) lang project where I am trying to verify the risc0 proof inside the Wasm contract. Having this dependency doesn't allow the contract to compile.

## Proposed solution
Simply remove the crate or replace with `no_std` equivalent

flaub:
I was under the impression that wasm targets do support `std`, is that not possible here?

We haven't had any explicit requirement yet that `risc0-zkp` or `risc0-zkvm` works under `no_std`, so this is a good data point that we can use to adjust our assumptions. It should be really easy to make verify run in `no_std` but the prover would be much harder.

SkymanOne:
I know that this was not the original design. ink! smart contracts are compiled in `no_std` environment and target `wasm32-unknown-unknown`. 

I have come across several problems while trying to verify the proof on-chain:
* `thiserror` dependency on `std` which was mentioned earlier
* Since `0.16` the `risc0-zkvm` causes the contract to compile with floating-point operations that are not supported and discouraged to be used on-chain.
* Dowgrading to `0.15.3` solved the issue above, and allowed to deploy the contract on-chain. However, it failed to deserialise the slice into `SessionReceipt` ([example](https://github.com/SkymanOne/ink-risc0/blob/e87b121c048cf50b5fb3ce8589e3753f2330ab17/flipper/lib.rs#L42)) causing the contract to panic. I could not investigate the cause of this, but the same code worked in `std` environment

I am very excited about this project and would love to see some support for ink! smart contracts

flaub:
> Since 0.16 the risc0-zkvm causes the contract to compile with floating-point operations that are not supported and discouraged to be used on-chain.

I don't know of any reason floating-point would be used or related to version 0.16. Maybe there's some new dependency that makes this happen. Could you provide more context on how this is manifesting?

PR #704, opened 2023-07-16
Remove thiserror from risc0-zkp

flaub:
Fix for #694


Brando753:
Have you considered instead of out right removing thiserror to create a std and no_std option? 

For example in code that I want to support no_std environments I would add:
```TOML
[features]
default = ["std"]
std = ["thiserror"]

[dependencies]
thiserror = { version = "1.0.40", optional = true}
```

And then I would use the std features like so:
```rust

#[cfg(not(feature = "std"))]
extern crate alloc;

#[cfg(feature = "std")]
use thiserror::Error;

#[derive(Debug, PartialEq)]
#[cfg_attr(feature = "std", derive(Error))]
pub enum CustomError {
    /// The Prover was unable to construct a proof with your given inputs
    #[cfg_attr(
        feature = "std",
        error("The prover was unable to construct a proof with your given inputs")
    )]
    UnableToProve,
}
```

flaub:
@Brando753 I'd like to have support for error messages in `no_std` as well. It's not much boilerplate and if it becomes onerous we could switch to `snafu`.

Brando753:
Understood; looking over the changes, it's not that bad. If you don't have too many error structs, then it's probably not worth worrying about. I don't see the no_std crate attribute. Are you planning to add that?

flaub:
> Understood; looking over the changes, it's not that bad. If you don't have too many error structs, then it's probably not worth worrying about. I don't see the no_std crate attribute. Are you planning to add that?

See: https://github.com/risc0/risc0/blob/main/risc0/zkp/src/lib.rs#L16

PR #712, opened 2023-07-17
Support SNARK seals in Bonsai relay and enable CI for Solidity code via Foundry

nategraf:
This PR:
* Adds support for supplying SNARKs to authenticate callbacks through the `BonsaiRelay` contract, thus enabling the fully-verifying Bonsai callback flow for use in local deployment and testnet environments such as Sepolia.
* Updates the Governance showcase to be based on the latest version of the Foundry template, including supporting the `run` and `upload` subcommands in the local relay CLI.
* Adds GitHub actions workflows for testing and linting Solidity code with Foundry. It includes actions to trigger `forge fmt` and `forge test` in `bonsai/ethereum` (the Solidity library for Bonsai) and `bonsai/examples/governance` (Governance showcase).

This PR is a dependency of https://github.com/risc0/bonsai-foundry-template/pull/29

Resolves https://github.com/risc0/risc0/issues/722


nategraf:
@imdaisyadams one unresolved issue is getting the end-to-end test flow setup hitting a Bonsai endpoint. I'm not 100% sure how to handle this from a standpoint of maintaining reliability in our CI workflows (i.e. maybe we don't actually want to include the end-to-end test at all?)

bonsai/examples/governance/Cargo.toml
```diff
@@ -1,12 +1,19 @@
...
 members = [
     "methods",
     "relay"
 ]
 
+[workspace.dependencies]
+# Outside of the risc0 monorepo, use the following dependency definitions.
+# risc0-build = { git = "https://github.com/risc0/risc0", branch = "release-0.16" }
+# risc0-zkvm = { git = "https://github.com/risc0/risc0", branch = "release-0.16", default-features = false }
+# bonsai-sdk = { git = "https://github.com/risc0/risc0", branch = "release-0.16" }
```
flaub, line 11: It's legal to put both `path = ` and `version =`. This allows in-tree and out-tree building to work.

nategraf:
@morgangibbs87 or @imdaisyadams, do you have any idea what might be going on here
https://github.com/risc0/risc0/actions/runs/5593389447/jobs/10226980254?pr=712

nategraf:
> @nategraf - The error seems to show that `DEP_RISC0_SYS_METAL_KERNELS_ZKP` is not defined somewhere.
> 
> ```
> --- stderr
>     thread 'main' panicked at 'CARGO_FEATURE_METAL is defined, but DEP_RISC0_SYS_METAL_KERNELS_ZKP is not: NotPresent', /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/build.rs:26:14
> ```

Ya, so what's unclear to me at the moment is how this PR would have triggered this issue. I will investigate.

risc0/zkp/src/core/hash/poseidon/mod.rs
```diff
@@ -15,6 +15,8 @@
 //! An implementation of Poseidon targeting the Baby Bear field with a security
 //! of 128 bits.
 
+extern crate alloc;
```
flaub, line 18: Why do we need these? If we do need it, it should be sufficient to specify it once at the top-level in `lib.rs`

bonsai/examples/governance/methods/guest/Cargo.toml
```diff
@@ -8,10 +8,11 @@ name = "finalize_votes"
 path = "src/bin/finalize_votes.rs"
 
 [dependencies]
+risc0-zkvm = { version = "0.16", path = "../../../../../risc0/zkvm", default-features = false, features = ["std"] }
```
flaub, line 11: Please sort deps

bonsai/examples/governance/methods/Cargo.toml
```diff
@@ -1,17 +1,22 @@
 [package]
-name = "bonsai-starter-methods"
+name = "methods"
```
flaub, line 2: I think we should still prefix this, maybe `governance-methods`

.github/workflows/main.yml
```diff
@@ -77,8 +78,7 @@ jobs:
       - run: cargo test -F $FEATURE -F profiler
       - run: cargo test -F $FEATURE --tests -- --ignored
       - run: cargo test -F $FEATURE --manifest-path examples/Cargo.toml
-      - run: cargo test --manifest-path bonsai/examples/governance/Cargo.toml
-        if: matrix.device == 'cpu'
```
flaub, line 81: Was this intentional? This is probably what's causing all the build failures. You need to specify `-F $FEATURE` if you want this to work on GPUs. But I also wonder what we gain by running this test on GPU.

flaub: I think we should also considering making the workspace `bonsai/examples` rather than `bonsai/examples/governance`

risc0/zkp/src/core/hash/poseidon/mod.rs
nategraf: Currently `cargo clippy` thinks the crate is unsued when specified in `mod.rs`, so on first pass I made this change, but looking at it again it makes more sense just to turn off the warning.

bonsai/examples/governance/methods/Cargo.toml
nategraf: Done in cf1c802c4328f19f3d6d32961dd985c1d6acb463

bonsai/examples/governance/methods/guest/Cargo.toml
nategraf: Done in cf1c802c4328f19f3d6d32961dd985c1d6acb463

.github/workflows/main.yml
nategraf: > I think we should also considering making the workspace bonsai/examples rather than bonsai/examples/governance

The main goal of the showcase is to demonstrate what a more-or-less full project looks like. If pull the workspace declarations out into `bonsai/examples` it makes it harder to copy this as a self-contained starting point/example. Maybe we can consider this again when we add a second Bonsai example? By then we will hopefully have also extracted the `relay` crate, which is not actually specific to this example.

flaub: I'm starting to think maybe we can extend the `risczero` tool to turn examples into standalone projects.

.github/workflows/foundry.yml
```diff
@@ -0,0 +1,57 @@
...
+          submodules: recursive
+
+      - name: install rust
+        uses: ./.github/actions/rustup
+
+      - name: Install Foundry
+        uses: risc0/foundry-toolchain@2fe7e70b520f62368a0e3c464f997df07ede420f
+
+      - name: check solidity code formatting in bonsai/ethereum
+        run: forge fmt --check
```
imdaisyadams, line 35: could we make this one line? `forget fmt --check bonsai/ethereum`. same with L39?

nategraf: It seems this actually yield different results. It seems in general `forge` expects the project root to be the current directory.

nategraf:
This PR is (finally) ready for review 🎉 

I did add one todo to the README. I think this should be addressed in the future as we make further improvements to the relay. https://github.com/risc0/risc0/pull/712/files#diff-2af8b9be5e649349cbac2e06a7f6dcd14c3903750e7ca741ea88640a68f0ac65R45

bonsai/examples/governance/methods/guest/Cargo.toml
```diff
@@ -17,9 +17,7 @@ k256 = { version = "=0.13.1", features = [
 ], default_features = false }
 # Directly import radium to silence warning about unused patch. See https://github.com/risc0/risc0/issues/549
 radium = "=0.7.1"
-risc0-zkvm = { path = "../../../../../risc0/zkvm", default-features = false, features = [
-  "std",
-] }
+risc0-zkvm = { version = "0.16", path = "../../../../../risc0/zkvm", default-features = false, features = ["std"] }
```
flaub, line 20: `version =` is only needed if we plan to publish the crate

nategraf: Removed in https://github.com/risc0/risc0/pull/712/commits/dc7473d831779b05ac6c657ce7ba399c3d030a0b

nategraf:
I've updated the control ID values to reflect that latest version of the "identity p254" circuit, and updated the tests for `RiscZeroGroth16Verifier.sol` to use a seal produced by the dev deployment of Bonsai proving service. End to end tests against the dev deployment of the Bonsai proving service now pass as well.

capossele (approved):

bonsai/ethereum/contracts/IBonsaiRelay.sol
```diff
@@ -38,4 +62,22 @@ interface IBonsaiRelay {
         bytes4 functionSelector,
         uint64 gasLimit
     ) external;
+
+    /// @notice Determines if the given athorization is valid for the image ID and journal.
+    /// @dev A (imageId, jounral) pair should be valid, and the respective callback authorized, if
```
tzerrell, line 67: Typos
```suggestion
    /// @notice Determines if the given authorization is valid for the image ID and journal.
    /// @dev A (imageId, journal) pair should be valid, and the respective callback authorized, if
```

tzerrell (approved):

PR #719, opened 2023-07-18
Update rust_crypto to support AssociatedOid

dobbobalina2:


flaub:
Would you mind showing some example code that requires this? We should also add a test to show it working properly.

flaub:
@dobbobalina2 This is a great start! Thank you. Could you try and add a test that uses RSA along with our Sha256 impl? I have a feeling that we're going to need deeper changes but it's a little unclear to me. The current changes fail to compile.

flaub:
@nategraf @mothran Any ideas here?

flaub:
See #726 for my attempt to solve this

nategraf:
@dobbobalina2 my recommended solution to this is to use our fork of the `sha2` crate. It lives in the [risc0/RustCrypto-hashes] repo. Adding the following line to your Cargo.toml either as a direct dependency (replacing `sha2`) or a patch should get full compatibility with RustCrypto.

```rust
sha2 = { git = "https://github.com/risc0/RustCrypto-hashes", tag = "sha2/v0.10.6-risc0" }
```

The [sha example] also includes this patch.

[sha example]: https://github.com/risc0/risc0/blob/main/examples/sha/methods/guest/Cargo.toml#L15C90-L15C90
[risc0/RustCrypto-hashes]: https://github.com/risc0/RustCrypto-hashes

PR #726, opened 2023-07-19
Add RSA support to sha2 compat

flaub:
@dobbobalina2 Here's my attempt to get `Pkcs1v15Sign` to work with the sha2 accelerator.


nategraf:
Looks like a good addition. In the general case, we can point people towards our RustCrypto fork for 100% compatibility. There was no test to run the `RsaCompat` so I added one in 2e48ada61c700dbb7359031c00c9dc9857dec263. (Although I suppose the failure was at build time anyway)

nategraf (approved):

nategraf:
This should be fixed by https://github.com/risc0/risc0/pull/726

flaub:
Is there any advantage to using `risc0_zkp::core::hash::sha::rust_crypto::Sha256` as opposed to the patched RustCrypto fork? One option is to just drop this entire `rust_crypto.rs` to make it less confusing for folks. Could we have `risc0-zkvm` rely on the patched fork also?

flaub:
@nategraf I added a more complete test, based on a reference from Discord.

@kevinnassery There's a private key here, but it's only for testing purposes.

Issue #672, opened 2023-07-06
fix panic when guest accesses higher regions of memory


SchmErik:
When we write to higher memory regions, the executor panics. Guard against this by returning a status from load/store functions

SchmErik:
This is slightly easier than I thought mainly because I was conflating 2 things into 1. Here's what I ran into:

1. zkVM panicking on out of bounds array access in the resident field of monitor
2. zkVM guest accessing unallowed regions of memory

Internally, the zkVM is allowed to access parts of memory that the guest cannot (the region starting at `0x0C00_0000` has a lot of important system state that the guest has no business tampering with for its own good). However, we need to enable the guest to understand which memory regions should not be accessed. In our effort to prevent panics we should definitely do 1 and I have a draft of this (which entails putting `Result<()>` as a function signature in a bunch of functions) but 2 only requires modifying the parts that rrs touches which is [this impl](https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/exec/monitor.rs#L395)

PR #727, opened 2023-07-19
Emulator: Improve usability by restricting memory region access for guest programs

SchmErik:
Memory regions outside of the address range TEXT_START to SYSTEM.start() - 1 is
used to store information related to the zkVM. Tampering with this region will
cause the prover to crash. Restrict the guest from writing to this region.

Fixes: #672


SchmErik:
I'm not sure if we want this commit https://github.com/risc0/risc0/pull/727/commits/bd843b837ff35e96f693f0ab8ecc502e82ec6bcf

I realized while adding all of those changes that returning Result on everything is a related but separate task

risc0/zkvm/src/lib.rs
```diff
@@ -16,9 +16,9 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![deny(rustdoc::broken_intra_doc_links)]
 #![deny(missing_docs)]
+#![feature(array_try_from_fn)]
```
SchmErik, line 19: This is a feature that's unstable that I used to return an error status when building an array. Not sure if this is ok but I thought I'd point it out.

flaub: So far the policy has been to not use any unstable features unless there's no other option. Since we're using a nightly version of the rust compiler, technically this would work. However, I'm hoping we can move to a stable compiler eventually. At least on the host-side, it should be totally fine to use a stable compiler, it's just a hassle to setup a separate toolchain for host vs guest.

SchmErik: Ok, I'll look at other options!

risc0/zkvm/src/exec/monitor.rs
```diff
@@ -83,6 +88,7 @@ impl MemoryMonitor {
         let pages = vec![None; num_pages];
         Self {
             image,
+            num_pages,
```
flaub, line 91: I think you can grab this from the `image.info`?

flaub: Ah ok, looking at more lines in the diff and I now see it is coming from `image.info`. We're just caching this value on the `MemoryMonitor` for easy access. OK, seems fine to me.

risc0/zkvm/src/exec/monitor.rs
```diff
@@ -394,20 +420,35 @@ impl MemoryMonitor {
...
+            return None;
+        }
         match size {
-            MemAccessSize::Byte => Some(self.load_u8(addr) as u32),
-            MemAccessSize::HalfWord => Some(self.load_u16(addr) as u32),
-            MemAccessSize::Word => Some(self.load_u32(addr)),
+            MemAccessSize::Byte => match self.load_u8(addr) {
+                Err(_) => None,
+                Ok(val) => Some(val as u32),
+            },
```
flaub, line 430:
```suggestion
            MemAccessSize::Byte => self.load_u8(addr).map(|x| x as u32).ok(),
```

flaub (approved):
Other than the unstable feature, I think this is great!

PR #737, opened 2023-07-23
Fix for large I/O

flaub:


risc0/zkvm/platform/src/syscall.rs
```diff
@@ -71,6 +71,9 @@ pub mod reg_abi {
 
 pub const DIGEST_WORDS: usize = 8;
 pub const DIGEST_BYTES: usize = WORD_SIZE * DIGEST_WORDS;
+pub const MAX_BUF_BYTES: usize = 64 * 1024;
+pub const MAX_BUF_WORDS: usize = MAX_BUF_BYTES / WORD_SIZE;
+pub const MAX_SHA_COMPRESS_ROUNDS: usize = 1000;
```
nategraf, line 76: `MAX_SHA_COMPRESS_ROUNDS` might be misread to imply that it controls the number of rounds within the compression function.
```suggestion
pub const MAX_SHA_COMPRESS_BLOCKS: usize = 1000;
```

nategraf, line 76: Do we also need to add a check to the build for the `Executor` that the segment limit is not set low enough that the max number of read or compress invocations does not exceed the segment limit?

flaub: WDYT of `MAX_SHA_COMPRESS_COUNT`? I can also add a comment

flaub: Good call, I'll review what the new minimum needs to be based on these parameters.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -277,15 +280,20 @@ pub unsafe extern "C" fn sys_sha_buffer(
...
-        );
+        let mut count_remain = count;
+        while count_remain > 0 {
+            let count = min(count_remain, MAX_SHA_COMPRESS_ROUNDS as u32);
+            asm!(
+                "ecall",
+                in("t0") ecall::SHA,
+                in("a0") out_state,
+                in("a1") in_state,
+                in("a2") buf,
```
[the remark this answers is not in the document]
flaub: Good find! I'll fix this and add a test

flaub:
@nategraf Hopefully addressed all your comments, thank you!

jbruestle (approved):
Looks good.  Notable that we will need to pull into WIP and since the changes to the set of PO2s will modify things like the recursion STARK->SNARK final hashes, etc.

risc0/zkp/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub mod hal {
 
 pub use risc0_core::field;
 
-pub const MIN_CYCLES_PO2: usize = 11;
+pub const MIN_CYCLES_PO2: usize = 13;
 pub const MIN_CYCLES: usize = 1 << MIN_CYCLES_PO2; // 1K
```
nategraf, line 48:
```suggestion
pub const MIN_CYCLES: usize = 1 << MIN_CYCLES_PO2; // 8K
```

nategraf (approved):

PR #738, opened 2023-07-24
Bonsai Relay with snarks

capossele:
- Refactor Bonsai Relay to run with the new contract interface, either with a Snark Verifier or without
- Add json artefacts to avoid the user to install forge as a dependency


nategraf:
> Add json artefacts to avoid the user to install forge as a dependency
With this we should add a CI rule that the artifacts must be up to date. It's worth noting a cost that this will increase the size of the Git repo, making it slower to clone overt time and take up more disk space.

bonsai/ethereum-relay/src/uploader/completed_proofs/complete_proof.rs
```diff
@@ -60,22 +65,47 @@ pub(crate) async fn get_complete_proof(
...
+                .segments
+                .last()
+                .ok_or(CompleteProofError::InvalidReceipt {
+                    id: bonsai_proof_id.clone(),
+                })?
+                .get_metadata()
+                .map_err(|_| CompleteProofError::InvalidReceipt {
+                    id: bonsai_proof_id.clone(),
+                })?;
+            Hash::from(<[u8; 32]>::from(metadata.post.digest())).into()
```
nategraf, line 90: It doesn't seem like to conversion to `Hash` is needed.
```suggestion
            <[u8; 32]>::from(metadata.post.digest())
```

bonsai/ethereum-relay/src/uploader/completed_proofs/manager.rs
```diff
@@ -14,7 +14,10 @@
 
 use std::sync::Arc;
 
-use bonsai_proxy_contract::{Callback, ProxyContract};
+use bonsai_relay_contract::{
+    bonsai_relay::Callback as RelayCallback, bonsai_test_relay::Callback as TestRelayCallback,
```
nategraf, line 18: As types, these should be the same.

bonsai/ethereum-relay/tests/e2e_test.rs
```diff
@@ -24,18 +24,23 @@ mod tests {
...
     use risc0_zkvm::{MemoryImage, Program, MEM_SIZE, PAGE_SIZE};
     use risc0_zkvm_methods::{SLICE_IO_ELF, SLICE_IO_ID};
     use tokio::time::{sleep, Duration};
 
     const BONSAI_API_URI: &str = "http://localhost:8081";
 
+    fn dev_mode(bonsai_url: String) -> bool {
+        bonsai_url.contains("localhost") || bonsai_url.contains("127.0.0.1")
+    }
+
```
nategraf, line 43: This does not need like a reliable way to tell if we are in `dev_mode`. Can we instead use whatever source of information sets the `bonsai_url` to `localhost` as the decide whether we are in dev mode instead? (This comment applies here, but is actually more relevant in `bonsai/ethereum-relay/src/lib.rs`)

bonsai/ethereum-relay/tests/solidity/contracts/ProxyTest.sol
```diff
@@ -19,44 +19,43 @@ contract Proxy {
...
     );
 
     event ProofsSubmitted();
 
-    // Callback
+    /// @notice Data required to authorize a callback to be sent through the relay.
+    struct CallbackAuthorization {
+        bytes seal;
+        bytes32 postStateDigest;
+    }
```
nategraf, line 35: Instead of redefining the same structs I would recommend declaring `Proxy is IBonsaiRelay` and importing the structures from that file.

bonsai/ethereum/contracts/callback_dummy.sol
```diff
@@ -15,29 +15,29 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.17;
 
-import "./proxy.sol";
+import "./BonsaiTestRelay.sol";
```
nategraf, line 18: I recommend importing and using the interface instead of `BonsaiTestRelay`.
```suggestion
import {IBonsaiRelay} from "./IBonsaiRelay.sol";
```

nategraf (commented):
In various places one or both of `BonsaiRelay` and `BonsaiTestRelay`. All the required interactions should be possible through the `IBonsaiRelay` interface. Does `abigen!` create Rust code for interface definitions? If so, we should use the function and `struct` definitions on the interface instead of the implementation contracts themselves.

bonsai/ethereum/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "bonsai-proxy-contract"
+name = "bonsai-relay-contract"
```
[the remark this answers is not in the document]
nategraf: Addressed in 4e124dbd1b5cc06c571a7478b54d59ba7305ed6d

bonsai/ethereum-relay/src/uploader/completed_proofs/manager.rs
nategraf: Resolved in [4449c68](https://github.com/risc0/risc0/pull/738/commits/4449c68d50d47e296bfb42e62c650b97baf9593b)

bonsai/ethereum/contracts/callback_dummy.sol
nategraf: Resolved in https://github.com/risc0/risc0/pull/738/commits/c354c6ffd4c25411860b768de83dc8266ea89da9

bonsai/ethereum-relay/tests/solidity/contracts/ProxyTest.sol
nategraf: After looking at this some more, this is part of a larger refactor that will need to happen, and not in this PR.

nategraf:
I've addressed the comments I made earlier (except with respect to `dev_mode()`) and am going to merge this into `victor/governance-showcase` now. Any further changes before merging to `main` can be made directly to that branch.

Issue #740, opened 2023-07-25
Add build instructions to ECDSA readme


pdg744:
The ECDSA readme doesn't include any instructions on how to run the example. 

Running `cargo run` or `cargo run --release` returns an error that references a `benchmark` binary and an `ecdsa-example` binary; would be nice if the README oriented users to these two binaries.

PR #742, opened 2023-07-26
Add default run to ECDSA example to make cargo run just work

nategraf:
Pointed out by @pdg744 in https://github.com/risc0/risc0/issues/740 the ECDSA example does not work
if a user goes to that folder and simply runs `cargo run`.

This is because there are two binaries defined, `ecdsa-example` and `benchmark`. `benchmark` is only
really defined for folks who want more details numbers on operation counts, and is not something a
first time viewer of the example should care about. As a result, it seems the default should
definitely by to run `ecdsa-example`.

This PR adds the `default-run` key to Cargo.toml to have `ecdsa-example` run when a user types
`cargo run`.

Resolves https://github.com/risc0/risc0/issues/740


tzerrell (approved):
Nice approach, thank you!

We may still want to add a tiny section to the readme along the lines of "Quickstart: cargo run -r", but that is more properly a topic for the example readme organizational work happening in https://github.com/risc0/risc0/pull/683 and doesn't need to go in this PR.

PR #748, opened 2023-07-28
Provide a higher-level API for most use-cases

flaub:
* Rename `SessionReceipt` to just `Receipt`.
* Drop the `RemoteExecutor`, this functionality exists purely in the `RemoteProver`.
* Drop the `Executor` trait and rename `LocalExecutor` to just `Executor`.
* Add the following higher-level functions to the `Prover` trait:
  * `prove`
  * `prove_elf`
  * `prove_elf_with_ctx`

The basic high-level API now looks like:

```
let receipt = default_prover().prove_elf(env, SOME_ELF).unwrap();
```


examples/factors/README.md
```diff
@@ -93,25 +93,27 @@ Use this command any time you'd like to check your progress.
...
       ExecutorEnv,
     };
 
     // First, we construct an executor environment
     let env = ExecutorEnv::builder().build().unwrap();
 
-    // Next, we make an executor, loading the (renamed) ELF binary.
-    let mut exec = default_executor_from_elf(env, MULTIPLY_ELF).unwrap();
-
     // We're not quite ready to run these steps yet
```
pdg744, line 109: We don't need to resolve this on this PR, but I'm wondering whether we should strike `we're not quite ready to run these steps yet`. I don't really follow what this comment is about.

examples/bevy/src/main.rs
```diff
@@ -29,15 +29,14 @@ fn main() {
...
-    let mut exec = default_executor_from_elf(env, BEVY_GUEST_ELF).unwrap();
+    // Obtain the default prover.
+    let prover = default_prover();
 
-    // Run the executor to produce a session.
-    let session = exec.run().unwrap();
-
-    // Prove the session to produce a receipt.
-    let receipt = session.prove().unwrap();
+    // Produce a receipt by proving the specified ELF binary.
```
pdg744, line 35: love this

examples/factors/README.md
flaub: It's part of the tutorial, we should probably do a deeper review of the entire README to make sure it still flows correctly. It's possible that certain steps aren't really needed anymore.

examples/bevy/src/main.rs
```diff
@@ -29,15 +29,14 @@ fn main() {
...
+    let receipt = prover.prove_elf(env, BEVY_GUEST_ELF).unwrap();
 
+    // The prover already runs a verify internally, but this is how other users
```
pdg744, line 38: hmm... this comment feels odd to me. mb this is better?
```suggestion
    // After running a user generates a receipt, other users can verify the receipt using: 
```

pdg744: is there an easy way to suggest a revision on github that spans multiple lines?

(i.e., if you hit commit on this it will leave line 39 hanging and that's annoying)

examples/browser-verify/tests/app.rs
```diff
@@ -21,6 +21,6 @@ wasm_bindgen_test_configure!(run_in_browser);
 // This runs a unit test in the browser, so it can use browser APIs.
```
pdg744, line 21: I think "it" could be clarified here. 
I tried to suggest a revision to fix this but found myself confused

examples/factors/README.md
```diff
@@ -93,25 +93,27 @@ Use this command any time you'd like to check your progress.
 
 Our next objective is to provide the guest program with input. Before we implement this, let's take a closer look at how we run and prove the guest program in `factors/src/main.rs`.
 
-In the starter template project, our host driver program creates an executor environment before constructing an executor.  When `Executor::run()` is called, it will produce a session. Calling `Session::prove()` will cause our guest program to execute, producing a receipt:
+In the starter template project, our host driver program creates an executor environment before constructing a prover.  When `Prover::prove_elf()` is called, it will produce a receipt:
```
pdg744, line 96: so much simpler!

pdg744 (approved):
Love to see these simplifications to the user-facing API! 
I left a couple in-line comments/suggestions, but nothing blocking.

examples/bevy/src/main.rs
tzerrell: When selecting the line to comment on, click and drag to select multiple lines to comment on. This works even if you don't suggest revisions, but if you do suggest revisions it lets you revise all the selected lines.

(see my multiline comment for my thoughts on the actual comment here)

examples/bevy/src/main.rs
```diff
@@ -29,15 +29,14 @@ fn main() {
...
+    // would verify the receipt.
```
tzerrell, line 39: I like that the original version indicates that the `verify` call isn't really necessary here. What about this 3rd option, which makes the redundancy more explicit?
```suggestion
    // The prover already runs a verify internally and so it's redundant to verify
    // again here. However, this is how other users would verify the receipt:
```

examples/browser-verify/tests/app.rs
tzerrell: Is this clearer? (I'm not personally find it very confusing, so I'm not sure what needs to be improved.)
```suggestion
// `test_verify` runs a unit test in the browser, so it can use browser APIs.
```

examples/factors/README.md
tzerrell: Agreed with both "do a deeper review" and "don't need to resolve on this PR"

examples/factors/README.md
```diff
@@ -217,10 +216,9 @@ Once more, the program won't do anything, but it should run successfully and bui
 For this step, we return to the main file for the host driver program at `factors/host/src/main.rs`, which currently has a placeholder comment asking to fill in with code for handling the [receipt]:
 
 ```no_compile
-    // Run the executor to produce a session.
-    let session = exec.run().unwrap();
+    // Run the prover to produce a receipt.
+    let receipt = prover.prove_elf(env, METHOD_NAME_ELF).unwrap();
```
tzerrell, line 221: Drop this line too
```suggestion
```

examples/factors/README.md
```diff
@@ -217,10 +216,9 @@ Once more, the program won't do anything, but it should run successfully and bui
...
 
-    // Prove the session to produce a receipt.
     let receipt = session.prove().unwrap();
```
tzerrell, line 222: Drop this line too: There's no session (that the user sees) anymore.
```suggestion
```

risc0/zkvm/src/prove/mod.rs
```diff
@@ -176,196 +171,45 @@ where
     pub eval: Rc<E>,
 }
 
-/// TODO
+/// A Prover can execute a given [MemoryImage] and produce a [Receipt] that can
+/// be used to verify correct computation.
 pub trait Prover {
```
tzerrell, line 176: :tada: I like the new `Prover` and the various `prove` wrappers, especially `prove_elf`, are great

tzerrell (approved):
I like it!

SchmErik (approved):
This is really nice!

mothran (approved):
This looks great! Also means we can dogfood lots of the `examples/` dir against bonsai by the looks of it!

PR #750, opened 2023-07-29
change non-standard uint to unsigned int to support Windows

dimitrysuen:
I tested locally on my Windows.
Here're the screenshots:
image
image


flaub:
These changes are to the CUDA kernels where a uint is available. I don't think this change has anything to do with Windows.

PR #751, opened 2023-07-29
Fixes for clippy and add lint to CI

flaub:


risc0/zkvm/platform/src/syscall.rs
```diff
@@ -541,59 +583,58 @@ pub unsafe extern "C" fn sys_getenv(
...
-        ptr
+    // Check to make sure we keep space between the heap and the
+    // stack so they don't accidentally step on each other.
+    let mut stack_pointer: usize;
+    #[cfg(target_os = "zkvm")]
+    unsafe {
+        asm!("add {stack_pointer}, sp, zero", stack_pointer = out(reg) stack_pointer)
+    };
+    #[cfg(not(target_os = "zkvm"))]
+    let stack_pointer: usize = crate::memory::STACK_TOP as usize;
```
SchmErik, line 632: I see that this is now enabled for `#[cfg(not(target_os = "zkvm"))]` what are the use cases for this function outside of the zkvm?

SchmErik (approved):

risc0/zkvm/platform/src/syscall.rs
flaub: This was simply to make tooling (like clippy and rust-analyzer) aware of more code. It's not meant to ever be executed except on the `zkvm` target.

SchmErik: ok thanks for the context!

PR #752, opened 2023-08-01
zkvm: add dev mode

SchmErik:
This PR adds dev mode: a mode for the zkvm to be able to generate Fake inner receipts. This mode can be enabled by defining the `DEV_MODE` environment variable. I've also added a `disable-dev-mode` feature flag so that dev mode cannot be used. Let me know what you think!

Another thought: instead of a `disable-dev-mode` flag, we could have an `enable-dev-mode` flag and force the user to set `DEV_MODE` to enable dev mode.


3lkn:
Early review comment but I would favor an `enable-dev-mode` flag set and like the idea of defaulting to having this mode disabled safely for out-of-box builds.

risc0/zkvm/src/prove/dev_mode.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    fn prove_session(&self, _ctx: &VerifierContext, session: &Session) -> Result<Receipt> {
+        log::info!("prove_session: {}", self.name);
+        eprintln!(
+            "WARNING: Proving in DevMode does not generate a valid receipt. \
+            Receipts generated from this process are invalid and should never be used in production."
+        );
+
+        if cfg!(feature = "disable-dev-mode") && std::env::var("DEV_MODE").is_ok() {
```
tzerrell, line 58: Should we be using a RISC Zero-specific variable name? e.g. `RISC_ZERO_DEV_MODE` instead of just `DEV_MODE`?

risc0/zkvm/src/prove/dev_mode.rs
```diff
@@ -0,0 +1,89 @@
...
+use anyhow::Result;
+use risc0_binfmt::MemoryImage;
+
+use crate::{
+    prove::Prover,
+    receipt::{InnerReceipt, Receipt, VerifierContext},
+    Executor, ExecutorEnv, Segment, SegmentReceipt, Session,
+};
+
+/// An implementation of a [Prover] for development and testing purposes.
```
tzerrell, line 26: Can we expand this documentation a bit? Since running intended-to-be-secure code in DevMode would be such a substantial failure, I want developers to have every opportunity to learn when to use it (and when not)
```suggestion
/// An implementation of a [Prover] for development and testing purposes.
///
/// When code is "proven" with this DevModeProver, no actual proof is produced.
/// Instead, the guest code is executed and a fake receipt with the true journal
/// but no cryptographic information is returned. This receipt will verify with
/// if DevMode is turned on, but will otherwise fail to verify.
///
/// This allows for rapid development by quickly iterating on code without
/// forcing the developer to wait for proving to complete. However, it must not
/// be used in production as it provides no security whatsoever.
///
/// DevMode is only used when the environment variable `DEV_MODE` is set. It can
/// also be fully disabled at compile time, regardless of environment variables,
/// by setting the feature flag `disable-dev-mode` on the `risc0_zkvm` crate.
```

risc0/zkvm/src/receipt.rs
```diff
@@ -298,6 +303,16 @@ pub struct SegmentReceipt {
 }
 
 /// Context available to the verification process.
+#[cfg(not(feature = "disable-dev-mode"))]
+pub struct VerifierContext {
+    /// A registry of hash functions to be used by the verification process.
+    pub suites: BTreeMap<String, HashSuite<BabyBear>>,
+    /// DevMode facilitates development by generating a fake receipt
```
tzerrell, line 310: Let's expand the docs here as well. Perhaps something like the below. I did notice I duplicated a fair amount of copy, if someone has an alternative with less duplication I'm open to that as well.
```suggestion
    /// DevMode facilitates development by generating a fake receipt
    ///
    /// DevMode facilitates development by generating a fake receipt
    ///
    /// When this is `true`, the verifier DOES NOT CHECK if the cryptographic
    /// information in the receipts is correct. This enables faster development
    /// by bypassing the need for honest receipts, but it provide NO SECURITY
    /// WHATSOEVER. Do not use in production.
    ///
    /// DevMode is only used when the environment variable `DEV_MODE` is set. It
    /// can also be fully disabled at compile time, regardless of environment
    /// variables, by setting the feature flag `disable-dev-mode` on the
    /// `risc0_zkvm` crate.
```

tzerrell (commented):
:tada:

DevMode is very useful but also something of a foot gun. I've made some suggestions trying to make this as clear as reasonable from a documentation perspective, but I'd love any further suggestions on how to make the usage and security profile of DevMode clearer and/or how to make the warnings less obnoxious while still being sufficient to keep users secure.

risc0/zkvm/src/prove/dev_mode.rs
```diff
@@ -0,0 +1,89 @@
...
+        let receipt = Receipt::new(InnerReceipt::Fake, session.journal.clone());
+        let image_id = session.segments[0].resolve()?.pre_image.compute_id();
+        receipt.verify_with_context(
+            &VerifierContext {
+                suites: BTreeMap::new(),
+                #[cfg(not(feature = "disable-dev-mode"))]
+                dev_mode: true,
+            },
+            image_id,
+        )?;
```
flaub, line 71: we can probably drop this since we don't actually need to run `verify` during fake proof construction.

risc0/zkvm/src/receipt.rs
```diff
@@ -252,8 +252,13 @@ impl InnerReceipt {
         match self {
             InnerReceipt::Flat(x) => x.verify_with_context(ctx, image_id.into(), journal),
             InnerReceipt::Succinct(x) => x.verify_with_context(ctx),
-            // TODO: add support for dev-mode
-            InnerReceipt::Fake => Err(VerificationError::InvalidProof),
+            InnerReceipt::Fake => {
+                #[cfg(not(feature = "disable-dev-mode"))]
+                if ctx.dev_mode {
```
flaub, line 257: Shouldn't we check the env var here too?

risc0/zkvm/src/receipt.rs
flaub: I'm not sure we need this bool. In my mind, verify would just check to see if the `InnerReceipt` was `Fake`.

risc0/zkvm/src/prove/dev_mode.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    fn prove_session(&self, _ctx: &VerifierContext, session: &Session) -> Result<Receipt> {
+        log::info!("prove_session: {}", self.name);
+        eprintln!(
+            "WARNING: Proving in DevMode does not generate a valid receipt. \
+            Receipts generated from this process are invalid and should never be used in production."
+        );
+
+        if cfg!(feature = "disable-dev-mode") && std::env::var("DEV_MODE").is_ok() {
+            panic!("zkVM: dev mode is disabled. unset DEV_MODE environment variable to produce valid proofs")
```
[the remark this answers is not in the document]
3lkn:
```suggestion
            panic!("zkVM: Inconsistent settings -- please resolve. Dev mode is disabled by feature flag, but the DEV_MODE environment variable is present. To produce valid proofs, unset the DEV_MODE environment variable. To enable dev mode and accept INSECURE, INVALID proofs, remove the `disable-dev-mode` feature flag from the `risc0-zkvm` crate.")
```
Taking a stab at this one to clarify settings and recommendations.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -213,6 +217,9 @@ pub trait Prover {
 }
 
 fn provers() -> HashMap<String, Rc<dyn Prover>> {
+    if cfg!(feature = "disable-dev-mode") && std::env::var("DEV_MODE").is_ok() {
+        panic!("zkVM: dev mode is disabled. unset DEV_MODE environment variable to produce valid proofs")
```
3lkn, line 221: Suggesting same addition as with other comment.
```suggestion
        panic!(""zkVM: Inconsistent settings -- please resolve. The DEV_MODE environment variable is set but dev mode has been disabled by feature flag. To produce valid proofs, unset the DEV_MODE environment variable. To enable dev mode and accept INSECURE, invalid proofs, remove the `disable-dev-mode` feature flag.")
```

risc0/zkvm/src/prove/dev_mode.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn get_peak_memory_usage(&self) -> usize {
+        // note, we don't actually run the prover, so very little memory is consumed
+        // while generating the receipt
+        0
+    }
+
+    fn prove_session(&self, _ctx: &VerifierContext, session: &Session) -> Result<Receipt> {
+        log::info!("prove_session: {}", self.name);
+        eprintln!(
+            "WARNING: Proving in DevMode does not generate a valid receipt. \
```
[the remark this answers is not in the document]
3lkn: I like `dev mode` or `Dev Mode` because `DevMode` looks like a variable name and we already have an env var and a feature flag thrown into the mix. Gently leaning towards `dev mode` for that reason.

risc0/zkvm/src/prove/mod.rs
```diff
@@ -308,6 +320,12 @@ pub fn default_prover() -> Rc<dyn Prover> {
             return prover.clone();
         }
     }
+    if cfg!(not(feature = "disable-dev-mode")) && std::env::var("DEV_MODE").is_ok() {
+        eprintln!("WARNING: proving in dev mode");
```
3lkn, line 324:
```suggestion
        eprintln!("WARNING: proving in dev mode. This will not generate valid, secure proofs.");
```

risc0/zkvm/src/prove/tests.rs
```diff
@@ -270,6 +270,33 @@ fn continuation() {
...
+    std::env::set_var("DEV_MODE", "1");
+    let receipt = prove_nothing("$devmode").unwrap();
+    assert_eq!(receipt.inner, InnerReceipt::Fake);
+    receipt.verify(MULTI_TEST_ID).unwrap();
+    std::env::remove_var("DEV_MODE");
+}
+
+#[test]
+#[should_panic(
+    expected = "zkVM: dev mode is disabled. unset DEV_MODE environment variable to produce valid proofs"
```
3lkn, line 284:
```suggestion
    expected = "zkVM: dev mode is disabled by feature flag. To produce valid, secure proofs, unset the DEV_MODE environment variable."
```

risc0/zkvm/src/prove/dev_mode.rs
3lkn:
```suggestion
/// An implementation of a [Prover] for development and testing purposes.
/// This DevModeProver does not produce an actual proof.
/// Instead, the guest code is executed and a fake receipt is returned with
/// accurate journal contents but no cryptographic information.
/// Because the receipt is fake, a verifier can only "verify" this receipt
/// if DevMode is turned on; verification will otherwise fail.
///
/// CONVENIENT, BUT NOT MEANT FOR PRODUCTION
/// Dev mode supports rapid development by allowing the developer to quickly iterate
/// on code without being forced to wait for proving to complete. 
/// However, it must not be used in production as it provides no security whatsoever.
///
/// How to enable and disable dev mode:
/// Dev mode is only used when the environment variable `DEV_MODE` is set. It can
/// be fully disabled at compile time, regardless of environment variables,
/// by setting the feature flag `disable-dev-mode` on the `risc0_zkvm` crate.
```

risc0/zkvm/src/receipt.rs
SchmErik: yeah, you're right. I realized that I made this more complicated than necessary. We can just check the env var because we aren't supporting dev mode in `no_std` anyway.

jbruestle:
I think the result of our long discussion earlier was: as long as dev mode only takes effect when an environment variable is set (true), it should be enabled by default.  Very few people run code without controlling the environment variables.  Heck, if LD_PRELOAD is set you can swap out libc, so game over anyway.  The truely paranoid should have a way to full disable runtime config, but for normal folks not have dev mode enabled by default (ie it reads the env variable) is annoying and doesn’t contribute to security in most threat models.

risc0/zkvm/src/prove/dev_mode.rs
tzerrell: I'm happy with your rewrite, but please add a line break between the first (summary) line and the in-depth documentation.

(If this isn't just a typo & you're not familiar, Rust will use everything before the first empty line as a summary, so it's important to put a full empty line in early. Best practices are to keep the summary single line. See https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components for more information.)

risc0/zkvm/src/prove/tests.rs
```diff
@@ -270,6 +270,34 @@ fn continuation() {
     }
 }
 
+fn run_dev_mode() {
+    use crate::receipt::InnerReceipt;
+    std::env::set_var("DEV_MODE", "1");
```
flaub, line 275: I think this is dangerous. If this test runs in parallel, it could break other tests. @mothran Mentioned this which seems apropos: https://stackoverflow.com/questions/35858323/how-can-i-test-rust-methods-that-depend-on-environment-variables/67433684#67433684

risc0/zkvm/src/receipt.rs
```diff
@@ -252,11 +252,23 @@ impl InnerReceipt {
...
+        if cfg!(not(feature = "disable-dev-mode")) && std::env::var("DEV_MODE").is_ok() {
+            return Ok(());
+        }
+        Err(VerificationError::InvalidProof)
+    }
+
+    #[cfg(not(feature = "std"))]
+    fn verify_fake() -> Result<(), VerificationError> {
+        Err(VerificationError::InvalidProof)
+    }
```
flaub, line 270:
```suggestion
    fn verify_fake() -> Result<(), VerificationError> {
        #[cfg(all(feature = "std"))]
        if cfg!(not(feature = "disable-dev-mode")) && std::env::var("DEV_MODE").is_ok() {
            return Ok(());
        }
        Err(VerificationError::InvalidProof)
    }
```

risc0/zkvm/src/prove/tests.rs
SchmErik: So... I looked into this and realized that any test that does proof generation or verification needs to use this. I looked at the implementation of that crate and it seems to use a mutex on the env var. I wonder if it would be easier to mark all proof tests and verification tests as `serial`. On the other hand, it would be nice to run these dev mode tests in isolation. I feel bad having to turn all proving tests serial because of this one test...

risc0/zkvm/src/prove/dev_mode.rs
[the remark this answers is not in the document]
SchmErik: I'll go with dev mode for now!

SchmErik:
@flaub so I decided to `#[ignore]` the dev mode test and run all ignored tests using one thread... Let me know if you have any thoughts on this approach...

risc0/zkvm/src/receipt.rs
```diff
@@ -252,11 +252,18 @@ impl InnerReceipt {
...
             InnerReceipt::Flat(x) => x.verify_with_context(ctx, image_id.into(), journal),
             InnerReceipt::Succinct(x) => x.verify_with_context(ctx),
-            // TODO: add support for dev-mode
-            InnerReceipt::Fake => Err(VerificationError::InvalidProof),
+            InnerReceipt::Fake => Self::verify_fake(),
         }
     }
 
+    fn verify_fake() -> Result<(), VerificationError> {
+        #[cfg(all(feature = "std"))]
```
flaub, line 260:
```suggestion
        #[cfg(feature = "std")]
```

flaub (approved):

risc0/zkvm/src/prove/mod.rs
```diff
@@ -213,6 +217,12 @@ pub trait Prover {
 }
 
 fn provers() -> HashMap<String, Rc<dyn Prover>> {
+    if cfg!(feature = "disable-dev-mode") && std::env::var("RISC0_DEV_MODE").is_ok() {
```
nategraf, line 220: @SchmErik sorry I'm a bit late to the party, but I am reading through this now and have a concern.

`std::env::var("RISC0_DEV_MODE").is_ok()` has some unfortunate semantics such as `RISC0_DEV_MODE=false` and `RISC0_DEV_MODE=''` both yielding the system in dev mode. My impression is most developers have never actually `unset` and environment variable because in most cases, setting it to empty string is almost always treated as semantically equivalent. Can we update this to avoid these sharp edges? One appraoch that would seem pretty clear would be to have `true` and `false` (and possibly others like `1` and `0`) turn dev mode on and off respectively, unset would generally be "off", and other values would be an error (e.g. `RISC0_DEV_MODE=sure` bails).
 
 One other reason for this suggestion is that in test environments it may actually make sense to default to dev mode, and set `RISC0_DEV_MODE=false` to get proving to occur. This suggestion is less important and I am not 100% convinced this is the right way to do it.

nategraf: (Oh, and another weirdness about this line is that if the `Err` is something other than `VarError::NotPresent` then this will treat it as unset)

SchmErik: @nategraf good catch! I'll address this in a follow-up PR

PR #757, opened 2023-08-03
workspace: patch cc crate

SchmErik:
The `cc` crate 1.0.80 breaks our build due to our use of our `-Z build-std` flag. This is a temporary fix to pin `cc` to a known working version to get guests to build and CI to run.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,3 +3,6 @@ xtask = "run --release --package xtask --"
 
 [build]
 rustflags = ["-Dclippy::all", "-Dwarnings"]
+
+[patch.crates-io]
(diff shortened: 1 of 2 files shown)


.cargo/config.toml
```diff
@@ -3,3 +3,6 @@ xtask = "run --release --package xtask --"
 
 [build]
 rustflags = ["-Dclippy::all", "-Dwarnings"]
+
+[patch.crates-io]
+cc = { git = "https://github.com/rust-lang/cc-rs", rev = "e5bbdfa" }
```
flaub, line 8: I think we should probably pin to a version rather than a commit

PR #762, opened 2023-08-04
Recover non processed blocks after websocket connection loss

### LINKED ISSUE #743: Store last processed block on Ethereum Relay
Currently if we lose connection with a node we might lose events while reconnecting; We should store the last processed block in order to start from this state after successfully reconnecting.

Cardosaum:
# Overview

Currently, in case of websocket connection loss, we lose all events emitted until we get a reconnection. This PR seeks to implement recovery logic to the running Relay. 

## Caveats 

It's important to not that the current Relay implementation does not provide permanent state, so if for any reason the Relay itself goes down we currently don't provide a way to recover prior emitted events, the only events that are currently possible to recover are the ones a current running Relay identifies it didn't process due to websocket connection loss. 

### Related issues

Closes #743


capossele (approved):

flaub:
FYI for future reference, when we squash & merge, it'd be good to try and cleanup the final commit message so that in the future, if someone is doing a bisect or looking thru the history, they don't see all of the intermediate commit messages. GitHub provides a way to edit the final commit message just during the squash & merge process.

PR #781, opened 2023-08-10
Commented out governance with-zkvm tests

mothran:
Temporary fix for CI pipeline speed until we can optimize this test suite


nategraf (approved):
Related to https://github.com/risc0/risc0/issues/783

Issue #783, opened 2023-08-11
Refactor Forge tests in the Governance example to take less time to run


nategraf:
As of `2012f208b2330de843579800d00c62f5414a4578`, it takes about an hour to run `forge test` in `bonsai/examples/governance`. This is too long and slows down development in the monorepo.

The tests for this example need to be refactored to take less time, before the can be re-enabled to run in CI

PR #784, opened 2023-08-11
Disable Governance test with zkVM in CI

nategraf:
Related to https://github.com/risc0/risc0/issues/783, the governance example tests with zkVM execution take too long to run. This PR disables those tests for now.

Note that a basic build and functionality test for the zkVM guest is included in `cargo test`. Tests for the Solidity code is included in `forge test` with `TEST_USE_ZKVM=false`. So both these components are individually covered by tests in CI. With this PR their integration will not be covered until https://github.com/risc0/risc0/issues/783 is fixed.

### DIFF
--- a/.github/workflows/foundry.yml
+++ b/.github/workflows/foundry.yml
@@ -47,15 +47,15 @@ jobs:
         run: forge test -vvv
         working-directory: bonsai/ethereum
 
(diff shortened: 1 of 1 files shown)


nategraf:
Oops, forgot about https://github.com/risc0/risc0/pull/781

PR #785, opened 2023-08-11
Convert dev mode unit tests to integration tests

flaub:
This avoids env var concurrency issues in unit tests


SchmErik:
thanks for doing this!

PR #786, opened 2023-08-11
Prepare for release v0.17.0

flaub:
The following crates require a newer version of rust, so these will be deferred until a later release:

* `bonsai-ethereum-contracts`
* `bonsai-ethereum-relay`
* `bonsai-rest-api-mock`


capossele:
Should we also include the crate `bonsai-ethereum-contracts` @nategraf ?

capossele (approved):

examples/bevy/README.md
```diff
@@ -24,14 +33,15 @@ To link gameplay to a particular player, you may want to pair this demo with the
...
+
 zkVM applications are organized into a [host program] and a [guest program].
 The host program can be found in [`src/main.rs`] and the guest program can be found in [`methods/guest/src/main.rs`].
 
 The [host] first [executes] the guest program and then [proves the execution] to construct a [receipt].
 The receipt can be passed to a third party, who can examine the [journal] to check the program's outputs and can [verify] the [receipt] to ensure the integrity of the [guest program]'s execution.
 
-[`src/main.rs`]: https://github.com/risc0/risc0/tree/v0.16.1/examples/bevy/src/main.rs
-[`methods/guest/src/main.rs`]: https://github.com/risc0/risc0/tree/v0.16.1/examples/bevy/methods/guest/src/main.rs
+[`src/main.rs`]: src/main.rs
```
pdg744, line 43: @tzerrell and I have recently been expressing some uncertainty about whether the relative paths or full url paths are preferable in this context

examples/wordle/README.md
```diff
@@ -2,7 +2,7 @@
 
 ## Quick Start
 
-First, make sure [rustup](https://rustup.rs) is installed. This project uses a [nightly](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html) version of [Rust](https://doc.rust-lang.org/book/ch01-01-installation.html). The [`rust-toolchain`](rust-toolchain) file will be used by `cargo` to automatically install the correct version.
+First, make sure [rustup](https://rustup.rs) is installed. This project uses a [nightly](https://doc.rust-lang.org/book/appendix-07-nightly-rust.html) version of [Rust](https://doc.rust-lang.org/book/ch01-01-installation.html). The [`rust-toolchain.toml`](../../rust-toolchain.toml) file will be used by `cargo` to automatically install the correct version.
```
pdg744, line 5: Can we align the recommendations here with the other examples?
```suggestion
First, [install Rust] if you don't already have it.

Next, install the `cargo-risczero` tool and install the toolchain with:
```bash
cargo install risczero
cargo risczero install
```

Then, run the example with:
```bash
cargo run --release
```

nategraf:
> Should we also include the crate `bonsai-ethereum-contracts` @nategraf ?

It's not necessary since Foundry uses submodules for dependency management. It does mean that the `bonsai-foundry-template` will use a git dependency for the relay instead of pulling from crates.io, but I think that's fine for now. Overall I think leaving it off crates.io for now seems like the right call. All we need right now is a git tag.

nategraf (approved):

PR #789, opened 2023-08-11
aligning quick start

pdg744:
Several examples are currently missing essential information that's necessary to run the examples. 
This PR resolves that issue by adding those instructions to the Quick Start sections. 

Edit: Frank points out that to reduce maintenance, it's probably better to have a single `installation` page and have all the READMEs point there. probably worth landing this as a stop-gap

### DIFF
--- a/examples/browser-verify/README.md
+++ b/examples/browser-verify/README.md
@@ -3,15 +3,17 @@
 Verify a RISC Zero program in your browser using WASM!
 
 ## Quickstart
-
-### Dependencies
-
-In addition to [Rust] and [Node.js], you will need
+In addition to [Rust] [Node.js] , you will need
 ```
 cargo xtask install
 cargo xtask gen-receipt
 ```
 
+Next, install the `cargo-risczero` tool and install the toolchain with:
+```bash
+cargo install cargo-risczero
+cargo risczero install
+```
 ### Running a test of in-browser verification
 
 From this directory, run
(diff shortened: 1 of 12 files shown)


examples/browser-verify/README.md
```diff
@@ -24,5 +26,6 @@ where `$BROWSER` is one of
 - `firefox`
 - `safari`
 
+[install Rust]: https://doc.rust-lang.org/cargo/getting-started/installation.html
```
tzerrell, line 29: This isn't used... did you want the intro to be more in line with the phrases of the other examples (which would use this link)?

tzerrell (commented):
Nice! the browser-verify README came out a bit mangled though

examples/browser-verify/README.md
pdg744: Ahh yes, good catch. I've removed the unused [install rust] link. 
The extra dependency of [Node.js] made the flow more natural to me with `install` outside of the link text.

flaub:
Is there anyway we could just link to an installation page or embed content? I'm imagining the install process eventually being able to specify versions and other things, and we'll have to update a bunch of pages each time.

pdg744:
> Is there anyway we could just link to an installation page or embed content? I'm imagining the install process eventually being able to specify versions and other things, and we'll have to update a bunch of pages each time.

I like that idea, but not going to work on it over the weekend... 
I think it's probably worth landing this in the mean time; current READMEs are insufficient to run the examples

flaub (approved):
Don't work on the weekend! I didn't mean to imply this was a blocker.

PR #791, opened 2023-08-13
Move bootstraps over to xtask

flaub:


xtask/src/install.rs
```diff
@@ -0,0 +1,33 @@
...
+pub struct Install;
+
+impl Install {
+    pub fn run(&self) {
+        install_wasm_tools();
+    }
+}
+
+fn install_wasm_tools() {
+    if which("wask-pack").is_err() {
```
mothran, line 29: `wask-pack` ?

SchmErik (approved):
I think this looks good aside from the typo

PR #799, opened 2023-08-15
Reproducible builds via Docker

### LINKED ISSUE #116: Make ELF binaries build deterministically

### LINKED ISSUE #755: Sending code or docker container to verifier
Due to issues with deterministic builds, it is currently not clear how to allow a verifier to be certain that an image ID comes from the stated guest code. Our attempts to serialize a receipt, send it to another device, and have that device verify it with the same risc0 project which they have compiled locally have all failed.

One approach which was suggested [here](https://github.com/risc0/risc0/issues/116#issuecomment-1403810479) is to build the project in a docker container and send the container for another party to verify. However, we have also not succeeded in running our risc0 projects (which run correctly on desktop) in docker containers. The most common error trace we get is shown below:

```   Compiling factors-methods v0.1.0 (/factors/methods)
error: failed to run custom build command for `factors-methods v0.1.0 (/factors/methods)

Caused by:
  process didn't exit successfully: `/factors/target/debug/build/factors-methods-7b921c936b2e62ef/build-script-build` (signal: 11, SIGSEGV: invalid memory reference)
  --- stdout
  Standard library /factors/target/debug/build/factors-methods-3c8cc1556b2bae16/out/rust-std_ad062567ca92a9b48981bf6d4ac3375b88abc750249689321f6ab3bbc996d11d does not exist; downloading 
```

Would it be possible to provide an example of how another party is able to verify a receipt from guest code? If the solution involves a docker container, than a Dockerfile / project which builds successfully would be very helpful.

capossele:
This PR extends the `cargo risczero` tool with a new `build` command.

It is based on a docker image built with this [Dockerfile](https://github.com/risc0/risc0/blob/59934ccad2735d154ad66bf8d4b00b809b69d9d1/risc0/cargo-risczero/docker/Dockerfile.release) and pushed to the [risczero](https://hub.docker.com/repository/docker/risczero/risc0-guest-builder/general) Docker hub account. 

## Installation
To install the tool just run:
```bash
cargo install --path risc0/cargo-risczero/
```

## Example usage
```bash
# Build the factors example
cargo risczero build --manifest-path examples/factors/methods/guest/Cargo.toml
```
This will generate an output similar to:
```
ELFs ready at:
./elfs/multiply/multiply - ImageID: 9ee1612b0a7e270f8df248e47dc85d9908ff4c1e7df42f398a65ca878b57a23d
```

Packages that generate multiple ELFs are supported as well: 
```bash
# Build the chess example
cargo risczero build --manifest-path risc0/zkvm/methods/guest/Cargo.toml      
```
This will generate an output similar to:
```
ELFs ready at:
./elfs/risc0_zkvm_methods_guest/hello_commit - ImageID: eb12f9b97d8759327f651afeb09ae9a5713e7dbc428284d453b8cf56e8dadd5a
./elfs/risc0_zkvm_methods_guest/multi_test - ImageID: 761900e766a4ae1d8edcb2b49dc9aee54b94e42c9b0d6421cfb112314c4e3efc
./elfs/risc0_zkvm_methods_guest/slice_io - ImageID: 3f2ad1a2d500ab4ab927eebe241d872d3f598065b8987b182410cf01f350f74c
```

Resolves #755 
Resolves #116


flaub:
This is great! I think `cargo risczero build` (drop the -guest) is a better name.

capossele:
Yeah I agree. I'm all open for testing suggestions. So far I came up with the following idea:
- Build a few guests on Linux and on Mac and check that both result into same values. This requires having Docker installed on both. I'm not sure though if would be feasible to compare a GitHub job (on Linux) result to another job (on Mac)

flaub:
What if we did a manual build locally on any platform, then we checked in the resulting IMAGE_ID. Then in CI, we just run the build and compare?

capossele:
Great idea! that would work! However we must be sure that the Cargo.lock used to do the local build is the same that gets checked in CI. I can try to put together a test for this 👍

.github/workflows/reproducible_build.yml
```diff
@@ -0,0 +1,33 @@
+name: Reproducible-build
+
+on:
+  push:
+    branches: [ main, 'release-*' ]
+  pull_request:
+    branches: [ main, 'release-*' ]
+
+  workflow_dispatch:
```
mothran, line 9: Generally I think this type of PR checker would live as `job` inside of the `main.yml`  like: https://github.com/risc0/risc0/blob/main/.github/workflows/main.yml#L195 

Any reason here to have it broken off?

risc0/cargo-risczero/Cargo.toml
```diff
@@ -12,17 +12,22 @@ keywords = ["risc0", "risczero", "tool", "cli", "generate"]
 [dependencies]
 anyhow = { version = "1.0", features = ["backtrace"] }
 cargo-generate = "0.18"
+cargo_metadata = "0.17"
 clap = { version = "4.0", features = ["derive"] }
 const_format = "0.2"
 dirs = "5.0"
+docker-generate = "0.1"
```
mothran, line 19: TIL about this crate! Super useful

risc0/cargo-risczero/docker/Dockerfile.release
```diff
@@ -0,0 +1,11 @@
+# To build run: docker build -f Dockerfile.release -t risczero/risc0-guest-builder:v0.17 .
+FROM ubuntu:20.04@sha256:3246518d9735254519e1b2ff35f95686e4a5011c90c85344c1f38df7bae9dd37
```
mothran, line 2: small nit 20.04 will EOL in 2024, vs 22.04 which is is EOL (base support tier)  in 2026

Should we jump to 22.04 right now to extend the lifecycle of these images?

risc0/cargo-risczero/src/commands/build_guest.rs
```diff
@@ -0,0 +1,131 @@
...
+use anyhow::{Context, Result};
+use cargo_metadata::MetadataCommand;
+use clap::Parser;
+use docker_generate::DockerFile;
+use risc0_zkvm::{sha::Digest, MemoryImage, Program, MEM_SIZE, PAGE_SIZE};
+use risc0_zkvm_platform::memory;
+
+use crate::utils::{ensure_binary, CommandExt};
+
+const DOCKER_IGNORE: &'static str = "**/elfs\n**/target\n**/Dockerfile\n**/.git";
```
mothran, line 31: Sometimes in situations like these I use the `r#""#` multi-line string syntax with rust. That way you don't have to manually put in the \n characters. [ref](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals)

risc0/cargo-risczero/src/commands/build_guest.rs
```diff
@@ -0,0 +1,131 @@
...
+            .run(r"find target/riscv32im-risc0-zkvm-elf/release -maxdepth 1 -type f -exec test -x {} \; -exec cp {} /tmp/ \;");
+
+        let binary: DockerFile<'_> = DockerFile::new()
+            .comment("binary stage")
+            .from_alias("binary", "scratch")
+            .env(pkg_env)
+            .copy_from("build", "/tmp", "/$PKG_NAME");
+
+        let file = DockerFile::new().dockerfile(build).dockerfile(binary);
+        let mut dockerfile = File::create("Dockerfile")?;
```
mothran, line 104: Do we cleanup these temp files anywhere?

risc0/cargo-risczero/README.md
```diff
@@ -64,3 +64,18 @@ Use the `build-toolchain` command to build the toolchain locally for your host.
 [risc-zero]: https://risczero.com
 [install-rust]: https://doc.rust-lang.org/cargo/getting-started/installation.html
 [rust-starter]: https://github.com/risc0/risc0/tree/main/templates/rust-starter
+
+## build
+
+Use the `build` command to build a guest code for the target `riscv32im-risc0-zkvm-elf`. The resulting ELF binary will be saved on `elfs/`. Warning: this requires having docker installed.
```
mothran, line 70: I think the other requirement is that the have run `cargo risczero install` first right?

SchmErik:
Here's an idea for testing - We have the fault checker guest code that will be a part of the zkvm when #736 lands. We could have some sort of check that runs the build on the fault checker and ensure that the elf and the image ID matches on each CI run.

risc0/cargo-risczero/docker/Dockerfile.release
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends ca-certificates clang curl libssl-dev pkg-config
+RUN curl --proto '=https' --tlsv1.2 --retry 10 --retry-connrefused -fsSL 'https://sh.rustup.rs' | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN cargo install cargo-risczero
+RUN cargo risczero install
```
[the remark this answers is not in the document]
flaub: Yeah, I think the next evolution of the install command is to have it specify version.

risc0/cargo-risczero/README.md
flaub: Isn't the toolchain only required in the docker image?

risc0/cargo-risczero/src/commands/build_guest.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn build(&self) -> Result<()> {
+        Command::new("docker")
+            .args(["build", "--output=elfs/", "--target=binary", "."])
+            .run_verbose()?;
+
+        Ok(())
+    }
+
+    /// Compute the image ID for a given ELF.
+    fn image_id(&self, elf_path: &PathBuf) -> Result<Digest> {
```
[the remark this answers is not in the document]
flaub: I actually just made it be exposed from the `risc0-binfmt` crate (since that's actually where the impl is). I'll submit a PR soon

risc0/cargo-risczero/README.md
mothran: ahhh yes thats correct.

.github/workflows/reproducible_build.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  reproducible-build:
+    runs-on: [self-hosted, prod, Linux, X64, docker]
+    steps:
+      # This is a workaround from: https://github.com/actions/checkout/issues/590#issuecomment-970586842
+      - run: "git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :"
+      - uses: actions/checkout@v3
+        with:
+          # Recursive checkout required to build Solidity contracts and bindings.
+          submodules: recursive
```
flaub, line 24: Do we need this for this job?

.github/workflows/reproducible_build.yml
```diff
@@ -0,0 +1,33 @@
...
+      - uses: ./.github/actions/rustup
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: echo "expected_image_id=9ee1612b0a7e270f8df248e47dc85d9908ff4c1e7df42f398a65ca878b57a23d" >> $GITHUB_ENV
+      - run: echo "branch_image_id=$(echo -n $(cargo run --bin cargo-risczero -- risczero build --manifest-path examples/factors/methods/guest/Cargo.toml 2>&1) | rev | cut -d ' ' -f 1 | rev )" >> $GITHUB_ENV
```
flaub, line 30: It'd be great if we could make this a rust unit or integration test so that devs locally can just do `cargo test` to see if we broke the docker stuff.

carterbrett:
@heavypackets had some concerns around potential entropy introduced by crates.io. @heavypackets would be great to hear from you on whether the particular approach here should be impacted?

mothran:
Talked to @hmrtn a bit, might be worth checking that the source we copy from the users manifest location includes a Cargo.lock (assuming its a lock file for the guest code). Then throwing a warning to the user if its not included to prevent crates.io drift.

capossele:
The docker container already uses `--locked` as a manifest-path option to ensure that the Cargo.lock (of the guest package) is up-to-date. We can surely add a check even before starting the docker build to make sure the Cargo.lock file is indeed present

risc0/cargo-risczero/src/commands/build_guest.rs
capossele: Added 👍

heavypackets:
My first thought it that we're using the term "reproducible build" in a non-standard way. For an example of historical context, the organization I've supported since my Docker days goes into the motivations: https://reproducible-builds.org.

A reproducible build implies: identical source code -> identical binary via a deterministic compiler, a guarantee that is portable across disparate environments into perpetuity. This is important for distributing applications that can be rebuilt anywhere with guaranteed security, a primary motivation behind reproducible builds in OSS.

This is not the contract we are implying with our tool. We are concerned about Image ID stability. As I understand, the Image ID ignores parts of the ELF that are unimportant to the program semantics, so binary comparison is unimportant in the zkVM security model. In our model, rustc could produce binaries who's binary hashes do not match but have the same Image ID -- that distinction might not be important in our VM security model, but that is not generalizable to meet the same bar implied by a "reproducible build" of a compiled program.

So, I'd recommend renaming this feature, preferably avoiding anything that implies a deterministic build. Then document it heavily, so people aren't confused what we are/aren't guaranteeing.

mothran:
> So, I'd recommend renaming this feature, preferably avoiding anything that implies a deterministic build. Then document it heavily, so people aren't confused what we are/aren't guaranteeing.

I am confused about this because we can emit a MemoryImage object reproduce-ably and deterministically using this system (ideally assuming we squash most of the entropy bugs). The ELF might not be, but you don't need a ELF, you can skip right to the memory image as your primary input on both the ZKVM and Bonsai.

.github/workflows/reproducible_build.yml
capossele: removed 👍

.github/workflows/reproducible_build.yml
capossele: Added 👍

mothran:
> > So, I'd recommend renaming this feature, preferably avoiding anything that implies a deterministic build. Then document it heavily, so people aren't confused what we are/aren't guaranteeing.
> 
> I am confused about this because we can emit a MemoryImage object reproduce-ably and deterministically using this system (ideally assuming we squash most of the entropy bugs). The ELF might not be, but you don't need a ELF, you can skip right to the memory image as your primary input on both the ZKVM and Bonsai.

I just double checked and while the ImageID is derived from most members of the MemoryImage object, I think its possible to have the data on disk change without the ImageID change because of the `PageTableInfo` but I would wanna double check with @flaub first. 

If that is the case then I think my above comment is probably not accurate.

flaub:
The ImageID is a function that has a type like: `fn compute_image_id(used_elf_pages, page_size, page_table_addr, pc)` where:
* `used_elf_pages`: the set of pages that are read by the ELF binary (this means that some ELF pages are not accessed and not needed)
* `page_size`: The circuit is currently hard coded to 1024 bytes.
* `page_table_addr`: The linear address where the page table exists in the runtime memory image. This is also hard coded in the circuit.
* `pc`: Program counter, this is the starting address to begin execution.

If any of the parameters change, you get a different ImageID. The part that we're concerned with here for a "reproducible build" is just the `used_elf_pages`.

With this approach (using docker) or any other approach we attempt, the goal is to be able to ensure that if multiple users submit the same source code, they get identical `use_elf_pages`. Thus, to me this satisfies the requirements of a "reproducible build system". So I do think we imply "identical source code -> identical binary via a deterministic compiler". It just so happens that the "deterministic compiler" has to be wrapped in a docker image until `rustc` and `cargo` itself can satisfy this requirement.

risc0/cargo-risczero/Cargo.toml
```diff
@@ -12,17 +12,22 @@ keywords = ["risc0", "risczero", "tool", "cli", "generate"]
...
 flate2 = "1"
 fs2 = "0.4"
+hex = "0.4"
 reqwest = { version = "0.11", default-features = false, features = [
   "blocking",
   "json",
   "rustls-tls",
   "gzip",
 ] }
+risc0-zkvm = { workspace = true }
```
flaub, line 29: We should be able to directly depend on `risc0-binfmt` here since all the types you're using from `risc0-zkvm` are just re-exports.

risc0/cargo-risczero/src/commands/build_guest.rs
```diff
@@ -0,0 +1,201 @@
...
+                    .to_string(),
+                self.expected_image_id
+            );
+        }
+    }
+    #[test]
+    #[ignore] // requires Docker to be installed
+    fn test_reproducible_multiply_method() {
+        let multiply_test = Tester {
+            manifest_path: "examples/factors/methods/guest/Cargo.toml".to_string(),
```
flaub, line 194: Something a bit more persistent would be `risc0/zkvm/methods/guest/Cargo.toml`

risc0/cargo-risczero/src/commands/build_guest.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        Ok(())
+    }
+
+    /// Build the dockerfile and ouputs the ELF.
+    ///
+    /// Overwrites if an ELF with the same name already exists.
+    fn build(&self) -> Result<()> {
+        Command::new("docker")
+            .args(["build", "--output=elfs/", "--target=binary", "."])
```
flaub, line 127: I think the output should land under `target/`, so maybe `target/riscv-guest/riscv32im-risc0-zkvm-elf/docker`?

.gitignore
```diff
@@ -3,11 +3,14 @@
...
 .DS_Store
 .idea/
 bazel-*
 Cargo.lock
 dist/
+Dockerfile
 rust-project.json
 target/
 tmp/
+elfs/
```
flaub, line 16: I'd prefer the ELFs to land under `target/` someplace

heavypackets:
> * `used_elf_pages`: the set of pages that are read by the ELF binary (this means that some ELF pages are not accessed 
> If any of the parameters change, you get a different ImageID. The part that we're concerned with here for a "reproducible build" is just the `used_elf_pages`.
> 

This highlights my concern. We aren't concerned about the resulting binary on disk being identical through separate builds and we aren't testing binary hashes in our build system. In this scenario, we could have different binaries on disk produce the same ID because parts are ignored -- that would be an example of a violation of reproducible builds as I've seen understood in OSS and in PL/compiler space.

For the Web3 example, the NEAR documentation linked in Slack does claim to build identical binaries: https://docs.near.org/sdk/rust/building/reproducible-builds

> To verify your contract user can build it themselves and check that the binaries are identical.

In the NEAR case, two slightly different on disk but identical semantically contracts would not meet their definition of reproducibility.

We are claiming that some derivative of those compiled binaries + other inputs will be identical. That's a different definition, and one that requires its own terminology IMO. Or, at least, very careful explanation how/why our definition is different.

flaub:
In fact, I think we should also tell users that they should check that their ELF binary is identical after using this docker build. I'm also fine with doing this comparison in our CI tests.

risc0/cargo-risczero/src/commands/build_guest.rs
capossele: modified 👍

risc0/cargo-risczero/src/commands/build_guest.rs
capossele: We are not committing the Cargo.lock for that method though

risc0/cargo-risczero/Cargo.toml
capossele: done 👍

flaub (approved):

jchilds872 (approved):

CLEILI:
Is there any difference between different operate system? I recently build the same guest code in my macos and linux server,  but I got two different ImageID of the code? However the ImageID should not be changed for the same code. Please tell me some solutions or debug method?

SchmErik:
> Is there any difference between different operate system? I recently build the same guest code in my macos and linux server, but I got two different ImageID of the code? However the ImageID should not be changed for the same code. Please tell me some solutions or debug method?

no, there shouldn't be any differences regardless of operating system if you're using the reproducible build system. What command are you using to build the guest?

CLEILI:
> > Is there any difference between different operate system? I recently build the same guest code in my macos and linux server, but I got two different ImageID of the code? However the ImageID should not be changed for the same code. Please tell me some solutions or debug method?
> 
> no, there shouldn't be any differences regardless of operating system if you're using the reproducible build system. What command are you using to build the guest?

Sorry for replying to you late. I just use "cargo run -F cuda" to run the host code and the guest code is built automatically. The guest image ID is found at target/debug/build/methods-.../out/method.rs. The image IDs are indeed different.

PR #801, opened 2023-08-15
Add start of ExternalProver 

flaub:
An ExternalProver invokes r0vm to do the actual proving. This first version is pretty limited.

* Make the persistent encoding of MemoryImage a bit more future proof
* Split SystemState into separate module
* Expose `risc0_binfmt::compute_image_id` as public
* Add --image to `r0vm`
* Use bincode for receipt format
* Rename `RemoteProver` to `BonsaiProver`


risc0/zkvm/src/prove/external.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let mut stdin = child.stdin.take().ok_or(anyhow!("Failed to open stdin"))?;
+        std::thread::spawn(move || stdin.write_all(&env.input));
+
+        let status = child.wait()?;
+        if !status.success() {
+            bail!("Failed to run r0vm");
+        }
+
+        let receipt_bytes = std::fs::read(receipt_path)?;
```
SchmErik, line 82: I think it would be nice to verify the receipt here using the verifier context

risc0/zkvm/src/prove/external.rs
```diff
@@ -0,0 +1,102 @@
...
+/// An implementation of a [Prover] that runs proof workloads via an external
+/// `r0vm` process.
+pub struct ExternalProver {
+    name: String,
+    r0vm_path: PathBuf,
+}
+
+impl ExternalProver {
+    /// Construct an [ExternalProver].
+    pub fn new<P: AsRef<Path>>(name: &str, r0vm_path: P) -> Self {
```
SchmErik, line 38: Here, I think we could add a hash function name as a parameter to `new`. Similar to the LocalProver, we can have the caller/user request a hash function to be used. When creating an instance of the External prover, we could call r0vm to see a list of supported hash functions. If the requested hash function doesn't exist, we can do  `bail("hash function {} is not supported by r0vm. Supported hash functions {}", requested_hash, supported_hash_list)`.

tzerrell (approved):

risc0/zkvm/src/prove/external.rs
flaub: Done

risc0/zkvm/src/prove/external.rs
flaub: done

PR #808, opened 2023-08-18
Reorg to support client/server architecture

flaub:
## Why

The zkVM is moving towards a client/server architecture for two main reasons:
1. Streamline the installation of the zkVM for most users.
2. Allow multiple versions of the zkVM (and hence circuits) to be used by clients

The first is achieved by using prebuilt binaries that include HW acceleration which is usually expensive to build from source. The second is to rely on a stable interface between client and server, so that newer clients can work with older servers. Both are supported by decoupling the expensive zkVM bits from the user bits so that the dependency graph of the client is minimized. It's also possible in the future to support clients written in other languages.

## Reorg

The biggest change is a reorganization of the code:
* zkvm
  * guest
  * host
    * client
    * server
      * exec
      * prove

## Prover Traits

The `Prover` trait has been split into `Prover` and `DynProverImpl`. There is now a distinction among client-side `Prover`s, and server-side `ProverImpl`s. A `Prover` is a higher-level interface which includes execution. A `ProverImpl` is only concerned with the proving phase.

A `ProverOpts` is added to the prove functions so that users can configure the prover. Right now it only supports setting the hashing function.

There are three client `Trait` impls:
* Bonsai
* External (calls `r0vm`, not yet complete)
* Local (only available when `prove` feature is enabled).

## Feature Flags

Currently the `prove` feature is enabled by default. In the future, once the `ExternalProver` is complete and it's easy for users to install the prebuilt `r0vm` via `cargo risczero install`, the `prove` feature will not be enabled by default and should only be used by `r0vm`.

A new `client` feature is added to toggle support for the client-side. The intention is for `default = ["client"]` once client/server is complete. Pure verify (useful for WASM) is supported with `default-features = false`.

## Changes

* `RISC0_DEV_MODE` now requires the value to be one of `["1", "true", "yes"]` to be considered enabled.
* The `provers()` registry is dropped, along with `get_prover()`. Use `get_prover_impl()` to select a `ProverImpl` based on `ProverOpts`.
* Drop support for low-level custom syscalls on the `ExecutorEnv`. It's difficult and potentially dangerous to expose this across the client/server interface.
* Dev mode is implemented as a `ProverImpl` rather than a `Prover` so that it works independent of the hash function (or other `ProverOpts`) that are selected.


risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+/// be used to verify correct computation.
+pub trait Prover {
+    /// Return a name for this [Prover].
+    fn get_name(&self) -> String;
+
+    /// Prove the specified [MemoryImage].
+    fn prove(
+        &self,
+        env: ExecutorEnv<'_>,
+        ctx: &VerifierContext,
```
nategraf, line 39: It seems like maybe this should be removed from this method since `VerifierContext` cannot be serialized, and so can't be sent to a remote prover. It seems like this would be set when constructing the prover, and the `opts` here acts as the way for client to indicate which hash suite should be used.

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+}
+
+/// Return a default [Prover] based on environment variables, falling back to a
+/// default CPU-based prover.
+pub fn default_prover() -> Rc<dyn Prover> {
+    if !is_dev_mode()
+        && std::env::var("BONSAI_API_URL").is_ok()
+        && std::env::var("BONSAI_API_KEY").is_ok()
```
nategraf, line 87: One of the slight annoying things about this factoring is that it means I need to `unset` these variables in my terminal if I want local proving. (In particular, by calling `unset BONSAI_API_URL` as `BONSAI_API_URL=''` will result in `std::env::var("BONSAI_API_URL").is_ok()` remaining as `true`)

risc0/zkvm/src/host/client/exec.rs
nategraf, line 1: It seems this file is an API definition for the client-server protocol. I might be a good idea to extract it into an `api` module. This would make the code layout a bit more clear and also allow the `server` modules to build without the `client` module (i.e. allow the `prove` feature without the `client` feature)

nategraf, line 1: Are these protocol messages something we want to define with protobuf? Maybe not in this PR, but it seems like to achieve the goal of client and server version interop, these messages are the key interface point.

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+
+use self::{dev_mode::DevModeProver, prover_impl::ProverImpl};
+use crate::{
+    is_dev_mode, Executor, ExecutorEnv, ProverOpts, Receipt, Segment, SegmentReceipt, Session,
+    VerifierContext,
+};
+
+/// A DynProverImpl can execute a given [MemoryImage] and produce a [Receipt]
+/// that can be used to verify correct computation.
+pub trait DynProverImpl {
```
nategraf, line 66: It took me a minute to understand that `Prover` is the client and `DynProverImpl` is the server. (Also it's slightly confusing that a type with the name `Impl` is a `trait` instead of a `struct`). Maybe this should be something like `ProverServer`?

risc0/zkvm/src/host/client/exec.rs
```diff
@@ -0,0 +1,208 @@
...
+
+#[derive(Serialize, Deserialize)]
+enum PosixCmd {
+    ReadAvail,
+    Read,
+    Write,
+}
+
+#[derive(Serialize, Deserialize)]
+enum IpcSyscall {
```
nategraf, line 63: Calling these `Ipc*` seems to imply they can only be sent over IPC mechanisms, whereas is seems quite possible that this interaction might occur over a network in the future. Should we just call this `Syscall` and the others respectively? The module should disambiguate that this is not the same as a syscall elsewhere.

nategraf (approved):

risc0/zkvm/src/host/client/exec.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+impl Default for IpcResponse {
+    fn default() -> Self {
+        Self {
+            from_host: Default::default(),
+            result: 0,
+        }
+    }
+}
```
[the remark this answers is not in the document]
flaub: Sure, by the way my next PR makes all this obsolete, this was just scratch code that I didn't want to lose. I can drop in this PR now.

risc0/zkvm/src/host/client/prove/mod.rs
flaub: However, if we want to call verify within prove on the client side, we'd still need it. I have a new branch which might make this obsolete, but this is a good point regardless.

risc0/zkvm/src/host/client/prove/mod.rs
flaub: I actually didn't change the semantics here from before, but I'll fix this so that it acts like a 'truthy' value (as it now does with dev mode).

risc0/zkvm/src/host/client/exec.rs
flaub: This was just prototype code which will get dropped. I'll push a new PR as soon as this one lands. I can just drop this whole file for this PR if it makes the review easier.

flaub: Protobuf brings the need to have a compiler and in general adds a lot of dependencies. I'm starting to lean in the direction of using it, but I think I want to get a first version done that doesn't require it and then see where we're at.

risc0/zkvm/src/host/server/prove/mod.rs
flaub: Naming is hard :) Maybe `ProverServer`

SchmErik: I like ProverSever!

SchmErik (approved):
I think this is a great start and I appreciate all of the clean up! I see that there are some things still in flight and you removed syscall context from the client to limit the scope and I think this client/server format will be really helpful in thinking about what we should allow/disallow. I think the ProverServer name change would be good but I'm think we can merge this and continue

PR #809, opened 2023-08-18
Print VerificationError message instead of variant

rlukata:
This PR changes how Debug behaves on the VerificationError struct. Instead of printing the variant of the struct, it uses the Display format to print the custom message.

### DIFF
--- a/risc0/zkp/src/verify/mod.rs
+++ b/risc0/zkp/src/verify/mod.rs
@@ -46,7 +46,7 @@ pub enum VerificationError {
 
 impl fmt::Debug for VerificationError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        Display::fmt(&self, f)
+        fmt::Display::fmt(&self, f)
     }
 }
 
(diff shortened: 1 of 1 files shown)


risc0/zkp/src/verify/mod.rs
```diff
@@ -44,6 +44,12 @@ pub enum VerificationError {
     InvalidHashSuite,
 }
 
+impl fmt::Debug for VerificationError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        Display::fmt(&self, f)
```
flaub, line 49:
```suggestion
        fmt::Display::fmt(&self, f)
```

risc0/zkp/src/verify/mod.rs
```diff
@@ -19,7 +19,11 @@ mod merkle;
 mod read_iop;
 
 use alloc::{vec, vec::Vec};
-use core::{cell::RefCell, fmt, iter::zip};
+use core::{
+    cell::RefCell,
+    fmt::{self, Display},
+    iter::zip,
+};
```
flaub, line 26:
```suggestion
use core::{cell::RefCell, fmt, iter::zip};
```

mothran (approved):

PR #810, opened 2023-08-23
Refactor Bonsai as workspace and use Foundry as external dependency

nategraf:
Currently on `main` the Foundry tools, including `forge` are declared as build dependencies in `bonsai/ethereum/Cargo.toml` and so Foundry is built as part of building and testing the `risc0` repo.
This turns out to be problematic, with the largest issue being dependency conflicts that keep breaking builds on different developer machines.
Foundry CLI is not meant to be imported as a Rust dependency, and it is becoming clear the that it should not be used as such.

This PR first refactors the build for `bonsai/ethereum` contracts to call `forge` as a subprocess rather than building and running it in-process.
If this is done alone, a consequence is that building `risc0` will require installing the Foundry tools, even if the developer has no interest in Ethereum integration or Bonsai.
In order to address this, the `bonsai` directory is refactored as its own workspace.


SchmErik (commented):
not sure what's going on exactly but disabling the test seems to be causing CI to fail

.github/workflows/bonsai.yml
```diff
@@ -12,8 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  # Install foundry and use it to lint and test parts of the codebase that use Solidity.
-  foundry:
+  test:
```
flaub, line 16: I wonder if this rename is the culprit for stalled jobs

flaub (approved):

PR #812, opened 2023-08-24
Add entropy to memory state to protect privacy

nategraf:
A Merkle root of the memory state is included in the `ReceiptMetadata`, and so is made public to the verifier. Without adding entropy, this leaks a commitment to the end memory state that may reveal information about the program state including inputs.

This PR adds 128 bits of entropy, provided by the host, to the memory state on guest initialization to prevent this privacy leakage.


nategraf:
PR is marked as draft pending a rebase on #810

nategraf:
In https://github.com/risc0/risc0/pull/812/commits/03399774d3104d7e3de2ad58c8fba7b5ec43c191, I did some refactoring to how `SliceIo` works to make it a bit more flexible. The direct purpose of this is to all hooking of `SYS_RANDOM` for testing this feature. Indirectly, I am hoping to make `SliceIo` more flexible so that it can be used to replace all the reasonable use-cases for `Syscall`.

I did not finish this refactor, but I need to sign off a bit early today, so I am committing it to this branch as is until I can return to it on Monday.

flaub:
I'm actually heavily modifying SliceIo to make it work in a client/server context. At the moment I'd like for sys_random to remain separate because I have no plans of making random values come from the client.

nategraf:
@flaub, in [8841c13](https://github.com/risc0/risc0/pull/812/commits/8841c13636d9faadecf47a76c42e66b3d5028ac7) I made the `Executor::syscall_table` field have visibility `pub(crate)`to allow for overriding in the test. This is definitly the most direct path if it seems reasonable to you.

nategraf:
@jbruestle, if you have a moment to look at this (and approve if it looks good) that would be helpful. It's a very small change in terms of lines of code.

jbruestle (approved):

PR #813, opened 2023-08-24
Make Control Id Parameters to Constructor

jjtny1:
This PR makes the control_id fields parameters of the constructor. The fields on the contract are immutable so when the contract is deployed, the provided values will be hardcoded into the deployed bytecode.

Closes: https://github.com/risc0/bonsai-foundry-template/issues/46


Cargo.toml
```diff
@@ -63,3 +63,15 @@ lto = true
 
 [profile.release.build-override]
 opt-level = 3
+
+[patch."https://github.com/gakonst/ethers-rs"]
```
jjtny1, line 67: The build is breaking because foundry-cli uses the latest ethers (v2.0.9) which has a dependency conflict with cargo-dependencies uses in cargo-risczero

Screen Shot 2023-08-24 at 7 49 16 PM

nategraf: I think with #810 now merged this should no longer be an issue. Sorry for introducing such a radioactive dependency to the repo 😅

bonsai/ethereum/contracts/BonsaiTest.sol
```diff
@@ -54,7 +54,9 @@ abstract contract BonsaiTest is Test, BonsaiCheats {
             bonsaiTestRelay = new BonsaiTestRelay(vm.envOr("TEST_BONSAI_TEST_RELAY_EXPECTED_CHAIN_ID", uint256(31337)));
             bonsaiRelay = new BonsaiRelayQueueWrapper(bonsaiTestRelay);
         } else {
-            IRiscZeroVerifier verifier = new RiscZeroGroth16Verifier();
+            uint256 control_id_0 = vm.envUint("CONTROL_ID_0");
+            uint256 control_id_1 = vm.envUint("CONTROL_ID_1");
```
nategraf, line 58: This factoring makes it impossible to run `forge test` without setting those environment variables. It also does not provide any instructions for how the user should locate these parameters. Are you still working on that? I think we should make sure these are set automatically to a reasonable default before merging this PR

jjtny1: Hey yes, im still working on that part. Wanted to see if anyone had a better way of doing thing

bonsai/ethereum/contracts/groth16/RiscZeroGroth16Verifier.sol
```diff
@@ -77,13 +77,25 @@ struct Seal {
     uint256[2] c;
 }
 
+ /// Control ID hash for the identity_p254 predicate decomposed as implemented by splitDigest.
+ /// New releases of RISC Zero's zkvm may require updating these values.
```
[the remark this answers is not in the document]
nategraf: Can we get these checked in to `risc0`? We have the control IDs for other recursion predicates there, but this predicate seems to me missing. IIRC, this predicate is referred to as `identity_p254`?

nategraf:
This looks good to me overall. Happy to approve when we have some instructions for developers (including ourselves) on how to get these values when they need to be updated.

bonsai/ethereum/contracts/groth16/RiscZeroGroth16Verifier.sol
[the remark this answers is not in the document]
capossele: #831 should address that

capossele:
#831 merged the ability to get those values by running: `cargo run --bin bonsai-ethereum-contracts -F control-id`. Maybe we can include these instruction on this PR?

capossele (approved):

PR #814, opened 2023-08-25
zkvm: Protobuf based client/server IPC

flaub:
This introduces an API based on `protobuf` messages passed over a TCP connection. The server is launched by the client as a sub-process. The child process intended to be `r0vm` or other compatible binary that executes for the duration of a request and then terminates. In this scenario, the client is the one that binds to the TCP socket on port 0 to allocate a port. This port is passed to `r0vm --port <PORT>` and the 'server' then connects back to the listener. This ensures that ports are unique for each caller in a race-free manner.

In this PR, the client/server messages are defined, while the core data types (receipts, segments, etc) are specified as bytes. This means there's an inner encoding that should be largely opaque to clients. There are instances where clients will want to deserialize these bytes (like a `Receipt`). Currently this uses `bincode` but the intention is to move to `protobuf` to allow for a better versioning story. Versioning will be fleshed out in a future PR. 

Therefore there will be two levels of protobufs, one for the API, and another for the core data types. This design follows best practices: https://protobuf.dev/programming-guides/api/#use-different-messages

TODO:
* more tests especially around I/O
* pluggable implementation for lift & join


flaub:
This isn't done yet but I pushed as a preview. I went ahead and used protobuf for the message definitions, this will set us up well for versioning in the future. At this time messages are still using the term IPC, but I'd like to come up with a better name. Comments are requested.

risc0/zkvm/src/host/ipc/server.rs
```diff
@@ -0,0 +1,280 @@
...
+        let proxy = SliceIoProxy {
+            channel: channel.clone(),
+        };
+        for (name, _) in request.slice_ios.iter() {
+            env_builder.slice_io(&name, proxy.clone());
+        }
+        let env = env_builder.build()?;
+
+        let mut exec = Executor::new(env, image)?;
+        let session = exec.run_with_callback(|segment| {
```
mothran, line 226: This is a can-of-worms note so consider it whackadoodle:

Would this be a place that it would make sense to make the IPC server async with something like tokio? In other places I have tried to keep everything in the executor callback inside a tokio Task so the main exec loop could continue unblocked by IO (up to a certain limit).

Maybe IPC IO bandwidth is high enough that this wont be an issue in average cases. I would recommend we do some stress testing with zeth because it's segments are generally the largest I encountered so far.

flaub: Definitely can-of-worms, but something I've been thinking about too. Want to gather more data, zeth is a good call for that.

Ultimately I think we'll want the server to write to disk as fast as it can in the background with multiple threads (just like Bonsai is doing), and send notifications to clients as things get flushed. My rough impression is that in this case async won't buy us much since I think most of the time we're either going to be computing or waiting for disk I/O and I don't know if switching between threads will be that much overhead (as opposed to something like green threads with a good async runtime).

mothran: agreed. I think well just have to do some data collection to work out what model gets us the best disk IO saturation.

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -0,0 +1,240 @@
...
+    uint32 paused = 2;
+    google.protobuf.Empty system_split = 3;
+    google.protobuf.Empty session_limit = 4;
+  }
+}
+
+message Asset {
+  oneof kind {
+    bytes inline = 1;
+    string path = 2;
```
nategraf, line 144: Is this `path` intended to be created and managed by the server, or is the client intended to be able to create an asset as a file (or S3 blob?) and then hand the server a path which it will read? If these paths are only created by the server, maybe we should treat them as opaque identifiers from the API point of view.

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -0,0 +1,240 @@
...
+}
+
+message AssetRequest {
+  oneof kind {
+    google.protobuf.Empty inline = 1;
+    string path = 2;
+  }
+}
+
+message ClientCallback {
```
nategraf, line 155: From reading this my assumption is that this is the message sent from the server to client each time a callback triggering event occurs. Is that accurate? If so, when would the `oneof` be an error?

nategraf: From reading further my assumption is that error would be returned as a final message in the stream if an error occurs. Maybe it would make sense to factor this as a final message that is either "ok" or "error"?

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -0,0 +1,240 @@
...
+    ELF = 2;
+  }
+
+  Kind kind = 1;
+  Asset asset = 2;
+}
+
+message ProverOpts {
+  string hashfn = 1;
+  bool lift = 2;
```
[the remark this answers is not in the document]
flaub: I'll drop this from `ProverOpts` because I also realize that we have a different receipt type for each case anyways.

risc0/zkvm/src/host/protos/api.proto
flaub: The `AssetRequest` is intended for clients to fill in the base path where they'd like assets to go. In the case of EFS, we have a mount that the client knows about and we're asking the server to ensure that assets show up there.

risc0/zkvm/src/host/protos/api.proto
flaub: The issue is, I need to know a priori which type to decode from the stream. I suppose we could do:
```
message ClientResult {
  oneof kind {
    ClientCallback ok = 1;
    GenericError error = 2;
  }
}
```

risc0/zkvm/src/host/protos/api.proto
nategraf: Ah ok that makes sense. I mentioned it because having the client direct the server to write a file to an arbitrary path feels dangerous if they are not on the same machine with the same privileges, but my understanding is that this revision of the API is only intended to be used in a single-machine context.

risc0/zkvm/src/host/protos/api.proto
nategraf: To me that has slightly more clear semantics, but looking at this again it seems simple enough that the steam will contain zero or more `ClientCallback` messages and that one of them may be an error if something went wrong.

risc0/zkvm/src/host/protos/api.proto
flaub: I've been concerned about the security ramifications too. In this case it's fine, but we have to be careful with it in other scenarios. Alternatively, we could make this a `--root_dir` argument to `r0vm` or something rather than allowing clients to set the base path in-band. This would also nicely restrict it to the sub-process case where permissions are reasonable since they inherit from the parent.

flaub:
@nategraf @mothran If there aren't anymore changes needed, I'd like to get this to land

mothran (approved):

nategraf (approved):

Issue #815, opened 2023-08-25
In 0.17 and some environments build fails with `error: toolchain 'risc0' is not installed`


choongng:
In certain configurations, at least including using the VS Code `rust:0-1-bullseye` image as a dev container, the risc0 build fails. We have a repro internally, recording here for visibility.

Exact error:
```
   Compiling ui v0.2.0 (https://github.com/foundry-rs/foundry.git?rev=8e365beee278975720ecd3eb529b5dd6d17cac3b#8e365bee)
   Compiling risc0-zkvm-methods v0.17.0 (/workspaces/risc0/risc0/risc0/zkvm/methods)
   Compiling risc0-zkvm-receipts v0.17.0 (/workspaces/risc0/risc0/risc0/zkvm/receipts)
risc0-zkvm-methods-guest: Starting build for riscv32im-risc0-zkvm-elfipts, risc0-zkvm-methods(build)                                                                                                                             
risc0-zkvm-methods-guest: error: toolchain 'risc0' is not installed
error: failed to run custom build command for `risc0-zkvm-methods v0.17.0 (/workspaces/risc0/risc0/risc0/zkvm/methods)`

Caused by:
  process didn't exit successfully: `/workspaces/risc0/risc0/target/debug/build/risc0-zkvm-methods-3f495822abaa63c5/build-script-build` (exit status: 1)
  --- stdout
  Building guest package risc0-zkvm-methods.risc0-zkvm-methods-guest
  cargo:rerun-if-env-changed=RISC0_SKIP_BUILD
  Building guest package: cargo +risc0 build --release --target riscv32im-risc0-zkvm-elf --manifest-path /workspaces/risc0/risc0/risc0/zkvm/methods/guest/Cargo.toml --target-dir /workspaces/risc0/risc0/target/riscv-guest
warning: build failed, waiting for other jobs to finish...
```

Some information about the environment:
```
vscode ➜ /workspaces/risc0/risc0 (main) $ cargo +risc0 build
error: 'cargo' is not installed for the toolchain 'risc0'
note: this is a custom toolchain, which cannot use `rustup component add`
help: if you built this toolchain from source, and used `rustup toolchain link`, then you may be able to build the component with `x.py`
vscode ➜ /workspaces/risc0/risc0 (main) $ which cargo
/usr/local/cargo/bin/cargo
vscode ➜ /workspaces/risc0/risc0 (main) $ env | grep PATH
[...]

PR #817, opened 2023-08-26
Support for stable host toolchain

flaub:
* Switch to `stable` channel in `rust-toolchain.toml`
* Adjust `risc0-build` to support any host-side rustc version.
* Download LLVM source for build-toolchain
* `risc0-build` uses env vars to make debugging toolchains easier
* Add `--version` to `risczero install`


flaub:
I think I stumbled on a possible fix for #815.

SchmErik (approved):

PR #823, opened 2023-08-29
Re-enable the Governance Example test with zkVM Integration

nategraf:
This PR re-enables the governance example test with zkVM integration. The test was disabled because it was taking too long and blocking CI. See https://github.com/risc0/risc0/issues/783 for more details.

Additionally this PR removes the `test_happy_path` test in `bonsai/ethereum/src/lib.rs` as it was getting stuck in CI and was not providing any readily apparent value.

Fixes: https://github.com/risc0/risc0/issues/783


nategraf:
`bonsai` workflow is no longer the slowest workflow. It takes ~30 minutes, or which 14 minutes come from the re-introduced test. If this seems acceptable, then this PR can be merged.

capossele (approved):

PR #824, opened 2023-08-30
Fault proof part 1

SchmErik:
This PR consists of part 1 of what appears in https://github.com/risc0/risc0/pull/736 This PR implements a fault checker but does not compute the post image ID. I will follow up with a PR that implements the computation of the post image ID within the fault checker guest.


risc0/zkvm/Cargo.toml
```diff
@@ -143,3 +142,6 @@ std = [
   "risc0-zkp/std",
   "serde/std",
 ]
+
+# this is a temporary feature that will be removed after proof of fault becomes more stable
+enable-fault-proof = []
```
flaub, line 147:
```suggestion
fault-proof = []
```

risc0/zkvm/src/lib.rs
```diff
@@ -18,17 +18,23 @@
...
+pub use fault_ids::{FAULT_CHECKER_ELF, FAULT_CHECKER_ID};
 
 pub mod guest;
 #[cfg(not(target_os = "zkvm"))]
 mod host;
 pub mod serde;
 pub mod sha;
 
+#[cfg(feature = "std")]
+mod fault_monitor;
```
flaub, line 31: We should avoid using "std" as a way to enable features I think.

SchmErik: I could change this to `feature = "fault-proof"`.

Backstory: the reason this exists is because we have a no-std test and without this it broke the guest-methods compilation test for no-std. If I exposed this without this cfg, it broke the test. the fault_monitor mod depends on rrs-lib which depends on std :(

The fault monitor struct is used by both fault-checker and the host. So we only really need to expose this for the fault checker. Maybe having fault-proof feature doesn't need to be temporary but we could use it to toggle this? Exposing this in a normal guest context might not be very useful even if you have std enabled.

risc0/zkvm/src/fault_ids.rs
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Warning: this code is generated by cargo xtask and should not be modified by
+// hand.
+
+/// The image ID of the fault checker guest program
+#[rustfmt::skip]
+pub const FAULT_CHECKER_ID: [u32; 8] = [73690331, 82127537, 5319644, 3512993428, 3332701599, 1824104521, 2226309199, 2308040824];
```
nategraf: Was this produced with the reproducible build system? Probably in part 2, but we should have some mechanism of CI to ensure the fault checker ELF and ID are built reproducibly and that they do not get stale.

risc0/zkvm/src/fault_monitor.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn address_is_within_range(&self, addr: u32) -> bool {
+        Self::LOWEST <= addr && addr < Self::HIGHEST
+    }
+}
+
+/// This implements a way for the rrs crate to read and write memory regions. If
+/// the execution gets to this point, the address has been checked to ensure
+/// that it's aligned. The only thing the fault checker needs to do is to ensure
+/// that the read and write addresses are within range. The reads and writes do
+/// not need to fetch/write to the memory map. One exception is the PC. The pc
```
nategraf, line 53: I was confused about this when I first read it.
```suggestion
/// not need to fetch/write to the memory map. One exception is the next instruction. The instruction at pc
```

risc0/zkvm/src/host/receipt.rs
```diff
@@ -206,14 +214,29 @@ impl SegmentReceipts {
         final_receipt.verify_with_context(ctx)?;
         let metadata = final_receipt.get_metadata()?;
         log::debug!("final: {metadata:#?}");
-        if prev_image_id != metadata.pre.digest() {
+        // If there's more than one segment, the last segment could be the fault checker
```
nategraf, line 217: This code seems to work as intended only if the fault checker takes exactly one segment. Does it fit within our minimum po2?

risc0/zkvm/src/host/receipt.rs
```diff
@@ -206,14 +214,29 @@ impl SegmentReceipts {
...
+        if prev_image_id != metadata.pre.digest()
+            && !(cfg!(feature = "enable-fault-proof")
+                && receipts.len() > 0
+                && metadata.pre.digest() == FAULT_CHECKER_ID.into())
```
nategraf, line 221: It took me a minute to parse this correctly. Breaking out the complicated inner logic would be helpful.

Also, if `receipts.len() == 0`, then `prev_image_id` will be equal to `metadata.pre.digest()`, so we don't need to check that condition I don't think.
```suggestion
        let final_receipt_is_fault_checker = cfg!(feature = "enable-fault-proof")
                && metadata.pre.digest() == FAULT_CHECKER_ID.into();
        if prev_image_id != metadata.pre.digest()
            && !final_receipt_is_fault_checker
```

risc0/zkvm/src/host/receipt.rs
```diff
@@ -206,14 +214,29 @@ impl SegmentReceipts {
...
 
+        // For receipts indicating proof of fault, the guest code should post the
+        // post-image ID of the original guest code to prove that the fault checker
+        // tried to execute the next instruction from the same state of the machine.
+        if metadata.pre.digest() == FAULT_CHECKER_ID.into() {
+            let digest: Digest = from_slice(&journal.clone()).unwrap();
+            if digest != prev_image_id {
+                return Err(VerificationError::FaultStateMismatch);
+            }
+        }
```
nategraf, line 238: I imagine this code should not run if the fault checker is not enabled. Also, there is a weird condition there is only one receipt. I imagine this is actually where the number of receipts should be checked.

nategraf, line 238: What happens when the fault checker exits with a state that is not `Halted(0)`?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -215,7 +217,32 @@ impl<'a> Executor<'a> {
...
+    where
+        F: FnMut(Segment) -> Result<Box<dyn SegmentRef>>,
+    {
+        let mut guest_session = self.run_guest_only_with_callback(callback)?;
+        match guest_session.exit_code {
+            ExitCode::Fault => {
+                let fault_checker_session = self.run_fault_checker()?;
+                for segment in fault_checker_session.segments {
+                    guest_session.segments.push(segment);
+                }
```
nategraf, line 230: This is inconsistent with the code above that assumes there is only one fault checker segment.

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -215,7 +217,32 @@ impl<'a> Executor<'a> {
...
+                guest_session.journal = fault_checker_session.journal;
```
nategraf, line 231: One thing I find uncomfortable about this factoring, of tacking on the fault checker segment and setting the journal, is that it it's clear to me what stops a verifier fr verifying a `Receipt`, getting `Ok(())` back, and then accessing the journal believing the just verified a successful execution. Is there a mechanism to prevent this mistake?

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -665,6 +665,17 @@ fn session_limit() {
 
 #[test]
 fn memory_access() {
+    fn session_faulted(session: Result<Session>) -> bool {
+        if cfg!(feature = "enable-fault-proof") {
+            let session = session.unwrap();
+            let last = session.segments.last().unwrap().resolve().unwrap();
+            last.pre_image.compute_id() == FAULT_CHECKER_ID.into()
```
nategraf, line 672: There should definitely be an easier way to determine if a session exited with fault.

risc0/zkvm/src/lib.rs
SchmErik: I ended up creating a new feature called fault-proof and kept enable-fault proof. the `feature = "fault-proof"` will be used to expose the `fault_monitor` to the prover and `enable-fault-proof` will be used as a temporary feature to toggle the executor to run the fault checker.

nategraf (changes requested):
Most of the tests in `risc0/zkvm/src/host/server/exec/tests.rs are` no longer effective because the `run_test` function checks that the `Session` is `Ok` as a way to check if the `Session` ended without issue (e.g. panic or fault). On it's face, this is a very reasonable way to check for success. I think these return semantics would also result in a lot of broken user code. I think we should address this before merging this PR.

risc0/zkvm/src/host/receipt.rs
SchmErik: thanks!

risc0/zkvm/src/host/receipt.rs
SchmErik: I was going to take care of this later but I see that it's not much to implement. It currently fits within a single segment but I've changed the verifier logic to account for the possibility of multiple segments.

risc0/zkvm/src/host/receipt.rs
SchmErik: It should be an error. Thanks for catching this! Fixed!

risc0/zkvm/Cargo.toml
SchmErik: I ended up using both enable-fault-proof and fault-proof... See https://github.com/risc0/risc0/pull/824/commits/cc743477da3eaf08a167a023a553267c6a198871

bonsai/rest-api-mock/src/prover.rs
```diff
@@ -102,7 +102,9 @@ impl Prover {
                         anyhow::anyhow!("failed to build executor environment: {:?}", e)
                     })?;
                 let mut exec = Executor::new(env, mem_img)?;
-                let session = exec.run()?;
+                let session = exec.run().map_err(|e| {
+                    anyhow::anyhow!("Executor failed to generate a successful session: {:?}", e)
+                })?;
```
nategraf, line 107: Anyhow provides the [context function](https://docs.rs/anyhow/latest/anyhow/trait.Context.html) to add additional information like this to an error. It also preserves more metadata.
```suggestion
                let session = exec.run().context("Executor failed to generate a successful session)?;
```

risc0/zkvm/src/host/receipt.rs
nategraf, line 1: When we were discussing on Monday, we realized that if/when we rebase fault proofs onto user recursion, the changes to this file will be removed. Do you think we should remove these changes from this PR and keep the other parts, such as the guest and memory monitor code changes?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -100,6 +101,39 @@ impl OpCodeResult {
...
+}
+
+impl std::fmt::Display for ExecutorError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            ExecutorError::Error(e) => write!(f, "{e}"),
+            ExecutorError::Fault(s) => write!(
+                f,
+                "Faulted Session [Image ID: {}]",
+                s.segments[0].resolve().unwrap().pre_image.compute_id()
```
nategraf, line 129: We probably should not call `resolve` in this context since it could result in a network call that may fail. Since we are already in an error context in this code, we should be careful to avoid inducing another error.

nategraf (approved):
Approved! I also have a comment that it seems we will end up replacing some of this code in the near future when user recursion is available. I think with the feature flags, we can go ahead and merge as is, but removing the code from this PR that we expect not to be used later would be cleaner.

flaub:
I'm OK with either landing now and removing the verify changes when user recursion lands, or removing it now.

risc0/zkvm/src/host/server/exec/executor.rs
SchmErik: good catch!

risc0/zkvm/src/host/receipt.rs
SchmErik: I need to add the computation of the merkle root in the fault checker so I'll leave it in for ease of testing. I'm definitely OK with dropping all of this once we can use user recursion though!

PR #826, opened 2023-08-31
zkp: HAL cleanups and perf bump

flaub:
* Parallelize a top 10 hotspot in the prover. TODO: consider moving this hotspot to a HAL function that stays on-device.
* Rename `EvalCheck` to `CircuitHal` to be more appropriate for what it is now and in the future. The plan is to add more circuit-specific HALs to `CircuitHal`.
* Drop unused `CircuitStepHandler::calc_prefix_products`.


SchmErik (approved):
I don't have a lot of knowledge on the specific prover change but everything else looks good to me. I'll approve but maybe @jbruestle or @shkoo could take a look at that portion?

PR #838, opened 2023-09-01
crate validator: harden cargo risczero install

SchmErik:
This is a suggestion by Frank. This chagne make this run command more resilient against github's IP throttling.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -27,7 +27,9 @@ jobs:
 
       - run: cargo install --force --path risc0/cargo-risczero
 
-      - run: cargo risczero install
+      - run: cargo run --bin cargo-risczero -- risczero install
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - run: cargo build --release
         working-directory: tools/crates-validator/


.github/workflows/nightly.yml
```diff
@@ -27,7 +27,9 @@ jobs:
 
       - run: cargo install --force --path risc0/cargo-risczero
```
mothran, line 28: If we install the cargo-risczero tool here do we still need to run `cargo run` below? 

Using the installed version would prevent a rebuild, and we need the cargo-risczero tool installed for the crate validator.

SchmErik: good point. I reverted to this command

mothran (approved):

flaub:
It should also use the local version instead of using the last released version.

flaub (changes requested):
Use local vs last released

.github/workflows/nightly.yml
flaub, line 28: We should drop this

mothran: The validator uses the installed version: https://github.com/risc0/risc0/blob/main/tools/crates-validator/src/lib.rs#L211-L212

mothran: I do see what you mean, we could call `cargo run` inside the validator or expect the local `./target/` exists to better match up with the local vs installed. I am not sure the right path though.

flaub (approved):

.github/workflows/nightly.yml
flaub: Ah, I didn't realize we used it like this. Maybe we could pass an env var from CI so that the validator could pick it up.

PR #844, opened 2023-09-06
Fix full segment bug

### LINKED ISSUE #820: Failure to split a computation consisting in iterating calls to sha256
One can get the following error when running a long computation iterating calls to sha256:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Cannot expand, max po2 of 20 reached.', host/src/benches/iter_sha2.rs:82:89
```
This was observed on 0.17.0.

Steps to reproduce:
- Clone this branch: https://github.com/igarnier/zk-benchmarking/tree/iter-sha2-failure
- `cd risczero`
- `cargo b -r`
- `target/release/host --out dummy.csv iter-sha2`

The guest code is there: https://github.com/igarnier/zk-benchmarking/blob/iter-sha2-failure/risczero/methods/guest/src/bin/iter_sha2.rs

The error was triggered for `num_iter = 10_000` (this is an upper bound, didn't try to bisect).

capossele:
This PR fixes a bug occurring when calling iteratively sha2, ending up calling the `expand` function. The fix increases the const `SHA_CYCLES` from 72 to 73. 
It also adds a unit test for that case.

Fixes #820

### DIFF
--- a/risc0/cargo-risczero/src/commands/build_guest.rs
+++ b/risc0/cargo-risczero/src/commands/build_guest.rs
@@ -213,7 +213,7 @@ mod test {
         let tester = Tester::new("risc0/zkvm/methods/guest/Cargo.toml");
         tester.compare_image_id(
             "risc0_zkvm_methods_guest/multi_test",
-            "2a1f586685d5d7ff121cacdeab79432ae6e872bc2b8ff99335cfebe7ea3987c4",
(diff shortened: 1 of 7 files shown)


flaub:
This is interesting because I was planning to drop the expand function altogether since it was written before we had an executor which could compute ahead of time how big the prover matrix needs to be. So it feels like the executor is still miscalculating something.

flaub (approved):
Seems reasonable 👍

PR #848, opened 2023-09-08
Remove Bonsai SDK ImageIdExists error

mothran:
Now upload_img() returns a boolean for a user to check if the image exists within bonsai instead of a Err variant.

### DIFF
--- a/bonsai/ethereum-relay/src/downloader/proxy_callback_proof_processor.rs
+++ b/bonsai/ethereum-relay/src/downloader/proxy_callback_proof_processor.rs
@@ -17,7 +17,7 @@ use std::sync::Arc;
 use bonsai_ethereum_contracts::i_bonsai_relay::CallbackRequestFilter;
 use bonsai_sdk::{
     alpha::Client,
-    alpha_async::{create_session, put_input},
+    alpha_async::{create_session, upload_input},
 };
 use tokio::sync::Notify;
 use tracing::info;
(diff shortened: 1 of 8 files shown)


tzerrell (approved):
Interesting! I'm definitely happy to see "already exists" moved out of an error condition.

I find the portion that uses the `ImageExistsOpt` approach clearer than a `Result<bool,...>` (not only does it clearly label the topic, it's not intuitive whether `true` should mean "true the image already existed" or "true I did an upload [because the image didn't exist]") -- are there are technical reasons a fancier type doesn't make sense on `upload_img_file`?

flaub (approved):

mothran:
> Interesting! I'm definitely happy to see "already exists" moved out of an error condition.
> 
> I find the portion that uses the `ImageExistsOpt` approach clearer than a `Result<bool,...>` (not only does it clearly label the topic, it's not intuitive whether `true` should mean "true the image already existed" or "true I did an upload [because the image didn't exist]") -- are there are technical reasons a fancier type doesn't make sense on `upload_img_file`?

Thats a very good point, I only use `ImageExistsOpt` internally (returns from `get_image_upload_url` which is non-pub) to ferry data from the "common base" function to the two different call points above them.

Because the `ImgUploadRes` is not needed on the public facing side of that API (I only really need to get the url out to continue the PUT http call) I am not sure I know a good clean way to continue forwarding that return up from the pub methods.

Any ideas?

tzerrell:
> Any ideas?

Not immediately :/ -- it probably makes sense to return to this in followup work, I'm opening an issue https://github.com/risc0/risc0/issues/854

PR #849, opened 2023-09-08
Fix broken Waldo links

tzerrell:
Add missing Waldo links.

This won't actually fix the receipt.journal link until 0.18 is released, as it's pointing to the current location of receipt.journal, and 0.17 (the latest release) had it in a different location. I think that's ok for now, given the link was broken anyway and 0.18 shouldn't be too far away, but I'm open to feedback that I should set this up for 0.17 now and switch to 0.18 in a separate PR.


examples/waldo/README.md
```diff
@@ -48,6 +48,8 @@ cargo run --release --bin verify -- -i waldo.webp -r receipt.bin
 Running the verifier proves that the contents of [receipt.journal] were indeed constructed by the binary file associated with the expected [ImageID].
 
 [install Rust]: https://doc.rust-lang.org/cargo/getting-started/installation.html
+[receipt.journal]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Receipt.html#structfield.journal
+[ImageID]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.MemoryImage.html#method.compute_id
```
pdg744, line 52: Maybe this link is better? Not strongly opinionated
```suggestion
[ImageID]: https://dev.risczero.com/terminology#image-id
```

pdg744 (approved):
Thank you!

examples/waldo/README.md
```diff
@@ -48,6 +48,8 @@ cargo run --release --bin verify -- -i waldo.webp -r receipt.bin
 Running the verifier proves that the contents of [receipt.journal] were indeed constructed by the binary file associated with the expected [ImageID].
 
 [install Rust]: https://doc.rust-lang.org/cargo/getting-started/installation.html
+[receipt.journal]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Receipt.html#structfield.journal
```
pdg744, line 51: iiuc we're aiming to use tagged releases rather than `latest/` 
```suggestion
[receipt.journal]: https://docs.rs/risc0-zkvm/v0.18.0/risc0_zkvm/struct.Receipt.html#structfield.journal
```

Cardosaum:
To avoid broken links in the future, maybe we could rewrite in the following way?

> Running the verifier proves that the contents of [receipt.journal] were indeed constructed by the binary file associated with the expected [ImageID].

to

> Running the verifier proves that the contents of the receipt's [journal] were indeed constructed by the binary file associated with the expected [ImageID].
> [journal]: https://dev.risczero.com/terminology#journal

tzerrell:
> To avoid broken links in the future, maybe we could rewrite in the following way?

Good suggestion, I've revised accordingly.

PR #850, opened 2023-09-09
feat: adds paths filter to actions

### LINKED ISSUE #734: Consider using paths-filter github action
https://github.com/dorny/paths-filter

This can be used to filter which jobs run from which directory changes.

samtvlabs:
This closes #734 

Opening this as a draft PR to be sure this is going the right direction. Also we dont neccesarily need it on the bonsai action. We can filter it directly at the top. 

Havent tested the main.yaml and most of it runs on self hosted runners

### DIFF
--- a/.github/workflows/bonsai.yml
+++ b/.github/workflows/bonsai.yml
@@ -16,53 +16,53 @@ jobs:
   bonsai:
     runs-on: [self-hosted, prod, Linux, cpu]
     steps:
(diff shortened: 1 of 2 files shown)


flaub:
Thanks for this!

This is an interesting experiment and I think using the `paths-filter` action will speed up our CI. However, I think the right approach to this is more like: https://github.com/getsentry/sentry/blob/2ebe01feab863d89aa7564e6d243b6d80c230ddc/.github/workflows/backend.yml#L20

PR #857, opened 2023-09-11
Add no_std support for risc0-binfmt

### LINKED ISSUE #833: `risc0_binfmt` breaks the `no_std` support
The #725 PR introduced some refactoring to the crate system of `risc0_zkvm` crate. 
It also explicitly replaced `alloc` versions of data structures with the `std` ones:
https://github.com/risc0/risc0/blob/9b2d494689b54fbbe51962276ac557f88c8b0746/risc0/binfmt/src/elf.rs#L15 

Because these imports are not feature-gated and `risc0_zkvm` uses `risc0_binfmt` by default for verification:
https://github.com/risc0/risc0/blob/9b2d494689b54fbbe51962276ac557f88c8b0746/risc0/zkvm/Cargo.toml#L54

This essentially breaks earlier support `risc0_zkvm` for `no_std` environment.

I am happy to work on the fix PR, if possible.

flaub:
This should fix #833

### DIFF
--- a/risc0/binfmt/Cargo.toml
+++ b/risc0/binfmt/Cargo.toml
@@ -18,6 +18,10 @@ tracing-subscriber = "0.3"
 anyhow = { version = "1.0" }
 elf = "0.7"
 log = "0.4"
-risc0-zkp = { workspace = true, features = ["std"] }
+risc0-zkp = { workspace = true }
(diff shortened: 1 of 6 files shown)


tzerrell (approved):
LGTM!

PR #868, opened 2023-09-12
Update docs for 0.18

tzerrell:
Updates the docs.rs docs for the upcoming 0.18 release. This has three main parts:
* Updating the introduction to the `guest` module
* Porting the receipts documentation to live on the `Receipt` object, so it is not hidden after the module restructure.
* Porting the prover documentation to live on the `Prover` object, so it is not hidden after the module restructure.

I also revised during these ports, in particular expanding the Prover example code.


risc0/zkvm/src/host/receipt.rs
```diff
@@ -159,11 +147,11 @@ pub struct Receipt {
     pub journal: Vec<u8>,
 }
 
-/// An inner receipt can take the form of a collection of [SegmentReceipts] or a
+/// An inner receipt can take the form of a collection of [SegmentReceipt]s or a
 /// [SuccinctReceipt].
 #[derive(Debug, Deserialize, Serialize, PartialEq)]
 pub enum InnerReceipt {
-    /// The [SegmentReceipts].
+    /// The [SegmentReceipt]s.
```
flaub, line 154: This is wrong, it actually is `[SegmentReceipts]` which is distinct from `[SegmentReceipt]`

risc0/zkvm/src/host/receipt.rs
```diff
@@ -269,6 +257,9 @@ impl InnerReceipt {
     }
 
     /// Returns the [InnerReceipt::Flat] arm.
+    ///
+    /// Returns this arm as a slice of [SegmentReceipt]s, as the
+    /// `SegmentReceipts` type used to hold them is a private type.
```
flaub, line 262: I don't think we need to mention that `SegmentReceipts` is a private type, and honestly I think `SegmentReceipts` needs to be a public type.

flaub (approved):

PR #869, opened 2023-09-12
risc0-binfmt: use no_std for anyhow and elf

SchmErik:
This change uses `default-features = false` for `anyhow` and `elf` crates. Using these features resulted in several compiler warnings about error type conversions. To fix the errors, I've added pattern matching for Ok(...) and Err(...) mainly for u64 -> u32 conversions.


risc0/binfmt/src/elf.rs
```diff
@@ -32,7 +32,10 @@ impl Program {
     /// Initialize a RISC Zero Program from an appropriate ELF file
     pub fn load_elf(input: &[u8], max_mem: u32) -> Result<Program> {
         let mut image: BTreeMap<u32, u32> = BTreeMap::new();
-        let elf = ElfBytes::<LittleEndian>::minimal_parse(input)?;
+        let elf = match ElfBytes::<LittleEndian>::minimal_parse(input) {
+            Ok(elf) => elf,
+            Err(e) => bail!("Elf parse error: {e}"),
+        };
```
flaub, line 38:
```suggestion
        let elf = ElfBytes::<LittleEndian>::minimal_parse(input).ok_or(|err| anyhow!("Elf parse error: {e}")?;
```

risc0/binfmt/Cargo.toml
```diff
@@ -15,13 +15,13 @@ tracing = "0.1"
...
 log = "0.4"
 risc0-zkp = { workspace = true }
 risc0-zkvm-platform = { workspace = true }
-serde = { version = "1.0", default-features = false, features = ["derive"] }
+serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
 
 [features]
 default = ["std"]
-std = ["log/std", "risc0-zkp/std", "serde/std"]
+std = ["log/std", "risc0-zkp/std", "serde/std", "elf/std", "anyhow/std"]
```
flaub, line 27:
```suggestion
std = ["anyhow/std", "elf/std", "log/std", "risc0-zkp/std", "serde/std"]
```

nategraf (approved):

risc0/cargo-risczero/src/commands/build_guest.rs
nategraf, line 1: General comment: When do expect or care that this changes? IIUC, binfmt is not imported in any guest, and this code should be functionally the same, so it's unclear to me why this PR effects the guest image IDs.

It seems that as long as your local machine and CI agree, and we do not get any reports of inconsistent results from users, it is fine for every commit to change the guest image ID, but I'm not confident this is the bar.

flaub: It's because of the way cargo computes a fingerprint for all crates in a given workspace. Since the spec of one part of the workspace changed, the fingerprints will be different which causes the ELF binary to be changed.

PR #871, opened 2023-09-13
Rename DynProverImpl to ProverServer

flaub:


tzerrell (approved):
LGTM.

There's probably some room for more detailed docs on what this is, but that's an 0.19 problem IMO (certainly that's how I've been treating API client/server docs which are very much incomplete).

PR #877, opened 2023-09-14
updating references to 0.18

pdg744:
Ran a quick ctrl-f on the risc0 repo to see if there are remaining 0.17 references to update. Caught and updated a few. 

Wasn't sure about these ones, but it sounds like they have been discussed and are OK as is.

image
image

### DIFF
--- a/risc0/cargo-risczero/src/commands/build_guest.rs
+++ b/risc0/cargo-risczero/src/commands/build_guest.rs
@@ -98,7 +98,7 @@ impl BuildGuest {
         let rustflags_env = &[("RUSTFLAGS", rustflags.as_str())];
 
         let build = DockerFile::new()
-            .from_alias("build", "risczero/risc0-guest-builder:v0.17")
+            .from_alias("build", "risczero/risc0-guest-builder:v0.18")
             .workdir("/src")
             .copy(".", ".")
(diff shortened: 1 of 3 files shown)


risc0/cargo-risczero/src/commands/build_guest.rs
```diff
@@ -98,7 +98,7 @@ impl BuildGuest {
         let rustflags_env = &[("RUSTFLAGS", rustflags.as_str())];
 
         let build = DockerFile::new()
-            .from_alias("build", "risczero/risc0-guest-builder:v0.17")
+            .from_alias("build", "risczero/risc0-guest-builder:v0.18")
```
tzerrell, line 101: I think this is related to the first screenshot you posted and should not be updated. @flaub or @mothran would know better than me though.
```suggestion
            .from_alias("build", "risczero/risc0-guest-builder:v0.17")
```

tzerrell (commented):
I think one of these shouldn't be changed, but I'm tagging in others who would know better.

flaub:
These are related to a docker image that hasn't been updated and we will likely do somethign else in the future. There was a convo on slack about it and we've decided we don't need to do anything on this yet.

PR #878, opened 2023-09-14
load_elf: change maximum memory address to SYSTEM.start()

SchmErik:
load_elf takes a max_mem parameter before this change, MEM_SIZE was used as the
parameter. Any memory region about `SYSTEM.start()` is intended for use by the
zkVM and is not to be used by the ELF. Rather than have maximum memory as a
parameter, use a hard-coded constant, GUEST_MAX_MEM as the upper bound for
memory during elf load. This PR also restricts the loader to no load above GUEST_MAX_MEM

### DIFF
--- a/risc0/binfmt/src/elf.rs
+++ b/risc0/binfmt/src/elf.rs
@@ -18,6 +18,7 @@ use alloc::collections::BTreeMap;
 
 use anyhow::{anyhow, bail, Context, Result};
 use elf::{endian::LittleEndian, file::Class, ElfBytes};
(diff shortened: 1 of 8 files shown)


flaub:
I'd prefer to use a parameter, so that we can use this same code for newer circuit versions that have a different max memory bound.

SchmErik:
> I'd prefer to use a parameter, so that we can use this same code for newer circuit versions that have a different max memory bound.

Ah, that's a good point

flaub (approved):

PR #883, opened 2023-09-15
Ethereum-Relay: fix empty journal on dev-mode

capossele:
- Return the journal within a snark receipt when using the `bonsai-mock-rest-api`
- Add local e2e test of the relay on CI


mothran (approved):
Do you think this warrants a version bump on 0.18? We could bump the bonsai-ethereum-relay crate pointer in `bonsai-foundry-template` to a specific commit instead once this merges.

capossele:
Yeah I think just bumping  the bonsai-ethereum-relay crate pointer in `bonsai-foundry-template` should be enough for now.

PR #886, opened 2023-09-15
Executor: bail for very large cycle counts

SchmErik:
Some instructions within the zkVM have variable cycle counts that are configured by the guest code. A malicious host can alter parameters to ecalls such as SHA so that a single instruction to this ecall can exceed the segment cycle limit. If a single instruction takes more cycles than a single segment, there is no way to split the instruction among multiple segments. Return an error for this case to avoid infinite spits.


flaub:
Would we want to have a proof of fault here instead of just bailing on the host-side?

flaub (approved):

PR #890, opened 2023-09-19
Enable full stack trace on profiler

capossele:
- Improve the guest profiler to support full stack trace that can be used by pprof to show graph and flamegraph
- Add profiling example
- Add profiling section on the docs


examples/ecdsa/methods/guest/Cargo.toml
```diff
@@ -20,4 +20,13 @@ crypto-bigint = { git = "https://github.com/risc0/RustCrypto-crypto-bigint", tag
 [profile.release]
 # Empirically observed to result in the best performance for this binary (in particular, better than
 # "fat"). It's often a good idea to experiment with different optimization levels and options.
-lto = "thin"
+# lto = "thin"
+opt-level = 0
+debug = true
```
nategraf, line 25: As mentioned in the call, we generally want to profile optimized code, even if the results are not as easy to interpret.

risc0/zkvm/src/host/server/exec/profiler.rs
```diff
@@ -27,33 +27,100 @@
...
+    Pop,
+    PopPush,
+}
+
+/// Partially decodes the given instruction to determine if it is a functon
+/// call, or the return from a function call.
+///
+/// Uses the rules in section 2.5 as guidelines for operations on the return
+/// address stack to determine whether a given JAL(R) instruction is a call, a
+/// return, or neither.
```
nategraf, line 64:
```suggestion
/// Uses the rules in section 2.5 of the [RISC-V ISA manual] as guidelines for operations on the return
/// address stack to determine whether a given JAL(R) instruction is a call, a
/// return, or neither.
///
/// [RISC-V ISA manual]: https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf
```

nategraf (commented):
>  In the latter case, please use RISC0_BUILD_DEBUG=1 when building the guest binary.
Generally it's best to profile the optimized version of the code, do folks actually should not set this envvar by default.

One thing I'd love to see is an easier way to enable profiling. Currently the code in `r0vm` and in the `benchmark` host is ugly and annoying to add to your project. I'd love for us to essentially move that code into the executor such that when the `profiler` feature is enabled and a `RISC0_PPROF_OUT` envvar is set, it writes the profile to the specified location. wI think there are some questions in the client server architecture on whether this should be  part of the client or the server. We'd have to make a decision there.

flaub:
> > In the latter case, please use RISC0_BUILD_DEBUG=1 when building the guest binary.
> > Generally it's best to profile the optimized version of the code, do folks actually should not set this envvar by default.
> 
> One thing I'd love to see is an easier way to enable profiling. Currently the code in `r0vm` and in the `benchmark` host is ugly and annoying to add to your project. I'd love for us to essentially move that code into the executor such that when the `profiler` feature is enabled and a `RISC0_PPROF_OUT` envvar is set, it writes the profile to the specified location. wI think there are some questions in the client server architecture on whether this should be part of the client or the server. We'd have to make a decision there.

I love this idea!

website/docs/zkvm/developer-guide/profiling.md
```diff
@@ -28,14 +28,14 @@ risc0-zkvm = { version = "0.18", features = ["profiler"] }
 
 2. Initialize the profiler with your guest code.
 
-```rust
+```no_compile
```
flaub, line 31:
```suggestion
```rust no_compile
```

flaub (approved):

PR #897, opened 2023-09-20
Benchmark loop gh workflow

capossele:


tzerrell (approved):
Awesome, thank you!

risc0/zkvm/Cargo.toml
```diff
@@ -104,7 +105,7 @@ client = [
 cuda = ["prove", "risc0-circuit-rv32im/cuda", "risc0-zkp/cuda"]
 dual = []
 metal = ["prove", "risc0-circuit-rv32im/metal", "risc0-zkp/metal"]
-default = ["client", "prove"]
+default = ["client", "prove", "json"]
```
flaub, line 108: I think we can avoid a new feature flag since the `loop` is an example, which falls under `dev-dependencies`. I think it's fine for `dev-dependencies` to depend on `serde_json` without it being optional.

.github/workflows/bench_loop.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Download benchmark apple_m2_pro result
+        uses: actions/download-artifact@v2
+        with:
+          name: benchmark-result-apple_m2_pro
+          path: datasheet
+      - name: Download benchmark nvidia_rtx_a5000 result
+        uses: actions/download-artifact@v2
+        with:
+          name: benchmark-result-nvidia_rtx_a5000
+          path: datasheet
```
flaub, line 88: We should be able to combine all of these into a single action, the download action is designed to download multiple assets.

.github/workflows/bench_loop.yml
```diff
@@ -0,0 +1,98 @@
...
+        os: [macOS]
+        feature: [default]
+        device: [cpu]
+        include:
+          - os: Linux
+            feature: cuda
+            device: nvidia_rtx_a5000
+          - os: macOS
+            feature: metal
+            device: apple_m2_pro
```
flaub, line 37:
```suggestion
        include:
          - os: Linux
            feature: cuda
            device: nvidia_rtx_a5000
          - os: macOS
            feature: metal
            device: apple_m2_pro
          - os: macOS
            feature: default
            device: cpu
```

risc0/zkvm/examples/loop.rs
```diff
@@ -26,6 +26,15 @@ use risc0_zkvm_methods::{
...
 use tracing_subscriber::{prelude::*, EnvFilter};
 
+#[derive(serde::Serialize, Debug)]
+struct PerformanceData {
+    cycles: String,
+    duration: String,
+    ram: String,
+    seal: String,
+    speed: String,
+}
```
flaub, line 36: Ideally the json data would use numbers rather than strings so that we can use this data in other ways in the future. We could make all the human readable stuff happen at a later phase perhaps in client-side javascript.

flaub (changes requested):

flaub (approved):

tzerrell (commented):
This is looking great, the printed version turned out well!

I agree w/ the Slack comment that it would be nice to add continuations info on the post-1024k cycle entries. Sounds like you're already on that 🙂

ghpages/datasheet/index.html
```diff
@@ -0,0 +1,447 @@
...
+      reasons, the
+      cycle count is always rounded up in length to the next highest power of two; thus, a program that requires 33000
+      cycles
+      will have the same performance as a program that requires 63000 cycles, as both will round up to 65536 (64 k)
+      cycles.
+      <br></br>
+      Initializing a zkVM guest requires over 16 k cycles. Combined with this power of two rounding, this means that no
+      program can be executed in fewer than 32 k cycles.
+      <br></br>
+      To give context for these numbers we have included a table indicating the size of each of our example programs in
```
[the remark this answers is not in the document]
capossele: I'll create a new PR for these points

PR #898, opened 2023-09-20
CI: add no_std build test and fix risc0-zkvm build issue

SchmErik:
I looked into various solutions and building using this command seems to be the easiest
way to build `risc0-zkvm` with `no_std`. I used `x86_64-unknown-none` explicitly
because it seemed like a widely available target (as opposed to smaller arm targets). I
didn't chose `aarch64-apple-darwin` because seems to allows building the zkvm
with the tempfile dependency which might be caused by inclusion of some item
from std from this target.

This PR also moves zkvm's tempfile as a dev-dependency (this was causing the newly
added CI command to fail).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,6 +96,8 @@ jobs:
       - run: cargo check -F $FEATURE -p risc0-sys
       - run: cargo check -F $FEATURE -p risc0-zkp
       - run: cargo check -F $FEATURE -p risc0-zkvm
+      # build risc0-zkvm verifier for a no-std environment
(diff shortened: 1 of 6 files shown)


SchmErik:
Looks like this works really well: https://github.com/mystor/cargo-no-std-check

However, this requires a nightly compiler... :(

SchmErik:
> Looks like this works really well: https://github.com/mystor/cargo-no-std-check
> 
> However, this requires a nightly compiler... :(

I cloned this and removed the nightly compiler requirement and it still works!

SchmErik:
superseded by #907

PR #900, opened 2023-09-20
cargo-risczero test with static library runtime

jjtny1:


bonsai/examples/governance/Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = ["methods", "relay"]
 [workspace.dependencies]
 risc0-build = { path = "../../../risc0/build" }
 risc0-zkvm = { path = "../../../risc0/zkvm", default-features = false }
+risc0-zkvm-platform = { path = "../../../risc0/zkvm/platform", default-features = false }
```
flaub, line 7: Why do we need to add this dep everywhere?

PR #907, opened 2023-09-25
CI: check that the verifier builds in no_std

SchmErik:
This uses a fork on the `cargo no-std-check` utility that does not require the use of a nightly compiler.


SchmErik:
We briefly talked about this needing a nightly compiler but I'm still getting the tool to run and get good results by using a stable compiler. My guess is that nightly is only required for building the `stdlib` parts of `sysroot` and this utility kicks out `stdlib` from the `sysroot`. If you prefer forking into the risc0 repo itself, feel free to fork https://github.com/SchmErik/cargo-no-std-check and I can update the url in the `main.yml`! For now, I forked it on my account

PR #908, opened 2023-09-26
Prepare for client-only zkvm usage

flaub:
This doesn't yet change any default feature flags, so should still be fairly minimal. 

The planned set of PRs include:
* add release process to build r0vm for all major targets
  * use cargo zigbuild on linux to build with a specific glibc for max compat with older linux distros
  * use something like https://github.com/cargo-bins/release-pr to automatically build/publish binaries on a per-release basis
* update `cargo risczero install` or adjust docs to have users do `cargo binstall cargo-risczero` (depending on whether we package `r0vm` as standalone or combined with `cargo-risczero`)


risc0/zkvm/src/host/receipt.rs
```diff
@@ -376,9 +376,9 @@ impl SegmentReceipt {
         ReceiptMetadata::decode_from_io(layout::OutBuffer(elems))
     }
 
-    /// Return the seal for this receipt, as a slice of bytes.
-    pub fn get_seal_bytes(&self) -> &[u8] {
-        bytemuck::cast_slice(&self.seal)
+    /// Return the seal for this receipt, as a vector of bytes.
+    pub fn get_seal_bytes(&self) -> Vec<u8> {
+        self.seal.iter().flat_map(|x| x.to_le_bytes()).collect()
```
flaub, line 381: Possibly breaking change (on big-endian systems)

risc0/zkvm/src/host/client/env.rs
```diff
@@ -48,8 +48,8 @@ pub type TraceCallback<'a> = dyn FnMut(TraceEvent) -> Result<()> + 'a;
 #[derive(Clone, Default)]
 pub struct ExecutorEnv<'a> {
     pub(crate) env_vars: HashMap<String, String>,
-    pub(crate) segment_limit_po2: Option<usize>,
-    pub(crate) session_limit: Option<usize>,
+    pub(crate) segment_limit_po2: Option<u32>,
+    pub(crate) session_limit: Option<u64>,
```
flaub, line 52: Possibly breaking change

risc0/zkvm/Cargo.toml
```diff
@@ -104,7 +105,7 @@ client = [
 cuda = ["prove", "risc0-circuit-rv32im/cuda", "risc0-zkp/cuda"]
 dual = []
 metal = ["prove", "risc0-circuit-rv32im/metal", "risc0-zkp/metal"]
-default = ["client", "prove"]
+default = ["client"]
```
flaub, line 108: Breaking change

flaub:
@mothran Once this PR lands, I think we can begin to test how to get Bonsai to use the client API ahead of (or in parallel to) the other planned PRs to get the binary distributed for end users.

risc0/zkvm/src/host/api/mod.rs
```diff
@@ -319,28 +289,45 @@ pub enum AssetRequest {
     Path(PathBuf),
 }
 
+/// Provides information about the result of execution.
+pub struct SessionInfo {
+    /// The number of segments.
+    pub segments: u32,
+
+    /// The data publicly committed by the guest program.
+    pub journal: Bytes,
```
nategraf, line 298: Maybe this should be `Option<Bytes>` since an exit code other than `Paused` or `Halted` will not have a journal. Empty bytes also seems reasonable.

risc0/zkvm/src/host/api/tests.rs
```diff
@@ -110,15 +114,25 @@ fn with_server<T, F: FnOnce() -> Result<T>>(addr: SocketAddr, f: F) -> Result<T>
...
 
 #[test]
-fn prove_basic() {
+fn execute_image() {
+    let input = to_vec(&MultiTestSpec::DoNothing).unwrap();
+    let env = ExecutorEnv::builder().add_input(&input).build().unwrap();
+    let program = Program::load_elf(&MULTI_TEST_ELF, GUEST_MAX_MEM as u32).unwrap();
+    let image = MemoryImage::new(&program, PAGE_SIZE as u32).unwrap();
+    let binary = image.into();
+    TestClient::new().execute(env, binary).unwrap();
```
nategraf, line 131: In CI, do we want to make sure these tests all run both with in-process execution and with sub-process execution? (Same question for prover tests).

risc0/zkvm/src/host/api/mod.rs
```diff
@@ -319,28 +289,45 @@ pub enum AssetRequest {
...
+
+    /// The [ExitCode] of the session.
+    pub exit_code: ExitCode,
```
nategraf, line 301: Instead of journal and exit_code, it'd be nice to have this be `ReceiptMetadata`, which essentially represents the set of public claims about an execution. For composition, being able to get the `ReceiptMetadata` for an execution is very useful because it allows starting execution and proving for one program as soon as all of it's dependencies have been executed, rather than waiting for dependencies to be proven.

One downside to this would be pushing the `journal` one layer down, which is the main field a developer is likely interested in. One option would be to have these fields be top-level in `SessionInfo` and have a `get_metadata` method that constructs the `ReceiptMetadata` struct if needed.

risc0/zkvm/src/host/api/mod.rs
```diff
@@ -319,28 +289,45 @@ pub enum AssetRequest {
     Path(PathBuf),
 }
 
+/// Provides information about the result of execution.
+pub struct SessionInfo {
+    /// The number of segments.
+    pub segments: u32,
```
nategraf, line 295: Segments seems less useful if the cycle count or po2 for the segments is not known. I may be known by context, if the client requested a specific po2. It would be nice if enough information is included here to additional know (at least an upper bound on) the number of cycles this represents.

nategraf (approved):
I like this factoring quite a bit

risc0/zkvm/src/host/api/tests.rs
flaub: There are existing tests for the in-process which is actually more extensive. What I was shooting for in these tests is to get coverage for the client/server message passing specifically. Admittedly these tests are incomplete; there should be more to cover more cases.

I've been pondering how we could get a common collection of tests to be run in in-process or sub-process mode, but the best I can come up with is to use an integration test to ensure that `r0vm` gets built. These tests are a compromise, it doesn't actually spin up a separate process, it uses a separate thread instead. But it does use all the same messages.

risc0/zkvm/src/host/api/mod.rs
flaub: This is a really good point, I like the idea of having a `get_metadata`. I'll play around with this.

risc0/zkvm/src/host/api/client.rs
```diff
@@ -87,8 +93,8 @@ impl Client {
         }
 
         let receipt_bytes = asset.as_bytes()?;
-        let receipt = bincode::deserialize(&receipt_bytes)?;
-        Ok(receipt)
+        let receipt_pb = pb::core::Receipt::decode(receipt_bytes)?;
+        receipt_pb.try_into()
     }
 
     /// TODO
```
tzerrell, line 100: Not something that needs to be solved within this PR, but I want to note that this architecture is not yet fully documented and we'll need to finish its docs before the next release.

risc0/zkvm/src/host/api/mod.rs
tzerrell: There was a Slack discussion about how "empty journal" hashes to `SHA256([])` whereas "no journal" "hashes" to `0`. If we want to retain this distinction we probably need a way to represent it here (which `Option<Bytes>` makes sense for that to me).

tzerrell (approved):
Nice!

risc0/zkvm/src/host/api/client.rs
tzerrell: Tracking in #929

risc0/zkvm/src/host/api/mod.rs
flaub: Actually the hashing distinction was only really relevant to the way the circuit generates the 'output' digest.

I guess `Option<Bytes>` is slightly better for the API, but we should make that change in all the other APIs as well (this is just the client/server API, which is kind of a private API).

risc0/zkvm/src/host/api/client.rs
flaub: I actually think client/server is largely transparent to end users. We do need to describe the public bits and how to install the server, but I don't think we need to (or necessarily should) document the internal part since it's undergoing changes and won't be relevant (and in fact confusing) to the majority of users.

tzerrell: :+1: makes sense. I do think that anything that shows up on docs.rs shouldn't have docs that just say "TODO". (Minimally something like "Internal component")

flaub: If you see a TODO, could you mark it or link to it? I probably missed those

tzerrell: The line this comment is on is one of them. I think they were introduced in a different client/server PR -- I noticed them when reviewing by reading through the generated docs, not on the PR proper. Also lines 66, 100, 137, 177, and 215 of this file, and lines 178, 183, and 189 of https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/api/server.rs (as of commit `7a1581f160729b544025afaca525ee3a39394b6e`).

risc0/zkvm/src/host/api/mod.rs
flaub: I tried making this change but seems to be non-trivial so I'm going to kick this to a future PR.

risc0/zkvm/src/host/api/mod.rs
flaub: I like this idea, but it turns out that the executor doesn't yet have the smarts to construct a `ReceiptMetadata`. Also, I don't know how we can construct one without running the entire program since only the final segment is guaranteed to have outputs and the final output digest.

examples/zkevm-demo/src/main.rs
```diff
@@ -82,22 +83,17 @@ async fn main() {
...
         .build()
         .unwrap();
-    let mut exec = Executor::from_elf(env, EVM_ELF).unwrap();
-    let segment_dir = tempdir().unwrap();
-    let session = exec
-        .run_with_callback(|segment| {
-            Ok(Box::new(FileSegmentRef::new(
-                &segment,
-                &segment_dir.path(),
-            )?))
```
SchmErik, line 97: If we want this again in the future, maybe we could use executor env as a way to configure `run_with_callback` to use files. We certainly don't need to implement this right now though!

What's nice about this is that it's saving memory for large computations and it would be nice to give the user the ability to opt-into this during development if they run a large proof during development.

.github/workflows/main.yml
```diff
@@ -225,5 +242,5 @@ jobs:
         with:
           node-version: 18
       - run: cargo run --bin cargo-risczero -- risczero install
-      - run: cargo test -p risc0-build -F docker
-      - run: cargo test -F default -F profiler -F test-exact-cycles -p risc0-zkvm
+      - run: cargo test -F docker -p risc0-build
+      - run: cargo test -F docker -p risc0-zkvm -- "docker::"
```
SchmErik, line 246: Thanks for doing this!

examples/zkevm-demo/src/main.rs
flaub: I've been thinking a similar thing. It'd be nice to just specify the kind of standard handling in the ExecutorEnv. I think we should switch the default to use temporary files but we should allow the user to configure where the directory lives.

SchmErik (approved):

flaub:
### Merge Activity

* **Oct 1, 8:38 PM**: @flaub merged this pull request with [Graphite](https://app.graphite.dev/github/pr/risc0/risc0/908).

PR #909, opened 2023-09-26
adding dev mode feature flag to README

3lkn:
The purpose of this update is to provide a brief high-level description of the dev mode feature flag for reference. The companion website PR, which adds more detailed explanation to the FAQ, can be found here: https://github.com/risc0/website/pull/295


README.md
```diff
@@ -161,6 +161,7 @@ The following feature flags are present in one or more of the crates listed abov
 | Feature  | Target(s)         | Implies    | Description                                                                           | Crates                                      |
 | -------- | ----------------- | ---------- | ------------------------------------------------------------------------------------- | ------------------------------------------- |
 | cuda     |                   | prove, std | Turns on CUDA GPU acceleration for the prover. Requires CUDA toolkit to be installed. | risc0-circuit-rv32im, risc0-zkp, risc0-zkvm |
+| disable-dev-mode | all | | Disables dev mode so that the proving stage may not be skipped. For the verifier, enforces that dev mode receipts generated sans proving stage are invalid. | risc0-zkvm |
```
tzerrell, line 164: I feel like adding some detail about why you might want this would be helpful, although I'm also a bit worried that this entry is too long.
```suggestion
| disable-dev-mode | all | | Disables dev mode so that the proving stage may not be skipped. For the verifier, enforces that dev mode receipts generated sans proving stage are invalid. Used to prevent a misplaced `RISC0_DEV_MODE` from breaking security in production systems. | risc0-zkvm |
```

tzerrell (approved):
Thanks! Suggested a possible revision, although I'm not certain whether it's an improvement.

README.md
flaub: This is purely an internal flag used for testing. I think we should mark it as such and recommend that users ignore/don't use it.

pdg744, line 164: I suggest we use the term "fake receipts" 
```suggestion
| disable-dev-mode | all | | Disables dev mode so that the proving stage may not be skipped. For the verifier, enforces that [fake receipts](https://docs.rs/risc0-zkvm/0.18.0/risc0_zkvm/enum.InnerReceipt.html#variant.Fake) are invalid. Used to prevent a misplaced `RISC0_DEV_MODE` from breaking security in production systems. | risc0-zkvm |
```

flaub (approved):
Eventually a link to a "dev mode" website page would be nice I think to describe what it is, what it's for, and how to use it.

flaub:
Is this ready to land?

tzerrell:
> Is this ready to land?

I'm going to add @pdg744's suggestion and then yes

PR #910, opened 2023-09-26
Use PBKDF2 in the password checker example

nategraf:
This PR changes the password checker example to use PBKDF2 SHA-256 in the password checker, using the RustCrypto crates.

I had the thought that we actually can simply run PBKDF2 in the guest, and furthermore could actually prove a PBKDF2 derivation with reasonably secure iteration counts (e.g. 500,000) with Bonsai.
Now, this would be likely never be worth doing, but it's possible and showing how easily we can important and use existing crates, which would be huge pain to implement in another zk system, is a boon to this exmaple.

On the other hand, this PR brings in Cargo patching and the crypto accelerated forks.
We may wish to avoid adding that cogantive overhead.
This PR took me 15 minutes to put together and I am happy to throw it away. :)


tzerrell (commented):
Ooo, interesting... This seems like a good idea (it makes the example more "real" and as you say shows off the value of importing & using existing crates).

In thinking about whether we want this, I'm wondering whether we want password checker to be more of a cookbook or showcase (in the taxonomy of https://github.com/risc0/rfcs/pull/29) -- I think there could be an argument for either. As a showcase, I'm pretty heavily in favor of including these changes -- the benefits of showing a more complex workload are brought to the forefront, and if it's trickier for users to reproduce, that's not so big a deal with a showcase.

As a cookbook, I'm a bit less certain. It reduces the clarity of purpose of the example (although it kinda lacks clarity of purpose as-is) and not everyone will want every part of it. That said, for the specific thing we're demonstrating, it shows people techniques they will need to use to be following best practices. So I still think including this is good, but I'm a bit more split.

@3lkn do you have thoughts on where password checker would fall between showcase/cookbook and whether to include these changes in the cookbook case?

tzerrell (approved):
I like this and no one else has spoken up to object or request changes.

Issue #441, opened 2023-03-15
Catch WASM/no-std build issues in CI


justinFrevert:
https://github.com/risc0/risc0/pull/438 highlights that the current CI process does not catch breaking changes to building for WASM. The `wasm` crate may need more dependencies to catch the issue. 

One [idea](https://github.com/risc0/risc0/pull/438#issuecomment-1470444047) is to use ecosystem-specific crates that would be used in a similar environment, if ecosystem support is desired. Though, that depends on whether the Risc0 team is open to including those dependencies, and there may be more general crates Im not thinking of that could be included.

justinFrevert:
https://github.com/risc0/risc0/pull/911 resolves this

PR #911, opened 2023-09-27
Add minimal Substrate WASM runtime for compatibility check

justinFrevert:
This PR aims to ensure ongoing compatibility between [Substrate](https://substrate.io/) and RISC Zero through a test including both. This is useful because:
- It serves as a check on `risc0-zkvm`'s compatibility in some no-std, wasm environments
- It assures the RISC Zero verifier functions with Substrate blockchains

This change stemmed from a conversation with @SchmErik and myself in Discord.

Changes:
- Inclusion of tiny Substrate runtime
  - This example of a Substrate runtime has a minimal amount of Substrate stuff necessary to be considered a runtime from a crates standpoint(it does not actually function, and does not need to; it's only for compatibility testing)
  - This runtime depends on the local `risc0-zkvm`
- CI command which tries to `cargo check` the Substrate runtime.

Notes:
- On future maintenance: Substrate is commonly versioned by release-branch. This would need to be manually updated every quarter-1/2 year or so. Maybe there's some approach involving a crate which lives separately, but I'm happy to be the "bumper" on this for now
- Relates to https://github.com/risc0/risc0/issues/441


external/substrate/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+frame-support = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/substrate.git", branch = "polkadot-v1.0.0" }
+risc0-zkvm = { workspace = true, default-features = false }
+
+[build-dependencies]
+substrate-wasm-builder = { version = "5.0.0-dev", git = "https://github.com/paritytech/substrate.git", optional = true , branch = "polkadot-v1.0.0" }
```
SchmErik, line 15: Looks like this has been archived. You mentioned this repo on discord - https://github.com/paritytech/polkadot-sdk/tree/master/substrate/utils/wasm-builder

Could you use this one instead since is more frequently maintained?

justinFrevert: Ah, yeap; good catch. That's the new monorepo they just switched everything over to

Cargo.toml
```diff
@@ -18,7 +19,7 @@ members = [
   "risc0/zkvm/receipts",
   "xtask",
 ]
-exclude = ["tools/crates-validator"]
+exclude = ["external/substrate", "tools/crates-validator"]
```
flaub, line 22: I think we should drop this, I'd like for a normal `cargo test` to include this test.

external/substrate/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+version = "0.1.0"
+description = "Minimal Substrate runtime for testing compatibility"
+publish = false
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+frame-support = { version = "4.0.0-dev", default-features = false, git = "https://github.com/paritytech/polkadot-sdk/", branch = "release-polkadot-v1.1.0" }
+sp-runtime = { version = "24.0.0", default-features = false, git = "https://github.com/paritytech/polkadot-sdk/", branch = "release-polkadot-v1.1.0", features = ["serde"] }
```
flaub, line 12:
```suggestion
frame-support = { version = "4.0.0-dev", default-features = false }
sp-runtime = { version = "24.0.0", default-features = false, features = ["serde"] }
```

justinFrevert: Before the recent shift to the monrepo, Substrate's been using polkadot release branches for versioning for years. I need to confirm whether, with the move to the monorepo we can version according to what's being suggested here.

flaub: It's OK if we go with only git, but I don't think we need *both `git` and `version`.

SchmErik:
@justinFrevert looks like it's failing a CI step on `cargo sort --workspace --check`

flaub (approved):

PR #912, opened 2023-09-27
zkVM tests: add feature flag to build multi-test with docker environment

SchmErik:
This PR adds a test-exact-cycles feature for risc0-zkvm. This is used to build test guest binaries like multi-test using the docker environment in the `cargo test` command. The intention here is to run the tests using a reproducible ELF binary to eliminate test failures across different architectures that are caused by entropy in the ELF binaries resulting from rust build tools.

I've gated 4 tests using this feature flag. Initially, what I had in mind was to isolate these tests to a different CI test step but I realized that it was cumbersome to isolate the 4 tests as integration tests and prevent other tests from running and I would have to proliferate a bunch of `#[cfg(feature = ...)]` directives. My solution is to create a new feature flag that the CI can use and it will run all tests including the ones that rely on exact cycles counts. If users do not wish to run the tests that require the reproducible binary, they can simply leave out the `test-exact-cycles` feature flag in their `cargo test` invocation.

Most of the code changes in this PR involves moving code from the `cargo risczero build` command to the `risc0-build` crate.


risc0/zkvm/methods/build.rs
```diff
@@ -34,5 +34,10 @@ fn main() {
         ),
     ]);
 
-    embed_methods_with_options(map);
+    if cfg!(feature = "test-exact-cycles") {
+        std::env::set_current_dir("../../").unwrap();
```
flaub, line 38: This part worries me. This is probably fine for now, but in the future I think we need to be able to configure the root of the docker context.

.github/workflows/main.yml
```diff
@@ -218,4 +218,4 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-      - run: cargo test -p cargo-risczero -F docker
+      - run: cargo test -p risc0-build
```
flaub, line 221: This might be a mistake. I think we need to continue to run tests for `cargo-risczero`. You can just drop the `-F docker`. Also, `risc0-build` should already be getting tested by `cargo test` above which is workspace-wide.

risc0/build/src/docker.rs
```diff
@@ -0,0 +1,207 @@
...
+fn compute_image_id(elf_path: &Path) -> Result<String> {
+    let elf = fs::read(elf_path)?;
+    let program = Program::load_elf(&elf, GUEST_MAX_MEM as u32).context("unable to load elf")?;
+    let image =
+        MemoryImage::new(&program, PAGE_SIZE as u32).context("unable to create memory image")?;
+    Ok(image.compute_id().to_string())
+}
+
+// requires Docker to be installed
+#[cfg(feature = "repro")]
```
flaub, line 166:
```suggestion
#[cfg(feature = "docker")]
```

.github/workflows/main.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - run: cargo xtask install
       - run: cargo xtask gen-receipt
-      - run: cargo test -F $FEATURE -F profiler
+      - run: cargo test -F $FEATURE -F profiler -F test-exact-cycles
```
flaub, line 85: We won't be able to run tests with `-F test-exact-cycles` on all runners, just on the ones with the `docker` tag.

.github/workflows/main.yml
```diff
@@ -218,4 +218,5 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-      - run: cargo test -p cargo-risczero -F docker
+      - run: cargo test -p risc0-build
+      - run: cargo test -F default -F profiler -F test-exact-cycles
```
flaub, line 222: I'd probably limit this to `risc0-zkvm` to help with build times

SchmErik: thanks for catching this!

risc0/zkvm/methods/build.rs
SchmErik: 100% agree! Tracking issue here: https://github.com/risc0/risc0/issues/913

.github/workflows/main.yml
SchmErik: Thanks for catching this. So I moved everything from cargo-risczero to risc0-build and that's why I changed it. I decided to keep the -F docker in my most recent change. Do you think we should still run the test from cargo-risczero? If so, I'm totally fine moving the test from risc0-build to cargo-risczero

flaub (approved):

PR #915, opened 2023-09-28
adding directory readmes

pdg744:
This PR adds READMEs to the `contracts` directory and the `methods` directory. 
These are copy-pasted from `bonsai-foundry-template`.


tzerrell (commented):
Hmm, in the context of a functional (and non-tutorial) example, I'm not sure these READMEs explain things a reader would be curious about (e.g. the how to write guest code section feels quite out of place). I also don't think they're entirely accurate anymore (e.g. there are more than 2 contracts in the contracts directory).

What about switching to very minimal READMEs? E.g. "This directory contains the Solidity contract for deploying the governance application through the Ethereum relay." and "This directory contains the [zkVM] portion of the governance application."

I'm curious to see what @nategraf or someone else more familiar with Bonsai contracts and zkVM guests thinks.

pdg744:
hmm... I agree that the reference to the starter template and specific comment about there being "two contracts" needs to be fixed, but I don't see the motivation for making the rest of the readme. This is currently the only example in `bonsai/examples`, so it feels worthwhile to offer users navigation paths to other relevant information. 

We're also soon going to be publishing a blog post pointing people to this example, and I think the more detailed readme will improve the new-user-onboarding-experience for those who end up seeing this demo before the foundry template.

bonsai/examples/governance/methods/README.md
```diff
@@ -0,0 +1,34 @@
+## zkVM Methods
+
+This directory contains the [zkVM] portion of your [Bonsai] application. 
+This is where you will define one or more [guest programs] to act as a coprocessor to your [on-chain logic].
+
+> In typical use cases, the only code in this directory that you will need to edit is inside [`guest/src/bin`].
```
tzerrell, line 6: This makes it sound like the user needs to modify code in `guest/src/bin` to use the governance example. I don't think that's true (but maybe I'm confused?)

bonsai/examples/governance/methods/README.md
```diff
@@ -0,0 +1,34 @@
+## zkVM Methods
+
+This directory contains the [zkVM] portion of your [Bonsai] application. 
+This is where you will define one or more [guest programs] to act as a coprocessor to your [on-chain logic].
+
+> In typical use cases, the only code in this directory that you will need to edit is inside [`guest/src/bin`]. 
+
+### Writing Guest Code
+To learn to write code for the zkVM, we recommend [Guest Code 101]. 
+Note that you don't need to write host code because the Bonsai Ethereum Relay acts as the zkVM host.
```
tzerrell, line 10: This makes it sound like writing guest code is necessary to use the governance example, and I don't think it is.

bonsai/examples/governance/methods/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Examples of what you can do in the guest can be found in the [RISC Zero examples].
```
tzerrell, line 12: This is confusing, as the reader is already looking at an example of what they can do with guest code

bonsai/examples/governance/methods/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+### From Guest Code to Binary File
+
+Code in the `methods/guest` directory will be compiled into one or more [RISC-V] binaries. 
+Build configuration for the methods is included in `methods/build.rs`.
+
+Each will have a corresponding image ID, which is a hash identifying the program.
```
tzerrell, line 19: This is general information about how guest code is compiled. I guess this is fine, but it feels a bit out of place, _especially_ in the absence of more detailed example-specific information.

bonsai/examples/governance/methods/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Uploading Binary to Bonsai
+When [deploying] your application, you will upload your binary to Bonsai where the guest will run when requested.
+The image ID will be included in the deployment of the smart contracts to reference your guest program living in Bonsai.
```
tzerrell, line 23: Do users need to upload their own binary of this to Bonsai? Or will we have one already uploaded? I'm unclear (in particular, I don't know whether governing different things requires different ImageIDs or just different inputs). In the former case we should rewrite this to make it clear we're talking about specifically for the governance application, not Bonsai apps in general, and to make clear under what conditions a new deployment is needed. In the latter case, this section should be dropped and we should write a section explaining how to use the existing deployment.

tzerrell (commented):
Sorry, I didn't say that clearly. I also don't think super short READMEs are good; instead, I think that most of this content is inaccurate, confusing, or misplaced in the context of this example. I was suggesting very short READMEs as something simple to start from and improve upon that we could be confident wasn't misleading readers until improvements could be made.

I put in inline comments this time to be more clear about what I mean. (Also I could be wrong about some of this, I'm hardly a Bonsai expert.)

bonsai/examples/governance/contracts/README.md
```diff
@@ -0,0 +1,16 @@
+# Bonsai Solidity Contracts
+
+This directory contains the Solidity contract for deploying a [Bonsai] application through the Ethereum relay. 
+There are two primary starter template contracts included.
+
+The Solidity libraries for Bonsai can be found at [github.com/risc0/risc0](https://github.com/risc0/risc0/tree/main/bonsai/ethereum).
+
+Contracts are built and tested with [forge], which is part of the [Foundry] toolkit.
+Tests are defined in the `tests` directory.
```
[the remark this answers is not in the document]
tzerrell: PR opened in https://github.com/risc0/bonsai-foundry-template/pull/65 to address this.

pdg744:
Thanks @tzerrell for catching that the original PR had lots of confusing remnants. 
I've revised to minimal readmes to try to get this landed quickly.

flaub (approved):

tzerrell (approved):

PR #926, opened 2023-09-29
[proof system] Reducing use of the term `plonk` 

### LINKED ISSUE #922: Reduce overloading of `plonk` term in cocdebase

pdg744:
The term `plonk` is overloaded and confusing in our codebase; this PR reduces use of that term to disambiguate meaning. 

Please review based on whether the suggested names make sense; we can worry about the implications in terms of breaking changes once we're in alignment about whether the names are good names. 

Resolves #922

### DIFF
--- a/risc0/zkp/src/prove/accum.rs
+++ b/risc0/zkp/src/prove/accum.rs
@@ -101,7 +101,7 @@ impl<'a, F: Field> CircuitStepHandler<F::Elem> for Handler<'a, F> {
     ) -> Result<()> {
         assert!(cycle < self.cycles);
         match name {
-            "plonkWriteAccum" => {
+            "WriteAccumulator" => {
                 assert_eq!(args.len(), F::ExtElem::EXT_SIZE);
                 let elem = F::ExtElem::from_subelems(args.iter().copied());
                 let ptr = self.get_ptr(extra);
@@ -109,7 +109,7 @@ impl<'a, F: Field> CircuitStepHandler<F::Elem> for Handler<'a, F> {
                 // buffer.
                 unsafe { ptr.add(cycle).write(elem) };
             }
-            "plonkReadAccum" => {
+            "ReadAccumulator" => {
                 assert_eq!(outs.len(), F::ExtElem::EXT_SIZE);
                 let ptr = self.get_ptr(extra);
                 // Already checked that our cycle number is in range, so this offset is in the
--- a/risc0/zkp/src/prove/accum.rs
+++ b/risc0/zkp/src/prove/accum.rs
@@ -19,12 +19,12 @@ use risc0_core::field::{Elem, ExtElem, Field};
 
 use crate::adapter::CircuitStepHandler;
 
-/// Tracks grand product accumulations for PLONK-style permutation arguments.
+/// Tracks grand product accumulations for permutation and lookup arguments
 pub struct Accum<E: Elem> {
     /// Total number of cycles in this run.
     cycles: usize,
 
-    // We use two PLONK-style grand product accumulation checks;
+    // We use two grand product accumulation checks;
     // one for the memory permutation and a second for a lookup table.
     // We have two `kinds`: memory and bytes.
     kinds: BTreeMap<String, Vec<E>>,
--- a/risc0/zkvm/src/host/server/prove/exec.rs
+++ b/risc0/zkvm/src/host/server/prove/exec.rs
@@ -62,7 +62,7 @@ pub struct MemoryState {
 
     // Tables for sorting using non-deterministic advice from the host.
     // The correctness of the sorting is checked using a permutation argument.
-    pub ram_table: plonk::MemoryTable,
+    pub ram_table: plonk::RamTable,
     pub bytes_table: plonk::BytesTable,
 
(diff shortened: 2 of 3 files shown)


risc0/zkvm/src/host/server/prove/exec.rs
```diff
@@ -245,8 +246,8 @@ impl CircuitStepHandler<Elem> for MachineContext {
 
     #[tracing::instrument(skip(self))]
     fn sort(&mut self, _: &str) {
-        self.memory.ram_plonk.sort();
-        self.memory.bytes_plonk.sort();
+        self.memory.memory_table.sort();
```
tzerrell, line 249: Someone closer to the circuit code than me should check whether the various places you switched "RAM" to "Memory" in this PR are ok -- it's possible the "Random Access" portion of the term is load-bearing.

tzerrell (commented):
Nice, reducing our dependence on people understanding which meaning of "PLONK" we have in mind is definitely a good thing!

One inline comment that needs another reviewer's attention.

risc0/zkvm/src/host/server/prove/exec.rs
pdg744: oh good call -- i don't have any opinion between, e.g., `ram_table` vs. `memory_table`

risc0/zkp/src/prove/accum.rs
```diff
@@ -101,15 +101,15 @@ impl<'a, F: Field> CircuitStepHandler<F::Elem> for Handler<'a, F> {
     ) -> Result<()> {
         assert!(cycle < self.cycles);
         match name {
-            "plonkWriteAccum" => {
+            "WriteAccumulator" => {
```
flaub, line 104: Unfortunately we can't change these until we make a corresponding change in the circuit.

risc0/zkvm/src/host/server/prove/exec.rs
flaub: Many of these names have corresponding components in the circuit. In some places it's required that the names match (like in the case of extern calls that are made from the circuit out to the host). In other cases, it would be confusing to a circuit dev.

I really like the idea of renaming things, but we need to make the circuit changes first.

tzerrell (changes requested):
See the other comments -- this is blocked until we can make corresponding naming changes on the circuit.

I'm glad we're figuring out what we want these names to be, though, so we are ready to change them both here & in the circuit when we have the opportunity.

risc0/zkvm/src/host/server/prove/exec.rs
pdg744: Yeah I figured many of the changes I propose here will necessitate other changes in order to be coherent; I guess first we can sort out whether the suggestions I make here actually make sense, and if so, we can make a plan for how to get all the relevant dependencies sorted.

flaub: I'm partial to `ram` rather than `memory` since `ram` is well known in CS and actually is more specific than memory.

pdg744:
Added this note to the PR description: 

*Please review based on whether the suggested names make sense; we can worry about the implications in terms of breaking changes once we're in alignment about whether the names are good names.*

tzerrell:
> Added this note to the PR description:
> 
> _Please review based on whether the suggested names make sense; we can worry about the implications in terms of breaking changes once we're in alignment about whether the names are good names._

I'm going to drop my request for changes, because I like these names with no objections beyond what's in already-seen comments, and CI will prevent us from accidentally landing this without making the corresponding circuit changes.

risc0/zkvm/src/host/server/prove/exec.rs
```diff
@@ -60,21 +60,22 @@ impl MemoryOp {
 pub struct MemoryState {
     pub ram: MemoryImage,
 
-    // Plonk tables for sorting plonks in proper order
-    pub ram_plonk: plonk::RamPlonk,
-    pub bytes_plonk: plonk::BytesPlonk,
+    // Tables for sorting using non-deterministic advice from the host.
+    // The correctness of the sorting is checked using a permutation argument.
+    pub memory_table: plonk::MemoryTable,
```
pdg744, line 65:
```suggestion
    pub ram_table: plonk::RamTable,
```

risc0/zkp/src/prove/accum.rs
pdg744: Marking this as resolved in order to focus attention on the question of whether the naming suggestion improves clarity; will worry about breaking changes later

risc0/zkvm/src/host/server/prove/exec.rs
pdg744: Updated to avoid making any changes `ram` and `memory`.

pdg744:
OK I've updated the PR to avoid making any changes from `ram` to `memory`. 

@jbruestle how do the terms here feel to you?

pdg744:
Discussed this with proof system team at October 16th sync -- closing this for now. We can revisit this again in the future as appropriate, perhaps in conjunction with audits and/or implementing LogUp

PR #930, opened 2023-09-29
[proof system] code to control

pdg744:
I think we're in agreement on changing the term `code` to `control` in the context of polyGoups. 

This PR starts implementing this change, although it seems like this change is going to have pretty wide-reaching impact and ctrl-f for `code` returns a lot of irrelevant results... 

(Getting this started but would definitely be happy to pass this off!)

### DIFF
--- a/risc0/zkp/src/hal/mod.rs
+++ b/risc0/zkp/src/hal/mod.rs
@@ -156,7 +156,7 @@ pub trait CircuitHal<H: Hal> {
     fn eval_check(
         &self,
         check: &H::Buffer<H::Elem>,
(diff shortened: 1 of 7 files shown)


flaub:
I'm thinking maybe we should do these kinds of reviews (that we know will cause build failures due to the circuit needing to be updated first) outside of PRs because these eat up CI machine resources on each commit that is pushed.

pdg744:
Discussed this at the Oct 16th proof system team sync. 

Given that we're planning to merge the `code/control` polyGroup into `data` soon anyway, this doesn't seem worth prioritizing.

PR #932, opened 2023-09-30
zkVM Private Multisig Wallet

blmalone:


examples/multisig/methods/guest/src/bin/zk_multisig.rs
```diff
@@ -0,0 +1,345 @@
...
+const TRUSTED_MERKLE_ROOT: &str =
+    "1962fbc17a4440d383f544724e9a943773c0cc6eade89b66b724c31cf33d3f62";
+
+pub fn main() {
+    let signature_one_input: String = env::read(); // <ecdsa_sig_address_one + recovery_id>
+    let signature_two_input: String = env::read(); // <ecdsa_sig_address_two + recovery_id>
+    let merkle_proof_one_input: String = env::read(); // <proof_hashes + index_of_leaf>
+    let merkle_proof_two_input: String = env::read(); // <proof_hashes + index_of_leaf>
+    let signed_message_input: String = env::read(); // <signed_message>
+    let all_leaves_input: String = env::read(); // <merkle_leaves>
```
flaub, line 34: We could use a single struct and have all of these be fields of the struct, that way this could would be `let input: Input = env::read()`

examples/multisig/methods/guest/src/bin/zk_multisig.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    if merkle_proof_one_input.eq(&merkle_proof_two_input) {
+        let error_message = "Error, merkle proofs cannot be the same!";
+        env::log(&error_message);
+        return;
+    }
+
+    /*
+     * Checking that the first merkle proof is valid
+     */
```
flaub, line 78: Please avoid using C-style comments, use `//` instead

blmalone: Thank you for the feedback. It's my first time writing Rust so still learning the ropes.

PR #936, opened 2023-10-02
Fix to allow risc0-zkvm to be published

flaub:
Using a feature flag to toggle a feature of a dev-dependency is not handled well by `cargo publish`. So we need to come up with a new way to optionally build the methods crate in a docker environment.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -241,6 +241,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18
-      - run: cargo run --bin cargo-risczero -- risczero install
       - run: cargo test -F docker -p risc0-build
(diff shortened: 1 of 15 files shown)


flaub:
Current dependencies on/for this PR:
* main
  * **PR #936** <a href="https://app.graphite.dev/github/pr/risc0/risc0/936" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/risc0/risc0/936?utm_source=stack-comment).

SchmErik:
I love the direction of this PR! I wonder if the test failure is resulting from some strange path-related issue. We replace '-' with '_' in some spots...

SchmErik (approved):
LGTM!

PR #950, opened 2023-10-05
guest memory: place stack below text/code

SchmErik:
The previous memory layout was prone to a security vulnerability because the
stack could collide with the heap. This change places the stack below
TEXT_START with enough space for 2 MB of stack. This also results in increased
locality of memory which will result in less page faults for small programs.


SchmErik:
@intoverflow I'm not sure if 2MB is large enough for zeth... Feel free to chime in and let us know if it's not enough!

flaub (approved):

Wollac:
> I'm not sure if 2MB is large enough for zeth... Feel free to chime in and let us know if it's not enough!

There is a test https://github.com/risc0/zeth/blob/main/testing/ef-tests/tests/executor.rs that attempts to run an artificial transaction that should require a lot of stack in the guest. It might be worth checking these changes against this test. If it passes, I think we are pretty sure that it should work for real Ethereum blocks as well.

PR #952, opened 2023-10-05
Automatize cycles count for datasheet

### LINKED ISSUE #940: Update cycle count examples on datasheet
- [x] Automatically update the cycle counts displayed for the examples 
- [x] Select a subset of the new examples to display

capossele:
This PR adds a new step in the `bench_datasheet` workflow to count the cycles of our examples.
Currently, `zkevm-demo` has been excluded as it requires a connection to some Ethereum node.

The new table in the ghpages will look something like this:
Screenshot 2023-10-05 at 10 19 50

The displayed examples can also be filtered by modifying:
https://github.com/risc0/risc0/blob/731a9793fa7f8e4350b06fd2df9e1e8465b7d32a/ghpages/dev/datasheet/index.html#L316-L331

Closes #940

### DIFF
--- a/.github/workflows/bench_datasheet.yml
+++ b/.github/workflows/bench_datasheet.yml
@@ -64,6 +64,9 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - run: cargo run --release -F $FEATURE --example loop -- --json | tee ${{ matrix.os }}-${{ matrix.device }}.json
         working-directory: risc0/zkvm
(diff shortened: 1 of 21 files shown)


tzerrell (approved):
Thank you, this will be a good upgrade!

I don't think it's necessary for this PR, but a possible place for future improvement would be to encapsulate a baseline execution for each example in some sort of `main` function in such a way that `cycle_count` could just "run main" rather than needing to duplicate appropriate inputs (or other host data).

flaub (approved):

PR #971, opened 2023-10-11
Migrate website to risc0 repository

flaub:


docs/readme.md
```diff
@@ -1,4 +0,0 @@
-# Documentation
```
tzerrell, line 1: This README file is here because of an external user who thought we had no docs because they found a nearly empty `docs` folder in our repo. I'd be happy to change it (e.g. to point to [docs.rs/risc0-zkvm](docs.rs/risc0-zkvm)) but I don't think we should drop it entirely.

tzerrell (commented):
This looks good locally, but I don't see the Vercel bot in-thread, nor do I see a preview deployment when I look on the Vercel website. That makes me think this version of the website won't actually deploy at present. Is that intended to be followup work, or should it be addressed before landing this PR?

pdg744:
Does this PR affect what's actually happening on https://dev.risczero.com, or is this just one of the preparation steps? 
Wondering about whether we'll still have Vercel previews in this approach, or whether we're planning on using `yarn start` to do local reviews?

flaub:
There's a chicken and egg problem. I need to get this landed so I can configured Vercel.

docs/readme.md
flaub: Thanks, I'll drop it for now but plan to deal with this in the future.

tzerrell (approved):

PR #974, opened 2023-10-12
adding info for upcoming study club sessions

pdg744:


tzerrell (approved):
Looks good to me, thanks for adding in both places!

pdg744:
hmm; I guess I need to run `prettier` for formatting... 
will check the prettier docs and see if i can figure it out

pdg744:
I followed [these instructions](https://prettier.io/docs/en/install) and it seems to have worked.

pdg744:
hmm... looks like there's something related to the term "status" that's not quite right yet...?

PR #976, opened 2023-10-12
Some corrections / clarifications for hello-world example

tbro:
There are a few typos and confusing text that may cause some confusion to someone exploring this ground for the first time. It also fixes the the link to the `examples ` directory at the bottom of the tutorial. Note that the **terminology** links there are also broken, but apparently due to how they are bieng interpreted by github.

### DIFF
--- a/examples/hello-world/tutorial.md
+++ b/examples/hello-world/tutorial.md
@@ -138,7 +138,8 @@ They will be added to the top of our `main` function:
 fn main() {
     let a: u64 = 17;
     let b: u64 = 23;
-    //...snip...
+    // ...snip...
+}
 ```
 
 We'd like the host to make the values of `a` and `b` available to the guest prior to execution. We can do this by adding them to the the executor environment, which is responsible for managing guest-readable memory. When the prover executes the program, it will have access to these guest inputs.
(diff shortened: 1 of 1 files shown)


examples/hello-world/tutorial.md
```diff
@@ -131,33 +131,36 @@ In the starter template project, our host driver program creates an executor env
...
-Let's start by picking some aesthetically pleasing primes:
+In this step, we'll be continuing to modify `hello-world/host/src/main.rs`.
+Let's start by picking some aesthetically pleasing primes such as `17` and `23`.
+They will be added to the top of our `main` function:
 ```
 fn main() {
     let a: u64 = 17;
     let b: u64 = 23;
-}
+    //...snip...
```
tzerrell, line 141: We need to leave the code snippets testable (or add `no_compile` if that's not practical)
```suggestion
    // ...snip...
}
```

tzerrell (commented):
Thanks! See my inline comment

examples/hello-world/tutorial.md
tbro: should be good in latest commit

flaub (approved):

PR #977, opened 2023-10-12
Begin Poseidon2 implementation

jbruestle:


risc0/build_kernel/kernels/metal/fp.h
```diff
@@ -87,6 +87,9 @@ class Fp {
   /// Construct an FP from a uint32_t, wrap if needed
   constexpr Fp(uint32_t val) : val(encode(val)) {}
 
+  /// Construct an FP from a raw value
```
tzerrell, line 90: Maybe clarify how this is different from `Fp(val)` (since "raw" might be interpreted as just "raw int not Fp object" not also "encoded in Montgomery form").
```suggestion
  /// Construct an Fp from an already-encoded raw value
```

risc0/zkp/src/hal/mod.rs
```diff
@@ -440,6 +440,7 @@ mod testutil {
     }
 
     pub(crate) fn hash_rows<H: Hal<Elem = BabyBearElem>>(hal_gpu: H) {
+        log::debug!("Hello\n");
```
tzerrell, line 443: Presumably left behind by mistake?
```suggestion
```

tzerrell (approved):

flaub (approved):

PR #978, opened 2023-10-13
Setup github pages preview

flaub:
Uses this action to provide previews hosted by GitHub Pages:
https://github.com/marketplace/actions/pages-preview

* Make markdown links consistent, use relative file paths rather than relative URLs
* Fixes risc0/website#305
* For now there are duplicate previews, one for Vercel and one hosted by Github Pages. We may migrate completely to github pages if this works well. Currently Vercel charges on a per-user basis which prevents non-member contributors from seeing the previews. Also, github pages is free.


morgangibbs87 (approved):

tzerrell:
Nice!

Issue #983, opened 2023-10-13
Support dynamic resolution of composition assumptions during execution


nategraf:
In addition to assumptions added to the added to the `ExecutorEnv` directly, it is desirable to have a callback by which assumptions can be resolved at runtime to provide my dynamic execution. One use-case for this would be to check a database of previously proven `ReceiptMetadata` structs with their `Receipt` and allow lookups to this database.

Here is the proposed interface:
```rust
impl ExecutorEnv {
    /// Provide an [AssumptionResolver] which will be used to lookup assumptions that cannot be
    /// found in the local assumptions cache.
    ///
    /// This can be used to allow for the creation of a database of past receipts, such that
    /// any previously proven claims can be used in new proofs. The local cache, populated by
    /// `add_assumption` will always be checked before forwarding a query to the
    /// [AssumptionResolver]. If no [AssumptionResolver] is provided, only the local cache of
    /// assumptions will be used.
    pub fn assumption_resolver(&mut self, resolver: Rc<dyn AssumptionResolver>) -> &mut Self {}
}
```

Related: #980, #981

Issue #982, opened 2023-10-13
Support conditional composite receipts in proof composition


nategraf:
With proof composition, there is the concept of a "conditional receipt". A conditional receipt is a receipt that is valid only if any assumptions made by the guest during execution can be proven by other corroborating receipts.

This issue is to implement conditional receipts through the prover and `CompositeReceipt` struct.

Related: #980, #981

flaub:
@nategraf Did this make it into 1.0?

nategraf:
No, because it requires the support for unresolved assumptions. The API is designed to support it, but the actual support is not implemented.

saefstroem:
Any update on this?

nategraf:
@saefstroem, just to bring the context from Discord to here: we scoped this feature, but so far have not found a use case that really needs it. Because we are always short on time, this is one that has never reached the top of priority list.

saefstroem:
> @saefstroem, just to bring the context from Discord to here: we scoped this feature, but so far have not found a use case that really needs it. Because we are always short on time, this is one that has never reached the top of priority list.

Right makes sense and is understandable. We have found other ways to go about our problem so it is fine. This is no longer a priority for me either.

PR #992, opened 2023-10-17
Initial implementation of proof composition

### LINKED ISSUE #980: Implement the proof composition guest API
Proof composition is a new feature of the zkVM allowing verification of receipts produced by other zkVM guest programs inside of a guest program. This allow for composition of guest programs, in which the journal, or other output state, of one guest can be consumed as input to another.

In this guest API, this is supported by the introduction of two new functions `env::verify(image_id: Digest, journal: &[u8])` and `env::verify_integrity(metadata: &ReceiptMetadata)`. `env::verify` is the high-level function most developers are expected to use, designed to make it easy to verify that a given guest by image ID produced the given journal, with an exit code of `Halted(0)` or `Paused(0)`. `env::verify_integrity` is a more flexible API for use cases where the developer wants to verify a receipt for a program that exited with an error, or some other uncommon condition.

Related: [#981](https://github.com/risc0/risc0/issues/981)

### LINKED ISSUE #981: Implement the proof composition host API and local prover for composite receipts
Proof composition is a new feature of the zkVM allowing verification of receipts produced by other zkVM guest programs inside of a guest program. This allow for composition of guest programs, in which the journal, or other output state, of one guest can be consumed as input to another.

In the host API, the prover can add receipts to the execution context as assumptions. During execution, when the guest calls `env::verify` or `env::verify_integrity`, the executor will look for a matching assumption and use the information to respond to the guest, and additionally to attach to the composite receipt.

Related: [#980](https://github.com/risc0/risc0/issues/980)

nategraf:
Proof composition is a new feature of the zkVM allowing verification of receipts produced by other zkVM guest programs inside of a guest program. This allow for composition of guest programs, in which the journal, or other output state, of one guest can be consumed as input to another.

This PR provides a basic implementation of proof composition. In particular, it implements composition over the `CompositeReceipt` struct (formerly `SegmentSeceipts` and `InnerReceipt::Flat`). It does not implement the recursion predicates needed to produce a single `SuccintReceipt` for a receipt of execution for a proof composition (#991).

TODO:

- [x] Implement more test as noted by `TODO(victor)`
- [x] Allow for `env::verify` of receipts ending in `Paused(0)`.
- [x] Resolve concern in Session about exit code and post state digest

Other changes:
* This PR clarifies and makes more consistent the semantics of the `verify` method on `Receipt` and related structs. Semantics are now to require an exit code of `Halted(0)` or `Paused(0)`, which aligns with how receipts are using in all examples, and what seems most likely to be intended by users. It adds the `verify_integrity_with_context` method as the core method for verifying cryptographic integrity and internal consistency, allowing use cases that do indeed want to verify receipts for non-success executions.
* Relatedly, the `prove_guest_errors` options is added to `ProverOpts` to continue and prove sessions that end with an error status (e.g. `ExitCode::Fault` or `ExitCode::Halted(1)`). When not set, the prover will stop after executor on guest error, such that the default is to only return receipts that of successful execution.

Resolves #980
Resolves #981
Related #982, #983


nategraf:
NOTE: Some of the diff is changes from formatting. If/when https://github.com/risc0/risc0/pull/993 I can rebase onto that to clean it up.

risc0/binfmt/src/image.rs
```diff
@@ -42,6 +42,7 @@ pub struct MemoryImage {
     /// Metadata about the structure of the page table
     pub info: PageTableInfo,
 
+    // TODO(1.0): Include the pc in the MemoryImage digest and remove it from SystemState.
```
flaub, line 45: I don't think we can include the `pc` in the digest unless we make a circuit change, and that's not planned for 1.0.

risc0/r0vm/src/lib.rs
```diff
@@ -35,6 +35,13 @@ struct Cli {
     #[arg(long, value_enum, default_value_t = HashFn::Sha256)]
     hashfn: HashFn,
 
+    /// When set to false, it is considered an error if the guest exists in an
+    /// unsuccessfully state, such as returning an exit code of `1` or
```
flaub, line 39:
```suggestion
    /// unsuccessful state, such as returning an exit code of `1` or
```

risc0/r0vm/src/lib.rs
```diff
@@ -35,6 +35,13 @@ struct Cli {
...
+    /// encountering a fault. When set to true, this is not considered an
+    /// error and proving will continue. Default is false.
```
flaub, line 41:
```suggestion
    /// error and proving will continue. Default is false.
    
```

risc0/r0vm/src/lib.rs
```diff
@@ -163,6 +170,7 @@ impl Cli {
         };
         let opts = ProverOpts {
             hashfn: hashfn.to_string(),
+            prove_guest_errors: false,
```
flaub, line 173: I feel like maybe we should punt on this feature, unless it's really necessary for proof composition.

Cargo.toml
```diff
@@ -49,6 +49,7 @@ risc0-r0vm = { version = "0.19.0-alpha.1", default-features = false, path = "ris
 risc0-sys = { version = "0.19.0-alpha.1", default-features = false, path = "risc0/sys" }
 risc0-zkp = { version = "0.19.0-alpha.1", default-features = false, path = "risc0/zkp" }
 risc0-zkvm = { version = "0.19.0-alpha.1", default-features = false, path = "risc0/zkvm" }
+risc0-zkvm-methods = { version = "0.19.0-alpha.1", default-features = false, path = "risc0/zkvm/methods" }
```
flaub, line 52: This breaks the release process, I use the `cargo release` tool and I have not yet found a way to publish methods crates.

bonsai/ethereum-relay/src/api/request_extractor.rs
```diff
@@ -112,12 +113,13 @@ fn json_content_type(headers: &HeaderMap) -> bool {
...
 mod tests {
-    use super::*;
-    use crate::sdk::CallbackRequest;
     use axum::http::{Method, Request};
     use hyper::{header::HeaderValue, Body};
     use serde::Serialize;
 
+    use super::*;
+    use crate::sdk::CallbackRequest;
+
```
flaub, line 122: I think we should try to avoid making any `cargo fmt` changes combined with logical changes. It makes it harder for a review to focus on the important changes.

This likely stems from unstable rules in the `rustfmt.toml` that no longer work on stable rust.

bonsai/sdk/Cargo.toml
```diff
@@ -15,7 +15,7 @@ tokio = { version = "1", features = ["full", "sync"], optional = true }
 
 [dev-dependencies]
 env_logger = "0.10"
-httpmock = "0.6"
+httpmock = "0.6.8"
```
flaub, line 18: Why is this change needed? Typically I ignore patch versions

risc0/cargo-risczero/src/lib.rs
```diff
@@ -19,13 +19,12 @@ mod commands;
...
 
-#[cfg(feature = "experimental")]
-pub use self::commands::build::BuildSubcommand;
-
 use clap::{Parser, Subcommand};
 
 #[cfg(feature = "experimental")]
 use self::commands::build::BuildCommand;
+#[cfg(feature = "experimental")]
+pub use self::commands::build::BuildSubcommand;
```
flaub, line 27: I like to put re-exports before normal uses so that it's clear what's public.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -678,6 +680,75 @@ pub unsafe extern "C" fn sys_alloc_aligned(bytes: usize, align: usize) -> *mut u
...
+            to_host.as_ptr() as u32,
+            2 * DIGEST_BYTES as u32,
+        )
+    };
+
+    // Check to ensure the host indicated success by returning 0.
+    // This should always be the case. This check is included for
+    // forwards-compatiblity.
+    if a0 != 0 {
+        const MSG: &[u8] = "sys_resolve_guest_journal returned error result".as_bytes();
```
flaub, line 719:
```suggestion
        const MSG: &[u8] = "sys_verify returned error result".as_bytes();
```

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -678,6 +680,75 @@ pub unsafe extern "C" fn sys_alloc_aligned(bytes: usize, align: usize) -> *mut u
...
+            metadata_digest as u32,
+            DIGEST_BYTES as u32,
+        )
+    };
+
+    // Check to ensure the host indicated success by returning 0.
+    // This should always be the case. This check is included for
+    // forwards-compatiblity.
+    if a0 != 0 {
+        const MSG: &[u8] = "sys_verify returned error result".as_bytes();
```
flaub, line 747:
```suggestion
        const MSG: &[u8] = "sys_verify_integrity returned error result".as_bytes();
```

Cardosaum:
@nategraf, quick not with regards to the PR description; When marking issues to be closed, we should follow the pattern `<keyword> <issue> [, <keywork> <issue>]` if we want to close multiple issues at the same time. In the current state we won't close `#981` when merged.

[reference](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

risc0/r0vm/src/lib.rs
jbruestle: So mostly this is useful to separate 'proof is invalid' from 'guest returned an error', which right now is mostly about panics, but also sets us up well for more general proof-of-fault support.  I'm not sure if we ended up deciding that proof-of-fault was a 1.0 feature, but even if it's not, this feature will reduce eventual API changes stemming from proof-of-fault, so given that it's already implemented, I think it's a good thing to keep.

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -594,32 +656,21 @@ fn oom() {
...
 
-    assert!(session_faulted(access_memory(0x0000_0000)));
-    assert!(session_faulted(access_memory(0x0C00_0000)));
-    assert!(!session_faulted(access_memory(0x0B00_0000)));
+    assert_eq!(access_memory(0x0000_0000).unwrap(), ExitCode::Fault);
+    assert_eq!(access_memory(0x0C00_0000).unwrap(), ExitCode::Fault);
+    assert_eq!(access_memory(0x0B00_0000).unwrap(), ExitCode::Halted(0));
 }
 
 /// The post-state digest (i.e. the Merkle root of the memory state at the end
```
Cardosaum, line 676: There is a typo in line 677: `program`

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -413,3 +411,137 @@ mod docker {
...
+            .unwrap();
+        // TODO(#982) Conditional receipts currently return an error on verification.
+        assert!(get_prover_server(&opts)
+            .unwrap()
+            .prove_elf(env, MULTI_TEST_ELF)
+            .is_err());
+
+        // TODO(#982) With conditional receipts, implement the following cases.
+        // verify with proven corraboration in verifier success.
+        // verify with unresolved corraboration in verifier success.
```
Cardosaum, line 494: typo `corraboration` in both lines.

risc0/zkvm/src/host/server/session.rs
```diff
@@ -135,6 +153,76 @@ impl Session {
...
+                self.assumptions.is_empty(),
+                "Session with exit code {:?} has encoded assumptions",
+                self.exit_code
+            );
+            None
+        };
+
+        // DO NOT MERGE: post should match the final segment system state, but doesn't
+        // always. Prior to merging figure out what is happening here and what
+        // to do about it (possibly nothing).
```
Cardosaum, line 208: Should we keep an open comment in this PR to track this?

risc0/zkvm/src/receipt_metadata.rs
```diff
@@ -0,0 +1,449 @@
...
+            ],
+            &[sys_exit, user_exit],
+        )
+    }
+}
+
+/// Indicates how a Segment or Session's execution has terminated
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq)]
+pub enum ExitCode {
+    /// This indicates when a system-initiated split has occured due to the
```
Cardosaum, line 243:
```suggestion
    /// This indicates when a system-initiated split has occurred due to the
```

risc0/zkvm/src/sha.rs
```diff
@@ -61,6 +61,19 @@ cfg_if::cfg_if! {
     }
 }
 
+/// Defines a collision resistant hash for the typed and structured data.
+pub trait Digestable {
```
Cardosaum, line 65: Would it be desirable to change from `Digestable` to `Digestible`?

Cardosaum (commented):
Awesome work! 🤘💯

I just made a few suggestions, mostly on small typos.

jbruestle (commented):
Did a somewhat detailed review:  Overall very nicely done, and coding is quite clear (although there seemed like some possible confusing parts in the PR description, 'Semantics are not to require' maybe is 'Semantics are now to require'?)

Some minor points: I'm not sure if there is a better name for 'MaybePruned', but it might be good to document it a bit.  I found it confusing initially.  I eventually figured out that it's a way of either holding actual data which is 'merklizable', or just a commitment to it (i.e. you don't know the actual value, just the hash).  Also, protobuf handling there is (perhaps unresolvably?) ugly due to the need for the 'any'.  Generally, this is a good example of the complexity arising from the combination of Merklization and serializability.  I feel like this is an interesting area for research someday.

Second point: Is there an implied rule that an assumption can only be used once?  I'm not very up to speed on rust memory management, but the extracting of assumption from the cache when used (rather than copying) seems like a nice memory optimization (proofs are of course non-trivially large), but I wonder if it adds weird user constraints.  Maybe I'm misreading what is happening in the code though.

risc0/zkvm/src/sha.rs
nategraf: Yes, haha 😅  Sorry, I barely passed spelling in the third grade 😆

risc0/r0vm/src/lib.rs
nategraf: I assume this is because the paragraph clutters the help output? I address this in 2eb803d1a9af86266b034d2da1b4c5667b99ce5f

Cargo.toml
nategraf: It breaks the release process to include this as a workspace level dependency? In it's Cargo.toml `release` is set to false, so it shouldn't try to be released if I understand correctly. Maybe you could clarify the issue?

bonsai/ethereum-relay/src/api/request_extractor.rs
nategraf: File was removed from diff after merging #993

> But won't CI fail if formatting isn't run?

So some rules are unstable. We have them activated right now, but CI uses stable Rust, and so the rules are not turned on. I ran `cargo +nightly fmt --all` which them applied all the stable rules, which is why there are extra changes. In #993 I suggested we might run with `+nightly` in CI, but instead I think we will just turn off those rules probably.

bonsai/sdk/Cargo.toml
nategraf: Removed the patch version in 4d19fe86412bca8d17f87f872c0808d0074511a3

risc0/cargo-risczero/src/lib.rs
nategraf: Change reverted in 4d19fe86412bca8d17f87f872c0808d0074511a3

risc0/zkvm/platform/src/syscall.rs
nategraf: Resolved in 5d13bbf0335dcc114b13d85c03462da73fcce11d

risc0/zkvm/platform/src/syscall.rs
nategraf: Resolved in 5d13bbf0335dcc114b13d85c03462da73fcce11d

nategraf:
> @nategraf, quick not with regards to the PR description; When marking issues to be closed, we should follow the pattern `<keyword> <issue> [, <keywork> <issue>]` if we want to close multiple issues at the same time. In the current state we won't close `#981` when merged.
> 
> I went ahead and updated the description. it's more verbose but should do the trick 🤷‍♂️
> 
> [reference](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

Thanks for fixing this!

risc0/zkvm/src/sha.rs
nategraf: Fixed this in 0e5f89aec525422ec605d83e730925f120f575f5

risc0/zkvm/src/host/server/session.rs
nategraf: Yes, thank you

risc0/zkvm/src/guest/env.rs
```diff
@@ -71,6 +94,132 @@ pub fn syscall(syscall: SyscallName, to_host: &[u8], from_host: &mut [u32]) -> s
...
+            assumptions: MaybePruned::Pruned(Digest::zero()),
+        })
+        .into(),
+    };
+    unsafe { ASSUMPTIONS_DIGEST.add(assumption_metadata.into()) };
+
+    Ok(())
+}
+
+/// Error encountered during a call to [crate::verify_metdata].
```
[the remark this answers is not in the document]
nategraf: Fixed in 813bf617be730feefdfcc7ae4fb8af1e8cd1539d

nategraf:
> Some minor points: I'm not sure if there is a better name for 'MaybePruned', but it might be good to document it a bit. I found it confusing initially. I eventually figured out that it's a way of either holding actual data which is 'merklizable', or just a commitment to it (i.e. you don't know the actual value, just the hash).

Good call. I've added a paragraph of explanation in 677da5cebf05b2ce6d2066ce42ebb39a9cc55f3d

>  Also, protobuf handling there is (perhaps unresolvably?) ugly due to the need for the 'any'. Generally, this is a good example of the complexity arising from the combination of Merklization and serializability. I feel like this is an interesting area for research someday.

Ya, this is a bit of a sore spot of this implementation. I'd say it's moreso a fact of protobuf not supporting generics and my choice to use generics here (protobuf _does_ support Rust-style enums, which is nice). An alternative would be to simply make an explicit message type for each of the types we use with MaybePruned. It's only a handful, so the bloat would be manageable, especially with some clever use of traits. I'm definitly open to alternatives to bring in `Any` here.

> Second point: Is there an implied rule that an assumption can only be used once? I'm not very up to speed on rust memory management, but the extracting of assumption from the cache when used (rather than copying) seems like a nice memory optimization (proofs are of course non-trivially large), but I wonder if it adds weird user constraints. Maybe I'm misreading what is happening in the code though.

There is nothing preventing an assumption from being used twice, but I imagine 95% of guests will only verify a given assumption once. We could move the `Assumption` instead of copying, and if someone wants to use the same assumption twice, they can add it twice to the cache. Overall though, I don't have any evidence this is a performance or memory issue and my intuition says it's probably not. Each program will likely deal with a small list of assumptions, and assumptions can be `SuccintReceipts`. My intuition here is to only change this if it appears to be a performance issue in practice. If that happens, we can maintain the same semantics as today by using `Cow` or `Rc` types internally.

risc0/binfmt/src/hash.rs
```diff
@@ -0,0 +1,125 @@
...
+    let tag_digest: Digest = *S::hash_bytes(tag.as_bytes());
+    let mut all = Vec::<u8>::new();
+    all.extend_from_slice(tag_digest.as_bytes());
+    for digest in down {
+        all.extend_from_slice(digest.borrow().as_ref());
+    }
+    for word in data.iter().copied() {
+        all.extend_from_slice(&word.to_le_bytes());
+    }
+    let down_count: u16 = down.len().try_into().unwrap();
```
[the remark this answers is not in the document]
nategraf: A developer would need to define a "struct" with more than 2^16 fields. If so, this is an error by the developer.

risc0/zkvm/src/guest/env.rs
```diff
@@ -71,6 +94,132 @@ pub fn syscall(syscall: SyscallName, to_host: &[u8], from_host: &mut [u32]) -> s
...
+impl std::error::Error for VerifyError {}
+
+/// Verify there exists a receipt for an execution with the given `image_id` and
+/// `journal`.
+///
+/// In order to be valid, the [Receipt] must have have `ExitCode::Halted(0)`, an
+/// empty assumptions list, and an all-zeroes input hash. It may have any post
+/// [SystemState].
+pub fn verify(image_id: Digest, journal: &[u8]) -> Result<(), VerifyError> {
+    let journal_digest: Digest = Sha256::digest(journal).as_slice().try_into().unwrap();
```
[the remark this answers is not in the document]
nategraf: Changed this to avoid a potential panic

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -435,14 +458,30 @@ impl TryFrom<pb::core::Digest> for Digest {
...
+
 impl From<ReceiptMetadata> for pb::core::ReceiptMetadata {
     fn from(value: ReceiptMetadata) -> Self {
         Self {
             pre: Some(value.pre.into()),
             post: Some(value.post.into()),
             exit_code: Some(value.exit_code.into()),
             input: Some(value.input.into()),
-            output: Some(value.output.into()),
+            // Translate MaybePruned<Option<Output>>> to Option<MaybePruned<Output>>.
```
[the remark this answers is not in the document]
nategraf: This comment is saying what the code is doing. I'm not sure how to make it more clear. It's not really what I think of as a NOTE either. Feel free to suggest a change!

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -435,14 +458,30 @@ impl TryFrom<pb::core::Digest> for Digest {
...
+            output: match value.output {
+                MaybePruned::Value(x) => match x {
+                    Some(output) => Some(MaybePruned::Value(output).into()),
+                    None => None,
+                },
```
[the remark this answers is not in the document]
nategraf: Thanks

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -456,11 +495,27 @@ impl TryFrom<pb::core::ReceiptMetadata> for ReceiptMetadata {
             post: value.post.ok_or(malformed_err())?.try_into()?,
             exit_code: value.exit_code.ok_or(malformed_err())?.try_into()?,
             input: value.input.ok_or(malformed_err())?.try_into()?,
-            output: value.output.ok_or(malformed_err())?.try_into()?,
+            // Translate Option<MaybePruned<Output>> to MaybePruned<Option<Output>>.
```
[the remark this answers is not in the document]
nategraf: Same as above

risc0/zkvm/src/host/server/exec/tests.rs
nategraf: Oof ya 😅

risc0/zkvm/src/host/server/prove/tests.rs
nategraf: Thanks

risc0/zkvm/src/host/receipt.rs
```diff
@@ -35,59 +35,15 @@ use risc0_zkp::{
 use risc0_zkvm_platform::WORD_SIZE;
 use serde::{Deserialize, Serialize};
 
-use super::{
-    control_id::{BLAKE2B_CONTROL_ID, POSEIDON_CONTROL_ID, SHA256_CONTROL_ID},
-    recursion::SuccinctReceipt,
-};
+use super::control_id::{BLAKE2B_CONTROL_ID, POSEIDON_CONTROL_ID, SHA256_CONTROL_ID};
+// Make succint receipt available through this `receipt` module.
```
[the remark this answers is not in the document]
nategraf: Thanks

risc0/zkvm/src/receipt_metadata.rs
nategraf: Thanks

Cargo.toml
flaub: You can try it out with:
```
cargo release publish
```

This is safe to do because it will just do a dry run by default.

risc0/zkvm/src/guest/env.rs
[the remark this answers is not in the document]
flaub: Generally I think we should try to avoid `expect` in the guest because it's very costly. Maybe this is a bad idea, but I haven't found much use for `expect` generally since I can get the line of code the panic occurs and the Error type itself should contain information about the error.

risc0/zkvm/src/host/api/convert.rs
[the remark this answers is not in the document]
Cardosaum: Maybe implementing a `transpose` method for `MaybePruned`? 
Feel free to just ignore this though, as it's a very nitpick point.

Cargo.toml
nategraf: So if I comment out the `package.metadata.release.release  = false` setting in the `Cargo.toml` for `risc0-zkvm-methods` and run `cargo release publish -p risc0-zkvm-methods` it fails with a build error, and I assume this is the error that prevents publishing methods packages.

When that line is left intact, `cargo release publish -p risc0-zkvm-methods` gives `error: no packages selected`. Additionally `cargo release publish` makes no mention of the methods package. So I _think_ the methods crate is being properly ignored, and this change will not change that?

risc0/zkvm/src/host/api/convert.rs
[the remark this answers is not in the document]
nategraf: I thought about that, but `transpose` is not always well defined. In particular, it's unclear what `MaybePruned::Pruned(Digest(_)).transpose()` should be when digest is non-zero. I think this is the factoring to use for the moment.

risc0/zkvm/src/host/receipt.rs
```diff
@@ -156,185 +113,400 @@ pub struct Receipt {
...
+                    self.journal_digest
+                        .ok_or(VerificationError::ReceiptFormatError)?,
+                ),
+                assumptions: self.assumptions_metadata()?.into(),
+            };
+
+            // If these digests do not match, this receipt is internally inconsistent.
+            if self_output.digest() != metadata.output.digest() {
+                let empty_output = metadata.output.is_none()
+                    && self.journal_digest.unwrap() == Vec::<u8>::new().digest();
```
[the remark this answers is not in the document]
nategraf: Because it was already checked above to be `Some(_)`. I went ahead and changed this though in e23ad8fa787b2566e730d1d9eff7d47919cabe0d

risc0/zkvm/src/host/receipt.rs
```diff
@@ -436,6 +562,70 @@ impl SegmentReceipt {
     }
 }
 
+/// An assumption associated with a guest call to `env::verify` or
+/// `env::verify_metdata`.
```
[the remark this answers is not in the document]
nategraf: Fixed in e23ad8fa787b2566e730d1d9eff7d47919cabe0d

risc0/zkvm/src/host/receipt.rs
```diff
@@ -452,61 +642,44 @@ fn decode_system_state_from_io(
...
-            _ => Err(VerificationError::ReceiptFormatError),
-        }
-    }
+fn decode_receipt_metadata_from_io(
+    io: layout::OutBuffer,
+) -> Result<ReceiptMetadata, VerificationError> {
+    let body = layout::LAYOUT.mux.body;
+    let pre = decode_system_state_from_io(io, body.global.pre)?;
+    let mut post = decode_system_state_from_io(io, body.global.post)?;
+    // In order to avoid extra logic in the rv32im circuit to perform arthimetic on
```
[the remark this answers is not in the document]
nategraf: Fixed in e23ad8fa787b2566e730d1d9eff7d47919cabe0d

risc0/zkvm/src/host/receipt.rs
```diff
@@ -156,185 +113,400 @@ pub struct Receipt {
...
+            output: output.into(),
+        })
     }
 
-    /// Verify the integrity of this receipt.
-    pub fn verify_with_context(
+    /// Check that the output fields in the given receipt metadata are
+    /// consistent with the exit code, and with the journal_digest and
+    /// assumptions encoded on self.
+    fn verify_output_consitency(
```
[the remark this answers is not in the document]
nategraf: Fixed in e23ad8fa787b2566e730d1d9eff7d47919cabe0d

risc0/zkvm/src/host/receipt.rs
```diff
@@ -156,185 +113,400 @@ pub struct Receipt {
...
+        } else {
+            Err(VerificationError::ReceiptFormatError)
+        }
+    }
+
+    /// Extract the [ReceiptMetadata] from this receipt.
+    pub fn get_metadata(&self) -> Result<ReceiptMetadata, VerificationError> {
+        match self {
+            InnerReceipt::Composite(ref receipt) => receipt.get_metadata(),
+            InnerReceipt::Succinct(ref succint_recipt) => Ok(succint_recipt.meta.clone()),
```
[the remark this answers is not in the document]
nategraf: Fixed in e23ad8fa787b2566e730d1d9eff7d47919cabe0d

flaub:
I haven't had a chance to do a deeper review of this, but I'd like to. Could we hold landing it?

nategraf:
> I haven't had a chance to do a deeper review of this, but I'd like to. Could we hold landing it?

Sure, that sounds good to me 👍

risc0/zkvm/src/guest/env.rs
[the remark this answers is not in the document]
nazar-pc: It is actually a good practice to always use `.expect("Proof why this never ever happens")` instead of `.unwrap()` (except in tests where crashes are fine). It significantly helps with auditability and generally quality of the software, especially in blockchain context. And with careful use of type system most of these should compile away.

`.expect()` is not to print a better message to developer than `.unwrap()` when things crash, it is to prove to the reader that code will not crash under any circumstances (or in rare cases that it might, but explain why it is acceptable), otherwise error handling must be used instead.

nategraf:
I started to implement a test for non-success exit codes such as `Halted(1)` and `Paused(1)` but realized there is not a standard way to exit the guest with a non-zero code. In [ea5f321](https://github.com/risc0/risc0/pull/992/commits/ea5f321aa381263c6962a61e8dd00a96e20345e4) I added `env::exit` and made `env::pause` accept and exit code. I can take those out though if we'd rather not include this in this PR.

.github/workflows/main.yml
```diff
@@ -186,6 +186,7 @@ jobs:
       - run: cargo doc --no-deps --exclude=risc0-zkvm-methods --workspace
 
   check-benchmarks:
+    if: false # Currently broken. See https://github.com/risc0/risc0/issues/1013
```
flaub, line 189: I think it's OK to still run this and have it fail because currently this status check isn't required to allow PRs to land.

.github/workflows/bench_applications.yml
```diff
@@ -24,6 +24,7 @@ concurrency:
 
 jobs:
   bench:
+    if: false # Currently broken. See https://github.com/risc0/risc0/issues/1013
```
flaub, line 27: I think it's OK for this to fail

Cargo.toml
```diff
@@ -49,6 +49,7 @@ risc0-r0vm = { version = "0.19.0-rc.1", default-features = false, path = "risc0/
 risc0-sys = { version = "0.19.0-rc.1", default-features = false, path = "risc0/sys" }
 risc0-zkp = { version = "0.19.0-rc.1", default-features = false, path = "risc0/zkp" }
 risc0-zkvm = { version = "0.19.0-rc.1", default-features = false, path = "risc0/zkvm" }
+risc0-zkvm-methods = { version = "0.19.0-rc.1", default-features = false, path = "risc0/zkvm/methods" }
```
flaub, line 52: This command fails by having this be in the workspace:

```
$ cargo package -p risc0-zkvm
error: failed to prepare local package for uploading

Caused by:
  no matching package named `risc0-zkvm-methods` found
  location searched: registry `crates-io`
  required by package `risc0-zkvm v0.19.0-rc.1 ($SRC_ROOT/risc0/risc0/zkvm)`
```

risc0/zkp/src/verify/mod.rs
```diff
@@ -33,6 +33,7 @@ use crate::{
 };
 
 #[derive(PartialEq)]
+#[non_exhaustive]
```
flaub, line 36: What does this do?

risc0/zkvm/src/host/client/prove/bonsai.rs
```diff
@@ -85,7 +85,18 @@ impl Prover for BonsaiProver {
...
+                    receipt.verify_integrity_with_context(ctx)?;
+                    ensure!(
+                        receipt.get_metadata()?.pre.digest() == image_id,
+                        "received unexpected image ID: expected {}, found {}",
+                        hex::encode(&image_id),
+                        hex::encode(&receipt.get_metadata()?.pre.digest())
+                    );
+                } else {
+                    receipt.verify_with_context(ctx, image_id)?;
+                }
```
flaub, line 99: I wonder why this doesn't belong in `verify_with_context` on the receipt, rather than for each caller to do?

risc0/zkp/src/verify/mod.rs
nategraf: It indicates that more enum variants (i.e. more error conditions) may be added in the future, and so Rust will give a compile error to developers that try to write an exhaustive `match` statement in their code. Adding this attribute prevents adding a new enum variant from being a breaking change, since all callers are required to handle the "unknown variant" case.

risc0/zkvm/src/guest/env.rs
```diff
@@ -109,6 +109,7 @@ pub fn syscall(syscall: SyscallName, to_host: &[u8], from_host: &mut [u32]) -> s
 /// In order to be valid, the [crate::Receipt] must have `ExitCode::Halted(0)`, an
 /// empty assumptions list, and an all-zeroes input hash. It may have any post
 /// [crate::SystemState].
+#[must_use]
```
flaub, line 112: My understanding is the `Result` is already marked as `#[must_use]` and so this is redundant. I also thought there was a clippy lint for this:
https://std-dev-guide.rust-lang.org/policy/must-use.html

flaub: https://rust-lang.github.io/rust-clippy/stable/index.html#/double_must_use

risc0/zkvm/src/host/client/prove/bonsai.rs
nategraf: I wouldn't expect most callers to have a if statement like this. It's in this location because the Bonsai client is implementing `Prover` and this is where it is checking the Bonsai's work. So the conditions that need to be enforced here depend on whether the `Prover` interface user is expecting a successful execution or doesn't care.

.github/workflows/main.yml
nategraf: Removed this change in 6b167135198f97d7b4ad829d00c699e6f026ab16

.github/workflows/bench_applications.yml
nategraf: Removed this change in 6b167135198f97d7b4ad829d00c699e6f026ab16

risc0/zkvm/src/guest/env.rs
nategraf: Oh, nice! I did not know that. Removed the new `#[must_use]` attributes in 444d8710d50e809bde089f830bddb0eebbf9be4c

risc0/zkp/src/verify/mod.rs
flaub: Wouldn't we want new enum variants to represent a breaking change?

Cargo.toml
nategraf: Oh interesting! I'm surprised this makes a difference. I've reverted this change in 616c42ca446e7f1243edb1fd18bcc033b609bcb8

risc0/zkp/src/verify/mod.rs
nategraf: My intuition in this case is "no". If your intuition is "yes", then I'll go ahead and remove this one.

flaub: I'm not sure, let's leave it and see what happens over time

flaub:
I think once we adjust the Any type this looks good to go!

nategraf:
> I think once we adjust the Any type this looks good to go!

@flaub, I've removed the usage of google.Any and replaced it with `bytes` in 6e555264691bbc29423c44e4c914105a16cb2de3. I started implementing an additional `tag` field on the `MaybePruned` message, but it started to seem unclear to me what value it provides. Do you have thoughts, after seeing this, if we need a tag to indicate the type encoded? The type should always be known at compile-time, just like for the other protos we decode.

flaub:
I guess if the type is always fixed based on context, then we don't need the tag.

flaub (approved):

nategraf:
It looks like there are two test failures I need to investigate:
```
cargo test -p risc0-zkvm -F docker -F prove -- "docker::"

running 4 tests
test host::server::exec::tests::docker::trace ... ok
test host::server::exec::tests::docker::session_limit ... FAILED
test host::server::prove::tests::docker::pause_continue ... FAILED
test host::server::prove::tests::docker::continuation ... ok

failures:

---- host::server::exec::tests::docker::session_limit stdout ----
R0VM[30073] Busy loop starting!
R0VM[31178] Busy loop complete
R0VM[30073] Busy loop starting!
R0VM[30073] Busy loop starting!
R0VM[77748] Busy loop complete
R0VM[30073] Busy loop starting!
R0VM[77748] Busy loop complete
R0VM[85523] Busy loop starting!
R0VM[85523] Busy loop starting!
R0VM[110507] Busy loop complete
thread 'host::server::exec::tests::docker::session_limit' panicked at 'assertion failed: run_session(1 << 16, 15, 10).is_ok()', risc0/zkvm/src/host/server/exec/tests.rs:1153:9

---- host::server::prove::tests::docker::pause_continue stdout ----
R0VM[31176] before
thread 'host::server::prove::tests::docker::pause_continue' panicked at 'called `Result::unwrap()` on an `Err` value: received unexpected metadata digest: expected 2d110acfcf4066889ba72969c9edcc9646101256764fb4f4974afa109a535a66, found 154f1184403e2556c31f1403625c2f24049116544567e8fd5c9db55a4f1a6abc', risc0/zkvm/src/host/server/prove/tests.rs:389:39


failures:
    host::server::exec::tests::docker::session_limit
    host::server::prove::tests::docker::pause_continue

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 115 filtered out; finished in 59.86s

error: test failed, to rerun pass `-p risc0-zkvm --lib`
```

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -839,6 +1150,6 @@ mod docker {
         let err = run_session(1 << 16, 15, 3).err().unwrap();
         assert!(err.to_string().contains("Session limit exceeded"));
 
-        assert!(run_session(1 << 16, 15, 10).is_ok());
+        assert!(run_session(1 << 16, 15, 16).is_ok());
```
nategraf, line 1153: @flaub it appears that the extra work in the guest, presumably the slight increase in data hashed for the output plus any paging, increased the minimum number of cycles are reduced the amount of work that can be done in a 2^15 segment by a significant amount. I want to investigate this further, but would also like to get this PR merged.

Should we go ahead and merge this now?

PR #994, opened 2023-10-17
Cleanup of template README

### LINKED ISSUE #986: Structure of a zkVM app link is broken
In the ‘rust-starter’ section at https://github.com/risc0/risc0/tree/main/templates/rust-starter, the structure for a zkVM app is highlighted but the link seems to be broken.

flaub:
* Fixes #986

### DIFF
--- a/templates/rust-starter/README.md
+++ b/templates/rust-starter/README.md
@@ -1,51 +1,62 @@
 # RISC Zero Rust Starter Template
 
-Welcome to the RISC Zero Rust Starter Template! This template is intended to give you a starting point for building a project using the RISC Zero zkVM. Throughout the template (including in this README), you'll find comments labelled `TODO` in places where you'll need to make changes.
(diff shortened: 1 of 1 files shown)


flaub:
A few things to note for markdown files:
* Always use markdown links, put all links at the bottom as footnotes. This makes it much easier to maintain links as they change over time.
* Link names are distinct from the representation and can be overridden. Every footnote link should be unique.
* Word wrap at 80-columns for users reading markdown as a text file (from a terminal/console).
* Always annotate code blocks with a format marker. For code blocks that don't have a useful language syntax, use `text`

pdg744 (approved):
Great! Thanks for pushing this. 

It occurred to me while reviewing that the `Video Tutorial` referenced here is almost a year old. Not really related to this PR, but I wonder whether we're better off removing that reference at this point...

PR #1002, opened 2023-10-17
Replace add_input with write and write_slice

### LINKED ISSUE #845: The dev ex of the `add_input`
https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.ExecutorEnvBuilder.html#method.add_input

I hope the problem is described by the two interactions below:

> The add_input API is not intuitive. Imagine you need to pass a Vec<u8> via add_input and env::read it in the guest. I would expect it would work if I just pass a raw vector. However, I have to always go through to_vec. This part is surprising  and the docs don't clarify nor have an example. 

> However, imagine that you are a developer that wants to start asap, with the starter template. You have a byte buffer, and assume that since it's a byte buffer no special serialization is needed, since in a lot of different places this is how it usually work. It doesn't work with some cryptic error (such are the serialization/deserialization errors). You go to the documentation, don't see any note on this, but see the example, which kinda doesn't tell you much to fix the wrong mental model, if anything it only confirms it: well, I have a byte vec, why do I need to to_vec it?
i think the signature with the Pod give a hint that something is fishy, but once again, you are an eager developer who wants to get the hands dirty!

flaub:
* Fixes #845
* Consistent with corresponding guest API (`read` and `read_slice`).


flaub:
A followup PR should be to have a guide on how to use host/guest I/O.

Is this an improvement?

risc0/zkvm/src/host/client/env.rs
```diff
@@ -179,27 +181,56 @@ impl<'a> ExecutorEnvBuilder<'a> {
         self
     }
 
-    /// Add initial input that can be read by the guest from stdin.
+    /// Write input data to a buffer that the zkVM guest can read from.
     ///
-    /// Calling `add_input` iteratively concatenates inputs; the guest can
-    /// access each input using consecutive reads.
+    /// This function will serialize `data` using a zkVM-optimized codec that
+    /// can be deserialized in the guest with a correponding `env::read` with
```
tzerrell, line 187: Typo
```suggestion
    /// can be deserialized in the guest with a corresponding `env::read` with
```

tzerrell (approved):
Great! I think this is a substantial improvement.

mothran (approved):
So nice to have this improvement.

SchmErik (approved):
I think this is much better over all! I wonder if `read` and `write` are too vague. Would it be even more clear if we renamed these functions as `read_input` / `read_input_slice` and `write_input` / `write_input_slice`? I know we've had `read` for a while and my mind has gotten so used to it but I wonder if this improves the ability for new people to understand this better.

PR #1003, opened 2023-10-17
Use poseidon as default

### LINKED ISSUE #1000: Replace SHA-256 with Poseidon for default hashfn
Bonsai only works with the Poseidon hashing function, since recursion relies on this. Change the default to match.

capossele:
Closes #1000


flaub (approved):
Thank you!

PR #1005, opened 2023-10-17
Collapse CircuitCoreDef and CircuitProveDef into CircuitDef

SchmErik:
When I published the recursion verifier, I split the CircuitDef trait into two separate traits. Now that the recursion prover and verifier are in the same repo, we can merge the two traits back into one again!


SchmErik:
Combining the traits results in build errors. The only way to fix the error is to include the cpp module even when the prove feature is disabled. Let's keep the traits separate as-is. Closing

PR #1010, opened 2023-10-18
Add session ID to the Relay API response

### LINKED ISSUE #1006: Return session_uuid in `/v1/callbacks` REST API route
When it comes to offchain proving (code for context below), I think it would make sense to return the bonsai session ID as part of the response. This would allow a client to track the status of the requested callback

```rs
#[async_trait::async_trait]
impl<S: Storage + Sync + Send> EventProcessor for ProxyCallbackProofRequestProcessor<S> {
    type Event = CallbackRequestFilter;

    async fn process_event(
        &self,
        event: CallbackRequestFilter,
    ) -> Result<(), crate::api::error::Error> {
        let input_id =
            upload_input(self.bonsai_client.clone(), event.input.clone().to_vec()).await?;
        let bonsai_session_id = create_session(
            self.bonsai_client.clone(),
            hex::encode(event.image_id),
            input_id.clone(),
        )
        .await?;

        // Store the request in storage
        self.storage
            .add_new_bonsai_proof_request(ProofRequestInformation {
                proof_request_id: bonsai_session_id,
                callback_proof_request_event: event,
            })
            .await?;

        if let Some(notifier) = self.notifier.clone() {
            notifier.notify_one()
        }

        info!(?input_id, "sent new callback event to bonsai");
        Ok(())
    }
}
```

Unless there is other mechanism that I do not know of, currently the only way to track an offchain proof would be to listen the blockchain and await the transaction, which would miss possible issues.

capossele:
Fixes #1006


nategraf (approved):
This looks good to me 👍

General question we should consider on the Relay more generally, are users of a relay server expected to need to talk only to the relay, or to the relay and to Bonsai directly? In this case, we are giving them the means to talk to Bonsai directly, which definitely seems like a good step. More generally, is this the expectation of the developer?

PR #1014, opened 2023-10-19
Default Executor with FileSegmentRef

### LINKED ISSUE #998: Replace SimpleSegmentRef with FileSegmentRef as default
To avoid out of memory conditions when running larger executions

capossele:
- Add a new `segment_path` option field to the `ExecutorEnv` struct. Currently this is only public to the crate and no public functions to modify it are provided. Its purpose is to keep the tempdir path in the same scope as the `ExecutorEnv`
- Executor `run` and `run_guest_only` use `FileSegmentRef` with a temporary directory to store segments
- tempfile dependency has been added as a consequence.

Resolves #998

### DIFF
--- a/risc0/zkvm/Cargo.toml
+++ b/risc0/zkvm/Cargo.toml
@@ -80,7 +80,7 @@ num-derive = "0.4"
 num-traits = { version = "0.2", default-features = false, optional = true }
 prost = { version = "0.12", optional = true }
 sha2 = { version = "0.10", optional = true }
-tempfile = "3"
+tempfile = { version = "3", optional = true }
 tracing = { version = "0.1", default-features = false, features = [
   "attributes",
 ] }
@@ -163,6 +163,7 @@ prove = [
   "dep:protobuf-src",
   "dep:rayon",
   "dep:sha2",
+  "dep:tempfile",
   "dep:typetag",
   "risc0-circuit-recursion/prove",
   "risc0-circuit-rv32im/prove",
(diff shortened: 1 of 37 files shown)


risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -241,7 +242,10 @@ impl<'a> ExecutorImpl<'a> {
     /// This will run the executor to get a [Session] which contain the results
     /// of the execution.
     pub fn run(&mut self) -> Result<Session, ExecutorError> {
-        self.run_with_callback(|segment| Ok(Box::new(SimpleSegmentRef::new(segment))))
+        let temp_dir = TempDir::new().map_err(|e| ExecutorError::Error(e.into()))?;
```
flaub, line 245: It seems like we only want to set the tempdir if the Option is None?

risc0/zkvm/Cargo.toml
```diff
@@ -80,6 +80,7 @@ num-derive = "0.4"
 num-traits = { version = "0.2", default-features = false, optional = true }
 prost = { version = "0.12", optional = true }
 sha2 = { version = "0.10", optional = true }
+tempfile = "3"
```
SchmErik, line 83: I think this will break the `no_std` test for the verifier. I suggest setting this as optional = true and then enabling only for the prove feature.

capossele: good catch!

SchmErik (approved):
Looks good to me!

Issue #995, opened 2023-10-17
Connect lift/join/identity to recursion zkvm


flaub:
The server needs to call the recursion functions from here:
https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/api/server.rs#L231

to here:
https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/recursion/prove/mod.rs#L70
and
https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/recursion/prove/mod.rs#L84

It also looks like `api.proto` and both client & server are missing an `identity_p254` which would connect to here:
https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/recursion/prove/mod.rs#L98

PR #1015, opened 2023-10-20
Server/Client: connect lift, join, and identityP254 to zkvm

SchmErik:
This PR adds support for lift/join/identityp254 to the client/server infrastructure so that it can be used by the zkvm and eventually r0vm.

Closes: #995


flaub (approved):
Nice

PR #1017, opened 2023-10-20
Fix warning about using features in patch for smartcore

nategraf:
Currently on `main` building the examples triggers a warning about the non-effect of setting features on the `smartcore` crate, in the `smartcore-ml` example. I looked into this and realized that additionally this example is using the patch mechanism when it could just be referencing the git dependency directly. I've modified the example to dependency directly, silencing the warning and avoiding unnecessary use of the patch mechanism.

### DIFF
--- a/examples/smartcore-ml/Cargo.toml
+++ b/examples/smartcore-ml/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2021"
 risc0-zkvm = { path = "../../risc0/zkvm" }
 rmp-serde = "1.1"
 serde_json = "1.0"
-smartcore = { git = "https://github.com/risc0/smartcore.git", features = ["serde"] }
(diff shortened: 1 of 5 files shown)


Roee-87 (approved):

examples/smartcore-ml/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition = "2021"
 risc0-zkvm = { path = "../../risc0/zkvm" }
 rmp-serde = "1.1"
 serde_json = "1.0"
-smartcore = { version = "0.3", features = ["serde"] }
+smartcore = { git = "https://github.com/risc0/smartcore.git", features = ["serde"] }
```
flaub, line 10: we should use a git reference

nategraf: I've not pinned to a particular git commit and added a note about why the git reference is used instead of using the published crate

flaub (approved):

Issue #470, opened 2023-03-24
Make METHOD_NAME parameter of rust-starter template configurable


mothran:
Reference: https://github.com/risc0/risc0/pull/453/files#r1148030108 

When a user creates a new rust-starter project from the template, we should provide a default option but make it a optional template param to change the METHOD_NAME file name + const var names.

PR #1032, opened 2023-10-24
 cargo-risczero new: add --guest-name option

SchmErik:
The guest name parameter is used as the name of the guest package
as well as method ID and ELF constants in the guest code. Therefore, this
parameter is required to be a valid rust identifier. If this is not set, the
`cargo risczero new` command will prompt the user to enter a guest name.
If no value is supplied by the command line parameter or the prompt, the
default value for this parameter will be "method".

closes: https://github.com/risc0/scratch/issues/102
closes: #470


templates/rust-starter/host/src/main.rs
```diff
@@ -1,7 +1,8 @@
-// TODO: Update the name of the method loaded by the prover. E.g., if the method
-// is `multiply`, replace `METHOD_NAME_ELF` with `MULTIPLY_ELF` and replace
-// `METHOD_NAME_ID` with `MULTIPLY_ID`
-use methods::{METHOD_NAME_ELF, METHOD_NAME_ID};
+// These constants represent the RISC-V ELF and the image ID generate by risc0-build.
```
flaub, line 1:
```suggestion
// These constants represent the RISC-V ELF and the image ID generated by risc0-build.
```

flaub (approved):

it09:
Do users need to rename `name = methods` to `name = project-name-methods` in the `methods/Cargo.toml` and in `host/Cargo.toml`, rename `methods = { path = "../methods" }` to `project-name-methods = { path = "../methods" } `

it09:
Do we need to add `serde::{from_slice, to_vec}` into `host/scr/main.rs`?

SchmErik:
> Do users need to rename `name = methods` to `name = project-name-methods` in the `methods/Cargo.toml` and in `host/Cargo.toml`, rename `methods = { path = "../methods" }` to `project-name-methods = { path = "../methods" } `

So there's a guest crate within the methods crate and this renames the guest so there's no need to change the methods

SchmErik:
> Do we need to add `serde::{from_slice, to_vec}` into `host/scr/main.rs`?

no, the `ExecutorEnvBuilder::write()`calls to_vec so that should be taken care of... I think we need from_slice but the example doesn't use it at this time.. we expect the user to add code like this to deserialize the journal

flaub:
I'm working on a PR to remove the need for `from_slice`.

PR #1033, opened 2023-10-25
docs(website): Add dev-mode instructions

### LINKED ISSUE #434: [DevEx] Document how to use Dev Mode feature

Cardosaum:
Closes #434


website/docs/faq.md
```diff
@@ -245,8 +245,34 @@ Because the data structures supporting all three of these need to match very car
...
+<details closed>
+<summary>
+Q:
+What is developer mode and how can I use it safely?
+</summary>
+A: We support a development-only mode for standalone risc0 projects in which code is executed but not proven. This adds efficiency to the development cycle during development stages where proving correct execution is not yet critical.
+
+A risc0 project in dev mode supports ([fake](https://docs.rs/risc0-zkvm/0.18.0/risc0_zkvm/enum.InnerReceipt.html#variant.Fake)) receipt creation and pass-through (fake) 'verification' workflows, so that dev mode may be switched on and off without impacting project workflows. In particular, receipts generated in dev mode still include public outputs written to the [journal](https://dev.risczero.com/terminology#journal).
+
+However, because the proving process is bypassed, receipts generated in dev mode will fail a standard receipt verification check. Only when the verifier is also built in dev mode will it perform pass-through 'verification' of the fake receipt.
```
nategraf, line 259: This flag is an environment variable checked at runtime rather than at build-time

website/docs/faq.md
```diff
@@ -245,8 +245,34 @@ Because the data structures supporting all three of these need to match very car
...
+<a class="anchor" id="dev-mode"></a>
+
+<details closed>
+<summary>
+Q:
+What is developer mode and how can I use it safely?
+</summary>
+A: We support a development-only mode for standalone risc0 projects in which code is executed but not proven. This adds efficiency to the development cycle during development stages where proving correct execution is not yet critical.
+
+A risc0 project in dev mode supports ([fake](https://docs.rs/risc0-zkvm/0.18.0/risc0_zkvm/enum.InnerReceipt.html#variant.Fake)) receipt creation and pass-through (fake) 'verification' workflows, so that dev mode may be switched on and off without impacting project workflows. In particular, receipts generated in dev mode still include public outputs written to the [journal](https://dev.risczero.com/terminology#journal).
```
nategraf, line 257: Suggested change of "workflow" to "function" so the word is not used twice in one sentence. Also, inserting a newline at the sentence break.
```suggestion
A risc0 project in dev mode supports ([fake](https://docs.rs/risc0-zkvm/0.18.0/risc0_zkvm/enum.InnerReceipt.html#variant.Fake)) receipt creation and pass-through (fake) 'verification' function, so that dev mode may be switched on and off without impacting project workflows.
In particular, receipts generated in dev mode still include public outputs written to the [journal](https://dev.risczero.com/terminology#journal).
```

website/docs/faq.md
```diff
@@ -245,8 +245,34 @@ Because the data structures supporting all three of these need to match very car
...
+Only projects built without this flag may run dev mode. Enabling dev mode requires also that the environment variable `RISC0_DEV_MODE` be set.
+
+As additional protection, if the dev mode [environment variable](https://docs.rs/risc0-zkvm/0.18.0/risc0_zkvm/fn.is_dev_mode.html) is present alongside a project built with the "disable-dev-mode" feature flag, the project will panic.
+
+For further reference, take a look at the table below. To learn more about usage, see the [risc0 project README](https://github.com/risc0/risc0/#readme). For a closer look at implementation, take a look at the [dev mode source](https://github.com/risc0/risc0/blob/v0.19.0-rc.1/risc0/zkvm/src/host/server/prove/dev_mode.rs).
+
+| Dev mode env var              | Feature flag off         | Feature flag on          |
+| ----------------------------- | ------------------------ | ------------------------ |
+| RISC0_DEV_MODE=true           | Developer mode activated | Prover panic             |
+| RISCO_DEV_MODE={false, unset} | Default project behavior | Default project behavior |
```
nategraf, line 272: Will the verifier panic when the `disable-dev-mode` flag is on? Also it's not clear just looking at this table which feature flag you are referring to. Directly referencing `diable-dev-mode` would be helpful to make this table self-contained.

website/docs/zkvm/quickstart.md
```diff
@@ -42,14 +42,27 @@ Don't worry -- each `TODO` comes with instructions for what you need to do, and
...
+### Running your project
+
+To run your project locally, going through the full proof generation process, run:
 
 ```bash
 cargo run --release
 ```
 
-> **Congratulations!** <br/>_That's all it takes to build and run a minimal RISC Zero application._
+#### Tip: Leveraging Developer Mode
```
SchmErik, line 55:
```suggestion
#### Tip for Quick Development: Leveraging Developer Mode 
```
I think it would make this easier to read if we tell the readers what they can accomplish using this mode

risc0/zkvm/src/host/receipt.rs
```diff
@@ -167,6 +167,15 @@ pub enum InnerReceipt {
...
     /// A fake receipt for testing and development.
+    ///
+    /// This receipt is not valid and will fail verification unless the
+    /// environment variable `RISC0_DEV_MODE` is set to `true`, in which case a
+    /// pass-through 'verification' will be performed, but it *does not*
+    /// represent any meaningful attestation of receipt's integrity.
+    ///
+    /// This type solely exists to improve development experience, for further
+    /// information about development-only mode see our [dev-mode
+    /// documentation](https://dev.risczero.com/tech_faq#dev-mode).
```
flaub, line 178: I think we should point to a dedicated page for dev-mode.

website/docs/faq.md
flaub: The website is now versioned, and so the `website/docs` directory is intended for the 'next' version that is unreleased (which will be 0.19). So we should not refer to `0.18` versions here.

flaub:
A reminder to rename the PR before landing. :)

website/docs/zkvm/quickstart.md
```diff
@@ -42,14 +42,27 @@ Don't worry -- each `TODO` comes with instructions for what you need to do, and
...
+
```
SchmErik, line 56:
```suggestion
Depending on the length of execution, the `cargo run --release` command can take a long time to run. This command spends the majority of its time generating valid proofs. While it is important to produce proofs, waiting for while running code can slow down iteration during development. Rather than waiting for proofs on each iteration, developers can disable proof generation for development purposes and re-enable proving when they are done with each iteration.
```

I think we should add something like this to frame the problem that developers might encounter and maybe edit the beginning of the next paragraph for a smoother transition

SchmErik (commented):
We discussed this a while ago but I think we should consider referring to "developer mode" as "dev-mode" throughout the document so that we have consistency in terminology

website/docs/zkvm/quickstart.md
Cardosaum: Completely agree, thanks for the input!

website/docs/faq.md
Cardosaum: Thanks for the suggestions, Victor!

Cardosaum: Also, thanks for pointing out a versioning issue here, Frank!

website/docs/faq.md
Cardosaum: It will panic, we test for this property [here](https://github.com/risc0/risc0/blob/70b637cb4cd31d22e370498f0fe8fa27fab4ff50/risc0/r0vm/tests/dev_mode.rs#L63-L70).

I changed the table to reference `disable-dev-mode` directly, thanks for pointing out!

website/docs/zkvm/quickstart.md
Cardosaum: Thank you for the suggestion, Erik!

Cardosaum:
> We discussed this a while ago but I think we should consider referring to "developer mode" as "dev-mode" throughout the document so that we have consistency in terminology

You're correct, thanks for calling that out. 
Addressed it in [this commit](https://github.com/risc0/risc0/pull/1033/commits/2f68ed92664d5289d30f505de509cc0c7bdb0dbb)

Cardosaum:
> A reminder to rename the PR before landing. :)

Oh, definitely! 
Done ✅

risc0/zkvm/src/host/receipt.rs
Cardosaum: Done! ✅

SchmErik (approved):
Looks good but before you merge, could you update the links to reflect the latest version? I see some `0.18`

flaub:
For prerelease, you can use `*`, see: https://docs.rs/about/redirections

PR #1036, opened 2023-10-25
Improve devex for Journal decoding

flaub:
* Replace journal bytes with a `Journal` type that has a `decode()`.
* Adjust release build to support building variants of `r0vm`.

Example:

```rust
let output: Output = from_slice(&receipt.journal).unwrap();
```

becomes:

```rust
let output: Output = receipt.journal.decode().unwrap();
```


SchmErik (approved):
Awesome! Between this and the new `write` functionality, I think it'll make things smoother for developers

tzerrell (approved):
Nice! 🎉

it09:
So smooth :))

PR #1044, opened 2023-10-26
template: add input to guest and write to journal from guest

SchmErik:


.github/workflows/main.yml
```diff
@@ -232,7 +232,7 @@ jobs:
             --guest-name test_method \
             template-test
         shell: bash
-      - run: cargo build --release --manifest-path ${{ runner.temp }}/template-test/Cargo.toml
+      - run: cargo run --release --manifest-path ${{ runner.temp }}/template-test/Cargo.toml
       - run: ${{ runner.temp }}/template-test/target/release/host
```
flaub, line 236: Do we need this line if we're now going to rely on `cargo run`?

flaub (approved):

PR #1047, opened 2023-10-27
feat: Add execution statistics in trace

### LINKED ISSUE #1027: Enable easy access to cycle count, segment count, and executor time stats from running dev-mode
A feature is proposed to simplify the process for users to view cycle count, segment count, and executor time statistics when running in development mode. By setting `RISC0_DEV_MODE=1` and `RUST_LOG="<module>=debug"`, a clear log of executor stats should be enabled. This feature aims to enhance the user experience by providing an easier way to obtain statistics about the program running.

Cardosaum:
Closes #1027


risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -246,28 +276,46 @@ impl<'a> ExecutorImpl<'a> {
...
+                segment_count: session.segments.len(),
+                execution_time: start.elapsed(),
+            };
+
+            tracing::debug!(total_cycles = ?execution_stats.total_cycles);
+            tracing::debug!(session_cycles = ?execution_stats.session_cycles);
+            tracing::debug!(segment_count = ?execution_stats.segment_count);
+            tracing::debug!(execution_time = ?execution_stats.execution_time);
+            tracing::debug!(raw_execution_stats = ?execution_stats);
+            tracing::debug!(%execution_stats);
```
Cardosaum, line 315: Added redundant traces in different lines so users can filter exactly what they want to show in the logs. 

In order to print this information selectively, it's necessary to set `RUST_LOG` in accordance with [Tracing Subscriber Filter Directives](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives)

Example: 

```bash
RUST_LOG=[{execution_time}],[{total_cycles}],[{session_cycles}],[{segment_count}],[{raw_execution_stats}],[{execution_stats}]
```

Although we have the flexibility to print those logs only selecting specific fields, the recommended way for users to get the information would be to use only the `execution_stats` directive: 

```bash
$ RUST_LOG=[{execution_stats}] cargo run --release

2023-10-27T18:43:24.604781Z DEBUG run_with_callback: risc0_zkvm::host::server::exec::executor: execution_stats=Execution Statistics:
Total Cycles: 33981
Session Cycles: 23252
Segment Count: 1
Execution Time: 3.492208ms
```

Cardosaum: @mothran , @SchmErik , I'd like to ask your opinion on the data retrieval, as I'm not sure if I'm obtaining it appropriately. 

Could you please confirm or suggest changes?

flaub: I'd like to avoid using `tracing::debug` until we consistently do so across the entire codebase in one NFC PR. For now, would it be OK to just use `log::debug`?

Cardosaum: Oh definitely! Thanks for the quick reply, I'll change it now ✅

Cardosaum: @flaub , I changed from `tracing` to `log` (and `tracing_subscriber` to `env_logger`). 

It seems we lose the ability to filter the logs more accurately, though. 
What I did was to add a prefix for the logs we want to show, I'm not aware of a better way to do it.

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -257,24 +287,43 @@ impl<'a> ExecutorImpl<'a> {
...
+                segments
+                    .into_iter()
+                    .for_each(|s| guest_session.segments.push(s));
+                guest_session.journal = journal;
+                Err(guest_session.into())
             }
             _ => Ok(guest_session),
+        };
+
+        if log::log_enabled!(target: "RISC0_EXECUTION_STATS", log::Level::Debug) {
```
SchmErik, line 308:
```suggestion
        if log::log_enabled!(target: "RISC0_EXECUTION_STATS", log::Level::Info) {
```

Cardosaum: Oh, good catch, I forgot to update it. 
Thanks!

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -257,24 +287,43 @@ impl<'a> ExecutorImpl<'a> {
...
+                    .for_each(|s| guest_session.segments.push(s));
+                guest_session.journal = journal;
+                Err(guest_session.into())
             }
             _ => Ok(guest_session),
+        };
+
+        if log::log_enabled!(target: "RISC0_EXECUTION_STATS", log::Level::Info) {
+            if let Ok(ref session) = result {
+                let execution_stats = ExecutionStats {
```
mothran, line 310: Why have a struct for these stats if we just print them after constructing the struct?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -257,24 +287,43 @@ impl<'a> ExecutorImpl<'a> {
...
+                segments
+                    .into_iter()
+                    .for_each(|s| guest_session.segments.push(s));
+                guest_session.journal = journal;
+                Err(guest_session.into())
             }
             _ => Ok(guest_session),
+        };
+
+        if log::log_enabled!(target: "RISC0_EXECUTION_STATS", log::Level::Info) {
```
mothran, line 308: Any reason to use a dedicated var vs `RUST_LOG="executor=info"` syntax?

templates/rust-starter/README.md
```diff
@@ -23,6 +23,16 @@ cargo run
...
 the code).
 
+### Executing the project locally in development mode
+
+During development, faster iteration upon code changes can be achieved by leveraging [dev-mode], we strongly suggest activating it during your early development phase. Furthermore, you might want to get insights into the execution statistics of your project, and this can be achieved by specifying the environment variable `RUST_LOG=RISC0_EXECUTION_STATS` before running your project.
+
+Put together, the command to run your project in development mode while getting execution statistics is:
+
+```bash
+RUST_LOG=RISC0_EXECUTION_STATS RISC0_DEV_MODE=1 cargo run
```
mothran, line 33: Can you still set your logging level when using `RISC0_EXECUTION_STATS` ?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -257,24 +287,43 @@ impl<'a> ExecutorImpl<'a> {
...
+            if let Ok(ref session) = result {
+                let execution_stats = ExecutionStats {
+                    total_cycles: self.total_cycles(),
+                    session_cycles: self.session_cycle(),
+                    segment_count: session.segments.len(),
+                    execution_time: start.elapsed(),
```
mothran, line 314: .elapsed() should be called right after the executor completes. Maybe move it to line 294?

mothran (changes requested):

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: The struct creation was just to enable logging of all stats in a single log line if the user so desires, so they could print logs for specific items or print all of them together. (I can drop the struct as well and print individual items in a single line as well, just though using a struct could make sense - open to dropping it though)

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: Oh I see, yeah I can definitely move it.

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: It's not a dedicated var, it's actually a string value we should pass. similar to `RUST_LOG="executor=info"`. I chose a longer value though as originally, when dealing with `tracing` filters we were matching module names as well (and executor was one of them). 

Moving to `env_logger` we can use `executor` as a shorter filter though :)

templates/rust-starter/README.md
Cardosaum: Yes, it's possible. And I changed to conform to your suggestion of `RUST_LOG="executor=info"` :)

carterbrett:
Reminder: Add callout of how to do this in step 6 of zkVM quickstart

templates/rust-starter/README.md
```diff
@@ -88,6 +98,7 @@ We'd love to hear from you on [Discord][discord] or [Twitter][twitter].
 [cargo-risczero]: https://docs.rs/cargo-risczero
 [crates]: https://github.com/risc0/risc0/blob/main/README.md#rust-binaries
 [dev-docs]: https://dev.risczero.com
+[dev-mode]: https://dev.risczero.com/zkvm/dev-mode
```
flaub, line 101:
```suggestion
[dev-mode]: https://dev.risczero.com/api/zkvm/dev-mode
```

flaub (approved):

templates/rust-starter/README.md
Cardosaum: Thanks!

Cardosaum:
@flaub , with this change we'll also need to bump `r0vm` and have users to update it using `cargo binstall`, as the logs will not work otherwise (would work for `--feature prove` - so, when running locally, but not when using `ipc`).

flaub:
Does this PR include the fix for forwarding the logging environment to the server? I didn't see it here, but maybe I'm missing something.

Cardosaum:
> Does this PR include the fix for forwarding the logging environment to the server? I didn't see it here, but maybe I'm missing something.

So, what happens is that we already pass all the defaults as we should when we call `.spawn()`, but we do need to have a new version of the `r0vm` binary that contains the changes made in this PR, by doing this we'll be able to get the statistics as expected, hence [this previous comment](https://github.com/risc0/risc0/pull/1047#issuecomment-1792146714). 

For context, while doing the development, when executing the `hello-world` example with `RUST_LOG=executor=info RISC0_DEV_MODE=1 cargo r` I didn't get any logs from the executor, but when doing `RUST_LOG=executor=info RISC0_DEV_MODE=1 cargo r --feature prove` I did got the logs. This happens because the feature `prove` will contain the new code for the `executor.rs` file, whereas the one without specifying it will go through `ipc`, using the `r0vm` binary installed via `cargo binstall cargo-risczero@0.19.0` won't contain such changes as the binary was prior to them. 

I did manage to get the logs when going through `ipc` by running the examples with the following: 

```bash
cargo build --release --package risc0-r0vm --manifest-path risc0/r0vm/Cargo.toml
export RISC0_SERVER_PATH=target/release/r0vm
RUST_LOG=executor=info RISC0_DEV_MODE=1 cargo r
```

Basically, compiling the new `r0vm`, pointing to it by setting the `RISC0_SERVER_PATH` variable, and then executing the example again. 

That's why I said we'd need to bump the version for the `r0vm` in order to get the logs working :) 

cc/ @flaub

PR #1049, opened 2023-10-27
Fix missing LFS issue

flaub:
* Add smoke-test that can be used for e2e integration testing of binary installations


mothran (approved):
The map + reduce trick is nice!

PR #1050, opened 2023-10-30
fix: avoid blocking on stdio with no input

### LINKED ISSUE #1016: Executor hangs when guest attempts to read more input than is provided
When a guest in run in the Executor, and not enough input is provided, the process hangs. In particular this is caused when the guest calls `env::read()` and no more input is provided in the `ExecutorEnv` to pass to the guest.

I've not yet investigated the root cause, or have a suggestion of how to fix this. It also relates to a more general question of input handling, especially in the client-server model where it is perfectly valid for the client to provide to the server input to the guest in a streaming manner.

austinabell:
Closes #1016 

The workflow I tested was just a simple template program with an `env::read` added in. I wasn't able to replicate the issue of it hanging when not enough input provided, only when no input was provided, so maybe there is another workflow that I'm not seeing.

If it's intentional not to overwrite the stdin reader on empty input, possibly this API should be changed to express the difference between empty input and no overwriting.

As mentioned in the issue comment, it seems like https://github.com/risc0/risc0/blob/e48f779d2ce66b0af274ca4881c8da5df405cebe/risc0/zkvm/src/host/client/posix_io.rs#L37 is a bit out of place, and should either be documented why that's the default, or have more explicit initialization to avoid this ambiguity.

### DIFF
--- a/risc0/zkvm/src/host/client/env.rs
+++ b/risc0/zkvm/src/host/client/env.rs
@@ -89,13 +89,11 @@ impl<'a> ExecutorEnvBuilder<'a> {
     pub fn build(&mut self) -> Result<ExecutorEnv<'a>> {
         let inner = self.inner.clone();
 
-        if !inner.input.is_empty() {
-            let reader = Cursor::new(inner.input.clone());
-            inner
-                .posix_io
-                .borrow_mut()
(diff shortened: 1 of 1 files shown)


flaub:
Fixed in #1148

PR #1055, opened 2023-10-30
Initial implementation of proof composition [re-submission of #992]

nategraf:
This PR is a re-submission of #992, which was reverted in #1052 due to release sequencing consideration. This PR will be taken out of draft status when a `0.19` branch has been cut, and `main` is ready to accept changes for release in `0.20`.

Reverts risc0/risc0#1052


flaub:
I think this is now safe to land!

nategraf:
> I think this is now safe to land!

Would you go ahead and give this an approval?

flaub (approved):

PR #1063, opened 2023-10-31
Workaround for git lfs issue

flaub:


.github/workflows/release.yml
```diff
@@ -86,17 +36,23 @@ jobs:
...
+      - run: tar cv * | gzip -9 > ../cargo-risczero-$TARGET.tgz
+        working-directory: tmp/pkg
       - uses: svenstaro/upload-release-action@v2
         with:
+          overwrite: true
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           file: tmp/cargo-risczero-${{ env.TARGET }}.tgz
-          # tag: ${{ env.TAG }} # only for testing
+          tag: ${{ env.TAG }} # only for testing
+      - run: sleep 90
```
mothran, line 55: ?

.github/workflows/release.yml
```diff
@@ -6,77 +6,27 @@ on:
...
-      # - run: cargo run
-      #   working-directory: tools/smoke-test
-
-  macos:
-    runs-on: [self-hosted, prod, macOS, apple_m2_pro]
-    needs: release
+    runs-on: [self-hosted, prod, "${{ matrix.os }}", cpu]
+    strategy:
+      fail-fast: false
+      matrix:
```
mothran, line 21: Good use of matrix builds!

.github/containers/release/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN apt update && apt -y install curl libssl-dev pkg-config
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+WORKDIR /src
+
+FROM base AS builder
+COPY . .
+RUN --mount=type=cache,target=/root/.cache \
```
mothran, line 13: Generally its a good idea to include: `,id=<key_name>` other wise the `target=` param will be used as your cache key and you could have cache mixups if running on the same machine. But I think in this case its unlikely.

Unfortantly with the CI machines being short lived, the cache-mounts might not buy a ton of performance since the caches are blown away with the machine existing at end-of-day.

You might consider the s3 layer caching mode we are using in bonsai as a alternative, but it tends to throw out the cache a little too frequently.

mothran (approved):

.github/containers/release/Dockerfile
flaub: This was purely a thing for doing local development, it's not intended to work well in AWS at the moment. I might revisit this later once we figure out how to get S3/MemoryDB backend working for caching.

.github/workflows/release.yml
flaub: This is needed because otherwise the test that follows tries to download too soon and causes intermittent 404s. I don't know if `90` is long enough but seems to be more stable this way.

.github/workflows/release.yml
flaub: Only possible when avoiding the `container:` feature of github. Invoking docker directly avoided a whole mess of issues.

.github/workflows/release.yml
mothran: That make sense, might be worth a comment inline to that effect.

mothran (approved):

PR #1065, opened 2023-10-31
Support profiling without code changes via RISC0_PPROF_OUT env var

nategraf:
This PR provides support for generating `pprof` profiles of guest code, without any code changes to the guest or host. It accomplishes this by first piping `TraceEvent`s across the client-server connection and then automatically instantiating the profiler in the client when the `RISC0_PPROF_OUT` is set and the `profiler` feature is enabled.


flaub:
I think the goal should be to remove the need for  "prove" for all of the examples. To do this, we need to refactor the profiling so that it doesn't rely on "prove". Users should be able to enable profiling simply by setting an environment variable and this should happen thru the client.

flaub:
The issue that I see is that if we enable "prove" in any example, it will require all examples to bring in prove because of the way cargo unifies features on a per-workspace basis. This will then cause users to build the entire prover which is what we're trying to avoid with binary distributions.

nategraf:
@flaub, I took a stab at finishing up the piping of TraceEvent messages to the client, and indeed it looks like there wasn't much left that needed doing. If this looks right to you, I'll move forward with this implementation.

nategraf:
Hitting an odd panic on this profiling example now:
```
╰─ RUST_BACKTRACE=1 RISC0_PPROF_OUT=profiling.pprof cargo run --release
   Compiling fibonacci-methods v0.1.0 (/Users/victorgraf/risc0/risc0/examples/profiling/methods)
   Compiling risc0-zkvm v0.19.0-rc.4 (/Users/victorgraf/risc0/risc0/risc0/zkvm)
fibonacci-guest: Starting build for riscv32im-risc0-zkvm-elfmethods(build), risc0-zkvm(build)
fibonacci-guest:    Compiling risc0-zkvm v0.19.0-rc.4 (/Users/victorgraf/risc0/risc0/risc0/zkvm)
fibonacci-guest:    Compiling fibonacci-guest v0.1.0 (/Users/victorgraf/risc0/risc0/examples/profiling/methods/guest)
fibonacci-guest:     Finished release [optimized + debuginfo] target(s) in 0.43s
   Compiling profiling v0.1.0 (/Users/victorgraf/risc0/risc0/examples/profiling)
    Finished release [optimized + debuginfo] target(s) in 21.29s
     Running `/Users/victorgraf/risc0/risc0/examples/target/release/profiling`
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `4`: Guest requested more data than was available', risc0/zkvm/src/host/server/exec/syscall.rs:391:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:67:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/panicking.rs:229:5
   4: risc0_zkvm::host::server::exec::syscall::<impl risc0_zkvm::host::client::posix_io::PosixIo>::sys_read
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/syscall.rs:391:9
   5: <risc0_zkvm::host::client::posix_io::PosixIo as risc0_zkvm::host::server::exec::syscall::Syscall>::syscall
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/syscall.rs:318:13
   6: <alloc::rc::Rc<core::cell::RefCell<risc0_zkvm::host::client::posix_io::PosixIo>> as risc0_zkvm::host::server::exec::syscall::Syscall>::syscall
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/syscall.rs:334:9
   7: risc0_zkvm::host::server::exec::executor::ExecutorImpl::ecall_software
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/executor.rs:731:17
   8: risc0_zkvm::host::server::exec::executor::ExecutorImpl::ecall
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/executor.rs:562:32
   9: risc0_zkvm::host::server::exec::executor::ExecutorImpl::step
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/executor.rs:452:13
  10: risc0_zkvm::host::server::exec::executor::ExecutorImpl::run_guest_only_with_callback::{{closure}}
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/executor.rs:313:42
  11: risc0_zkvm::host::server::exec::executor::ExecutorImpl::run_guest_only_with_callback
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/executor.rs:362:25
  12: risc0_zkvm::host::server::exec::executor::ExecutorImpl::run_with_callback
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/server/exec/executor.rs:260:39
  13: risc0_zkvm::host::api::server::Server::on_execute
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/api/server.rs:289:23
  14: risc0_zkvm::host::api::server::Server::run
             at /Users/victorgraf/risc0/risc0/risc0/zkvm/src/host/api/server.rs:257:17
  15: risc0_r0vm::run_server
             at /Users/victorgraf/risc0/risc0/risc0/r0vm/src/lib.rs:176:5
  16: risc0_r0vm::main
             at /Users/victorgraf/risc0/risc0/risc0/r0vm/src/lib.rs:86:9
  17: core::ops::function::FnOnce::call_once
             at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Error: Child finished with: 101
```

This is surprising because the I/O code looks correct at first glance, and I'm not sure how I might have broken it.

flaub:
I was thinking we could keep the profiler on the server-side, so as to reduce the amount of stuff that users need to compile on the client-side. My thinking was that we could ask the server to do profiling and request that it dumps the results to the specified path.

nategraf:
> I was thinking we could keep the profiler on the server-side, so as to reduce the amount of stuff that users need to compile on the client-side. My thinking was that we could ask the server to do profiling and request that it dumps the results to the specified path.

Ok, ya that can definitely be done. My intuition would be to factor that as an asset request and add a field to the execution request to allow the client to ask for a profile? Do we still want to pipe TraceEvents back to the client when they request it?

nategraf:
Fixed the issue I had above, and believe I incidentally fixed `SliceIo` when running with client/server mode. 5dd5a6d55c7571ab89b85bd1ae502d44bd508cc3

I've confirmed that with the current implementation I can build profiles. I'll work on a refactor next.

nategraf:
@flaub I tried a a few different ways to pipe through all the pieces needed to request a receive a complete profile from the server. I never found a clean way to do it. The two biggest roadblocks were 1) the profiler requires the ELF, but we usually convert the ELF to a memory image before sending a request to the server 2) I want to make sure that profiles work with both `Prover::prove_elf` and `Executor::execute_elf`. Currently when proving, the server does not send any information about the session back to the client, so there is no good place to pipe back the profile data. It's also not obvious to me what happens when a client requests a profile, but the server can't provide it (i.e. not compiled with profile, or it did not receive the elf).

I threw together a version of this PR that handles the profile building on the client side, but does not require any code changes to get a profile, provided that `profiler` feature flag is enabled. The changes changes seem clean to me at least, and I tested it out, getting profiles from various programs including Waldo.

What do you think of this factoring?

nategraf:
Example usage:

```bash
RISC0_PPROF_OUT=profile.pprof cargo run --manifest-path=examples/profiling/Cargo.toml
```

flaub (approved):
I think this is fine for now. I think we can try to move it back to the server later, but this is a nice forward step for devs to enable profiling without having to rebuild.

nategraf:
Should we go ahead and make the `profiler` a `default` feature flag?

flaub:
Maybe we can make it default once we move it to the server.

examples/waldo/Cargo.toml
```diff
@@ -15,6 +15,7 @@ waldo-core = { path = "core" }
 waldo-methods = { path = "methods" }
 
 [features]
-cuda = ["risc0-zkvm/cuda"]
 default = []
-metal = ["risc0-zkvm/metal"]
+cuda = ["risc0-zkvm/cuda"]
+prove = ["risc0-zkvm/cuda"]
```
flaub, line 20: I don't understand this one

flaub:
Do we need to update the website docs for the next version?

examples/waldo/Cargo.toml
nategraf: In [e975275](https://github.com/risc0/risc0/pull/1065/commits/e97527580f96131e1eecd62cb78eff6468cc5134) I added a passthrough `profiler` feature flag to all the examples so it is easy for folks to try taking a profile of any of them. I saw that the `metal` is still there even though the feature is broken, so I removed it.

Happy to revert this change if it doesn't seem like the right call.

nategraf:
> Do we need to update the website docs for the next version?

With the optimization guide, that would be good. I am working on changes to the profiler docs, and will change the profiling example README, to go along with this.

flaub:
I'm thinking specifically of: https://dev.risczero.com/api/zkvm/developer-guide/profiling

nategraf:
I've changed the merge target for https://github.com/risc0/risc0/pull/1080 to be this branch. The plan is to fix the profiler doc tests in that PR, where the docs can be reviewed with this PR as the base. When finished, that PR will be merged here and then both into `main`.

examples/chess/Cargo.toml
```diff
@@ -12,6 +12,7 @@ serde = "1.0"
 shakmaty = "0.22"
 
 [features]
-cuda = ["risc0-zkvm/cuda"]
 default = []
-metal = ["risc0-zkvm/metal"]
+cuda = ["risc0-zkvm/cuda"]
+prove = ["risc0-zkvm/cuda"]
```
flaub, line 17:
```suggestion
prove = ["risc0-zkvm/prove"]
```

flaub: This issue appears to be duplicated in many examples

nategraf: Good catch. Fixed for all examples in 8b6f47e57f2fd9497532b4b1661075afd5a498b7

examples/Cargo.lock
```diff
@@ -2452,6 +2505,21 @@ dependencies = [
  "syn 0.11.11",
 ]
 
+[[package]]
+name = "metal"
```
flaub, line 2509: I'm surprised this is showing up for the first time in this file.

Issue #1035, opened 2023-10-25
Docs: write explainer or reference on zkVM execution environment


SchmErik:
This is a document that describes the zkVM's execution environment and the characteristics of the zkVM with descriptions of memory layout, ISA, and etc.

PR #1067, opened 2023-10-31
docs: add zkVM technical specification

SchmErik:
This is the beginning of a technical specification of the zkVM as an execution environment. In this document, I assume that the reader to understands systems programming terminology but I've added a reference to OsDevWiki as a resource for those who may not be familiar with the terminology.

closes: https://github.com/risc0/risc0/issues/1035


website/docs/zkvm/developer-guide/zkvm-specification.md
```diff
@@ -0,0 +1,56 @@
...
+terminates without executing exception handlers.
+
+## zkVM Memory Layout
+
+The following table summarizes the layout of the zkVM memory
+
+| Address Range             | Size   | Name          | Description                                                                                   |
+| ------------------------- | ------ | ------------- | --------------------------------------------------------------------------------------------- |
+| `0x00000000 - 0x000003ff` | 1 KB   | Invalid page  | This page of memory is invalid, so that dereferencing a NULL address will result in a failure |
+| `0x00000400 - 0x0BFFFFFF` | 192 MB | Guest Memory  | Contains code, heap, and stack of the guest program                                           |
```
flaub, line 45:
```suggestion
| `0x00000400 - 0x0BFFFFFF` | 192 MB | User Memory  | Contains code, heap, and stack of the guest program                                           |
```

website/docs/zkvm/developer-guide/zkvm-specification.md
```diff
@@ -0,0 +1,56 @@
+# zkVM Technical Specification
+
+Like many other execution environments, the zkVM it has unique characteristics
```
flaub, line 3:
```suggestion
Like many other execution environments, the zkVM has unique characteristics
```

website/docs/zkvm/developer-guide/zkvm-specification.md
```diff
@@ -0,0 +1,56 @@
...
+| Address Range             | Size   | Name          | Description                                                                                   |
+| ------------------------- | ------ | ------------- | --------------------------------------------------------------------------------------------- |
+| `0x00000000 - 0x000003ff` | 1 KB   | Invalid page  | This page of memory is invalid, so that dereferencing a NULL address will result in a failure |
+| `0x00000400 - 0x0BFFFFFF` | 192 MB | Guest Memory  | Contains code, heap, and stack of the guest program                                           |
+| `0x0C000000 - 0x0CFFFFFF` | 16 MB  | System Memory | Contains RISC-V registers and non-leaf Merkle tree nodes for the page table                   |
+| `0x0D000000 - 0x0DFFFFFF` | 16 MB  | Page Table    | Contains the Page Table                                                                       |
+
+## zkVM Memory Model
+
+The zkVM executes instruction in the order they appear in the ELF. In other
```
flaub, line 51:
```suggestion
The zkVM executes instructions in order; in other
```

flaub (approved):
Small suggestions, but otherwise LGTM.

I think we should move this to `api` so that it gets versioned. Most of this content is not version-specific. However the memory map might change over time and it'd be nice to expand this document with say the syscall interface which would also be version-specific.

We also need to add a sidebar entry to link to this.

SchmErik:
> Small suggestions, but otherwise LGTM.
> 
> I think we should move this to `api` so that it gets versioned. Most of this content is not version-specific. However the memory map might change over time and it'd be nice to expand this document with say the syscall interface which would also be version-specific.

Good call. Yeah, I was getting this out as a first cut and I was going to add the syscalls in a follow-up PR

> 
> We also need to add a sidebar entry to link to this.

Will do!

PR #1070, opened 2023-10-31
docs website updates

flaub:
* Upgrade docusaurus to 3.0
* Use multi-instance to split versioned from unversioned content
* Split unversioned files into resources directory
* Snap version 0.19


tzerrell (commented):
Great! I'm not sure if this is just a Vercel error, but it looks like the root page gives a 404 -- i.e., going to plain [dev.risczero.com](dev.risczero.com) would return a 404, so perhaps something was lost in setting a "main" landing page.

tzerrell (commented):
While the preview link still gives a 404, I think it may be an artifact of how it previews; it redirects from https://risc0.github.io/website-preview/risc0/risc0/pr/1070 to https://risc0.github.io/api, and the latter doesn't exist. However, https://risc0.github.io/website-preview/risc0/risc0/pr/1070/api _does_ exist, and hopefully better models what this would act like in production.

flaub:
Note that the Vercel preview does work as expected. It looks like the ghpages preview isn't being rebuilt on each push to the PR branch.

tzerrell (approved):

PR #1079, opened 2023-11-01
Drop prover feature from bonsai-rest-api-mock

capossele:


bonsai/rest-api-mock/Cargo.toml
```diff
@@ -7,16 +7,13 @@ license = { workspace = true }
...
-release = false
-
 [dependencies]
 anyhow = "1.0"
 axum = { version = "0.6", features = ["macros", "headers"] }
 bincode = "1.3"
 bonsai-sdk = { workspace = true }
 hex = "0.4"
-risc0-zkvm = { workspace = true, features = ["prove"] }
+risc0-zkvm = { workspace = true, default-features = false, features = ["client"] }
```
flaub, line 16:
```suggestion
risc0-zkvm = { workspace = true }
```

bonsai/Cargo.toml
```diff
@@ -9,12 +9,12 @@ homepage = "https://risczero.com/"
...
 [workspace.dependencies]
-bonsai-ethereum-contracts = { path = "ethereum" }
-bonsai-ethereum-relay = { default-features = false, path = "ethereum-relay" }
-bonsai-rest-api-mock = { default-features = false, path = "rest-api-mock" }
-bonsai-sdk = { default-features = false, path = "sdk" }
-risc0-build = { default-features = false, path = "../risc0/build" }
-risc0-zkvm = { default-features = false, path = "../risc0/zkvm" }
+bonsai-ethereum-contracts = { version = "0.1.0", path = "ethereum" }
+bonsai-ethereum-relay = { version = "0.1.0", default-features = false, path = "ethereum-relay" }
+bonsai-rest-api-mock = { version = "0.1.0", default-features = false, path = "rest-api-mock" }
```
flaub, line 14: Now that we plan to publish these, it would be convenient to keep them aligned to whatever `bonsai-sdk` has.

.github/workflows/bonsai.yml
```diff
@@ -40,6 +40,15 @@ jobs:
...
         uses: risc0/foundry-toolchain@2fe7e70b520f62368a0e3c464f997df07ede420f
 
+      - name: Install cargo-install
+        uses: risc0/cargo-install@v1
+        with:
+          crate: cargo-binstall
+          version: "1.4"
+
+      - name: Cargo binstall cargo-risczero
+        run: cargo binstall -y cargo-risczero
```
flaub, line 50:
```suggestion
        run: cargo binstall -y cargo-risczero@0.19.0
```

flaub: Actually, I think we should use the current version from source instead of the latest published one so that we can detect breaking changes in CI.

flaub (approved):

PR #1080, opened 2023-11-01
Add Guest Optimization Guide to docs

nategraf:
This PR includes a new Guest Optimization Guide for developers.
See [./website/docs/zkvm/developer-guide/optimization.md](/risc0/risc0/tree/victor/optimizaton-docs/website/docs/zkvm/developer-guide/optimization.md)

For a full local preview run `yarn && yarn start` in the website directory. Make sure to select version as next in the site options on the page.


nategraf:
Currently marked as draft because the profiling instructions need to be adjusted once the profiling examples are fixed https://github.com/risc0/risc0/pull/1065. Everything else is ready for review.

website/docs/zkvm/developer-guide/optimization.md
```diff
@@ -0,0 +1,436 @@
...
+
+**The zkVM is essentially a CPU.**
+
+In particular, it is an implementation of the [RISC-V architecture] (specifically riscv32im), similar to how your laptop’s CPU may be an implementation of an [x86] or [ARM] architecture.
+This biggest difference is that the zkVM is implemented with [arithmetic circuits], in software, instead of circuitry made from silicon and copper.
+
+[RISC-V architecture]: /reference-docs/about-risc-v
+[x86]: https://en.wikipedia.org/wiki/X86
+[ARM]: https://en.wikipedia.org/wiki/ARM_architecture_family
+[arithmetic circuits]: https://dev.risczero.com/proof-system/stark-by-hand
```
pdg744, line 23:
```suggestion
[arithmetic circuits]: /reference-docs/about-arithmetic-circuits
```

website/docs/zkvm/developer-guide/optimization.md
```diff
@@ -0,0 +1,436 @@
...
+
+### What is a “cycle”?
+
+Both in the zkVM and on physical CPUs, the cost of an operation is measured in “clock cycles”.
+
+Intuitively, a "clock cycle" is the smallest unit of time in a CPU's operation, representing one tick of the CPU's internal clock and the time it takes to execute a basic CPU operation (e.g. adding two integers).
+We’ll refer to this as a “cycle”. [^1]
```
pdg744, line 30: hmm... this footnote doesn't seem to be working in the preview

website/docs/zkvm/developer-guide/optimization.md
```diff
@@ -0,0 +1,436 @@
...
+
+**Proving times for the zkVM are directly related to the number of cycles in an execution.**
```
pdg744, line 32: Seems like a nice spot to include a link to our datasheet for more information

website/docs/zkvm/developer-guide/optimization.md
```diff
@@ -0,0 +1,436 @@
...
+If you are new to optimization, or new to Rust, we recommend you read this guide.
+
+[The Rust Performance Book]: https://nnethercote.github.io/perf-book/
+
+### Don’t assume, measure.
+
+Performance is complex, in the zkVM as on a physical CPU.
+Don’t assume you know what the bottlenecks are. Measure and experiment.
+
+If you make a function 100x faster (or ∞x faster) but it only takes 1% of your execution time, you’ll see a &lt;1% improvement in performance.
```
pdg744, line 49: The `<` felt confusing to me on first read. 
```suggestion
If you make a function 100x faster (or ∞x faster) but it only takes 1% of your execution time, you’ll see less than 1% improvement in performance.
```

website/docs/zkvm/developer-guide/optimization.md
```diff
@@ -0,0 +1,436 @@
...
+
+### Measuring by printing to console
+
+Starting simple, measure by adding an `eprintln!` line to your guest code to measure how long an operation takes, and how many times it is called.
+
+Using [`env::get_cycle_count()`] will tell you the current number of execution cycles that have occurred in your program.
+
+As an example:
+
+```rust title="methods/guest/src/main.rs"
```
pdg744, line 60: In the past we were including comments with versioning tags like `v0.19` to make it easier to handle version maintenance via `ctrl-f`. @flaub may have other thoughts on how to handle tagging code to help manage maintenance

pdg744:
Fantastic work, Victor! I'm very excited to see how comprehensive of a guide this turned out to be! I made a handful of minor in-line suggestions

website/docs/zkvm/developer-guide/optimization.md
nategraf: Oops, I forgot to add a `:` after the footnote target. Fixed in b030c12f8c4fd7c7d6984adb533f1a5c7500842f

website/docs/zkvm/developer-guide/optimization.md
nategraf: My current understanding is we do not have a version of that sheet that we want to share widely right now. If we do, I definitely want to link to it.

capossele (approved):
Awesome!

website/api/zkvm/developer-guide/optimization.md
```diff
@@ -0,0 +1,441 @@
...
+
+In particular, it is an implementation of the [RISC-V architecture] (specifically riscv32im), similar to how your laptop’s CPU may be an implementation of an [x86] or [ARM] architecture.
+This biggest difference is that the zkVM is implemented with [arithmetic circuits], in software, instead of circuitry made from silicon and copper.
+
+[RISC-V architecture]: /reference-docs/about-risc-v
+[x86]: https://en.wikipedia.org/wiki/X86
+[ARM]: https://en.wikipedia.org/wiki/ARM_architecture_family
+[arithmetic circuits]: /reference-docs/about-arithmetic-circuits
+
+### What is a “cycle”?
```
tzerrell, line 25: Maybe we should add a link to this section as a "see also" in the "clock cycle" entry on the Terminology page.

tzerrell (commented):
This is great!

website/api/zkvm/developer-guide/optimization.md
nategraf: I'm not sure that would improve clarity. I worry the reader would click the link, and be confused to see the same definition repeated. (I also don't want to only have a link, and no inline definition, because this is a really core concept and I don't want them to miss it)

examples/profiling/README.md
```diff
@@ -1,59 +1,91 @@
...
+Use the environment variable `RISC0_PPROF_OUT` to set to the desired output path for the profiling data.
 
 ### Step 3: Visualization
-To visualize the profiling data using `pprof`, run:
+
+To visualize the profile using `pprof`, run:
+
 ```bash
-pprof -http=127.0.0.1:8089 ../target/riscv-guest/riscv32im-risc0-zkvm-elf/release/fibonacci profile.pb
+go tool pprof -http=127.0.0.1:8000 profile.pprof
```
flaub, line 50: This won't work unless someone has `go` installed.

I think we should avoid bringing in go, are there other installation methods that don't require go?

SchmErik: This seems useful: https://github.com/tikv/pprof-rs

nategraf: I've been searching for a way to use pprof without installing Go, but can't find one that seems reasonable [1]. The method recommended in this doc before was to use `go install github.com/google/pprof@latest`, which then puts the `pprof` binary in `PATH`, but that is an extra step beyond simply installing Go, which already bundles pprof in.

[1] There exists a package _someone_ maintains called `gperftools` that includes it... but it's not widely used

nategraf: We could build and bundle it with `cargo-risczero` binstall

website/api/zkvm/developer-guide/profiling.md
```diff
@@ -1,192 +1,96 @@
...
 ```
 
 ### Step 2: Running
 
-Then, run the example with:
+Run the Fibonacci profiling example with:
 
 ```bash
-RISC0_PPROF_OUT=./profile.pb cargo run --release
+RISC0_PPROF_OUT=./profile.pprof cargo run
```
flaub, line 42: I wonder why we're dropping `--release`

nategraf: It's not needed in the client-server scenario and `cargo run` is faster when not compiling in `release` mode. I can add it back though. I can definitely see the downside that if `prove` gets included, then building in debug mode would be an issue.

website/api/zkvm/developer-guide/optimization.md
tzerrell: I was proposing a link in the other direction: from Terminology to this article. In that direction, it seems like having this long performance article is giving a "why you care" context to the cycle definition.

nategraf: Ah yes! I went to add it but realized that the issue we face is that this doc will land on the `next` API version. I don't think we want to link from Terminology to docs under `next`, so I think we should add that link when the page is under a released version.

Added the line as a comment in 84b3c95d3aa9038ed5739d9c1f28eb80d1d9cdde

website/api/zkvm/developer-guide/profiling.md
flaub: Gotcha, yes linking can take a lot longer with `--release`. I'm fine with this, thanks.

examples/profiling/README.md
flaub: I suppose this is fine given the context, maybe we should add a quick blurb about installing go.

nategraf: > This seems useful: https://github.com/tikv/pprof-rs

I don't think this crate can help us. It's a way to add profiling outputs to Rust programs by including the profiler in the binary. As it is designed, this would need to compiled into the guest binary, and would try to use Linux syscalls from within the guest to create a profile. So this tool won't generate a profile.

What we actually need is the `pprof` tool binary on the users machine. The tool is what does the visualization. It you look at `pprof-rs` docs, it actually points the user to use `pprof` installed via Go for visualization https://github.com/tikv/pprof-rs#use-with-pprof

nategraf: Oh, I just noticed (from the `pprof-rs` page) that the `gperftools` package is the old C++ implementation of pprof, which I don't think is being maintained anymore. IIUC, it also does not support the `http` server for interactively exploring profiles.

examples/profiling/README.md
```diff
@@ -1,59 +1,91 @@
...
-```
-
 ### Step 2: Running
-Then, run the example with:
+
+Run the Fibonacci profiling example with:
+
 ```bash
-RISC0_PPROF_OUT=./profile.pb cargo run --release
+RISC0_PPROF_OUT=./profile.pprof cargo run
```
capossele, line 39: Is there a reason to go with the extension .pprof? I originally used .pb as it is the same extension (aside from being .gz) used by pprof (https://github.com/google/pprof#generate-a-graph-in-an-svg-file-and-open-it-with-a-web-browser)

nategraf: I just thought it sounded good. I'll go with `.pb` instead 👍 
9e8f10032be5b38e6dde7d2cf8c53fed4b8b52a9

tzerrell (approved):

PR #1089, opened 2023-11-02
Fix benchmarks

### LINKED ISSUE #1013: Re-enable Benchmarks in CI which are broken in #992
Changes to the `Session`, `Receipt` and `ExecutorEnv` types in #992 broke the benchmarks uses to produce the datasheet. After an initial attempt to quickly fix them, and talking to @capossele, we decided it would be best to land #992 with the benchmarks disabled in CI, and allow @capossele to work on refactoring and then re-enabling the benchmarks in CI next week.

capossele:
Refactor benchmarks after the changes introduced by #1055 

Resolves #1013


benchmarks/Cargo.toml
```diff
@@ -42,6 +42,10 @@ zeth-guests = { git = "https://github.com/risc0/zeth", package = "zeth-guests" }
 zeth-lib = { git = "https://github.com/risc0/zeth", package = "zeth-lib" }
 zeth-primitives = { git = "https://github.com/risc0/zeth", package = "zeth-primitives" }
 
+[patch.crates-io]
+risc0-build = { path = "../risc0/build" }
+risc0-zkvm = { path = "../risc0/zkvm" }
+
```
capossele, line 48: This is my attempt at patching risc0 dependencies of zeth

benchmarks/src/main.rs
```diff
@@ -129,7 +124,9 @@ fn main() {
         run_jobs::<sudoku::Job>(&cli.out, sudoku::new_jobs());
     }
 
-    if cli.command == Command::All || cli.command == Command::Zeth {
+    // TODO: re-enable once zeth gets updated
+    // if cli.command == Command::All ||
+    if cli.command == Command::Zeth {
```
capossele, line 129: Disabled zeth from the benchmark until we figure how to properly patch or zeth gets updated to a compatible version

capossele: Using an updated branch on zeth worked

flaub (approved):

PR #1095, opened 2023-11-03
[WIP] `zkp::prove::soundness`: Soundness error calculator

therealyingtong:


risc0/zkp/Cargo.toml
```diff
@@ -29,7 +29,7 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 tracing = { version = "0.1", default-features = false, features = [
   "attributes",
 ] }
-
+libm = "0.2"
```
flaub, line 32: This might cause issues for guest and no-std builds. I'm thinking we can make this dep optional and enable with the `soundness` flag.

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+            d: 1.0,
+            biggest_combo: 1.0,
+            ext_size: 4,
+            ext_field_size,
+            trace_domain_size,
+            lde_domain_size,
+            num_folding_rounds,
+        };
+
+        pub fn toy_proven(params: Params) -> f32 {
```
pdg744, line 240: I'm confused about this part. 
In my mind these are the relevant regimes: 

- proven UDR
- proven LDR
- toy problem conjecture
- conjecture 8.4 & conjecture 2.3

This `toy_proven` thing seems to be a 5th scenario that I think we can drop? 

If I understand correctly, ethSTARK 5.10.1 describes parameteres for the toy problem conjecture, and ethSTARK 5.10.2 describes parameters for proven soundness (no toy problem).

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
+use crate::{hal::Hal, taps::TapSet, FRI_FOLD, FRI_MIN_DEGREE, INV_RATE};
+use risc0_core::field::{baby_bear, ExtElem};
+
```
pdg744, line 3: Would be helpful to add a code comment giving a high-level orientation to what this file contains. something like: 

A soundness calculator for the RISC Zero STARK protocol that secures the RISC Zero zkVM. 

Soundness for STARK protocols can be analyzed under a number of different cryptographic assumptions. RISC Zero targets 100 bits of conjectured soundness, using the Toy Problem Conjecture. 

For completeness, we also include analysis in three other regimes:
- Conjectured soundness using Conjecture 8.4 from [Proximity Gaps] and Conjecture 2.3 from [DEEP-FRI]
- Proven soundness in the list-decoding regime
- Proven soundness in the unique-decoding regime

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+/// Johnson's bound
```
pdg744, line 4:
```suggestion
/// Johnson parameter. See https://eprint.iacr.org/2022/1216
```
m here is a parameter which controls proximity to the johnson bound

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+        let rho_plus = (params.trace_domain_size + params.biggest_combo) / params.lde_domain_size;
+        let m_plus = 1.0 / (params.biggest_combo * (alpha / rho_plus.sqrt() - 1.0));
+        let m_plus = m_plus.ceil();
+
+        (m_plus + 0.5) / rho_plus.sqrt()
+    };
+    soundness_error::<H>(taps, coeffs_size, theta, e_proximity_gap, l_plus)
+}
+
+pub fn conjectured<H: Hal>(taps: &TapSet, coeffs_size: usize) -> f32 {
```
pdg744, line 31: there are two relevant conjectured regimes -- one based on the toy problem and one based on conjecture 8.4 and conjecture 2.3. 

at least for right now, we're moving forward using the toy problem conjecture as our cryptographic foundations, so we should be careful to clarify that this is not "our" conjectured result

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        w_accum + w_code + w_data
+    };
+    // Max degree of the constraint system, i.e. no. of segment polys
+    // FIXME get from circuit instead of hard-coding
+    let d = 4.0;
+
+    let biggest_combo = taps.combos().map(|combo| combo.size()).max().unwrap() as f32;
+
+    // e, field extension degree
```
pdg744, line 120: is this a reference to the variable name in one of the other versions of this analysis?

pdg744: if so, i can revise the python notebook to use `ext_size`

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+}
+
+impl Params {
+    fn plonk_plookup_error(&self) -> f32 {
+        let n_columns = self.n_sigma_mem + self.n_sigma_bytes;
+        self.ext_size as f32 * n_columns as f32 * self.trace_domain_size / self.ext_field_size
+    }
+
+    /// (m + 1/2)^7 / (3 * sqrt(ρ)^3) * |D|^2 / |K|
```
pdg744, line 149: probably worth adding a citation for the math here

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
+use crate::{hal::Hal, taps::TapSet, FRI_FOLD, FRI_MIN_DEGREE, INV_RATE};
+use risc0_core::field::{baby_bear, ExtElem};
+
+/// Johnson's bound
+const M: f32 = 16.0;
+
+/// Rate
+const RHO: f32 = 1.0 / crate::INV_RATE as f32;
+
+/// η in Conjecture 8.4 [BCIKS21]
```
pdg744, line 10: Seems not immediately obvious how to track down a reference for BCIKS21. 

Maybe makes sense to use an eprint link instead? or i guess in this case mb [this link? ](https://dl.acm.org/doi/pdf/10.1145/3614423)

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+        let lde_domain_size = trace_domain_size * INV_RATE as f32;
+
+        let num_folding_rounds = num_folding_rounds(coeffs_size, ext_size);
+
+        let params = Params {
+            n_sigma_mem: 0,
+            n_sigma_bytes: 0,
+            n_trace_polys: 1.0,
+            d: 1.0,
+            biggest_combo: 1.0,
```
pdg744, line 232: hmm... this line seems surprising to me at first glance; will circle back to this...

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        fn toy_soundness_error(
+            params: Params,
+            theta: f32,
+            e_proximity_gap: f32,
+            l_plus: f32,
+        ) -> f32 {
+            let plonk_plookup_error = params.plonk_plookup_error();
+            let fri_error = params.e_fri(theta, e_proximity_gap);
+            let deep_ali_error = params.e_deep_ali(l_plus);
```
pdg744, line 279: It looks to me like the analysis here is more complicated than what is done in ethSTARK and miden's toy problem analysis. 

in particular, I don't believe the other projects include `deep_ali_error` in the analysis for the toy problem conjecture.

pdg744: from ethstark: image

pdg744: from miden: 
image

https://github.com/Al-Kindi-0/winterfell/blob/main/air/src/proof/mod.rs

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+            trace_domain_size,
+            lde_domain_size,
+            num_folding_rounds,
+        };
+
+        pub fn toy_proven(params: Params) -> f32 {
+            let e_proximity_gap = params.e_proximity_gap_proven();
+
+            // α = (1 + 1/2m) * sqrt(ρ)
+            let alpha = (1.0 + 1.0 / (2.0 * M)) * RHO.sqrt();
```
pdg744, line 244: lines 18-27 seem redundant with this part. 

does it make sense to consolidate to a function that computes the list size?

risc0/zkp/src/prove/soundness.rs
```diff
@@ -0,0 +1,292 @@
...
+        let ext_size = baby_bear::ExtElem::EXT_SIZE;
+        let field_size = baby_bear::P as f32;
+        let ext_field_size = libm::powf(field_size, ext_size as f32);
+        let trace_domain_size = (coeffs_size / ext_size) as f32;
+        let lde_domain_size = trace_domain_size * INV_RATE as f32;
+
+        let num_folding_rounds = num_folding_rounds(coeffs_size, ext_size);
+
+        let params = Params {
+            n_sigma_mem: 0,
```
pdg744, line 228: the main open question on my mind is about if/how to model the lookup/permutation argument in the toy problem analysis. 

would be nice to give some care to this question, as I think the question of whether or not to omit these arguments will be the difference between 97 bits and 100 bits of soundness.

risc0/zkp/src/prove/soundness.rs
therealyingtong, line 3:
```suggestion
/// A soundness calculator for the RISC Zero STARK protocol that secures the
/// RISC Zero zkVM. Soundness for STARK protocols can be analyzed under a
/// number of different cryptographic assumptions. RISC Zero targets 100 bits
/// of conjectured soundness, using the Toy Problem Conjecture. For
/// completeness, we also include analysis in three other regimes:
/// - Conjectured soundness using Conjecture 8.4 from [Proximity Gaps] and
///   Conjecture 2.3 from [DEEP-FRI]
/// - Proven soundness in the list-decoding regime
/// - Proven soundness in the unique-decoding regime

use crate::{hal::Hal, taps::TapSet, FRI_FOLD, FRI_MIN_DEGREE, INV_RATE};
use risc0_core::field::{baby_bear, ExtElem};

```

risc0/zkp/src/prove/soundness.rs
therealyingtong, line 10:
```suggestion
/// η in Conjecture 8.4 of the Proximity Gaps paper [BCIKS21](https://eprint.iacr.org/2020/654.pdf)
```

risc0/zkp/src/prove/soundness.rs
therealyingtong, line 120:
```suggestion
```

risc0/zkp/src/prove/soundness.rs
pdg744: here's the numerical summary if we don't set n_sigma_mem and n_sigam_bytes to 0

image

flaub:
Superseded by #1593

PR #1096, opened 2023-11-03
Capture and forward server errors to clients

### LINKED ISSUE #1093: r0vm failures (guests faults in Executor) should report rich error data to the Client.
While running in bonsai the Client::execute() method's Result::Err contained:
```
Child finished with: 101
```

But the stderr from the r0vm output in the log:
```
thread 'main' panicked at risc0/r0vm/src/lib.rs:176:18:
called `Result::unwrap()` on an `Err` value: Guest panicked: source slice length (0) does not match destination slice length (32)
```

Being able to report that inner error to the Client would be great for Bonsai error reporting UX.

flaub:
Resolves #1093


mothran (approved):
LGTM

SchmErik:
> looks good to me! I'm not sure what in this PR is causing the benchmark check to fail though...

Oh, I'm assuming this is the broken benchmarks referenced by https://github.com/risc0/risc0/issues/1013

PR #1097, opened 2023-11-03
Extend Benchmark Metrics

RomanBoegli:
Notable changes:
- add peak physical/virtual memory to metrics (credits to [benchy](https://docs.rs/crate/benchy/latest/source/src/memory.rs)'s implementation) 
- add operations per second to metrics
- new method` Metrics.to_csv_row()`
- express durations in micro seconds instead nano seconds


RomanBoegli:
Maybe wise to wait for PR #1089 to be completed first.

PR #1099, opened 2023-11-03
Backporting commits for upcoming 0.19.1

flaub:


mothran (approved):
This looks like the two PRs I need. Which PRs are including in this backport?

Issue #1071, opened 2023-10-31
Docs: write docs on continuations


SchmErik:
Write a doc for continuations. Describe the problem, and how it's solved, etc. This is an internal feature that's enabled automatically but it's good to have a description. Target audience is people who might be familiar with other proving systems.

tzerrell:
It's not the same content as documentation, but my news post on this might be helpful to draw from in places when writing these docs: https://www.risczero.com/news/continuations

SchmErik:
Sorry for the noise - so I Brett asked me to write a short overview of continuations as a part of the zkvm overview and when I opened this, I was intending on this being a tracking issue for that. The PR for this small overview is #1101 but let's just use this issue to track the continuations document anyway.

SchmErik:
In order to build up the notion of continuations in a way that people can understand at a deeper level, here's a list of concepts used:

- cycles
- page fault

nategraf:
Do we believe this addressed by https://github.com/risc0/risc0/pull/1314 and https://github.com/risc0/risc0/pull/1274? Together, the example, terminology, and docs on the relevant Rust methods and types consistent the implemented documentation. The only other planned item of documentation is a study club talk.

tzerrell:
> Do we believe this addressed by https://github.com/risc0/risc0/pull/1314 and https://github.com/risc0/risc0/pull/1274?

Personally, between those and #1101 (and the blog post #1101 references) I think our docs here are probably adequate. Open to suggestions about what's not sufficiently covered if I'm wrong though.

SchmErik:
I think we could close this. @carterbrett what do you think?

SchmErik:
Looked more closely at @nategraf's PRs and I think this is sufficient to close. @carterbrett Feel free to reopen if there's more discussion needed though.

PR #1101, opened 2023-11-03
docs: zkvm overview: add short explanation of continuations

SchmErik:
This adds a short section that explains the zkVM's key features in the overview page. Continuations are one of the key features so this change adds a short overview of it and references an article. We'll fill out this section with more features such as composition.


website/api/zkvm/zkvm_overview.md
```diff
@@ -65,6 +65,15 @@ Once you've built a zkVM application, you can generate proofs on your own machin
 For local proof generation, we recommend a machine with at least 16GB of RAM.
 Options for generating proofs using GPU acceleration and skipping proof generation are documented in the [feature flags].
 
+## Key Features
+
+There are several key features that distinguish RISC Zero's zkVM from others. These features enable developers to more easily develop zk powerful applications.
```
tzerrell, line 70: It feels a bit awkward to allude to "several key features" and then only list one. I would either drop this sentence, or add another subsection, maybe for instance this bit adapted from our website?
> Micro Architecture
>
> The zkVM is a verifiable computer that works like a real embedded RISC-V micro-processor. Write ZK proofs just like any other code.

tzerrell (commented):
Seems like a good idea to include!

website/api/zkvm/zkvm_overview.md
SchmErik: good point... I was thinking that this would be the first heading of a few points but it does read a little awkward at this point..

website/api/zkvm/zkvm_overview.md
```diff
@@ -65,6 +65,15 @@ Once you've built a zkVM application, you can generate proofs on your own machin
...
+
+### Continuations for limitless computations
+
+Continuations allow the zkVM to do large computations by splitting the work into small, independent segments of the same size.
```
tzerrell, line 74: Just saw your comment on concepts in https://github.com/risc0/risc0/issues/1071#issuecomment-1793161092. If you wanted, you could link to the terminology page here.
```suggestion
Continuations allow the [zkVM] to do large computations by splitting the work into small, independent [segments] of the same size.
```
You would need to add `[zkVM]: /terminology#zero-knowledge-virtual-machine-zkvm` and `[segments]: /terminology#segment` to the end of the article too if you do this.

website/api/zkvm/zkvm_overview.md
tzerrell: yeah it makes sense as soon as we flesh this out with more points, but until that happens its weird. You could always drop for now and grab the language out of this PR again once we have more points ready.

website/api/zkvm/zkvm_overview.md
flaub, line 1: We should probably duplicate this over into the 0.19 version as well.

website/api/zkvm/zkvm_overview.md
```diff
@@ -99,10 +109,12 @@ Read the [article](https://risczero.com/news/zeth-release).
 [proofs]: /terminology#validity-proof
 [prover]: /terminology#prover
 [publicly shared outputs]: /terminology#journal
+[segments]: /terminology#segment
 [verify]: /terminology#verify
 [zero-knowledge virtual machine]: /terminology#zero-knowledge-virtual-machine-zkvm
 [zkvm]: https://github.com/risc0/risc0#readme
 [zkVM Quick Start]: ./quickstart.md
 [zkVM example applications]: https://github.com/risc0/risc0/tree/v0.18.0/examples
```
flaub, line 117: For `api/`, this link should point to `main`.
For `api_versioned_docs`, this should point to the corresponding version.

tzerrell (approved):

website/api/zkvm/zkvm_overview.md
flaub:
```suggestion
[zkVM example applications]: https://github.com/risc0/risc0/tree/main/examples
```

flaub (approved):

PR #1106, opened 2023-11-06
Fix a typo in guest-code-101.md regarding journal

weikengchen:
It seems that journal is where we put the public outputs, rather than the private outputs, based on the contexts?

This is an edit to guest-code-101.md.


tzerrell (commented):
Whoops! Good catch, you are correct!

This also needs to be fixed in the other versions of the docs (in `website/api_versioned_docs/version-0.18/zkvm/developer-guide/guest-code-101.md` and `website/api/zkvm/developer-guide/guest-code-101.md`) -- do you mind getting those changes as well? Or I can make a separate PR for those if needed.

weikengchen:
@tzerrell Added those changes.

tzerrell:
Thank you for both finding this and fixing it!

PR #1107, opened 2023-11-06
XGBoost example using Spice AI query to train a model.

Roee-87:


flaub:
> I am getting a strange error when I attempt `cargo run -r`: `thread 'main' panicked at 'called `Result::unwrap()`on an`Err`value: Journal digest mismatch detected', risc0/r0vm/src/lib.rs:176:18 note: run with`RUST_BACKTRACE=1`environment variable to display a backtrace thread 'main' panicked at 'called`Result::unwrap()`on an`Err`value: failed to fill whole buffer', xgboost/src/main.rs:62:54 note: run with`RUST_BACKTRACE=1` environment variable to display a backtrace`
> 
> I believe I made an error at some point when I updated the format to make it compatible with the other examples. When I run the exact same code using the starter template, it executes just fine. See here for reference: https://github.com/risc0/scratch/tree/ml-models/dev/rotstein/xgboost-example

examples/xgboost/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    // We transfer theinput  data, the rmp array length, and the serialized model over to the guest.
+    let env = ExecutorEnv::builder()
+        .write(&data)
+        .unwrap()
+        .write(&rmp_array_length)
+        .unwrap()
+        .write_slice(&rmp_model)
+        .build()
+        .unwrap();
```
flaub, line 52:
```suggestion
    // We transfer the input data and the model over to the guest.
    let env = ExecutorEnv::builder()
        .write(&data)
        .unwrap()
        .write(&model)
        .unwrap()
        .unwrap();
```

examples/xgboost/methods/guest/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+    let length: usize = env::read();
+
+    // create a buffer for the byte array
+    let mut model_bytes: Vec<u8> = vec![0;length];
+
+    // read in the byte array
+    env::read_slice(&mut model_bytes);
+
+    // We deserialize the byte array into the trained model.  
+    let xgboost_model: GradientBooster = rmp_serde::from_slice(&model_bytes).unwrap();
```
flaub, line 36:
```suggestion
    // We deserialize the byte array into the trained model.  
    let xgboost_model: GradientBooster = env::read();
```

examples/xgboost/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "xgboost"
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
+
+[dependencies]
+risc0-zkvm = { path = "../../../../risc0/zkvm", default-features = false, features = ["std"] }
+forust-ml = "0.4.2"
+rmp-serde = "1.1.2"
```
flaub, line 11: This can be dropped, we should always prefer to use the internal risc0 serde codec since it's optimized for the guest.

examples/xgboost/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    // We return the inference value committed the journal.
+    receipt.journal.decode().unwrap()
+}
+
+#[cfg(test)]
+mod test {
+    use forust_ml::GradientBooster;
+    use methods::XGBOOST_ELF;
+    use risc0_zkvm::{default_executor, ExecutorEnv};
```
flaub, line 72: These all appear to be unused.

Roee-87: deleted.

examples/xgboost/methods/guest/Cargo.toml
Roee-87: The issue with passing the GradientBooster struct to the guest is that it causes the following error:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: DeserializeBadBool', /Users/royrotstein/risczero/risc0/risc0/zkvm/src/guest/env.rs:415:54
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Guest panicked: called `Result::unwrap()` on an `Err` value: DeserializeBadBool', risc0/r0vm/src/lib.rs:186:18
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: failed to fill whole buffer', xgboost/src/main.rs:61:54
```

flaub (approved):

PR #1109, opened 2023-11-06
Adjust log verbosity

capossele:


mothran (approved):
LGTM, but I will defer to @flaub on logging levels for our internal zkvm components. 

For bonsai its nice to break off these modules to a lower level for our logging.

flaub (approved):

PR #1110, opened 2023-11-06
replace markdown links with reference links

SchmErik:
This was done by using a simple rust program that utilizes the regex create and it captures more patterns than the js program I found on github.


SchmErik:
I'm not currently working on this so I'm going to reopen when it's being actively worked on. Closing

PR #1113, opened 2023-11-07
Improve CI build/test perf

flaub:
* Use SHA-256 for internal proving tests (2x faster than poseidon)
* Avoid hanging cargo-risczero tests waiting on stdin
* Avoid building `risc0-zkvm` twice with and without the "prove" feature

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -129,9 +129,8 @@ jobs:
       - run: cargo run --bin cargo-risczero --no-default-features -- risczero install --version $RISC0_TOOLCHAIN_VERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      - run: cargo test -F $FEATURE -F profiler --workspace --exclude doc-test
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -129,9 +151,8 @@ jobs:
       - run: cargo run --bin cargo-risczero --no-default-features -- risczero install --version $RISC0_TOOLCHAIN_VERSION
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      - run: cargo test -F $FEATURE -F profiler --workspace --exclude doc-test
+      - run: cargo test -F $FEATURE -F fault-proof -F profiler -F prove --workspace --exclude doc-test
       - run: cargo test -p risc0-r0vm -F $FEATURE -F disable-dev-mode
-      - run: cargo test -p risc0-zkvm -F $FEATURE -F fault-proof -F prove
```
SchmErik, line 134: NICE! I feel like we should have caught something like this during code review. I'll try my best to be aware of something like this in the future

SchmErik: This as in this hunk and the previous one

PR #1115, opened 2023-11-07
sys_log: write to stdout

SchmErik:
Previously, sys_log printed the message via println!. This change writes sys_log to whatever stdout is pointing to so that the output can be captured in the same manner in as how user specifies stdout to be handled.

Closes: #1028


risc0/zkvm/src/host/server/exec/syscall.rs
```diff
@@ -624,4 +610,24 @@ impl<'a> PosixIo<'a> {
...
+        // write to stdout, but be sure to point it to where the file descriptor is pointing
+        let fd = ctx.load_register(REG_A3);
+        let writer = self
+            .write_fds
+            .get_mut(&fileno::STDOUT)
+            .ok_or(anyhow!("Bad write file descriptor {fd}"))?;
+
+        log::debug!("Writing {buf_len} bytes to file descriptor {fd}");
+
+        let msg = format!("R0VM LOG[{}] ", ctx.get_cycle().to_string());
```
flaub, line 627:
```suggestion
        let msg = format!("R0VM[{}] ", ctx.get_cycle().to_string());
```

In case we want to keep the same format as before.

risc0/zkvm/src/host/server/exec/syscall.rs
```diff
@@ -624,4 +610,24 @@ impl<'a> PosixIo<'a> {
         writer.borrow_mut().write_all(from_guest_bytes.as_slice())?;
         Ok((0, 0))
     }
+
+    fn sys_log(&mut self, ctx: &mut dyn SyscallContext) -> Result<(u32, u32)> {
+        let buf_ptr = ctx.load_register(REG_A3);
+        let buf_len = ctx.load_register(REG_A4);
+        let from_guest = ctx.load_region(buf_ptr, buf_len)?;
+        // write to stdout, but be sure to point it to where the file descriptor is pointing
+        let fd = ctx.load_register(REG_A3);
```
flaub, line 619: I don't understand this line or the comment. It looks like we're not using `fd` for anything

risc0/zkvm/src/host/server/exec/syscall.rs
SchmErik: yeah.. I could go either way. I'll keep it the same for now

flaub (approved):

Issue #842, opened 2023-09-06
tcp errors from relay indicate proofs are invalid


austinabell:
When a tcp error happens with a local bonsai process, the logs indicate that a proof has failed, even though the proof is successful and can be queried directly. This leads to the proof not being submitted back on-chain.

Logs from a failed proof (note I added a debug log to show the TCP error, that isn't in code):

```
2023-09-06T00:59:28.901646Z  INFO bonsai_rest_api_mock::prover: Receiver: ProverMessage::RunSession: { task: Task { session_id: "3b83c18b-f44a-48e6-9e7a-b676df270c01", image_id: "4eea7db067e793b36ef6ef3a8a4cc8f7f43730e658263c2d23130252e1452d13", input_id: "340a8e0e-88b6-432d-9a4c-3fce9f5c1785" } }
2023-09-06T00:59:28.901692Z  INFO bonsai_rest_api_mock::prover: Running task...
2023-09-06T00:59:28.901731Z  INFO bonsai_rest_api_mock::prover: Loading guest image form ELF file
2023-09-06T00:59:28.901728Z  INFO bonsai_ethereum_relay::downloader::proxy_callback_proof_processor: sent new callback event to bonsai input_id="340a8e0e-88b6-432d-9a4c-3fce9f5c1785"
2023-09-06T00:59:28.901769Z  INFO bonsai_ethereum_relay::uploader::pending_proofs::manager: processing new pending proof log_id=SessionId { uuid: "3b83c18b-f44a-48e6-9e7a-b676df270c01" }
2023-09-06T00:59:29.822253Z DEBUG bonsai_ethereum_relay::uploader::pending_proofs::manager: pending proof result: pending_proof_result=Err(ClientAPI { source: Bonsai(HttpErr(reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(8081), path: "/sessions/status/3b83c18b-f44a-48e6-9e7a-b676df270c01", query: None, fragment: None }, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })) })), id: SessionId { uuid: "3b83c18b-f44a-48e6-9e7a-b676df270c01" } })
[...]

PR #1118, opened 2023-11-07
fix: Add exponential backoff to bonsai proof polling

austinabell:
Currently, this future will essentially spin loop and spam requests to the service while the proof while waiting to be completed. This doesn't error for small proofs, but in my testing, this would consistently error if there was some combination of requests being sent too fast or there was a proof that took a long time to execute. This caused, on my OS, the error documented in the issue below as there were too many connections being opened.

Original issue: https://github.com/risc0/risc0/issues/842

Opening as a draft, as it's opinionated how this is fixed. I can also add an issue to this if it's helpful, but I figured in this case a hacky solution might make it more clear what the issue is.

What these changes are is just adding an exponential backoff to this request, which intuitively makes the most sense to me. Adding a `sleep` also works, but this is better to minimize latency for small proofs.

### DIFF
--- a/bonsai/Cargo.lock
+++ b/bonsai/Cargo.lock
@@ -402,6 +402,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "backoff"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b62ddb9cb1ec0a098ad4bbf9344d0713fa193ae1a80af55febcff2627b6a00c1"
+dependencies = [
+ "futures-core",
+ "getrandom 0.2.10",
+ "instant",
+ "pin-project-lite",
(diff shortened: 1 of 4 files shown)


Cardosaum (approved):
Thank you, @austinabell! 

Your changes make sense and should help to minimize latency for small proofs and prevent the system from being overwhelmed by too many requests. I’ve added an upper limit to the exponential interval to ensure that the backoff doesn’t become too large. This should provide a balance between reducing spam requests and maintaining reasonable latency. Let’s continue to monitor and adjust as necessary. Your efforts are greatly appreciated!

flaub (changes requested):
I'd like to get some CI coverage before we let this land.

flaub:
@austinabell Could you update the `Cargo.lock` file and push again?

PR #1123, opened 2023-11-08
Minor revisions

tzerrell:
A few minor revisions from going through the code in the hackathon. These are 3 unrelated changes and they can be decoupled if there are problems with one of them.

### DIFF
--- a/templates/rust-starter/host/src/main.rs
+++ b/templates/rust-starter/host/src/main.rs
@@ -36,7 +36,7 @@ fn main() {
     // For example:
     let _output: u32 = receipt.journal.decode().unwrap();
 
-    // Optional: Verify receipt to confirm that recipients will also be able to
-    // verify your receipt
+    // The receipt was verified at the end of proving, but the below code is an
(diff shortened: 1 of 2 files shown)


templates/rust-starter/host/src/main.rs
```diff
@@ -22,7 +22,7 @@ fn main() {
     // ExecutorEnvBuilder::build().
 
     // For example:
-    let input: u32 = 15*2^27 + 1;
+    let input: u32 = 15 * u32::pow(2, 27) + 1;
```
flaub, line 25: Does this fail to compile? If so, I wonder why CI didn't catch it

tzerrell: Compiles just fine! It just is `2`, not `2013265921`, because `^` is bitwise xor rather than pow

flaub: Haha, I see, good call!!

flaub (approved):

PR #1124, opened 2023-11-08
Fix cargo publish for bonsai-ethereum-contracts

flaub:
Adjust `forge build` so that output goes to the output directory. cargo publish disallows writing to the source tree in a build script.


mothran (approved):
lgtm

PR #1128, opened 2023-11-08
Improve zkvm Hello World tutorial & Rust Starter Template

agryaznov:
Hello RISC Zero Team!

Going through the [zkVM Hello World Tutorial](https://dev.risczero.com/api/zkvm/tutorials/hello-world) I tried to improve a couple of things to provide a smoother learner experience:

1. Updated Rust Starter Template to 
   + Do not commit *input* into the *journal* as it appears to give a vicious practice example:
      that *input* is basically the only thing which should be kept in secret, right?
   + Generate some `rustdoc` which might help to understand it better.

2. Updated the Tutorial and fixed a couple of broken links in Quickstart.


risc0/cargo-risczero/src/commands/new.rs
```diff
@@ -138,11 +138,10 @@ impl NewCommand {
...
             Some(name) => name.clone(),
             None => {
-                eprint!(
-                    "Guest name was not supplied through the --guest-name option. Please enter\x20\
-                    package name for your template or press [enter] to use default guest package\x20\
-                    name \"method\"\n\
-                    Enter package name > "
+                eprint!("Guest name was not supplied through the --guest-name option. Please enter \
+                         package name for your template or press [enter] to use default guest package \
+                         name \"method\".\n\
```
tzerrell, line 143: We just had problems with whitespace not rendering with lines ending with ` \` (see #1123), does this revision work correctly or does it display as `enterpackage` and `packagename`?

tzerrell (commented):
@agryaznov thank you!

@it09 I've only briefly skimmed this as it's much more your domain than mine, but let me know if you would like me to do a careful review.

risc0/cargo-risczero/src/commands/new.rs
agryaznov: That was the reason I changed it. Now it looks good.

flaub:
Thank you, we'll take this into consideration in the next improvements made to the template.

PR #1141, opened 2023-11-11
Fix hello world example links

estensen:
Current links returns 404


flaub:
I think this has been fixed

PR #1143, opened 2023-11-13
Fix issues in CI

flaub:
* Avoid GitHub rate limiting for CI machines
* Drop `vendored` feature flag from `openssl` dependency. This works around some strange issue related to building openssl from source.
* Patch `protoc-prebuilt` to allow the use of GITHUB_TOKEN


morgangibbs87 (approved):
Approved

PR #1144, opened 2023-11-13
Avoid exhausting disk space by dropping segments in dev mode prover

nategraf:
A user on Discord "proving" a very large program in dev mode ran into an issue where their disk space was exhausted, requiring a restart to make they system usable again.

This PR addresses this issue by specializing the implementation of the `Prover::prove` method on the `DevModeProver` to pass a drop-segment callback to the executor. This avoids saving the segments to disk when they will not end up being used, saving disk space and potentially improving execution performance.


nategraf:
Moving to draft: I just realized more changed are needed to make this work.

nategraf:
Closing this PR as there is work in progress to address this issue in a more robust way

PR #1146, opened 2023-11-14
refactor(binfmt): Return errors instead of panics

### LINKED ISSUE #887: risc0-binfmt: remove asserts and return Err instead
There are several asserts in image.rs which results in panics. Change these to something like `anyhow::bail!` so that callers can provide better error messages rather than getting panics. While there might be many more cases, this issue specifically tracks risc0-binfmt

Cardosaum:
Closes #887

### DIFF
--- a/bonsai/examples/governance/.cargo/config.toml
+++ b/bonsai/examples/governance/.cargo/config.toml
@@ -17,7 +17,6 @@ rustflags = [
   "-Doverflowing_literals",
   "-Dpath_statements",
   "-Dpatterns_in_fns_without_body",
-  "-Dprivate_in_public",
   "-Dtrivial_numeric_casts",
   "-Dunconditional_recursion",
   "-Dunused_allocation",
--- a/risc0/binfmt/src/image.rs
+++ b/risc0/binfmt/src/image.rs
@@ -282,7 +282,7 @@ impl MemoryImage {
             let expected = self.hash_page(page_idx)?;
             let entry_addr = self.info.get_page_entry_addr(page_idx);
             let mut entry = [0_u8; DIGEST_BYTES];
-            self.load_region_in_page(entry_addr, &mut entry);
+            self.load_region_in_page(entry_addr, &mut entry)?;
             let actual = Digest::try_from(entry)?;
             tracing::debug!(
                 "page_idx: {page_idx}, page_addr: 0x{page_addr:08x} entry_addr: 0x{entry_addr:08x}"
@@ -296,7 +296,7 @@ impl MemoryImage {
         let root_page_addr = self.info.root_page_addr;
         let root_page_bytes = self.info.num_root_entries * DIGEST_BYTES as u32;
         let mut root_page = vec![0_u8; root_page_bytes as usize];
-        self.load_region_in_page(root_page_addr, &mut root_page);
+        self.load_region_in_page(root_page_addr, &mut root_page)?;
         let expected = hash_page_bytes(&root_page)?;
         let root = self.compute_root_hash()?;
         if expected != root {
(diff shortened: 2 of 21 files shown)


Cardosaum:
Shifting to draft status as we wait clarification on design considerations and determine the optimal approach for API export. Should the caller be aware of potential panics due to bad inputs, or would it be preferable to implement an error return mechanism for instances where properties deviate from expected invariants?

flaub:
I think this approach is probably reasonable.

bonsai/examples/governance/.cargo/config.toml
```diff
@@ -17,7 +17,6 @@ rustflags = [
   "-Doverflowing_literals",
   "-Dpath_statements",
   "-Dpatterns_in_fns_without_body",
-  "-Dprivate_in_public",
```
SchmErik, line 20: Ah, thanks for catching this!

risc0/binfmt/src/image.rs
```diff
@@ -71,10 +71,19 @@ pub struct PageTableInfo {
 
 impl From<PersistentPageTableInfo> for PageTableInfo {
     fn from(value: PersistentPageTableInfo) -> Self {
-        PageTableInfo::new(value.page_table_addr, value.page_size)
+        PageTableInfo::new(value.page_table_addr, value.page_size).expect("Invalid page table")
     }
 }
 
+// TODO: Decide between TryFrom and From
```
SchmErik, line 78: @Cardosaum were there any disadvantages to `TryFrom`? I think you mentioned it to me but I forgot...

SchmErik (approved):
looks good to me. Maybe we could decide on the TryFrom and From in a separate PR to keep the PR sizes smaller... @flaub do you have any thoughts about this PR?

bonsai/examples/governance/.cargo/config.toml
flaub: This was the rust toolchain was pinned, because this caused a build breakage in the newest version. I think we should try to stay up to date with the latest, but we should control that instead of having builds break suddenly.

risc0/binfmt/src/image.rs
flaub: `expect` causes a panic, so I don't think we can use `From` here.

risc0/binfmt/src/image.rs
```diff
@@ -213,44 +225,48 @@ impl MemoryImage {
...
-                &page[(addr - page_start) as usize..(addr - page_start) as usize + bytes.len()],
-            );
-        } else {
-            assert!(
-                addr as usize <= MEM_SIZE,
-                "address {addr:08X} outside MEM_SIZE ({MEM_SIZE:08X})"
-            );
-            bytes.fill(0);
-        }
+        match self.pages.get(&page_idx) {
```
flaub, line 232: Personally, I prefer `if let Some` over `match`

risc0/build/src/lib.rs
```diff
@@ -82,7 +82,8 @@ impl Risc0Method {
         let elf = fs::read(&self.elf_path).unwrap();
         let program = Program::load_elf(&elf, memory::GUEST_MAX_MEM as u32).unwrap();
         let image = MemoryImage::new(&program, PAGE_SIZE as u32).unwrap();
-        image.compute_id()
+        // TODO: Should we panic here?
+        image.compute_id().expect("Failed to compute image ID")
```
flaub, line 86: Yes, this is happens at compile time, not run time

risc0/binfmt/src/image.rs
Cardosaum: @SchmErik , the problem I was having was because the struct definition was specifying `serde(from = "PersistentPageTableInfo")` instead of `serde(try_from = "PersistentPageTableInfo")`, a simple detail I was overlooking.

risc0/binfmt/src/image.rs
Cardosaum: Reverted in #1226 👍

PR #1148, opened 2023-11-15
Change default PosixIo to use an empty vec for stdin

nategraf:
This PR changes the default `PosixIo` struct to hook the guest `stdin` to an empty vector, instead of the current default which is to use the host process `stdin`. This resolves issue #1016, where it is easy to accidentally cause a difficult to debug hand on the process by forgetting to add an `ExectorEnvBuilder::write` call, resulting in the executor performing a blocking read on stdin.

Note that if a developer does indeed want the host process `stdin` to be used as guest `stdin`, they can accomplish this by calling `builder.stdin(std::io::stdin())` while building the `ExectorEnv`.

Related: #1050
Resolves: #1016

### DIFF
--- a/risc0/zkvm/src/host/client/posix_io.rs
+++ b/risc0/zkvm/src/host/client/posix_io.rs
@@ -15,7 +15,7 @@
 use std::{
     cell::RefCell,
     collections::BTreeMap,
-    io::{stderr, stdin, stdout, BufRead, BufReader, Write},
+    io::{stderr, stdout, BufRead, Cursor, Write},
     rc::Rc,
 };
 
@@ -34,7 +34,7 @@ impl<'a> Default for PosixIo<'a> {
             read_fds: Default::default(),
             write_fds: Default::default(),
         };
-        new.with_read_fd(fileno::STDIN, BufReader::new(stdin()))
+        new.with_read_fd(fileno::STDIN, Cursor::new(vec![]))
             .with_write_fd(fileno::STDOUT, stdout())
             .with_write_fd(fileno::STDERR, stderr());
         new
--- a/risc0/r0vm/src/lib.rs
+++ b/risc0/r0vm/src/lib.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{fs, path::PathBuf, rc::Rc};
+use std::{fs, io, path::PathBuf, rc::Rc};
 
 use clap::{Args, Parser, ValueEnum};
 use risc0_zkvm::{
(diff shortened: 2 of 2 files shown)


risc0/zkvm/src/host/client/posix_io.rs
```diff
@@ -34,7 +34,7 @@ impl<'a> Default for PosixIo<'a> {
             read_fds: Default::default(),
             write_fds: Default::default(),
         };
-        new.with_read_fd(fileno::STDIN, BufReader::new(stdin()))
+        new.with_read_fd(fileno::STDIN, Cursor::new(vec![]))
             .with_write_fd(fileno::STDOUT, stdout())
             .with_write_fd(fileno::STDERR, stderr());
```
flaub, line 39: So it seems like we should probably not use default stdout and stderr as well if we're going to stop using stdin.

flaub:
Should we update `r0vm` to retain its previous behavior? I suppose if no one was relying on it and we don't have any tests that break, perhaps we don't need to worry about it.

nategraf:
@austinabell I spent a bit looking into this, and looking at your PR #1050, and this is the change set I think we should implement. One of the main differences is that the change set in #1050 actually breaks the users ability to override stdin on the `ExecutorEnvBuilder` because it will unconditionally reset the `stdin` fd when `build()` is called.

risc0/zkvm/src/host/client/posix_io.rs
nategraf: So I thought about this, and my inclination is the indeed have the default be empty vec for stdin and pipe to stdout. It is inconsistent, but I think it's what a new user will find least surprising / most useful. In particular, this preserves the behavior that `println!` and other guest output goes to users terminal. Do you agree with that logic?

nategraf:
> Should we update `r0vm` to retain its previous behavior? I suppose if no one was relying on it and we don't have any tests that break, perhaps we don't need to worry about it.

I think this will be the right behavior, and it turns out this behavior is indeed used in one of our tests
https://github.com/risc0/risc0/blob/e461c5d463c117c9b6946d86a1e14d55dddee686/risc0/r0vm/tests/dev_mode.rs#L25-L31

In e461c5d463c117c9b6946d86a1e14d55dddee686, I've gone ahead and made the default input source be the process stdin when running the "cli mode"(not in "server mode").

flaub (approved):

austinabell:
> @austinabell I spent a bit looking into this, and looking at your PR #1050, and this is the change set I think we should implement. One of the main differences is that the change set in #1050 actually breaks the users ability to override stdin on the `ExecutorEnvBuilder` because it will unconditionally reset the `stdin` fd when `build()` is called.

I agree! That's why I was suggesting having the input be `Option<Vec<u8>>` to avoid ambiguity, or have the constructor be explicit. 

This seems to be the same, except that there might be a weird pattern when input is provided and it overwrites the stdin when it expects it for some workflow like in this test. Might be a small foot gun, but also seems like you'd have to try to hit that case

PR #1150, opened 2023-11-15
Integrate sppark

flaub:
SPPARK: Supranational's performance primitives for arguments of knowledge

https://github.com/supranational/sppark


SchmErik (approved):
at a high level I understand what's going on but maybe @jbruestle might want to take a look as well?

flaub:
Looks like SPPARK requires CUDA 11.4+. I'll need to update our CI machines before this can land.

jbruestle (approved):

PR #1152, opened 2023-11-15
Add Poseidon2 support to CUDA

jbruestle:


tzerrell (approved):
Looks good, in particular everything up to and including `poseidon2_mix` I can give a very solid review on given my own work at the moment. The `poseidon2_rows` code has me thinking I'm missing something between the poseidon2_mix and the actual hash, but I think that's a flaw in my understanding not this code.

flaub (approved):

PR #1154, opened 2023-11-16
Fix panic on RUST_LOG=<filter> during tests

Cardosaum:
I'm opening this PR to address a bug report encountered when running a test with `RUST_LOG=info`. The bug causes a panic due to a regex parse error related to Unicode-aware case insensitivity matching - exactly like the issue described in [tracing-subscriber](https://github.com/tokio-rs/tracing/issues/2565). I have tried to fix it by specifying `test-log` feature to both `log` and `trace` in the `Cargo.toml` file. 

I can't validate the fix on my machine yet, since this seems to arise only when `-F cuda` is passed, and I don't have the required hardware to test it.


flaub:
It doesn't seem like this is related to `-F cuda`, but I guess I could be wrong.

examples/digital-signature/Cargo.toml
```diff
@@ -14,7 +14,8 @@ tracing = { version = "0.1", features = ["log"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
 [dev-dependencies]
-test-log = { version = "0.2", default-features = false, features = ["trace"] }
+env_logger = "0.10"
+test-log = { version = "0.2", features = ["trace", "log"] }
```
flaub, line 18: The default already contains `log`

```suggestion
test-log = { version = "0.2", features = ["trace"] }
```

risc0/binfmt/Cargo.toml
```diff
@@ -8,8 +8,9 @@ homepage = { workspace = true }
 repository = { workspace = true }
 
 [dev-dependencies]
+env_logger = "0.10"
```
SchmErik, line 11: why is both env_logger and tracing subscriber a part of this? I thought that we only needed `tracing` and `tracing-subscriber`

flaub:
I found the underlying issue that some of us have run into, it's likely because we were using an older version of `tracing-subscriber` and we don't have all our lock files checked in.

https://github.com/tokio-rs/tracing/issues/2565

PR #1156, opened 2023-11-16
Compute recursion control IDs in `cargo xtask bootstrap`

nategraf:
This finishes the process of cleaning up recursion code from the circuit repo by moving logic for computing recursion program (zkr) control IDs in `cargo xtask bootstrap`. It also includes the results of running bootstrap from the circuit from this PR and from the circuit repo at commit `ed9af04b1b18dc29ee179b73d335f830216bdad1`

Additionally, a number of comments are added to parts of the the API surface I touched in the process of writing this PR.


risc0/binfmt/Cargo.toml
```diff
@@ -26,3 +26,4 @@ tracing = { version = "0.1", default-features = false }
 [features]
 default = ["std"]
 std = ["anyhow/std", "elf/std", "tracing/log", "tracing/std", "risc0-zkp/std", "serde/std"]
+recursion = ["std", "risc0-zkp/prove"]
```
flaub, line 29: I'm not sure we need to feature flag the `Program`

risc0/zkvm/Cargo.toml
```diff
@@ -171,6 +171,7 @@ prove = [
   "risc0-circuit-recursion/prove",
   "risc0-circuit-rv32im/prove",
   "risc0-zkp/prove",
+  "risc0-binfmt/recursion",
```
flaub, line 174: Can you sort this?

flaub (approved):
Looks good, but I think we don't need the `recursion` feature flag

risc0/zkvm/Cargo.toml
nategraf: Fixed in 22b7e10e68ae473ad62ea714655663feded17bef

risc0/binfmt/Cargo.toml
nategraf: In 14fb44d847c432302566aef44d18bf5b1e1fef51 I refactored to drop that feature flag and move the control ID calculation into `risc0-zkvm`. Recursion doesn't have anything comparable to `Loader` for `rv32im`, so I simply placed it in the module as a function.

flaub: Does the WIP bootstrap need it? I thought it called it to compute the ControlIDs.

nategraf: Ok, I actually realized that doesn't work out. I've reverted that commit and instead just hid the function behind a `prove` flag. Not a huge fan of introducing a `prove` flag to `binfmt`, but its the best option I could think of.

nategraf:
I am going to rewrite this PR to move the control ID calculation for the recursion predicates from WIP bootstrap to xtask bootstrap.

alipostaci2001:
thanks for info

nategraf:
> I am going to rewrite this PR to move the control ID calculation for the recursion predicates from WIP bootstrap to xtask bootstrap.

Finished this refactor. This PR is fully ready for review.

flaub (approved):
Awesome!

PR #1158, opened 2023-11-16
Groth16 verifier

capossele:
- adds groth16 verifier in the zkvm, with support to run on the guest with no std
- adds support for verifying proof from Circom/SnarkJS json format
- adds a groth16 verifier example
- adjust ethereum-relay, bonsai-sdk, bonsai-rest-api-mock to the new changes

## Todo
- adjust bonsai-alpha to return a `Receipt` encoded with bincode instead of the deprecated `SnarkReceipt`


examples/groth16-verifier/README.md
flaub, line 1: How about we have the verifier example just run on the host? I don't know what value we get for having this verifier run in the guest since it takes up loads of cycles. We might be able to have proof composition be able to verify groth16 receipts in the future.

risc0/zkvm/Cargo.toml
```diff
@@ -61,6 +64,7 @@ addr2line = { version = "0.21", features = ["rustc-demangle"], optional = true }
 bincode = { version = "1.3", optional = true }
 bonsai-sdk = { workspace = true, optional = true }
 bytes = { version = "1.4", features = ["serde"], optional = true }
+ethereum-types = { version = "0.14", default_features = false }
```
flaub, line 67: Is there anyway we can avoid pulling in these types on the zkvm?

risc0/zkvm/src/host/receipt.rs
```diff
@@ -281,6 +281,9 @@ pub enum InnerReceipt {
     /// The [SuccinctReceipt].
     Succinct(SuccinctReceipt),
 
+    /// The [SnarkReceipt].
+    Snark(SnarkReceipt),
```
flaub, line 285:
```suggestion
    /// The [Groth16Receipt].
    Snark(Groth16Receipt),
```

.github/workflows/main.yml
```diff
@@ -362,12 +362,12 @@ jobs:
       - run: "git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :"
       - uses: actions/checkout@v4
       - uses: ./.github/actions/rustup
-      - uses: ./.github/actions/sccache
-        with:
-          key: Linux-default
+      # - uses: ./.github/actions/sccache
+      #   with:
+      #     key: Linux-default
```
flaub, line 367: This will slow things down, why?

capossele: Just testing, that test is passing for me locally both on Mac or Linux (with using docker) so I thought maybe was the cache. I’ll revert

risc0/zkvm/methods/build.rs
```diff
@@ -38,7 +38,13 @@ fn main() {
...
 
     let map = HashMap::from([
-        ("risc0-zkvm-methods-guest", GuestOptions::default()),
+        (
+            "risc0-zkvm-methods-guest",
+            GuestOptions {
+                features: vec![],
+                use_docker: use_docker.clone(),
+            },
+        ),
```
capossele, line 47: Added this as the docker test on the server exec was not using the docker generated elf for the `multi_test`

risc0/cargo-risczero/Cargo.toml
```diff
@@ -38,6 +38,9 @@ tokio = { version = "1", features = ["rt"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 zip = { version = "0.6", optional = true }
 
+[patch.crates-io]
+faster-hex = { git = "https://github.com/nervosnetwork/faster-hex.git", ref = "7ae3424" }
+
```
capossele, line 43: A new version of `faster-hex` (v0.8.2) broke cargo-generate. Added temporary patch until that get fixed

risc0/zkvm/src/host/receipt.rs
```diff
@@ -339,12 +353,43 @@ impl InnerReceipt {
...
 
+/// A receipt composed of a Groth16 over the BN_254 curve
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct Groth16Receipt {
+    /// the Grtoh16 seal of this receipt
+    pub seal: Vec<u8>,
+
+    /// the receipt metadata containing states of the system during the segment
+    /// executions
```
nategraf, line 371:
```suggestion
    /// [ReceiptMetadata] containing information about the execution that this receipt proves.
```

risc0/zkvm/src/host/receipt.rs
```diff
@@ -339,12 +353,43 @@ impl InnerReceipt {
...
+    pub meta: ReceiptMetadata,
+}
+
+impl Groth16Receipt {
+    /// Verify the integrity of this receipt, ensuring the metadata is attested
+    /// to by the seal.
+    pub fn verify_integrity_with_context(
+        &self,
+        _ctx: &VerifierContext,
```
nategraf, line 380: The `VerifierContext` is specific to our STARK system. It's fields are not relevant here, so it would be best to simply not include it in this method's signature. (It's also ok that this method name not match the others exactly)
```suggestion
    pub fn verify_integrity(
        &self
```

risc0/zkvm/src/groth16.rs
nategraf, line 1: This is probably not the right place for this file. `risc0_zkvm::host::receipt` and `risc0_zkvm::host::recursion` both seem like reasonable candidates to nest `groth16` under. (Re-exporting the functionality at a higher level seems reasonable)

nategraf: This would also help to have not so many `cfg` statements in this file, as the entire module could be excluded. (More generally, you can also exclude an entire file by adding a `#![cfg(...)]` statement at the top.

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+
+#[cfg(not(target_os = "zkvm"))]
+use crate::groth16::constants::*;
+#[cfg(not(target_os = "zkvm"))]
+use crate::sha::{Digest, Impl, Sha256};
+#[cfg(not(target_os = "zkvm"))]
+use crate::ALLOWED_IDS_ROOT;
+
+#[cfg(not(target_os = "zkvm"))]
+pub(crate) mod constants {
```
nategraf, line 74: How does this get updated? Is this entire file generated, or just these parameters?

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+/// Groth16 proof object encoded in big endian
+#[cfg(not(target_os = "zkvm"))]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct Groth16Seal {
+    /// Proof 'a' value
+    pub a: Vec<Vec<u8>>,
+    /// Proof 'b' value
+    pub b: Vec<Vec<Vec<u8>>>,
+    /// Proof 'c' value
+    pub c: Vec<Vec<u8>>,
```
nategraf, line 135: There are each fix-sized arrays. It would be better to encode that explicitly by using.

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+
+#[cfg(not(target_os = "zkvm"))]
+impl Groth16Seal {
+    const ELEMENT_SIZE: usize = 32;
+    const G1_GROUP_SIZE: usize = Self::ELEMENT_SIZE * 2;
+    const G2_GROUP_SIZE: usize = Self::ELEMENT_SIZE * 4;
+    const SIZE: usize = Self::G1_GROUP_SIZE * 2 + Self::G2_GROUP_SIZE;
+
+    /// Serialize the Groth16 seal into Vec<u8>
+    pub fn to_vec(&self) -> Vec<u8> {
```
nategraf, line 146: Is this encoding specified by some standard, or is this just defined here? If the latter, why no use `serde` derives instead?

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+    pub fn digest(&self) -> Digest {
+        let hash = Impl::hash_bytes(&self.to_vec());
+        *hash.as_ref()
+    }
+}
+
+/// Groth16 Proof as encoded by Circom/SnarkJS
+#[cfg(not(target_os = "zkvm"))]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct RawProof {
```
nategraf, line 402: "Raw" is rarely a helpful descriptor. Can we improve the naming here? Where does this come from? Is this format specified by Circom?

nategraf: Also, does this need to be `pub`? It feels like an implementation detail, and not one we intended to support with backwards compatibility into the future.

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+            from_u256(&raw_proof.pi_c[1])?,
+        ];
+
+        Ok(Groth16Seal { a, b, c })
+    }
+}
+
+/// Groth16 instance over the BN_254 curve encoded in little endian
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Groth16 {
```
nategraf, line 246:
```suggestion
pub struct Groth16Proof {
```

When I first saw `Groth16` I assumed this was a proof system implementation. Instead it appear to be a proof.

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+        Ok(Self {
+            pvk: pvk_bytes,
+            proof: proof_bytes,
+            prepared_inputs: prepared_inputs_bytes,
+        })
+    }
+
+    /// Creates a Groth16 instance from the raw material generated by Circom/SnarkJS
+    #[cfg(not(target_os = "zkvm"))]
+    pub fn from_raw(
```
nategraf, line 294: Maybe instead of "raw", this should be "circom"?
```suggestion
    pub fn from_circom(
```

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+        Ok(Self {
+            pvk: pvk_bytes,
+            proof: proof_bytes,
+            prepared_inputs: prepared_inputs_bytes,
+        })
+    }
+
+    /// Computes the prepared verifying key used by Bonsai
+    #[cfg(not(target_os = "zkvm"))]
+    pub fn pvk() -> Result<PreparedVerifyingKey<Bn254>, Error> {
```
nategraf, line 334: This feels like an implementation detail.
```suggestion
    fn pvk() -> Result<PreparedVerifyingKey<Bn254>, Error> {
```

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,802 @@
...
+        result.extend(&self.prepared_inputs);
+        result
+    }
+
+    /// Compute the SHA256 digest of the Groth16 instance
+    #[cfg(not(target_os = "zkvm"))]
+    pub fn digest(&self) -> Digest {
+        let hash = Impl::hash_bytes(&self.to_vec());
+        *hash.as_ref()
+    }
```
nategraf, line 396: Specifically, this is the hash of a particular encoding. I would suggest removing this method and making the encoding explicit where you need a hash.

```suggestion
```

risc0/zkvm/src/groth16.rs
nategraf: Another option for this module would be to live outside of `risc0-zkvm`. In many ways, this module is more similar to `risc0-zkp`. Having a `risc0-groth16` crate that defines these cryptographic details may make sense (it would also be a reasonable way to isolate and manage the dependencies brought in by Groth16 verification).

examples/chess/methods/guest/Cargo.lock
```diff
@@ -2,12 +2,198 @@
...
+ "zerocopy",
+]
+
 [[package]]
 name = "anyhow"
 version = "1.0.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "224afbd727c3d6e4b90103ece64b8d1b67fbb1973b1046c2281eed3f3803f800"
 
+[[package]]
```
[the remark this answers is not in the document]
capossele: The problem is that the verifier should be included in the `default-features=false`. Since groth16 now is part of a receipt, that would not be possible to make it optional without breaking the verifier completeness

risc0/zkvm/src/groth16.rs
capossele: My impression was that we wanted to be able to support verification of any circom/snarkjs groth16 proof. If that is not the case anymore, I can completely remove this

risc0/zkvm/src/groth16.rs
capossele: that seems a good idea

examples/chess/methods/guest/Cargo.lock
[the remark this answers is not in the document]
nategraf: The `Receipt` datatype is not included when `target_os = "zkvm"`

https://github.com/risc0/risc0/blob/bcfc9c6a55d00eb4696d3c9e20aff0bcff11f8ef/risc0/zkvm/src/lib.rs#L82-L90

More generally, we don't actively support directly verifying `Receipts` in the guest.

risc0/zkvm/src/groth16.rs
nategraf: To make sure I understand, this is the struct you get when you de-serialize the circom JSON outputs?

nategraf: On the question of verifying arbitrary Circom Groth16 proves. We we offload that concern to https://github.com/arkworks-rs/circom-compat, and here provide only routines specialized to our needs?

flaub: I don't think we need to worry about arbitrary groth16 proofs, only those that are created by our stark2snark process.

risc0/zkvm/src/groth16.rs
flaub: I think this is just an enum arm type for a particular `Receipt`, so it seems to me like the `Proof` suffix is inconsistent and redundant.

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -381,6 +382,32 @@ impl TryFrom<pb::core::SegmentReceipt> for SegmentReceipt {
     }
 }
 
+impl From<Groth16Receipt> for pb::core::Groth16Receipt {
```
flaub, line 385: I don't think we need to worry about Groth16 yet for client/server, we can take a separate pass on that later.

risc0/zkvm/src/groth16.rs
flaub: I'm not yet sold on the idea of having this appear in a separate crate. I do think we should move it into `host` since we don't plan to have this run in the guest.

bonsai/sdk/README.md
```diff
@@ -93,7 +93,8 @@ fn run_stark2snark(session_id: String) -> Result<()> {
                 continue;
             }
             "SUCCEEDED" => {
-                let snark_receipt = res.output;
+                let snark_receipt: Receipt = bincode::deserialize(&res.output
+                .expect("API error, missing receipt on completed snark session"));
```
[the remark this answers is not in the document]
flaub: Let's deal with protobuf in a different setting, I think we should be consistent if we're going to start using protobuf for anything external. At the moment it's only used internally for client/server.

examples/groth16-verifier/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+use byteorder::{BigEndian, WriteBytesExt};
+use hello_world_methods::{MULTIPLY_ELF, MULTIPLY_ID};
+use risc0_zkvm::{
+    default_prover, groth16::Groth16Seal, serde::to_vec, ExecutorEnv, Groth16Receipt, MemoryImage,
+    Program, Receipt, GUEST_MAX_MEM, PAGE_SIZE,
+};
+use std::time::Duration;
+
+// Run Bonsai for the hello world example
+pub fn run_bonsai(input_data: Vec<u8>) -> Result<Receipt> {
```
flaub, line 26: I think for the initial example, we should leave Bonsai out of it. This is because I think we need to unify and integrate the new Groth16 receipt into Bonsai and then hide all these details behind the `Prover` trait that zkvm already exposes.

examples/groth16-verifier/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    // create the memoryImg, upload it and return the imageId
+    let img_id = {
+        let program = Program::load_elf(MULTIPLY_ELF, GUEST_MAX_MEM as u32)?;
+        let image = MemoryImage::new(&program, PAGE_SIZE as u32)?;
+        let image_id = hex::encode(image.compute_id());
+        let image = bincode::serialize(&image).expect("Failed to serialize memory img");
+        client.upload_img(&image_id, image)?;
+        image_id
+    };
```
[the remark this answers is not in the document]
flaub: This is another reason to drop Bonsai from this example, I think it will be a lot easier to just deal with zkvm first and get that all working then we can deal with integrating this into the `Prover` trait and then we can have an example that works for Bonsai or Local in just the same way that all other examples work, you just specify Bonsai env vars.

examples/groth16-verifier/src/main.rs
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    // we build a groth16 proof from the circom material collected from SnarkJS
+    let proof = Groth16Proof::from_circom(circom_vkey, circom_proof, circom_public).unwrap();
+
+    // groth16 proof verification
+    proof.verify().unwrap();
+    println!("Verified the Groth16 proof from Circom");
+    Ok(())
+    // -------------------------------------------------------------
```
flaub, line 60: What's the use case for this? It seems like the above `groth16_receipt.verify` is sufficient.

risc0/zkvm/src/host/receipt.rs
flaub: The `VerifierContext` is meant to be any context that any verifier might want to have on hand to perform it's job. I don't think we've yet thought about what a `VerifierContext` might mean for `Groth16`. I think we should just call it `verify` to be consistent, but I also wonder, where is the `image_id`?

risc0/zkvm/src/groth16.rs
```diff
@@ -0,0 +1,798 @@
...
+pub struct Groth16Proof {
+    pvk: Vec<u8>,
+    proof: Vec<u8>,
+    prepared_inputs: Vec<u8>,
+}
+
+impl Groth16Proof {
+    /// Creates a Groth16 instance from a `Groth16Seal` and the metadata digest
+    /// of the original RISC Zero receipt
+    #[cfg(not(target_os = "zkvm"))]
```
flaub, line 258: We should avoid `#cfg[]` all over the place and instead mark this entire file as being 'host-only'.

examples/sha/methods/guest/Cargo.lock
```diff
@@ -2,12 +2,198 @@
...
+[[package]]
+name = "ark-bn254"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a22f4561524cd949590d78d7d4c5df8f592430d221f7f3c9497bbafd8972120f"
+dependencies = [
+ "ark-ec",
+ "ark-ff",
+ "ark-std",
+]
```
flaub, line 32: The goal with this initial PR is to avoid any changes to guest-side lock files.

flaub:
At this time, I think we should avoid trying to get Groth16 to work in the guest. We can perhaps do that in a future pass, but I'm also not sure it makes sense. I also think we should avoid dealing with any changes to client/server in this initial pass. I think we can also avoid dealing with Bonsai, and use some pre-baked Groth16 receipts for testing. I think once this is all in place, then we can go back and figure out how to integrate with Bonsai.

examples/groth16-verifier/src/main.rs
capossele: The use case for this was to be able to run the verification of any groth16 proof generated outside of our zkvm in the guest, to potentially be the integration point for multiple proving system (e.g., halo2, groth16 and the likes). However, as it seems we are not interested in that for the moment, I will drop it

capossele:
> At this time, I think we should avoid trying to get Groth16 to work in the guest. We can perhaps do that in a future pass, but I'm also not sure it makes sense. I also think we should avoid dealing with any changes to client/server in this initial pass. I think we can also avoid dealing with Bonsai, and use some pre-baked Groth16 receipts for testing. I think once this is all in place, then we can go back and figure out how to integrate with Bonsai.

While I disagree on the utility to prove the verification of external proving system, I understand we want to focus on just being able to verify RISC Zero groth16 receipt for the moment. We might revisit in the future then 👍

risc0/zkvm/src/host/api/convert.rs
capossele: Without that we would not be able to cover the case: `InnerReceipt::Groth16` for the `impl From<InnerReceipt> for pb::core::InnerReceipt`.  I'm not sure what's your suggestion here? Remove all the photo stuff related to Groth16? Would that work even if that enum is now part of the InnerReceipt?

examples/sha/methods/guest/Cargo.lock
capossele: I've tried several things to get this in but for some reason, if we use the std feature this is getting in even though we specifically say `[target.'cfg(not(target_os = "zkvm"))'.dependencies]` in the Cargo.toml.
The only thing that worked was to make these deps optional and enable them as part of the client, however, in that case the `external/substrate` test fails as the receipt should be part of `default-features=false` independently from the client or not. Any suggestions for getting around this?

risc0/zkvm/src/host/groth16.rs
```diff
@@ -0,0 +1,348 @@
...
+
+impl Groth16Seal {
+    const ELEMENT_SIZE: usize = 32;
+    const G1_GROUP_SIZE: usize = Self::ELEMENT_SIZE * 2;
+    const G2_GROUP_SIZE: usize = Self::ELEMENT_SIZE * 4;
+    const SIZE: usize = Self::G1_GROUP_SIZE * 2 + Self::G2_GROUP_SIZE;
+
+    /// Serialize the Groth16 seal into Vec<u8>
+    pub fn to_vec(&self) -> Vec<u8> {
+        let mut result = Vec::new();
```
hmrtn, line 89: We can avoid re-allocation here when appending the proof data by using the sum group size `Vec::with_capacity(Self::SIZE)` instead of dynamically allocating on the fly.

risc0/zkvm/src/host/receipt.rs
```diff
@@ -310,7 +310,7 @@ impl InnerReceipt {
     ) -> Result<(), VerificationError> {
         match self {
             InnerReceipt::Composite(x) => x.verify_integrity_with_context(ctx),
-            InnerReceipt::Groth16(x) => x.verify_integrity_with_context(ctx),
+            InnerReceipt::Groth16(x) => x.verify(),
```
nategraf, line 313: So I have been using `verify` to mean "verify a successful execution of the zkVM" and `verify_integrity` to "verify an execution of the zkVM, successful or not". I started doing this because most users will reach for `.verify`, since it sounds like what they want. My belief is that most users are only interested, or think about, successful executions, and that they will be surprised if `.verify` passes for an execution that failed (ended in `Faulted` for example).

In this taxonomy, this method would be `verify_integrity`.
```suggestion
            InnerReceipt::Groth16(x) => x.verify_integrity(),
```

risc0/zkvm/src/lib.rs
```diff
@@ -84,6 +82,7 @@ pub use self::host::{
 #[cfg(not(target_os = "zkvm"))]
 pub use self::host::{
     control_id::POSEIDON_CONTROL_ID,
+    groth16::{Groth16Proof, Groth16Seal},
```
nategraf, line 85: I might suggest not putting `Groth16Proof` and `Groth16Seal` in the global namespace. They feel like lower level details.
```suggestion
    groth16
```

examples/sha/methods/guest/Cargo.lock
nategraf: Very weird. This feels like a weird edge case for the `cfg` system in Cargo.toml (Logically, ark-cc shouldn't even be a decency when `target_os = "zkvm"` so it should be invalid to reference it in a feature). It's really not ideal to drag the arkworks codebase into the guest build :/

risc0/zkvm/src/host/receipt.rs
flaub: How about we have a single `verify()` that takes a parameter for users to decide whether success is desired? That way it's explicit and users don't have to guess which function to call.

risc0/zkvm/src/host/api/convert.rs
flaub: I would just panic with `unimplemented!` for now and then do a followup PR with just the client/server changes.

examples/groth16-verifier/src/lib.rs
flaub: I think we can still have a test somewhere that does the verification. Perhaps we don't need an example until it's all integrated with Bonsai.

examples/groth16-verifier/src/main.rs
flaub: I think the issue here is that it's not very feasible yet because it takes something like 100s of segments to do a verify in the guest. I think once we have some better bigint optimizations in place to do the EC ops we can make external verification in-guest much more feasible.

flaub:
> > At this time, I think we should avoid trying to get Groth16 to work in the guest. We can perhaps do that in a future pass, but I'm also not sure it makes sense. I also think we should avoid dealing with any changes to client/server in this initial pass. I think we can also avoid dealing with Bonsai, and use some pre-baked Groth16 receipts for testing. I think once this is all in place, then we can go back and figure out how to integrate with Bonsai.
> 
> While I disagree on the utility to prove the verification of external proving system, I understand we want to focus on just being able to verify RISC Zero groth16 receipt for the moment. We might revisit in the future then 👍

It's not a question of utility in my mind, I think it makes a ton of sense to verify external proofs! I think the issue is more one of feasibility. I think it's just a matter of phasing things so that we release once we have something that has better optimizations for common EC ops.

risc0/zkvm/src/lib.rs
capossele: This does not work as then we would need to make the `groth16` mod public, something like
```
pub mod groth16;
```

but then it would be the exception. I think keeping it the way it is now is more consistent with the way we are controlling what to re-export

capossele:
> > > At this time, I think we should avoid trying to get Groth16 to work in the guest. We can perhaps do that in a future pass, but I'm also not sure it makes sense. I also think we should avoid dealing with any changes to client/server in this initial pass. I think we can also avoid dealing with Bonsai, and use some pre-baked Groth16 receipts for testing. I think once this is all in place, then we can go back and figure out how to integrate with Bonsai.
> > 
> > 
> > While I disagree on the utility to prove the verification of external proving system, I understand we want to focus on just being able to verify RISC Zero groth16 receipt for the moment. We might revisit in the future then 👍
> 
> It's not a question of utility in my mind, I think it makes a ton of sense to verify external proofs! I think the issue is more one of feasibility. I think it's just a matter of phasing things so that we release once we have something that has better optimizations for common EC ops.

Yeah that makes super sense

examples/sha/methods/guest/Cargo.lock
capossele: To be fair it looks like the `[target.'cfg(not(target_os = "zkvm"))'.dependencies]` is not really doing what we want, because by looking at main, we also import in the cargo.lock of any methods using `std` all of the dependencies that are not listed as optional, independently from the `target_os ` (e.g., `risc0-circuit-recursion`, `risc0-circuit-rv32im`).

https://github.com/risc0/risc0/blob/cc92fa90903d1e9ebeb5dcb93cb812085efd1037/risc0/zkvm/methods/std/Cargo.lock#L441-L461

capossele: I guess some potential solutions could be:
1. Fix the `[target.'cfg(not(target_os = "zkvm"))'.dependencies]`
2. Create a `verifier` feature that is not part of `default-features` and add that feature as part of the `client` feature.  In the substrate case, we would need to excplicitly add the `verifier` feature to have a full `no-std` verifier.

flaub: So far, my strategy on this has been:
* To get the prover, use `prove` flag, but only available on host.
* The guest-side is `default-features = false`, but `std` is supported there.
* The verifier is thus supported only on the host and requires `std`.
* BTW, `client` and `prove` requires `std`.

capossele: > The verifier is thus supported only on the host and requires std
But that would brake the `wasm` `no-std` verifier we test under the `substrate` test, no?

capossele: I ended up setting all the host features that were not marked as optional to be optional and grouped them into the `verify` feature. That cleaned up a lot from the guest binaries

risc0/zkvm/Cargo.toml
```diff
@@ -96,14 +100,27 @@ tempfile = "3"
...
+  "ark-bn254",
+  "ark-groth16",
+  "ark-serialize",
+  "num-bigint",
+  "num-derive",
+  "risc0-circuit-recursion",
+  "risc0-circuit-rv32im",
+  "sha2",
+  "tracing"
+]
```
flaub, line 113:
```suggestion
verify = [
  "dep:ark-bn254",
  "dep:ark-groth16",
  "dep:ark-serialize",
  "dep:num-bigint",
  "dep:num-derive",
  "dep:risc0-circuit-recursion",
  "dep:risc0-circuit-rv32im",
  "dep:sha2",
  "dep:tracing"
]
```

flaub: Also can we keep the features sorted? Let's put `verify` at the end.

risc0/zkvm/Cargo.toml
```diff
@@ -148,20 +165,25 @@ prove = [
...
+std-verify = [
+  "ark-bn254/std",
+  "ark-groth16/std",
+  "ark-serialize/std",
+  "num-bigint/std",
+  "num-traits/std",
   "risc0-circuit-recursion/std",
   "risc0-circuit-rv32im/std",
+  "sha2/std"
+]
```
flaub, line 184: I don't think we need this.

We should be able to do:

```
std = ["ark-bn254?/std", ...]
```

flaub:
I'd like to avoid adding a new `verify` feature flag, but I think we could try this for now and maybe try to clean it up in a future PR. However, I do think we need to add docs about this feature flag. We'll need to update the feature flag table in a few places: 

* https://github.com/risc0/risc0/blob/main/README.md#feature-flags
* https://github.com/risc0/risc0/blob/main/risc0/zkvm/README.md#crate-feature-flags

risc0/zkvm/methods/build.rs
flaub: Why are we setting `use_docker` for `risc0-zkvm-methods-std` also?

risc0/zkvm/Cargo.toml
capossele: If we do this though, the `?` doesn't really seem to work. It will import every dep specified within the feature

risc0/zkvm/methods/build.rs
capossele: super good catch

risc0/zkvm/Cargo.toml
flaub: The cargo docs say:
> The "package-name/feature-name" syntax will also enable package-name if it is an optional dependency. Often this is not what you want. You can add a ? as in "package-name?/feature-name" which will only enable the given feature if something else enables the optional dependency.

flaub: I don't think we can use `Cargo.lock` as a guide to determine if it's working or not, because this is on a per-workspace basis and is the aggregate of all `Cargo.toml` files in the workspace.

flaub (approved):

PR #1159, opened 2023-11-17
Update versions

flaub:
* Bump version of client in main to provide better error message
* cargo update
* Update zkevm-demo to revm 3.5
* Pin sppark to specific git commit
* Split external/substrate to separate workspace
* Add serde support for u128/i128 data types

### DIFF
--- a/examples/zkevm-demo/core/src/lib.rs
+++ b/examples/zkevm-demo/core/src/lib.rs
@@ -12,20 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub use primitive_types::{H256, U256};
-// Re-export revm members for external usage.
-pub use revm::{
-    primitives::{
-        result::{Eval, ExecutionResult},
-        Env,
-    },
-    EVM,
-};
 use revm::{
-    primitives::{AccountInfo, Bytecode, State, B160, B256},
+    primitives::{AccountInfo, Address, Bytecode, State, B256},
     Database,
 };
 use serde::{Deserialize, Serialize};
+
+// Re-export for external usage.
+pub use primitive_types::H256;
+pub use revm::{
+    primitives::{
+        result::{Eval, ExecutionResult},
+        Env, U256,
+    },
+    EVM,
+};
 
 #[derive(Debug, Default, Deserialize, Serialize)]
 struct ResTrack<T> {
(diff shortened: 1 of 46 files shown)


flaub:
Tagging @Wollac to take a look at the zkevm-demo updates.

examples/zkevm-demo/core/src/lib.rs
```diff
@@ -94,31 +91,36 @@ impl Database for ZkDb {
...
+            gas_price: from_ethers_u256(tx.gas_price.unwrap()),
+            gas_priority_fee: tx.max_priority_fee_per_gas.map(|x| from_ethers_u256(x)),
             transact_to: if let Some(to_addr) = tx.to {
                 TransactTo::Call(to_addr.0.into())
             } else {
                 TransactTo::Create(CreateScheme::Create) // TODO: create2
             },
-            value: tx.value.into(),
+            value: from_ethers_u256(tx.value),
             data: Bytes::from(tx.input.to_vec()), // TODO: gotta be a cleaner way
```
Wollac, line 118:
```suggestion
            data: tx.input.0.into(),
```
That should work now

examples/zkevm-demo/core/src/lib.rs
```diff
@@ -94,31 +91,36 @@ impl Database for ZkDb {
...
             chain_id: tx.chain_id.map(|elm| elm.as_u64()),
             nonce: Some(tx.nonce.as_u64()),
-            access_list: Vec::new(), // TODO
+            access_list: Vec::new(),
```
Wollac, line 121: Not having the access list of the transaction will lead to the EVM computing the wrong gas cost. The way how this example is designed, the gas cost is not used anywhere, so I am not sure whether it should be added especially since the conversion is kinda ugly:
```rust
  .map(|item| {
    (
      item.address.0.into(),
      item.storage_keys
        .into_iter()
        .map(|key| U256::from_be_bytes(key.into())).collect(),
      )
  }).collect()
```

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "stable"
+channel = "1.73"
```
SchmErik, line 2: Why are we pinning to 1.73? Is there some issue with the current stable (1.74)?

flaub: To prevent future build breakage

SchmErik (approved):

PR #1168, opened 2023-11-27
Drop profiler flag and move to server side

flaub:


benchmarks/Cargo.toml
```diff
@@ -38,9 +38,6 @@ starknet-crypto = "0.6"
 tokio = { version = "1", features = ["full"] }
 tracing = { version = "0.1", default-features = false }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
-zeth-guests = { git = "https://github.com/risc0/zeth", branch = "capossele/bump-risc0-f265059f", package = "zeth-guests" }
-zeth-lib = { git = "https://github.com/risc0/zeth", branch = "capossele/bump-risc0-f265059f", package = "zeth-lib" }
-zeth-primitives = { git = "https://github.com/risc0/zeth", branch = "capossele/bump-risc0-f265059f", package = "zeth-primitives" }
```
capossele, line 43: What's the reason for removing zeth?

benchmarks/src/benches/mod.rs
```diff
@@ -26,4 +26,3 @@ pub mod iter_pedersen;
 pub mod iter_sha2;
 pub mod membership;
 pub mod sudoku;
-pub mod zeth;
```
capossele, line 29: If we remove this, then:

https://github.com/risc0/risc0/blob/29c3cd6bdd68d96fa908cab0401efc71602daa9b/benchmarks/src/main.rs#L127C5-L130C6

would fail to compile

nategraf (approved):
Looks good to me 👍 I'm excited for folks to have easy access to this in 0.20

benchmarks/Cargo.toml
flaub: I've disabled the benchmark apps workflow in CI for now since it takes a very long time to run and currently fails because of other issues in CI. I think we can add zeth back at some point, perhaps when zeth has released crates. I'd like to be able to make breaking changes to our code without having to also keep zeth up to date.

PR #1171, opened 2023-11-28
Patch ethers crates

capossele:


flaub:
I was able to resolve this by pinning the foundry toolchain in #1168

PR #1175, opened 2023-11-29
Add files for p0tion ceremony

tzerrell:
Add circom files and ceremony metadata in preparation for running a p0tion ceremony. Note that the commit hash listed in the metadata is necessarily something in the past; once this PR lands I'll make another tiny PR to update that hash to whatever the hash of this landed commit is, so that it points to a commit where the circom files actually exist.


flaub (approved):
LGTM

WDYT of renaming `stark2snark` in general to be more like `compact_proof/groth16`. Basically I've been toying with the term 'compact' to mean a proof optimized for storage.

tzerrell:
> LGTM
> 
> WDYT of renaming `stark2snark` in general to be more like `compact_proof/groth16`. Basically I've been toying with the term 'compact' to mean a proof optimized for storage.

I'm on board with that. Renamed.

PR #1176, opened 2023-11-29
Use built-in modular exponentiation in didactic_stark.py

zeegomo:
Numpy arrays are not great when it comes to finite field arithmetic as there is no built-in support for modulo operations. While the small size of the field used in this example guarantees that multiplications and additions are never overflow, we can't guarantee the same for exponentiation.
As a stopgap solution, I've added a 'mod_pow' function that uses Python built-in modular exponentiation to avoid this occurrence.

Another option could have been to use galois finite field arrays throughout the code but it would have been a much more invasive change.


pdg744 (approved):
Great, thank you so much!

(Also, TIL about using `pow()` with three arguments for modular exponentiation.)

PR #1179, opened 2023-11-30
Add the `resolve` recursion program

nategraf:
This PR adds the `resolve` recursion program. With this program it becomes possible to produce succinct proofs for zkVM executions that utilize proof composition through the `env::verify` API for logically verifying RISC Zero receipts within the RISC Zero zkVM.

As part of this PR, these is additional plumbing of `ReceiptMetadata` structs such that the all required information is available for the `resolve` program. In particular, the `seal` of a `SegmentReceipt` includes the top level fields of the `ReceiptMetadata`, but does not include the full `Output`. In order to run `resolve` the head of the assumptions list must be available. So this PR adds a an explicit `metadata` field to `SegmentReceipt` and ensures metadata is preserved when passing through `lift` and `join` operations. The `pub(crate) Merge` trait is added to `ReceiptMetadata` to help facilitate this and also check for consistency through the steps of recursion. (This actually identified a consistency issue with how the PC was being handled in `risc0-zkvm` and in the recursion circuit.)

Resolves: https://github.com/risc0/risc0/issues/991

### DIFF
--- a/risc0/zkvm/src/host/recursion/prove/mod.rs
+++ b/risc0/zkvm/src/host/recursion/prove/mod.rs
@@ -408,7 +408,7 @@ impl Prover {
     ///
     /// This set of control IDs forms the closure of recursion programs that can be applied to the
     /// receipts. Verifiers will use the root of this tree to constrain the Prover to only apply
-    /// those programs in it's processing of receipts.
+    /// those programs in its processing of receipts.
     pub fn make_allowed_tree() -> MerkleGroup {
         MerkleGroup {
             depth: ALLOWED_CODE_MERKLE_DEPTH,


nategraf:
This PR is marked as "draft" because it contains some `TODO` comments I still need to address. All of those are minor, and so the PR is ready to review.

risc0/zkvm/src/host/receipt.rs
```diff
@@ -383,31 +383,43 @@ impl CompositeReceipt {
...
                 return Err(VerificationError::UnexpectedExitCode);
             }
-            if !metadata.output.is_none() {
+            if !receipt.metadata.output.is_none() {
                 return Err(VerificationError::ReceiptFormatError);
             }
-            prev_image_id = Some(metadata.post.digest());
+            expected_pre_state_digest = Some({
+                // Post state PC is stored as the "actual" value plus 4. This matches the join
+                // predicateb implementation. See ReceiptMetadata for more detail.
```
flaub, line 403:
```suggestion
                // predicate implementation. See [ReceiptMetadata] for more detail.
```

risc0/zkvm/src/host/recursion/prove/mod.rs
```diff
@@ -72,31 +75,110 @@ pub struct RecursionReceipt {
...
     Ok(SuccinctReceipt {
         seal: receipt.seal,
         control_id: receipt.control_id,
-        meta,
+        metadata: meta_decoded.merge(&segment_receipt.metadata)?,
     })
 }
 
-/// TODO
+/// Run the join program to compress two receipts of the same continuation into one.
```
flaub, line 102:
```suggestion
/// Run the join program to compress two receipts of the same session into one.
```

risc0/zkvm/src/host/recursion/receipt.rs
```diff
@@ -41,20 +41,24 @@ pub fn valid_control_ids() -> Vec<Digest> {
     all_ids
 }
 
-/// This struct represents a receipt for one or more [crate::SegmentReceipt]s
-/// joined through recursion.
+/// A succinct receipt, produced via recursion, proving the execution of the zkVM.
+///
+/// Using recursion a [CompositeReceipt] can be compressed to form a [SuccinctReceipt]. In this
```
flaub, line 46:
```suggestion
/// Using recursion, a [CompositeReceipt] can be compressed to form a [SuccinctReceipt]. In this
```

risc0/zkvm/src/host/protos/core.proto
```diff
@@ -47,13 +47,14 @@ message SegmentReceipt {
...
   string hashfn = 4;
+  ReceiptMetadata metadata = 5;
 }
 
 message SuccinctReceipt {
   protos.base.CompatVersion version = 1;
   bytes seal = 2;
   Digest control_id = 3;
-  ReceiptMetadata meta = 4;
+  ReceiptMetadata metadata = 4;
```
flaub, line 57: I don't think this is an issue now, but for future reference, if we were using JSON, this would be considered a breaking change.

risc0/zkvm/src/host/recursion/prove/mod.rs
```diff
@@ -66,53 +68,136 @@ const RECURSION_CODE_SIZE: usize = 21;
...
     pub seal: Vec<u32>,
     pub output: Vec<u32>,
 }
 
-/// TODO
-pub fn lift(receipt: &SegmentReceipt) -> Result<SuccinctReceipt> {
-    let mut prover = Prover::new_lift(&receipt.seal, ProverOpts::default())?;
+/// Run the lift program to transform an rv32im segment receipt into a recursion receipt.
+///
+/// The lift program is verifies the rv32im circuit STARK proof inside the recursion circuit,
```
flaub, line 77:
```suggestion
/// The lift program verifies the rv32im circuit STARK proof inside the recursion circuit,
```

risc0/zkvm/src/host/protos/core.proto
nategraf: Oh, that's a good point 😅

risc0/zkvm/src/host/recursion/prove/mod.rs
```diff
@@ -411,12 +493,58 @@ impl Prover {
...
+            .value()
+            .context("cannot resolve conditional receipt with pruned assumptions")?;
+        assumptions_tail.resolve(&corr.metadata.digest())?;
+
+        prover.add_input_digest(&assumptions_tail.digest(), DigestKind::Sha256);
+        prover.add_input_digest(&journal.digest(), DigestKind::Sha256);
+        Ok(prover)
+    }
+
     /// TODO
```
flaub, line 548: Ooops, looks like I missed this one when I ported over the recursion from private.

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -294,34 +297,40 @@ impl<'a> ExecutorImpl<'a> {
                             .context("Too many segments to fit in u32")?,
                         cycles,
                     );
-                    let segment_ref = callback(segment)?;
-                    self.segments.push(segment_ref);
```
flaub, line 298: Why not do this for all arms? I would expect the callback to happen for all cases (including Halted and Fault)

flaub (approved):

risc0/zkvm/src/host/server/exec/executor.rs
flaub: I think I see now that there is special processing for the final segment. I guess this is fine but makes me feel like a deeper refactor might be needed in the future.

nategraf: I agree. It's definitely not the cleanest code I've ever written 😅

risc0/zkvm/src/host/recursion/prove/mod.rs
nategraf: Added a comment to this in 06555ff5a4dcf86c1be0f3444c8692c6bbe69406

PR #1189, opened 2023-12-04
Update license-check.py

mantle0x:
The script currently assumes successful execution of all subprocess calls. It might be beneficial to add try-except blocks to handle potential errors in these calls, especially when dealing with external commands like git. and provide full source code. This updated script includes try-except blocks around subprocess calls and file reading operations. It handles errors by printing an error message and, in some cases, exits the script with a non-zero status if critical errors (like failing to find the repo root or list tracked files) occur. This makes the script more robust and user-friendly.

### DIFF
--- a/license-check.py
+++ b/license-check.py
@@ -30,7 +30,6 @@ EXTENSIONS = [
 
 SKIP_DIRS = [
     str(Path.cwd()) + "/templates/rust-starter",
-    # Groth16 verifier implementaion uses circom generated code under GPL3.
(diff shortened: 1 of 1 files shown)


flaub:
The thing is `check_output` already checks if the command failed and prints a message and then throws. Failures will turn into a non-zero exit code already. So I'm not sure what benefit we get from doing extra raises. If they were recoverable errors then I could see it.

https://docs.python.org/3/library/subprocess.html#subprocess.check_output

PR #1195, opened 2023-12-06
Drop _elf suffix from API

flaub:
This is a breaking change which causes the older doc-tests to fail. A refactor is made so that each doc-test version has its own workspace for deps.


mothran (approved):
I can't say I know enough about the doctests, but the API renaming looks good to me.

weikengchen:
Would it be useful to allow the old functions but with `#[deprecated = "message"]`?

PR #1203, opened 2023-12-10
feat: Profiling the exec's memory and elapsed time

hugo-blue:
cd examples/hello-world
cargo run --features "prove"
The log will be like:
Memory used before exec: Ok(37126144)
Memory used after exec: Ok(53264384)
exec elapse = 105 ms
Memory usage after proving: Ok(475447296)
proving peak mem = 472385488, prove elapse = 37568 ms
I know the factors of 391, and I can prove it!

It means that, in exec, only ~20MB is used (However, the peak memory is not clear yet). The time is only 105 ms, much less than total proving time 37 sec.

### DIFF
--- a/risc0/zkvm/Cargo.toml
+++ b/risc0/zkvm/Cargo.toml
@@ -51,6 +51,7 @@ serde = { version = "1.0", default-features = false, features = [
   "alloc",
   "derive",
 ] }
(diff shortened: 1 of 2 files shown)


hugo-blue:
Just for self-test. Closed this PR.

PR #1206, opened 2023-12-10
chore: put template hints inside a MD comment

austinabell:
It's a bit annoying to have to delete the hint/description lines for issues

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -6,15 +6,15 @@ labels: bug
 ---
 
 ## Bug Report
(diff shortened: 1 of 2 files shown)


flaub (commented):
I kind of like the template as is so that it's easier to see what needs to be filled in

austinabell:
> I kind of like the template as is so that it's easier to see what needs to be filled in

You still do, it just doesn't get put in the rendered version when it gets posted. You only lose on the ability to see that text in the "preview" before you post. Fair if that's important, though! Feel free to close if so.

flaub:
Cool I like rust's template

flaub (approved):

PR #1211, opened 2023-12-11
fix(crate-validator): Persist Nightly CI runs correctly

Cardosaum:
We had a bug previously where the `grep` command on the `index.json` file containing the previous results could return an error code in case of empty file. We now fix the problem by ensuring the call to `grep` always returns success.

Minor changes to the CI were also made to improve logging.


morgangibbs87 (approved):

mothran (approved):

flaub:
It feels like we're starting to reach the limit of what a workflow yaml file should contain. For complex steps we can always call a bash script or python as needs be.

flaub (approved):

Issue #1214, opened 2023-12-11
[BUG] `integration_test_completed_proof_manager` fails


austinabell:
## Bug Report

```console
test tests::manager::integration_test_completed_proof_manager ... FAILED

failures:

---- tests::manager::integration_test_completed_proof_manager stdout ----
thread 'tests::manager::integration_test_completed_proof_manager' panicked at ethereum-relay/src/tests/manager.rs:112:10:
deployment should succeed: MiddlewareError { e: MiddlewareError(JsonRpcClientError(JsonError(Error("missing field `baseFeePerGas`", line: 1, column: 39)))) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The issue is related to the `anvil` version being mismatched with what ethers-rs is compatible with, but it probably shouldn't be required to downgrade the foundry install just to run the test.

## Steps to Reproduce
1. `cd ./bonsai`
2. `cargo test`

## Expected behavior

Test should pass

## Your Environment
- risc0-zkvm version: `main`/`0.20.0-alpha`
- Rust version: 1.73
- Platform/OS: mac

anvil version:
`anvil 0.2.0 (0ae39ea 2023-12-11T00:27:32.440127000Z)`

austinabell:
Issue is with call to `eth_feeHistory`, fixed by https://github.com/gakonst/ethers-rs/pull/2683 but has not been released yet

austinabell:
This was fixed a while ago

PR #1215, opened 2023-12-11
add contributing doc

austinabell:
Opening draft for visibility, would like to get #1214 completed before this comes in to avoid having to include the version pinning semantics into this doc.


austinabell:
Also noticed that the rustfmt config is nightly only so emitting warnings. Maybe it's preferred that we change the CI and the default to running `cargo +nightly fmt --all`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+> NOTE: It is only important that you install `cargo-risczero` with a matching version of the `zkvm` crate when interacting with the proof system as a separate, pre-built process ([`ExternalProver`](https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.ExternalProver.html)), which is currently the default. If you are using the `prove` feature on the `risc0-zkvm` crate for the host, this will compile the proving system into the host binary.
+
+#### [Git Large File Storage](https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage)
+
+Once installed, run:
+
+```console
+git lfs install
+git lfs pull
+```
```
flaub, line 51: This should probably be the very first step after cloning the repository.

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+#### [Foundry](https://book.getfoundry.sh/getting-started/installation)
+
+```console
+curl -L https://foundry.paradigm.xyz | bash
+foundryup
+```
```
flaub, line 58: I think we should eventually make this part of the `xtask install` command.

CONTRIBUTING.md
austinabell: I don't have the cloning step, as it's pretty intuitive for anyone contributing, but I can add it if preferred. Moved this to the start, though

flaub:
I wonder if this is now unblocked?

austinabell:
> I wonder if this is now unblocked?

Oops, forgot about this, sorry. Seems like this might be a helpful starting point, but I don't feel strongly about it

flaub (approved):

PR #1216, opened 2023-12-11
Fix `cargo vendor` for examples

nategraf:
Using `cargo vendor` with the examples currently does not work. It fails with an error about multiple versions of `smartcore`. This PR fixes that error by using one source for `smartcore`

NOTE: We do not generally guarantee that `cargo vendor` will work. I'm opening this PR because an simple and immediate fix seems to be available, and it improves code health by removing a duplicate dependency.


nategraf:
This does not actually fix `cargo vendor`

PR #1218, opened 2023-12-12
feat: jwt-validator example

hmrtn:
Adds a JSON Web Token validator to demonstrate validating RS256 signed tokens in the guest.


examples/jwt-validator/.gitignore
```diff
@@ -0,0 +1,4 @@
+.DS_Store
+Cargo.lock
+methods/guest/Cargo.lock
+target/
```
flaub, line 4: This file can be dropped

examples/jwt-validator/LICENSE
flaub, line 1: This file can be dropped

examples/jwt-validator/rust-toolchain.toml
flaub, line 1: This file should be dropped, it needs to be consistent with the workspace

examples/jwt-validator/core/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+pub struct Issuer {
+    alg: Rsa,
+    s_k: RsaPrivateKey,
+    pub p_k: RsaPublicKey,
+}
+
+impl Issuer {
+    pub fn new(jwk_str: &str) -> Self {
```
austinabell, line 48: maybe this should be as a `FromStr` to allow using `parse()`, in case someone wanted to build upon and wanted more graceful error handling or wanted their own custom deserialization of these components?

austinabell: for sure! Was just a suggestion, definitely your call. It wasn't intuitive to me when reading the uses of Issuer/Validator that the constructor was parsing the string.

examples/jwt-validator/Cargo.lock
```diff
@@ -0,0 +1,2143 @@
+# This file is automatically @generated by Cargo.
```
austinabell, line 1: You can delete this lockfile, as it is using the shared examples one

austinabell (approved):

flaub (approved):

flaub:
I think we can land this, but we're missing an end 2 end test to ensure this has CI coverage and allows for regression testing. This is important so that if we make a change to the API, we can ensure that this test will be updated when the tests break.

flaub:
I suggest we make `main` call a function. This exact same function is called by the end 2 end test. See for example: https://github.com/risc0/risc0/blob/main/examples/hello-world/src/lib.rs#L53

PR #1225, opened 2023-12-13
Remove unnecessary TODO comments

Cardosaum:
A simple change just removing unnecessary comments.


bonsai/ethereum-relay/src/uploader/pending_proofs/pending_proof_request_future.rs
```diff
@@ -110,10 +110,6 @@ impl Future for PendingProofRequest {
                             ) {
                                 ("SUCCEEDED", true) => Ok(receipt_response),
                                 ("SUCCEEDED", false) => {
-                                    // TODO: Should we consider this a failure
```
austinabell, line 113: There's also one below this that I added on the draft PR that was more stylistic and I forgot to remove, if you feel like removing another :) I can do it later if not

tools/crates-validator/src/gen_profiles/mod.rs
```diff
@@ -14,7 +14,6 @@
 
 pub mod args;
 
-// TODO(Cardosaum): Check how to export only the functions that are needed
```
austinabell, line 17: Curious where the exported types here are used, I don't see them in this repo?

bonsai/ethereum-relay/src/uploader/pending_proofs/pending_proof_request_future.rs
Cardosaum: Good point, @austinabell! 

I deliberately opted to keep this comment there for now as I agree with you that it should be refactored, but I consider that it'd be better to do it in a different PR. 

With regards to the actual effort put on refactoring, the Relay seems to be low priority for now so I'm not sure the effort would be the most recommended. Feel free to take it if you feel like it though! :)

tools/crates-validator/src/gen_profiles/mod.rs
Cardosaum: They're only used [here](https://github.com/risc0/risc0/blob/1005dca1e9943810a9523e56e47bbd5c696505f6/tools/crates-validator/src/bin/gen-profiles.rs#L29). 

The types need to be public since the methods are and they return those intermediary types. Not sure if this would be the best approach, feel free to recommend refactoring if you feel appropriate!

flaub (approved):

PR #1226, opened 2023-12-13
Refactor binfmt image.rs

Cardosaum:
Apply pending suggestions from [1](https://github.com/risc0/risc0/pull/1146#pullrequestreview-1743002974) and [2](https://github.com/risc0/risc0/pull/1146#discussion_r1401320191)

### DIFF
--- a/risc0/binfmt/src/image.rs
+++ b/risc0/binfmt/src/image.rs
@@ -53,7 +53,7 @@ struct PersistentPageTableInfo {
 }
 
 #[derive(Clone, Serialize, Deserialize)]
-#[serde(from = "PersistentPageTableInfo", into = "PersistentPageTableInfo")]
+#[serde(try_from = "PersistentPageTableInfo", into = "PersistentPageTableInfo")]
 pub struct PageTableInfo {
     pub page_size: u32,
     page_size_po2: u32,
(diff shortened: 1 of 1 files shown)


risc0/binfmt/src/image.rs
```diff
@@ -224,25 +217,23 @@ impl MemoryImage {
...
-                    &page[(addr - page_start) as usize..(addr - page_start) as usize + bytes.len()],
-                );
-            }
-        };
+        if let Some(page) = self.pages.get(&page_idx) {
+            bytes.clone_from_slice(
+                &page[(addr - page_start) as usize..(addr - page_start) as usize + bytes.len()],
+            );
+        } else {
+            assert!(
```
austinabell, line 230: What is the logical difference here if this check is panicked rather than handled as it was before?

Cardosaum: Oh, you mean from `assert!` to `ensure!`? that's actually a good catch!
Thanks!

Cardosaum: We were supposed to remove panics from this code section. But while reverting to the preferred style Frank mentioned I make the mistake of pasting the old version verbatim, whereas I should've replaced the panic for an error variant. Thanks for calling that out!

austinabell: Awesome! Wasn't sure if it was a typo or if it was intentional :)

flaub (approved):

PR #1228, opened 2023-12-13
Add check for github references in version api docs

nategraf:
Working on the docs, I noticed quite a few links to GitHub not using the correctly pinned version reference (I also contributed to this). I wrote up a quick script, starting from the `license-check.py` script, to look for github links in the versioned API docs and ensure they reference the release branch corresponding to their directory. I've added this rule as a check in CI.


flaub (approved):
Thanks a lot! this is great

Issue #1209, opened 2023-12-11
[BUG] `sys_alloc_aligned` does not seem to impose a minimal alignment


weikengchen:
## Bug Report

There is a very confusing line of code in the exported syscall of `sys_alloc_aligned`, which is behind RISC Zero's memory allocator.

Recall that the bump memory allocator is here:
https://github.com/risc0/risc0/blob/main/risc0/zkvm/platform/src/syscall.rs#L651

```rust
#[cfg(feature = "export-syscalls")]
#[no_mangle]
pub unsafe extern "C" fn sys_alloc_aligned(bytes: usize, align: usize) -> *mut u8 {
    extern "C" {
        static _end: u8;
    }
    static mut HEAP_POS: usize = 0;
    let mut heap_pos = unsafe { HEAP_POS };
    if heap_pos == 0 {
        heap_pos = unsafe { (&_end) as *const u8 as usize };
    }

    let offset = heap_pos & (align - 1);
    if offset != 0 {
        heap_pos += align - offset;
    }

    // Ensure all allocations are minimally aligned to a word boundary.
    let align = usize::min(align, WORD_SIZE);

    let ptr = heap_pos as *mut u8;
    heap_pos += bytes;
    
    if crate::memory::SYSTEM.start() < heap_pos {
        const MSG: &[u8] = "Out of memory!".as_bytes();
        unsafe { sys_panic(MSG.as_ptr(), MSG.len()) };
    }

    unsafe { HEAP_POS = heap_pos };
    ptr
}
```

The problem rests on the "align".

- First of all, is it `min` or `max`? If we want to make sure that the alignment is at least WORD_SIZE, it should be max.
- Second, "let align" has no meaningful purpose because the `offset` and `heap_pos` have been updated before "let align = usize::min(align, WORD_SIZE);"? `align` is never used after this.

## Expected behavior
Expected `min` to be `max` and the align line be put up before the computation of offset.

SchmErik:
Thanks for filing this. So it looks like we've been aligning to what has been passed in. IIUC, this function is meant to be called from rust's `alloc` crate and `align` is supposed to be a power of 2 that represents how memory should be aligned as a part of `alloc::Layout`. I'm wondering if we could simply remove the max/min and pass on the responsibility to the caller of the function. @flaub do you have any thoughts or info on this?

weikengchen:
[...]

PR #1239, opened 2023-12-15
allocator: fix dead code and honor alignment requests larger than WORD_SIZE

SchmErik:
There are cases where guest code might request allocations with alignments greater than WORD_SIZE. For example, aligning a struct to a page boundary could help with locality when accessing those structures.

Fixes: https://github.com/risc0/risc0/issues/1209

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -233,15 +233,15 @@ mod test {
         build("../../risc0/zkvm/methods/guest/Cargo.toml");
         compare_image_id(
             "risc0_zkvm_methods_guest/multi_test",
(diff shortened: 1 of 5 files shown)


flaub:
I wonder if there's anyway to test this behavior? We probably should have had a test for this beforehand.

SchmErik:
> I wonder if there's anyway to test this behavior? We probably should have had a test for this beforehand.

One thing that comes to mind is to define a struct with a special alignment, do some sort of pointer arithmetic in guest code to determine that the requested alignment is being honored by doing something like

```
let a = MyStruct::new();
let b = MyStruct::new();
let diff = b.as_ptr() - a.as_ptr(); // do pointer arithmetic
assert_eq(diff, /* my struct alignment */)
```

flaub (approved):

PR #1240, opened 2023-12-15
Clippy Action Fix

morgangibbs87:
Fix clippy action so it fails properly and lists errors that need to be fixed.


flaub:
I would expect the check to cause the build to fail since clippy hasn't been running for awhile and will require lots of code changes to get it all compliant.

morgangibbs87:
@flaub - This is ready to review and just includes:

- filter_mode: nofilter and fail_on_error: true additions so jobs check all files and fail
- Flags removed for now so it just runs `cargo clippy`

The only `clippy` error showing up that should be fixed before merging:

```
reviewdog: Reporting results for "clippy"
  Error: [clippy] reported by reviewdog 🐶
  error: incorrect implementation of `partial_cmp` on an `Ord` type
     --> risc0/core/src/field/baby_bear.rs:277:1
      |
  277 | /  impl PartialOrd for Elem {
  278 | |      fn partial_cmp(&self, rhs: &Self) -> Option<Ordering> {
      | | ___________________________________________________________-
  279 | ||         decode(self.ensure_valid().0).partial_cmp(&decode(rhs.ensure_valid().0))
  280 | ||     }
      | ||_____- help: change this to: `{ Some(self.cmp(rhs)) }`
  281 | |  }
      | |__^
      |
      = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incorrect_partial_ord_impl_on_ord_type
      = note: `-D clippy::incorrect-partial-ord-impl-on-ord-type` implied by `-D clippy::all`
  
  Raw Output:
  risc0/core/src/field/baby_bear.rs:277:1:e:error: incorrect implementation of `partial_cmp` on an `Ord` type
     --> risc0/core/src/field/baby_bear.rs:277:1
      |
  277 | /  impl PartialOrd for Elem {
  278 | |      fn partial_cmp(&self, rhs: &Self) -> Option<Ordering> {
      | | ___________________________________________________________-
  279 | ||         decode(self.ensure_valid().0).partial_cmp(&decode(rhs.ensure_valid().0))
  280 | ||     }
      | ||_____- help: change this to: `{ Some(self.cmp(rhs)) }`
  281 | |  }
      | |__^
      |
      = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incorrect_partial_ord_impl_on_ord_type
      = note: `-D clippy::incorrect-partial-ord-impl-on-ord-type` implied by `-D clippy::all`
  
  
  __END__
  reviewdog: found at least one result in diff
 ```

I will add a few commands you can use locally to evaluate the other linting libraries in the ticket.

@ssh-nkar

flaub:
I think clippy belongs in the `check` job rather than for each test.

morgangibbs87:
@flaub - Good call. Added to `check` job, removed from `test` and `crates-validator` jobs.

flaub:
Thanks, I'll add a commit to this PR to get the build to pass

flaub (approved):

Issue #1236, opened 2023-12-15
Composition fails to resolve assumption in the client/server API


rlukata:
## Report
Composition fails with `value: sys_verify_integrity: failed to resolve journal_digest and image_id` when used with client/server API.

## Steps to Reproduce
1. Cargo.toml
```
risc0-zkvm = { workspace = true, features = ["client"] }
```
2. In the host:
```
let env = ExecutorEnv::builder()
        .add_assumption(receipt.clone().into())
        .write(&compute_image_id(METHOD_NAME_ELF)?)?
        .write(&receipt.journal.bytes)?
        .build()
        .unwrap();
let exec = default_executor();
let session = exec.execute(env, METHOD_NAME_ELF).unwrap();
```
3. In the guest:
```
let image_id: Digest = env::read();
let journal: Vec<u8> = env::read();
env::verify(image_id, &journal).unwrap();
```

## Expected behavior
The verify should pass.

## Your Environment
- risc0-zkvm version: 0.20.1 alpha
- Rust version: 1.72.1
- Platform/OS: M2 MacOS

nategraf:
I've confirmed that I forgot to add assumptions to the protobuf definition of `ExecutorEnv` and so assumptions are not transferred from client to server 🤦‍♂️ . https://github.com/risc0/risc0/blob/05f46ef83d7be3bd209070e67b57a5b45b56bfb1/risc0/zkvm/src/host/api/client.rs#L326-L347

Working on adding assumptions to the protobuf types and fixing this now

nategraf:
Deleted the "[Enhancement]" from the tittle since we have tags to indicate what type of issue this is

PR #1241, opened 2023-12-16
Send composition assumptions from client to server

nategraf:
Composition currently does not work in client/server mode because receipts are not sent from client to server. This PR adds the required plumbing to send the assumptions from client to server.

Resolves: #1236


nategraf:
Note that I moved `ReceiptClaim` from `core` to `base`. My inuition is that this is the better place for it under the understanding that `api` is the surface with which external developers are expected to interact, and core contains less stable details such as the `MemoryImage`. Since the `ReceiptClaim` is part of the public API surface, and also used internally, it seems like it needs to be in `base`.

flaub:
The distinction between api, core, and base is:
* api: of the wire messages between client/server. 
* core: the core data types, these are meant to be persistent formats for objects that may be stored in disk, in a DB, and temporarily over the write in the api. Ideally, the API would treat these types as opaque blobs. The idea is to split wire from storage following this best practice: https://protobuf.dev/programming-guides/api/#use-different-messages
* base: the glue that works in both api and core.

Given this I think `ReceiptClaim` belongs in `core` rather than `base`.

nategraf:
Ok, I've refactored this and got test coverage by refactoring most of the executor tests to run in client/server mode in addition to local mode based on the `RISC0_EXECUTOR` env var. I'm not convinced this is the best option, but I haven't thought of a better one yet. @flaub , do you have any thoughts?

https://github.com/risc0/risc0/pull/1241/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R178-R181

nategraf:
In https://github.com/risc0/risc0/pull/1241/commits/b4d07bbc93470f65d82af89c536fbf6190d8d7e3 I added a step to build r0vm. Given caching, I'm not sure whether or not that will have a significant impact on CI run times.

Given the fact that tests are being run with `-F prove`, an alternative to this would be to provide some way of running the test in a sort of "loopback server mode" where, without starting a new process or in a `fork`, the test process creates an IPC connected to itself/ the fork. This would be a deeper change.

nategraf:
> In [b4d07bb](https://github.com/risc0/risc0/commit/b4d07bbc93470f65d82af89c536fbf6190d8d7e3) I added a step to build r0vm. Given caching, I'm not sure whether or not that will have a significant impact on CI run times. I will check the difference once CI completes here.
> 
> Given the fact that tests are being run with `-F prove`, an alternative to this would be to provide some way of running the test in a sort of "loopback server mode" where, without starting a new process or in a `fork`, the test process creates an IPC connected to itself/ the fork. This would be a deeper change.

Looks like the new steps add about 2-4 minutes

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -49,11 +50,8 @@ fn run_test(spec: MultiTestSpec) {
         .unwrap()
         .build()
         .unwrap();
-    let session = ExecutorImpl::from_elf(env, MULTI_TEST_ELF)
-        .unwrap()
-        .run()
-        .unwrap();
-    assert_eq!(session.exit_code, ExitCode::Halted(0));
```
flaub, line 56: I think at this level of testing, I was hoping to use `ExecutorImpl` directly... I have to think more about how this will effect things but wasn't the intent of these tests thus far.

nategraf: I've reverted these changes in https://github.com/risc0/risc0/pull/1241/commits/742143e47cd42335f0c2e38d9b23fe5c772a60fb

PR #1244, opened 2023-12-18
Move compute_image_id into binfmt

flaub:


SchmErik (approved):
looks good to me

examples/hello-world/README.md
nategraf, line 1: Is this the result of a linter that we could be using in our local workflows?

nategraf (approved):

examples/hello-world/README.md
flaub: I have a markdown plugin that formats tables so they are easier to read in plain text.

PR #1249, opened 2023-12-19
include link to client API

pdg744:
This path adds a navigation path from the feature flags on the risc0 readme toward the client API docs in the crate documentation.


rlukata (approved):

flaub:
I'd like to avoid pointing directly the the client API. Instead, I think the idea is that users use the `Prover` trait, and behind, based on env vars, the system chooses client vs direct interfaces.

flaub:
To add more context, the client API isn't really intended to be directly used. This is however:

* https://docs.rs/risc0-zkvm/0.19.1/risc0_zkvm/fn.default_prover.html
* https://docs.rs/risc0-zkvm/0.19.1/risc0_zkvm/trait.Prover.html

flaub:
Also, take a look at all the examples, none of them use the `ApiClient` directly.

PR #1250, opened 2023-12-19
Refactor the IRiscZeroVerifier smart contract interface

nategraf:
This PR refactors the `IRiscZeroVerifier` smart contract interface to:

* Match the latest changes to the Rust verifier API, including updates to the `ReceiptClaim` to support composition.
* Incorporate feedback from outside smart contract review.


nategraf:
PR is blocked on having an ability to update the Groth16 proofs and make tests pass. Opening the PR now for visibility.

flaub:
@nategraf @capossele This just landed: https://github.com/risc0/risc0/pull/1295

Is this enough to be unblocked?

nategraf:
> @nategraf @capossele This just landed: https://github.com/risc0/risc0/pull/1295
> 
> Is this enough to be unblocked?
> 

That's likely a path forward. I'll give that a look and see about making a good way to keep these tests up to date.

The other avenue is to get a new SNARK from Bonsai when they have a 0.20 prerelease in staging.

nategraf:
Note that I started moving around files in the contracts directory and deleted what I believe is dead code in https://github.com/risc0/risc0/pull/1250/commits/8cfb36ae3e0f8722b81fb612c1c24d37cdaef5bb. If it turns out to be a problem, it would be expedient to revert that commit as it is not required to be in this PR

capossele:
With https://github.com/risc0/risc0/pull/1250/commits/b438417909da5662f6f322549e633557da12cb60 I've pushed a newly generated TestReceipt compatible with this branch.

The bootstrap tool is still on https://github.com/risc0/risc0/pull/1336 for now, as this branch requires to use a different proving key not compatible with the circom files we have on the repo

risc0/zkvm/src/receipt_claim.rs
```diff
@@ -157,37 +156,47 @@ impl From<InvalidExitCodeError> for DecodeError {
...
-    /// initiate a split if the next instruction will trigger a fault, or if the session limit is
-    /// reached.
+    /// [continuations]: https://dev.risczero.com/terminology#continuations
+    /// [segments]: https://dev.risczero.com/terminology#segment
     SystemSplit,
 
+    /// This indicates that the guest exited upon reaching the session limit set by the host.
+    ///
+    /// NOTE: The current version of the RISC Zero zkVM will never exit with an exit code of SessionLimit.
+    /// This is because the system cannot currently prove that the session limit as been reached.
```
capossele, line 191:
```suggestion
    /// This is because the system cannot currently prove that the session limit has been reached.
```

.github/workflows/main.yml
```diff
@@ -382,18 +382,12 @@ jobs:
       # This is a workaround from: https://github.com/actions/checkout/issues/590#issuecomment-970586842
       - run: "git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :"
       - uses: actions/checkout@v4
-        with:
-          lfs: true
       - uses: ./.github/actions/rustup
-      - uses: ./.github/actions/sccache
-        with:
-          key: Linux-default
```
flaub, line 388: This is useful for speeding up the `cargo run --bin cargo-risczero` and also building the tests. I don't think we should drop this.

nategraf: I've reverted this

capossele (approved):

PR #1251, opened 2023-12-19
Add receipts and assumptions to Bonsai SDK

rlukata:
This PR changes the Bonsai SDK to incorporate composition:
- [x] Add the option to include a list of UUID representing receipts as part of the `create_session` payload
- [x] Add the `upload_receipt` route
- [x] Add the test to both new additions
- [x] Update alpha_async crate
- [x] Update Bonsai SDK crate
- [x] Update ethereum crate
- [x] Update rest-api-mock-crate


mothran (approved):
Looks good from the API side!

flaub (approved):
LGTM. Should we also update `upload_img` to not require the image_id to be passed?

rlukata:
> LGTM. Should we also update `upload_img` to not require the image_id to be passed?

We need the image_id because we use it as the key of [pre-signed URLs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html).

flaub:
I think the idea @mothran had was we could do the `compute_image_id` behind the `upload_img` function so that in the future we could change the primary key/identifier for the REST call later.

mothran:
> I think the idea @mothran had was we could do the `compute_image_id` behind the `upload_img` function so that in the future we could change the primary key/identifier for the REST call later.

We don't have access to the byte array in the REST handler because we have users submit directly to s3 to avoid the payload size requirements. There are currently a few factors that are making this pattern annoying so I am considering swapping it out, TDB on what that will look like though.

Issue #1243, opened 2023-12-18
[BUG] Entry point function cannot do a far jump


weikengchen:
## Bug Report

Based on the conversation in the Discord server (see: https://discord.com/channels/953703904086994974/1184915191411003463), if the guest program is sophisticated enough, the entry point function, showed below, cannot call the `__start` function, because it is too far. 

@DefiCake reported this problem, from their codebase https://github.com/DefiCake/fuel-risc0-prover/.

```rust
// Entry point; sets up global pointer and stack pointer and passes
// to zkvm_start.  TODO: when asm_const is stablized, use that here
// instead of defining a symbol and dereferencing it.
#[cfg(target_os = "zkvm")]
core::arch::global_asm!(
    r#"
.section .text._start;
.globl _start;
_start:
    .option push;
    .option norelax;
    la gp, __global_pointer$;
    .option pop;
    la sp, {0}
    lw sp, 0(sp)
    jal ra, __start;
"#,
    sym STACK_TOP
);
```

where the definition of the `__start` function is as follows. This is the function that starts the guest main.
```rust
#[cfg(target_os = "zkvm")]
#[no_mangle]
unsafe extern "C" fn __start() -> ! {
    env::init();

    {
        extern "C" {
            fn main();
        }
        main()
    }

    env::finalize(true, 0);
    unreachable!();
}
```

`jal` is an RISC-V raw instruction that jumps to `__start`, where `__start` is represented by a signed 20-bit offset from the pc. This means that it can jump either backward up to 1MB or forward up to 1MB. But, if it takes more than 1MB to go to `__start` in the memory from the current `pc`, this instruction does not work.

This would result in a compilation time error, shown as follows. 

> = note: rust-lld: error: /Users/cusgadmin/r0/fuel-risc0-prover/target/riscv-guest/riscv32im-risc0-zkvm-elf/release/deps/prover-8ebe1890752312c1.prover.cc888d54-cgu.6.rcgu.o:(.text._start+0x14): relocation R_RISCV_JAL out of range: 1508240 is not in [-1048576, 1048575]; references __start
prover:           >>> referenced by musl_memcpy.c
prover:           >>> defined in /Users/cusgadmin/r0/fuel-risc0-prover/target/riscv-guest/riscv32im-risc0-zkvm-elf/release/deps/prover-8ebe1890752312c1.prover.cc888d54-cgu.6.rcgu.o

## Solution

[...]

PR #1257, opened 2023-12-19
Change jal to call in zkVM entrypoint

nategraf:
As described in https://github.com/risc0/risc0/issues/1243, our use of the `jal` instruction in our entrypoint breaks when the jump distance becomes too far. This PR changes the instruction to the `call` psuedo-instruction to allow the assembler flexibility to solve this issue and implement a far jump.

Co-authored-by: @weikengchen
Resolves: https://github.com/risc0/risc0/issues/1243

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -233,15 +233,15 @@ mod test {
         build("../../risc0/zkvm/methods/guest/Cargo.toml");
         compare_image_id(
             "risc0_zkvm_methods_guest/multi_test",
-            "f95f1a062cc386bcebe950e093c05b2064d0938d63c5522e7762618200204b29",
+            "1179a5fefa42c0ffd9c9defacb007f0e5c3b91b008726f1c334333e62998b1fd",
         );
         compare_image_id(
             "risc0_zkvm_methods_guest/hello_commit",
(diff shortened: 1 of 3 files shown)


SchmErik:
@nategraf you'll likely need to update the image ID's for the guest methods

weikengchen:
(That was what I meant by difficulty to update the Image IDs, which makes me reluctant to submit a PR.)

flaub:
In general, ImageIDs change pretty easily and so this has never really been a determinant on whether we make a change.

flaub (approved):

SchmErik:
> (That was what I meant by difficulty to update the Image IDs, which makes me reluctant to submit a PR.)

Hi @weikengchen, thanks for your contributions. I really appreciate your thoroughness and care that you take in describing issues and solutions! We're working on a contributor guide that would describe this but here's how to generate the image ID's for the test:

1) run docker daemon (on mac os, launch the docker desktop app)
2) run `cargo risczero build --manifest-path risc0/zkvm/methods/guest/Cargo.toml` directions are [here](https://github.com/risc0/risc0/blob/aaa1349dd57dbd13b0ae99cd126a508e642d181d/risc0/build/src/docker.rs)
3) update generated output values [here](https://github.com/risc0/risc0/blob/2c516c367ec3c905a5164eb1795d2beb69fe56dc/risc0/build/src/docker.rs#L236)

I hope this explanation makes it less painful!

PR #1268, opened 2023-12-22
Simplify entrypoint and main for std guests

nategraf:
I spent a momment to look into the minimal guest programs under `std` and `no_std` configurations, and found that in the `std` case we can drop the non-standard lines we use today. This PR changes the minimal `std` guest in our examples and templates to the following:

```rust
use risc0_zkvm::guest::env;

fn main() {
    env::commit("hello");
}
```

By contrast, here was the `std` guest we were using before:

```rust
#![no_main]

use risc0_zkvm::guest::env;

risc0_zkvm::guest::entry!(main);

fn main() {
    env::commit(&0xf00u64);
}
```

In the case of `no_std`, I attempted to improve upon the current minimal guest, and did find an alternative. What I discovered is that removing `#![mo_main]` means we will need to define a `start` langauge symbol. I found that the standard way to do that is with the `start` feature, resulting in the following guest.

```rust
#![no_std]
#![feature(start)]

use risc0_zkvm::guest::env;

#[start]
fn main(_: isize, _: *const *const u8) -> isize {
    env::commit("hello");
    0
}
```

Comapred with what we have, this seems less readable. As a result, I am not suggesting we change the minimal prgram from `no_std`. It will remain as:

```rust
#![no_std]
#![no_main]

use risc0_zkvm::guest::env;

risc0_zkvm::guest::entry!(main);

fn main() {
    env::commit(&0xf00u64);
}

```

This PR additionally removes the `pub` visibility modifier from all `main` functions except `r0vm`, as it is not needed.

### DIFF
--- a/risc0/zkvm/methods/guest/src/bin/multi_test.rs
+++ b/risc0/zkvm/methods/guest/src/bin/multi_test.rs
@@ -32,8 +32,9 @@ use risc0_zkvm::{
 use risc0_zkvm_methods::multi_test::{MultiTestSpec, SYS_MULTI_TEST};
 use risc0_zkvm_platform::{
     fileno,
-    memory::{self, SYSTEM}, PAGE_SIZE,
+    memory::{self, SYSTEM},
     syscall::{bigint, sys_bigint, sys_log, sys_read, sys_read_words, sys_write},
+    PAGE_SIZE,
 };
 
 risc0_zkvm::entry!(main);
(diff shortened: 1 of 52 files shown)


risc0/zkvm/methods/guest/src/bin/multi_test.rs
nategraf, line 1: This file appears to have escaped formatting until now

flaub (approved):

flaub:
Is this intended for 0.20?

Issue #1194, opened 2023-12-06
`store_region_to_guest_memory` does not generate a `MemorySet` trace event


weikengchen:
When I am testing our profiler (https://github.com/l2iterative/profiler0), I noticed the following result about syscalls from the profiler. The profiler is supposed to capture memory operations and attribute it to the preceding instruction. 

**Problem:** It does not work for some syscalls.

**Cause:** Functions that write memory using `store_region_to_guest_memory` do not trigger a `MemorySet` trace event.

------------------------------

Observation as follows.

Syscalls for SHA:
```
············ Cycle: 46660 => 47975: ecall at 0x204310 takes 1315 cycles
············ Cycle: 51139 => 52454: ecall at 0x204310 takes 1315 cycles
```

Syscalls for read:
```
············ Cycle: 29078 => 30243: ecall at 0x2044fc takes 1165 cycles
············ Cycle: 30327 => 31728: ecall at 0x2044fc takes 1401 cycles
```

But, syscalls for bigint:
```
········ Cycle: 57940 => 59044: ecall at 0x204364 that accesses 0x1ff430, 0x1ff434, 0x1ff438, 0x1ff43c, ..., 0x1ff44c takes 1104 cycles
```

The reason that this profiler does not work for the other syscalls is because:
- syscall SOFTWARE (https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/server/exec/executor.rs#L755) uses `monitor.store_region_to_guest_memory`
- Same for syscall SHA (https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/server/exec/executor.rs#L649)
- But, bigint uses `store_u32_to_guest_memory` (https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/server/exec/executor.rs#L704)

The `store_region_to_guest_memory` calls `store_region`, which calls `raw_store_u8` that calls `store_bytes`. Although `raw_store_u8` remembers to load the page (through `load_u8`) and mark the page as dirty (which is important), it just does not create a trace event. 
```rust
    fn store_bytes(&mut self, addr: u32, bytes: &[u8]) -> Result<()> {
        let info = &self.image.info;
        let page_idx = self.get_page_index(addr)?;
        let offset = addr % info.page_size;
        self.pages[page_idx as usize]
            .get_or_insert_with(|| Page {
                buf: self.image.load_page(page_idx),
            })
            .store_bytes(offset, bytes);
        Ok(())
    }

[...]

PR #1273, opened 2023-12-22
monitor: trigger `MemorySet` trace event for `store_region`

SchmErik:
This PR fixes an issue where some memory operations done by ecalls do not show up in profiler outputs.

Fixes: #1194


flaub:
Is this intended for 0.20?

flaub:
The only issue I see with this is that the trace might get very large for an event for every byte in a large region. WDYT if we adjusted the `MemorySet` to take `addr` and `region` which would be a `Vec<u8>`? Maybe that's a bad idea...

weikengchen:
Btw a separate solution is just to allow the profiler to access the memory if the profiler wants (passing an interface that allows profilers to access the memory), so that you do not have to provide the bytes together.

SchmErik:
> Is this intended for 0.20?

Yeah that was my intention. My understanding of the rc- cycles is that once an rc is declared, we want to backport bug fixes but not any new features until 0.20 is cut but I think this was mostly centered around bonsai integration. I don't think we've talked about what should be done in this case where we fix a bug during the rc cycles specifically for the zkvm so I'm definitely open for further discussion/refinement of this process. @flaub do you have any thoughts on this?

flaub:
I just need an indication that a PR is intended for a stable release, I think a way to do that is set the milestone.

SchmErik:
> I just need an indication that a PR is intended for a stable release, I think a way to do that is set the milestone.

I see what you mean. I set the milestone on this PR about 2 weeks ago. I can see it on my end in this PR. Maybe there's an issue in github? Here's a screen shot:

Screenshot 2024-01-03 at 1 55 20 PM

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -239,7 +239,7 @@ message TraceEvent {
 
   message MemorySet {
     uint32 addr = 1;
-    uint32 value = 2;
+    bytes region = 2;
```
flaub, line 242: This is a breaking change to protobuf. We can make breaking changes to the API, but we can't make breaking changes to protobuf messages. If we're changing the data type, we need to say:
```suggestion
    uint32 value = 2;
    bytes region = 3;
```

flaub (approved):

PR #1274, opened 2023-12-22
Add a basic example of composition

nategraf:
This PR adds a basic example of composition. Building upon the proof of known factors, this example exponentiates a number mod a composite with known factors. This takes inspiration from RSA encryption, and is kind of a stripped down (and insecure) version of verifiable encryption such that the message is provably decrypt-able by some party.

Resolves: #1299


examples/composition/methods/guest/src/main.rs
```diff
@@ -0,0 +1,52 @@
...
+    let n: u64 = env::read();
+    let e: u64 = env::read();
+
+    // x value that will be kept private.
+    let x: u64 = env::read();
+
+    // Verify that n has a known factorization.
+    env::verify(
+        MULTIPLY_ID,
+        bytemuck::cast_slice(&serde::to_vec(&n).unwrap()),
```
flaub, line 28: I feel like users requiring to use `bytemuck` is a code-smell.

examples/composition/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+    // Here is where one would send 'receipt' over the network to a second prover that might not
+    // know the factors, but wants to do additional work with the proven composite number.
+    let env = ExecutorEnv::builder()
+        // add_assumption makes the receipt to be verified available to the prover.
+        .add_assumption(multiply_receipt.into())
+        .write(&n)
+        .unwrap()
+        .write(&9u64)
+        .unwrap()
+        .write(&100u64)
```
flaub, line 32: It seems like writing a struct might be nicer here rather than writing each individual value (which is being borrowed, which is also weird).

examples/composition/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+        .unwrap()
+        .build()
+        .unwrap();
+
+    let receipt = default_prover().prove(env, EXPONENTIATE_ELF).unwrap();
```
flaub, line 37: Don't we need to `resolve` this proof? Is this a conditional receipt?

examples/composition/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    // Here the receipt might be sent to a third party that is interested in the exponentiation
+    // under a known composite. This is similar to a verifiable RSA encryption under a public key.
+    receipt.verify(EXPONENTIATE_ID).unwrap();
```
flaub, line 41: I would expect a "conditional" receipt to fail to verify.

flaub:
I really like this simple example

examples/composition/src/main.rs
LiveDuo: @flaub We found using a struct more complex since it requires having an extra crate (if that still applies). Tuples seem great in that regard.

examples/composition/src/main.rs
nategraf: The result of this call is a `CompositeReceipt` with all assumptions proven by attached receipts.

examples/composition/src/main.rs
nategraf: This verify works because the assumptions in the output of the top-level receipt are proven by attached receipts in within the `CompositeReceipt`.

risc0/zkvm/src/host/client/env.rs
```diff
@@ -64,6 +64,8 @@ where
 pub(crate) struct Assumptions {
     pub(crate) cached: Vec<Assumption>,
     #[cfg(feature = "prove")]
+    // An ordered list of assumptions accessed during execution.
+    // Each time an assumption is used, it is cloned and pushed to the head of the list.
```
flaub, line 68:
```suggestion
    // An ordered list of assumptions accessed during execution.
    // Each time an assumption is used, it is cloned and pushed to the head of the list.
    #[cfg(feature = "prove")]
```

examples/composition/methods/guest/src/main.rs
nategraf: Resolved this by accepting `&[impl Pod]` in the `env::verify` signature.

examples/composition/src/main.rs
nategraf: I decided to go with a tuple here.

flaub (approved):

examples/composition/README.md
```diff
@@ -0,0 +1,45 @@
...
+```bash
+cargo run --release
+```
+
+[installation guide]: https://dev.risczero.com/api/zkvm/quickstart
+
+## Use Cases
+
+Composition is a feature of the zkVM supporting verification of RISC Zero [receipts] in a guest program.
+With this, multiply zkVM programs can be _composed_ and produce a single receipt that verifies all computation done to reach the final result.
```
flaub, line 23:
```suggestion
With this, multiple zkVM programs can be _composed_ and produce a single receipt that verifies all computation done to reach the final result.
```

examples/composition/README.md
```diff
@@ -0,0 +1,45 @@
+# Composition Example
+
+This example demonstrates a basic usage of composition feature.
```
flaub, line 3:
```suggestion
This example demonstrates a basic usage of the proof composition feature.
```

examples/composition/README.md
```diff
@@ -0,0 +1,45 @@
...
+It builds upon the [hello world example](../hello-world), where the [guest program] verified the prover knows a factorization of a composite number.
```
flaub, line 4:
```suggestion
It builds upon the [hello world example](../hello-world), where the [guest program] verified that the prover knows a factorization of a composite number.
```

examples/composition/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Examples
+
+Some use cases for composition include:
+
+* Splitting a program into multiple parts, proven by different parties, to preserve privacy and data ownership of each party.
+  * E.g. Produce a proof for a database query by joining receipts from the query over each privately held shard.
+* Aggregating many proofs into one for efficeint batch verification.
+  * E.g. Produce a proof for a block of transactions, where each trasaction is itself verified by a proof.
+* Creating a single receipt for a workflow that make be split into many different operations.
```
flaub, line 36:
```suggestion
* Creating a single receipt for a workflow that might be split into many different operations.
```

risc0/zkvm/src/guest/env.rs
```diff
@@ -110,9 +110,23 @@ pub fn syscall(syscall: SyscallName, to_host: &[u8], from_host: &mut [u32]) -> s
 
 /// Verify there exists a receipt for an execution with `image_id` and `journal`.
 ///
+/// Calling this function in the guest is logically equivalent to verying a receipt with the same
```
flaub, line 113:
```suggestion
/// Calling this function in the guest is logically equivalent to verifying a receipt with the same
```

risc0/zkvm/src/guest/env.rs
```diff
@@ -199,6 +213,11 @@ impl std::error::Error for VerifyError {}
 /// Verify that there exists a valid receipt with the specified
 /// [ReceiptClaim].
 ///
+/// Calling this function in the guest is logically equivalent to verying a receipt with the same
```
flaub:
```suggestion
/// Calling this function in the guest is logically equivalent to verifying a receipt with the same
```

nategraf: Resolved in https://github.com/risc0/risc0/pull/1274/commits/5365a056a93c8f585e9d6c8aea739d3d011a3291

PR #1276, opened 2023-12-23
fix: Add ExecutorStats struct and log correct values

Cardosaum:
We had a bug previously where execution statistics were reported with inaccurate values. This PR seeks to solve this issue.

### DIFF
--- a/risc0/zkvm/src/host/server/exec/executor.rs
+++ b/risc0/zkvm/src/host/server/exec/executor.rs
@@ -14,6 +14,7 @@
 
 //! This module implements the Executor.
 
+use core::time::Duration;
 use std::{cell::RefCell, fmt::Debug, io::Write, mem, rc::Rc};
 
 use addr2line::{
@@ -23,6 +24,7 @@ use addr2line::{
 };
 use anyhow::{anyhow, bail, Context, Result};
 use crypto_bigint::{CheckedMul, Encoding, NonZero, U256, U512};
+use human_repr::{HumanCount, HumanDuration};
 use risc0_binfmt::{MemoryImage, Program, SystemState};
 use risc0_zkp::{
     core::{
@@ -45,7 +47,7 @@ use rrs_lib::{instruction_executor::InstructionExecutor, HartState};
 use serde::{Deserialize, Serialize};
 use sha2::digest::generic_array::GenericArray;
 use tempfile::tempdir;
-use tracing::{level_filters::LevelFilter, Level};
+use tracing::{instrument, level_filters::LevelFilter, Level};
 
 use super::{monitor::MemoryMonitor, profiler::Profiler, syscall::SyscallTable};
 use crate::{
@@ -379,22 +381,34 @@ impl<'a> ExecutorImpl<'a> {
         let segment_ref = callback(final_segment)?;
         self.segments.push(segment_ref);
 
-        tracing::info!("total_cycles = {}", self.total_cycles());
-        tracing::info!("segment_count = {}", self.segments.len());
-        tracing::info!("execution_time = {:?}", elapsed);
-
         if let Some(profiler) = self.profiler.take() {
             let report = profiler.borrow_mut().finalize_to_vec();
             std::fs::write(self.env.pprof_out.as_ref().unwrap(), report)?;
         }
 
-        Ok(Session::new(
+        let session = Session::new(
             mem::take(&mut self.segments),
             session_journal,
             exit_code,
             post_image,
             assumptions,
-        ))
+        );
+
+        match ExecutorStats::try_from((&session, elapsed)) {
+            Ok(stats) => stats.log(),
+            Err(error) => tracing::error!("failed to log executor stats: {}", error),
+        }
+
+        Ok(session)
+
+        // Returns a tuple `(x, y)` where:
+        // * `x`: Total number of cycles that a prover experiences. This includes
+        //   overhead associated with continuations and padding up to the nearest
+        //   power of 2. -> "total prover cycles: "
+        // * `y`: Total number of cycles used for executing user instructions. -> "user instruction cycles: "
+
+        // ratio of available computation used
+        // (both for individual segments and for the entire session)
     }
 
     fn split(&mut self, pre_image: Option<Box<MemoryImage>>) -> Result<()> {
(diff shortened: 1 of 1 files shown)


risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -771,3 +785,58 @@ impl<'a> ExecutorImpl<'a> {
...
+            self.total_prover_cycles.human_count_bare()
+        );
+        tracing::info!(
+            "user instruction cycles: {}",
+            self.user_instruction_cycles.human_count_bare()
+        );
+        tracing::info!(
+            "available computation used: {}%",
+            self.cycles_used_ratio.human_count_bare()
+        );
```
Cardosaum, line 831: I thought it might be interesting to show users the total computation that the prover will undergo (considering continuations overhead - padding up to the nearest power of 2)  as well as the "real" computation we're executing. 

This would be the "global" value - for the whole session. 
Was thinking maybe in adding also the 50th, 75th, 95th and 99th percentile of "utilized computation" for the segments - so then we count how we're using or not each segment individually as well. 

Thoughts on this? 

(This might be out of the scope for this PR and I'm happy to open a new one for this if you see fit).

weikengchen: (users will ask for help on how to squeeze their code to fit into a certain power of 2 but become desperate.

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -771,3 +776,58 @@ impl<'a> ExecutorImpl<'a> {
...
+
+        tracing::debug!(executor_stats = ?self);
+    }
+}
+
+impl TryFrom<(&Session, Duration)> for ExecutorStats {
```
flaub, line 828: Why are we using `TryFrom`, it doesn't seem like it can possibly fail?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -771,3 +776,58 @@ impl<'a> ExecutorImpl<'a> {
...
+    total_prover_cycles: u64,
+    user_instruction_cycles: u64,
+    cycles_used_ratio: f64,
+}
+
+impl ExecutorStats {
+    fn new(session: &Session, elapsed: Duration) -> Result<Self> {
+        let (total_prover_cycles, user_instruction_cycles) = session.get_cycles()?;
+        let cycles_used_ratio = user_instruction_cycles as f64 / total_prover_cycles as f64 * 100.0;
+        let segment_count = session.resolve()?.len();
```
flaub, line 793: Resolve is expensive, we should not call this here

SchmErik: how about `session.segments.len()` instead?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -14,6 +14,7 @@
 
 //! This module implements the Executor.
 
+use core::time::Duration;
```
flaub, line 17:
```suggestion
use std::time::Duration;
```

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -771,3 +776,58 @@ impl<'a> ExecutorImpl<'a> {
...
 }
+
+#[derive(Debug, Serialize, Deserialize)]
+struct ExecutorStats {
+    execution_time: Duration,
+    segment_count: usize,
+    total_prover_cycles: u64,
+    user_instruction_cycles: u64,
+    cycles_used_ratio: f64,
+}
```
flaub, line 787: It doesn't seem like `ExecutorStats` is used anywhere else, so a struct seems like overkill when we could just have:

```
impl Session {
  fn log() {
    // compute values, print them here
    ...
}
```

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -771,3 +776,58 @@ impl<'a> ExecutorImpl<'a> {
...
+        Ok(Self {
+            total_prover_cycles,
+            user_instruction_cycles,
+            cycles_used_ratio,
+            segment_count,
+            execution_time: elapsed,
+        })
+    }
+
+    #[instrument(name = "executor", skip(self))]
```
flaub, line 804: What is this for?

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -771,3 +776,58 @@ impl<'a> ExecutorImpl<'a> {
...
+        tracing::info!(
+            "total prover cycles: {}",
+            self.total_prover_cycles.human_count_bare()
+        );
+        tracing::info!(
+            "user instruction cycles: {}",
+            self.user_instruction_cycles.human_count_bare()
+        );
+        tracing::info!(
+            "available computation used: {}%",
```
flaub, line 820: Perhaps call this the `"cycle efficiency"`

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: Oh, thanks for pointing this out!

Would it be okay to call `session.segments.len()`? I was thinking that, since `resolve` is a fallible call, it might be the case that the real `segment_count` might not be exactly the same as `session.segments.len()` if it happens to be the case that the `resolve()` call fails?

Looking into the implementation to `resolve()` it seems like it's not the case, and `session.segments.len()` should always be valid - but would like to confirm :)

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: `ExecutorStats::new` could fail since it calls `session.get_cycles()`, which in turn is a fallible method. To avoid panics I chose to return a `Result` instead of panicking.
(I saw in a previous comment that `.resolve()` is an expensive call, but I don't think it'd be possible to avoid incurring in its cost since `.get_cycles` uses it internally...)

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: Thanks!

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: Changed ✅

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: This simplifies log filtering, enabling us to get only the executor statistics and nothing more using simply `RUST_LOG=[executor]=<log-level>`. If we don't specify this, we'd get a log of other `info` lines in the trace as well. Filtering would require us to specify the path to the executor, which would be way longer. 

I'm not sure if there would be a better way to perform filtering though....

risc0/zkvm/src/host/server/exec/executor.rs
Cardosaum: Makes sense, thanks for the input!

risc0/zkvm/src/host/server/exec/executor.rs
flaub: `session.segments.len()` is the right thing to use here.

flaub (approved):

flaub:
I think this is OK for now, but I think we can do a refactor that removes the need to do `resolve` just to compute the cycle counters.

PR #1278, opened 2023-12-24
fix: migrate to using enum for directory

### LINKED ISSUE #1223: [BUG] temp segment file storage not deleted after use
## Bug Report

Currently, when a temp directory is initialized (as the default) for file storage, [`into_path`](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.into_path) is immediately called:

https://github.com/risc0/risc0/blob/cee4239273e827e9d0a9b12714fb7c4c3175d231/risc0/zkvm/src/host/server/exec/executor.rs#L234

Which bypasses the deletion of the file on `Drop`, and this can get quite leaky with storage space if running a lot of workflows.

A couple options to have these be deleted is:
1. Handle ownership of the tempdir through having the segment path location stored as an enum (or sth like this)
2. Add a new type, which would have a `Drop` impl to delete the file, use this as the default
    - Can use https://docs.rs/tempfile/latest/tempfile/fn.tempfile_in.html as an alternative
    - Allows files to be incrementally deleted as the segments are proven
    - This solution, in the case of a temp dir creation, would not remove the directory itself at the end of execution

## Steps to Reproduce
1. 
2.
3.

## Your Environment
- risc0-zkvm version: 0.20.0-alpha (main)
- Rust version: 1.73
- Platform/OS: mac

austinabell:
Closes #1223 

Going to have another look through later to see if there is a cleaner way to handle this, and see if there is a clean way to handle warn logging on delete fail. I didn't love the implications of adding the warn log to the error:

```diff
diff --git a/risc0/zkvm/src/host/client/env.rs b/risc0/zkvm/src/host/client/env.rs
index 0ac1e3c6..157c082b 100644
--- a/risc0/zkvm/src/host/client/env.rs
+++ b/risc0/zkvm/src/host/client/env.rs
@@ -101,6 +101,16 @@ pub struct ExecutorEnv<'a> {
     pub(crate) pprof_out: Option<PathBuf>,
 }
 
+impl Drop for ExecutorEnv<'_> {
+    fn drop(&mut self) {
+        if let Some(DirectoryPath::TempDir(dir)) = self.segment_path.take().as_deref() {
+            if let Err(e) = std::fs::remove_dir_all(dir.path()) {
+                tracing::warn!("failed to close tempdir: {:?}", e);
+            }
+        }
+    }
+}
+
 impl<'a> ExecutorEnv<'a> {
     /// Construct a [ExecutorEnvBuilder].
     ///
diff --git a/risc0/zkvm/src/host/client/prove/bonsai.rs b/risc0/zkvm/src/host/client/prove/bonsai.rs
index af1c7ada..bda1193e 100644
--- a/risc0/zkvm/src/host/client/prove/bonsai.rs
+++ b/risc0/zkvm/src/host/client/prove/bonsai.rs
@@ -57,7 +57,7 @@ impl Prover for BonsaiProver {
         client.upload_img(&image_id_hex, elf.to_vec())?;
 
         // upload input data
-        let input_id = client.upload_input(env.input)?;
+        let input_id = client.upload_input(env.input.clone())?;
 
         // upload receipts
         let mut receipts_ids: Vec<String> = vec![];
```

### DIFF
--- a/risc0/zkvm/src/host/client/env.rs
+++ b/risc0/zkvm/src/host/client/env.rs
@@ -28,6 +28,7 @@ use bytemuck::Pod;
 use bytes::Bytes;
 use risc0_zkvm_platform::{self, fileno};
 use serde::Serialize;
+use tempfile::TempDir;
 
 use crate::serde::to_vec;
 use crate::{
@@ -67,6 +68,20 @@ pub(crate) struct Assumptions {
     pub(crate) accessed: Vec<Assumption>,
 }
 
+pub(crate) enum DirectoryPath {
+    TempDir(TempDir),
+    Path(PathBuf),
+}
+
+impl DirectoryPath {
+    pub(crate) fn path(&self) -> &Path {
+        match self {
+            Self::TempDir(dir) => dir.path(),
+            Self::Path(path) => path.as_path(),
+        }
+    }
+}
+
 /// The [crate::Executor] is configured from this object.
 ///
 /// The executor environment holds configuration details that inform how the
@@ -82,7 +97,7 @@ pub struct ExecutorEnv<'a> {
     pub(crate) input: Vec<u8>,
     pub(crate) trace: Vec<Rc<RefCell<dyn TraceCallback + 'a>>>,
     pub(crate) assumptions: Rc<RefCell<Assumptions>>,
-    pub(crate) segment_path: Option<PathBuf>,
+    pub(crate) segment_path: Option<Rc<DirectoryPath>>,
     pub(crate) pprof_out: Option<PathBuf>,
 }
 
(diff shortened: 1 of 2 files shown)


austinabell:
This CI fails because the tempfile dep is only pulled in when not the zkvm target, but the `ExecutorEnv` is exposed also to the zkvm target. This seems like it shouldn't be, since the segment storage wouldn't work within the zkvm, but instead of changing the API surface or making a hacky workaround, I'll leave this PR as is until it's decided what to do with that.

flaub:
I'm thinking the underlying problem is that the SegmentRef isn't really needed anymore. I'd like to do some discovery work on seeing if we can drop this which I think will make the need for temporary files go away entirely.

PR #1279, opened 2023-12-24
feat: implement attribute entry macro to replace declarative macro

### LINKED ISSUE #1259: Refactor the `entry` macro as an atrribute
Currently the `entry` macro in `risc0_zkvm::guest` has the syntax of `entry!(main)`. It would be more stylish (😎) to have it be an attribute so it would be `#[entry] fn main() {}`.

This issue only a style concern, with no functional impact. If we do want to do this, we would want to do it before 1.0, but if we don't it's fine.

austinabell:
Closes #1259 

There's not really a way to make this a non-breaking change without deprecations unless we want this macro to be in a different namespace. Assuming this breaking change might be acceptable.

### DIFF
--- a/risc0/macros/src/entry.rs
+++ b/risc0/macros/src/entry.rs
@@ -1,3 +1,17 @@
+// Copyright 2023 RISC Zero, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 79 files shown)


risc0/macros/src/entry.rs
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+
+    let result = quote_spanned! {input.sig.span()=>
+        // Type check the given path
+        const ZKVM_ENTRY: fn() = #fn_name;
+
+        // Include generated main in a module so we don't conflict
+        // with any other definitions of "main" in this file.
+        mod zkvm_generated_main {
```
austinabell, line 47: noting I'd like to prefix any exposed tokens with `__` to minimize chance of conflict. Wanted to minimize any possible semantic changes in this PR though.

Also seems like the const should just be defined within the module? Unless it's used for more than type checking?

flaub (changes requested):
I'd like to consider a few things before landing this:
* The name of the crate
* Whether we want to introduce procedural macros to the build of guest which already we have build time issues.

PR #1280, opened 2023-12-25
fix: remove unnecessary loads in ecall_sha

austinabell:
Currently this will read past the first digest until the block is full (2*digest). In the case that the second block is directly after the first, this read is just redundant. In the case that they are not, when reading just one, there are unnecessary reads done past the first block/digest.

I might be missing something here, but checking manually and running some basic tests work, so just opening this to see if and what changes/breaks to see if I am missing something.

### DIFF
--- a/risc0/zkvm/src/host/server/exec/executor.rs
+++ b/risc0/zkvm/src/host/server/exec/executor.rs
@@ -621,7 +621,7 @@ impl<'a> ExecutorImpl<'a> {
         tracing::debug!("Initial sha state: {state:08x?}");
         for _ in 0..count {
             let mut block = [0u32; BLOCK_WORDS];
-            for (i, word) in block.iter_mut().enumerate() {
+            for (i, word) in block.iter_mut().take(DIGEST_WORDS).enumerate() {
                 *word = self
                     .monitor
                     .load_u32_from_guest_addr(block1_ptr + (i * WORD_SIZE) as u32)?;


flaub:
This feels like something a compiler should be able to optimize away, and since this is on the host side, we don't need to shave off cycles. Do we have any perf numbers that shows this is an improvement?

austinabell:
> This feels like something a compiler should be able to optimize away, and since this is on the host side, we don't need to shave off cycles. Do we have any perf numbers that shows this is an improvement?

It isn't something that is optimized away, since it potentially is loading parts of memory that would otherwise not be accessed (if the pointers were not aligned in memory), and is calling `load_u32_from_guest_addr` for the second digest twice for the second digest and I'm not sure how the compiler would be able to recognize that it is redundant in those cases.

I can find a way to benchmark this when I have time, but it seemed like an unopinionated improvement since it loads strictly less data. Currently this is reading a full block of u32s into the buffer, then overwriting the second digest with the second loop.

Do you have any preferred metrics for benchmarking this that would be necessary to change your mind on this?

austinabell:
Also, just so we're on the same page, are you referring to only this commit [`8cf96ad` (#1280)](https://github.com/risc0/risc0/pull/1280/commits/8cf96ade5bd7e95f66dc4237fc60d6c21619b851) or the whole PR? Admittedly the former is more likely optimized by the compiler, although looking through asm code for it seemed it wasn't (low confidence) and just moved the buffer init outside of the loop just to be safe. Happy to revert that commit if that's all you're referring to.

flaub:
I think when it comes to perf optimizations, it's possible that code becomes slight less readable and maintainable. Sometimes optimizations are worth the reduction in code readability, but it's hard to gauge if we don't have measurements. So I tend to prefer to make the code more readable even if there are micro-optimizations that are being left on the table. If the measurement shows this to be above 1% of the time we spend, then I think we could at least consider it. If the code is just logically more simple, then I would also be in favor. But if it's less than 1% and the code is slightly harder to understand, then it doesn't seem worth the trade.

austinabell:
> I think when it comes to perf optimizations, it's possible that code becomes slightly less readable and maintainable. Sometimes optimizations are worth the reduction in code readability, but it's hard to gauge if we don't have measurements. So I tend to prefer to make the code more readable even if there are micro-optimizations that are being left on the table. If the measurement shows this to be above 1% of the time we spend, then I think we could at least consider it. If the code is just logically more simple, then I would also be in favor. But if it's less than 1% and the code is slightly harder to understand, then it doesn't seem worth the trade.

This PR wasn't intended to be a micro-optimization. I assumed this was a bug, as it was reading more memory from the guest than the guest would expect. I also assumed that this would be more readable, as it was unclear what was happening with all the indexing and different syntax between loops.

To explain why I mean this is reading more from memory than expected, the first loop will read `16` words from the guest at the pointer for the first digest (which is only `8` words):
https://github.com/risc0/risc0/blob/6ab6a551c00d5b9c83141641178ef0589621f70a/risc0/zkvm/src/host/server/exec/executor.rs#L623-L624

Which will only do redundant reads for `sys_sha_buffer` since the guest ensures that the two blocks are next to each other in memory (and this I suppose is just removing unnecessary reads), but if you call `sys_sha_compress` or any higher-level API:
https://github.com/risc0/risc0/blob/486940740221fef76047b37daad0bb059bda01b8/risc0/zkvm/platform/src/syscall.rs#L311

you'd be reading past the bounds of the `Digest` in memory, which could incur an unnecessary page load. Seems also preventative to avoid any unnecessary side effects or memory unsafety.

Maybe I'm misunderstanding something about the system, so let me know if this reasoning isn't valid!

flaub (approved):
Thanks for walking me thru this, LGTM

PR #1281, opened 2023-12-26
Fix: prove -> prove_elf 

nlok5923:
- `prove` requires multiple arguments including binaries image, verifier context, options and env
- `prove_elf` works with `env` and elf.


austinabell:
See https://github.com/risc0/risc0/pull/1195, this is intended

PR #1282, opened 2023-12-26
refactor!: remove SYS_READ_AVAIL syscall and update read logic

### LINKED ISSUE #1254: Remove `sys_read_avail`
In the effort to stabilize our syscall interface, we have decided to remove the `sys_read_avail` syscall. It does not have an analog in POSIX ([read(2)](https://man7.org/linux/man-pages/man2/read.2.html)), and introduces some difficulty in cases where the amount of available bytes might be undefined or non-deterministic with respect to guest execution.

https://github.com/risc0/risc0/blob/5a4d1fcd45c60243440619705290fedff9c692a7/risc0/zkvm/src/host/server/exec/syscall.rs#L508-L519

austinabell:
Don't look at this yet, it's hacky and haven't tested enough, just opening for visibility. I'll come back to this later, and may not be actively working on it if someone else wants to take over.

Closes #1254 

What this does is allow requesting more bytes than available, and in those cases it short circuits the requests and returns the amount of read bytes in `sys_read`.

### DIFF
--- a/risc0/zkvm/methods/guest/src/bin/multi_test.rs
+++ b/risc0/zkvm/methods/guest/src/bin/multi_test.rs
@@ -32,9 +32,8 @@ use risc0_zkvm::{
 use risc0_zkvm_methods::multi_test::{MultiTestSpec, SYS_MULTI_TEST};
 use risc0_zkvm_platform::{
     fileno,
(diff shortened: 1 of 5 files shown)


flaub:
We will be dropping `sys_read_avail` soon

PR #1287, opened 2024-01-02
Update copyright to 2024

SchmErik:
Happy New Year! I've omitted the circuit from this PR but I will make a separate update after updating WIP and bootstrapping. It looks like doc-test/version-0.18 is failing due to the `risc0-zkvm-methods` crate failing to build. I'm going to omit changes to that directory for the time being.


capossele (approved):

flaub:
Do we really need to modify every file? I think there's a deeper issue here.

PR #1302, opened 2024-01-08
Add `cargo risczero benchmark` command

### LINKED ISSUE #1301: [Feature] Add `cargo risczero benchmark` command
## Feature
Add a new command under `cargo risczero` called `benchmark` that will be responsible to prove a predefined guest code ([`bench`][bench-example]) and report timings on execution, proving and lifting steps.

## Motivation
We have an internal way to show local proving timings by running the [`loop`][loop-example] example. 
Its usage is not well documented and there's not much visibility around it as well. By providing a specific command under `cargo risczero` we would allow users to quickly run a benchmark on computing time in their machines, without necessarily downloading and compiling from source.

## Implementation
Leverage the already existing [`bench`][bench-example] guest code, and the code on [`loop` example][loop-example] in order to create a standalone command `bench` under `cargo risczero`

---

Related issues: 
#1245

[bench-example]: https://github.com/risc0/risc0/blob/e0008fab53fb8fdd83d7d710df7157e7a968a3ce/risc0/zkvm/methods/std/src/bin/bench.rs
[loop-example]: https://github.com/risc0/risc0/blob/d7d86b78f4b3007dd3fe5dbd4f09f9716659abbd/risc0/zkvm/examples/loop.rs

Cardosaum:
Closes #1301


risc0/cargo-risczero/Cargo.toml
```diff
@@ -33,7 +33,8 @@ reqwest-middleware = "0.2"
 reqwest-retry = "0.3"
 risc0-build = { workspace = true }
 risc0-r0vm = { workspace = true, optional = true }
-risc0-zkvm = { workspace = true, optional = true }
+risc0-zkvm = { path = "../zkvm", default-features = false, features = ["std", "getrandom"] }
+risc0-zkvm-methods = { path = "../zkvm/methods", default-features = false }
```
flaub, line 37: Unfortunately this isn't publishable, we won't be able to publish this on crates.io.  We need to come up with some other clever way to embed the guest.

Cardosaum: Oh... Thanks for noting this! (and the quick review 😄)

risc0/build/src/lib.rs
```diff
@@ -500,6 +505,9 @@ pub struct GuestOptions {
     /// Features for cargo to build the guest with.
     pub features: Vec<String>,
 
+    /// Binary names to build.
+    pub bin: Vec<String>,
+
```
flaub, line 510: What do we need this for?

risc0/cargo-risczero/Cargo.toml
```diff
@@ -33,13 +33,17 @@ reqwest-middleware = "0.2"
 reqwest-retry = "0.3"
 risc0-build = { workspace = true }
 risc0-r0vm = { workspace = true, optional = true }
-risc0-zkvm = { workspace = true, optional = true }
+risc0-zkvm = { workspace = true, default-features = false, features = ["std", "getrandom"] }
+risc0-zkvm-methods = { workspace = true, default-features = false }
```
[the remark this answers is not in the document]
Cardosaum: Yes! 

I was thinking in moving the dependencies we have on `risc0-zkvm-methods` to a new crate `risc0-zkvm-methods-core` that we could publish, this way we would solve the issue. That way both `risc0-zkvm-methods` and `cargo-risczero` could depend on `risc0-zkvm-methods-core`.

risc0/build/src/lib.rs
Cardosaum: Oh, I changed this so we would be able to compile a single binary using docker, like specifying the `--bin` flag. (it would break the compilation step otherwise). 

But since today's meeting we discussed about not needing it, I think I'll just revert the commit that introduced changes to `risc0-build`.

risc0/cargo-risczero/Cargo.toml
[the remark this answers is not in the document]
flaub: I don't think any of the benchmark binaries need any 'core' crates because the interface is very simple. We pass in the number of iterations to run, so there's no types that need to be shared between host and guest.

Cardosaum:
@flaub , I tried to copy your logic from `gen-receipts` and apply the same here so we can include the elf binary in compile time. 

I have one question on how we ensure that the file generated by `xtask` will be present when we build `cargo-risczero` as now we have a dependency on running `xtask` before being able to compile `cargo-risczero`. 

Would you have some direction on that? 

Also, the whole logic for doing the `benchmarks` got way simpler this way, thanks for letting me know there was this option for doing it at compile time! ^^

risc0/cargo-risczero/Cargo.toml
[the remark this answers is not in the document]
Cardosaum: Oh I see... I followed this approach because this would be similar to what we do under `/examples` having a `core` that can be used both in `guest` and `host`. 

I did the change before seeing this comment. Please let me know if you'd like to revert the new `risc0-zkvm-methods-core` and I'll change it 🙇‍♂️

risc0/build/src/lib.rs
Cardosaum: [Reverted](https://github.com/risc0/risc0/pull/1302/commits/cff8bcd7f98728256955e8e340475f65ae238d05)

risc0/cargo-risczero/build.rs
```diff
@@ -37,7 +37,29 @@ mod runtime {
...
+    fn build_elf(path: &str) {
+        let is_success = std::process::Command::new("cargo")
+            .args(["xtask", "gen-benchmark", "--path", path])
+            .status()
+            .unwrap()
+            .success();
+        assert!(is_success);
+    }
+}
+
```
Cardosaum, line 59: I think I might have found a way to compile it automatically during build time! 😄

Cardosaum: Giving credit where credit is due, I found it already in our code base :)

SchmErik (approved):
Looks good to me - I think it would be good for @flaub to review and approve before merging

flaub (changes requested):
Thanks, I'll take a pass at this at some point to get it landed

flaub:
I still think this is a good idea, but @nvzqz could you take a pass at coming up with a design for this?

PR #1304, opened 2024-01-09
NFC: Rename Fp4 -> FpExt

flaub:


risc0/build_kernel/kernels/metal/fpext.h
```diff
@@ -26,83 +26,83 @@
 #define BETA Fp(11)
 #define NBETA Fp(Fp::P - 11)
 
-/// Intstances of Fp4 are element of a finite field F_p^4.  They are represented as elements of
+/// Intstances of FpExt are element of a finite field F_p^4.  They are represented as elements of
```
tzerrell, line 29: Not for this PR I don't think, but flagging that this is a location where docs will need to be updated.

tzerrell (approved):

PR #1309, opened 2024-01-10
 zkVM: remove SYS_READ_AVAIL and change SYS_READ

SchmErik:
This change removes SYS_READ_AVAIL because it's not POSIX compliant and
introduces some difficulty in cases where the number of available bytes might
be undefined or non-deterministic concerning guest execution.

This change removes the SYS_READ_AVAIL support from the host and guest. The
guest syscall libraries used SYS_READ_AVAIL to create a buffer of the correct
size. Due to its removal, we're allowing cases where EOF is encountered earlier
and a shorter read than requested is possible through the guest syscall
handlers provided. The underlying handler in the prover has changed to
reflect this reality by removing asserts. This is based on @austinabell's
solution with minor adjustments.

Summary of guest-side syscall handler changes:

The previous implementation relied on the number of bytes returned by
SYS_READ_AVAIL or has been requested. Logic in SYS_READ has been changed to use
the actual amount of bytes read from the syscall. Also, if the guest requests
more bytes than available, it short-circuits the requests and returns the
amount of read bytes.

Summary of host changes:

Changed the host to allow the guest to request more bytes than
available. In this case, the host will hit an EOF and return the buffer to the
caller while still allowing the last unaligned word to be read.

Fixes: #1254

Co-authored-by: Austin Abell <austinabell8@gmail.com>


weikengchen:
just a quick question: given that now sys_read works even if the guest requests more bytes, it is actually totally possible to implement, for example, BufReader, over the stdin?

austinabell:
Whoever merges this, can you remove the @ from my handle on the squashed commit so that I don't get notified every time someone forks the repo? :D

> just a quick question: given that now sys_read works even if the guest requests more bytes, it is actually totally possible to implement, for example, BufReader, over the stdin?

Just as possible as it was before, if I'm remembering correctly. From the guest side it would call to `SYS_READ_AVAIL` and use that value if it was lower than the requested amount passed into `sys_read`. I'd be curious what the perf comparison would be of using sth like https://doc.rust-lang.org/std/io/struct.BufReader.html with different capacities in the guest with different types.

flaub (approved):

Issue #1306, opened 2024-01-09
[BUG] composition: error when lifting assumption with ExitCode::Fault


SchmErik:
## Bug Report

I've made an initial attempt to try getting proof of fault to work using composition and to begin, I've copied parts of the code from the e2e recursion test to lift the assumption. I'm writing this down so that we can get to the root cause of this issue. We're going to be removing proof of fault from the zkVM and we might not need this so this issue might not need to be fixed. The following code test fails:

```rust
#[test]
fn proof_of_fault() {
    tracing::info!("running guest code");
    // This test does a proof of fault by running the fault checker.
    let (guest_receipt, fault_state) = prove_fault();

    tracing::info!("running fault checker");
    // Test that proving results in a success execution and unconditional receipt.
    let env = ExecutorEnv::builder()
        .write(&fault_state)
        .unwrap()
        .write(&guest_receipt.get_claim().unwrap())
        .unwrap()
        .add_assumption(guest_receipt.clone().into())
        .build()
        .unwrap();
    let receipt = get_prover_server(&prover_opts_fast())
        .unwrap()
        .prove(env, FAULT_CHECKER_ELF)
        .unwrap();

    let composition_receipt = receipt.inner.composite().unwrap().clone();
    assert_eq!(composition_receipt.segments.len(), 1);
    let conditional_segment_receipt = composition_receipt.segments[0].clone();

    assert_eq!(composition_receipt.assumptions.len(), 1);
    let assumption_receipt = composition_receipt.assumptions[0]
        .composite()
        .unwrap()
        .clone();
    assert_eq!(assumption_receipt.segments.len(), 1);
    assert_eq!(assumption_receipt.assumptions.len(), 0);
    let assumption_segment_receipt = assumption_receipt.segments[0].clone();

    // Lift and join them  all (and verify)
    tracing::info!("Lifting assumption");
    let lifted_assumption = lift(&assumption_segment_receipt).unwrap(); // PANIC: this line results in a panic due to the unwrap()
    lifted_assumption
        .verify_integrity_with_context(&VerifierContext::default())
        .unwrap();
    tracing::info!("Lift assumption claim = {:?}", lifted_assumption.claim);
}
```
Here's the code for the `prove_fault` function.
```rust
fn prove_fault() -> (Receipt, FaultState) {
    let env = ExecutorEnv::builder()
        .write(&MultiTestSpec::Fault)
        .unwrap()
        .build()
        .unwrap();
[...]

PR #1313, opened 2024-01-10
Allow merge of ReceiptClaim objects with system exit code 2

nategraf:
This PR is a potential solution to https://github.com/risc0/risc0/issues/1306

The root cause of this issue is that we currently have three `ExitCode` values that map to the system exit code value of 2. These are `SystemSplit`, `Fault` and `SessionLimit`. From the circuit "point-of-view" all three of these are the same, and result in the system exit code 2, and user exit code 0. The host is allowed to pick any of these three to report on the receipt claim without changing the digest.

The solution in this PR is to adjust the equality check in the `Merge` trait implementation to accept two exit codes are equal if they are the same when encoded as a pair. This aligns with how `ExitCode` is treated elsewhere.

This issue is a symptom of the fact that our code does not ditinsguish between guest-reported/proven and host-reported/unproven exit codes. A more robust solution would be to remove the `Fault` and `SessionLimit` exit codes from the `ExitCode` enum, as they are not actually checked by the guest. In the future, with the introduction of a guest kernel, it will be possible to provable check that the system exited in a fault state, or that it reached some specified cycle count.

Resolves: https://github.com/risc0/risc0/issues/1306

### DIFF
--- a/examples/composition/methods/build.rs
+++ b/examples/composition/methods/build.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 RISC Zero, Inc.
+// Copyright 2024 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/examples/composition/methods/src/lib.rs
+++ b/examples/composition/methods/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright 2023 RISC Zero, Inc.
+// Copyright 2024 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


SchmErik:
confirming that this fixes the issue!
```
   Compiling risc0-circuit-recursion v0.21.0-alpha.1 (/Users/erik/risc0/risc0/circuit/recursion)
   Compiling risc0-circuit-rv32im v0.21.0-alpha.1 (/Users/erik/risc0/risc0/circuit/rv32im)
    Finished test [optimized + debuginfo] target(s) in 7m 56s
     Running unittests src/lib.rs (target/debug/deps/risc0_zkvm-747aefa6025b566f)

running 1 test
test host::server::prove::tests::proof_of_fault has been running for over 60 seconds
test host::server::prove::tests::proof_of_fault ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 125 filtered out; finished in 91.62s

   Doc-tests risc0-zkvm

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 18 filtered out; finished in 0.02s
```

nategraf:
Unrelated, but odd that this copyright issue made it through CI before 4758dc43930215d8b5a21f82a2c2fb83631b752f

nategraf:
I'm going to refactor the solution to this issue after discussion with @flaub

nategraf:
Superseded by https://github.com/risc0/risc0/pull/1317

PR #1314, opened 2024-01-10
Add composition terminology to website and otherwise improve docs

nategraf:
This PR adds terminology entries to the website for terms/concepts related to composition. Additionally is adds a bit more documentation to types and methods related to composition, and additionally refactors the `ExecutorEnv::add_assumption` method to accept `impl Into<Assumption>` which allows the user to pass in a receipt without any transformation.


risc0/zkvm/src/host/receipt.rs
```diff
@@ -97,10 +99,8 @@ use crate::{
 /// ```
 ///
 /// The public outputs of the [Receipt] are contained in the
-/// [Receipt::journal]. We provide serialization tools in the zkVM
-/// [serde](crate::serde) module, which can be used to read data from the
-/// journal as the same type it was written to the journal. If you prefer, you
-/// can also directly access the [Receipt::journal] as a `Vec<u8>`.
+/// [Receipt::journal]. You can use [Journal::decode] to deserialize the journal are typed and
```
tzerrell, line 102:
```suggestion
/// [Receipt::journal]. You can use [Journal::decode] to deserialize the journal as typed and
```

risc0/zkvm/src/host/receipt.rs
```diff
@@ -659,11 +659,15 @@ impl SegmentReceipt {
     }
 }
 
-/// An assumption attached with a guest execution as a result of calling
+/// An assumption attached with to a guest execution as a result of calling
```
tzerrell, line 662:
```suggestion
/// An assumption attached to a guest execution as a result of calling
```

website/docs/terminology.md
```diff
@@ -142,11 +179,31 @@ See also: [Prover documentation], [Executor]
 ### Receipt
 
 [receipt]: #receipt
+[receipts]: #receipt
 
-A receipt attests to valid [execution] of a [guest program]. [Verifying] the receipt provides cryptographic assurance that the [journal] was indeed constructed using the expected [circuit] and the expected [imageID].
+A receipt attests to valid [execution] of a [guest program].
+[Verifying] the receipt provides cryptographic assurance that the [journal] was indeed constructed using the expected [circuit] and the expected [imageID].
 The receipt consists of a [journal] and a [seal].
```
tzerrell, line 186: Do we want to update this to point at `[receipt claim]` instead of `[journal]`? I'm a bit ambivalent but it seems like "journal + seal" is not quite what a receipt is these days...

tzerrell (approved):
I really like this! I think this is a good conceptual framework for thinking about composition, and I think you explain it well.

website/docs/terminology.md
nategraf: Good suggestion. I've addressed this in 049002808dad918709891283c55521f07000f405

nategraf:
Opened https://github.com/risc0/risc0/pull/1324 to address the issue with tests on `release-0.18`

PR #1317, opened 2024-01-11
Distinguish between and avoid use of ExitCode::Fault and SessionLimit

nategraf:
Currently our code does not distinguish between guest-reported/proven and host-reported/unproven exit codes. This PR addresses this issue by giving `ExitCode::SystemSplit`, `Fault` and `SessionLimit` distinct exit codes.

The rv32im circuit does not support exiting with with a verified status of `SessionLimit` or `Fault`. As a result, these exit codes should be unused and reserved for future revision of the circuit. This PR refactors the executor to reflect this and align with the fact that the prover will never write these statuses to the seal.

Supersedes: #1313
Resolves: #1306


risc0/zkvm/src/receipt_claim.rs
```diff
@@ -160,32 +160,34 @@ impl std::error::Error for DecodeError {}
...
-    /// execution ending in `SystemSplit`.
-    // TODO(1.0): Refine how we handle the difference between proven and unproven exit codes.
+    /// NOTE: The current version of the RISC Zero zkVM will never exit with an exit code of Fault.
+    /// This is because the system cannot currently prove that a fault has occured.
     Fault,
+
+    /// This indicates that the guest exited upon reaching the session limit set by the host.
+    ///
+    /// NOTE: The current version of the RISC Zero zkVM will never exit with an exit code of SessionLimit.
+    /// This is because the system cannot currently prove that a fault has occured.
```
flaub, line 189:
```suggestion
    /// This is because the system cannot currently prove that the session limit as been reached.
```

risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -304,11 +305,6 @@ impl<'a> ExecutorImpl<'a> {
                             self.segments.push(segment_ref);
                             self.split(Some(post_image.into()))?
                         }
-                        ExitCode::SessionLimit => {
-                            let segment_ref = callback(segment)?;
```
flaub, line 308: I'm wondering how we can get away with not having this part of the callback? I wonder if we could just leave the ExitCode stuff alone for the moment and I can refactor the SegmentRef stuff and then later revisit ExitCode?

nategraf: I'm not sure I understand what you are pointing to. Are you raising a concern about dropping the `callback` call that occurs as this location prior to this PR?

One option is simply to close this PR and leave #1306 open. I don't think it affects any known use cases, so maybe that's the right call. What do you think?

flaub (approved):

nategraf:
Somehow files are getting merged into `main` with incorrect copyright headers. I'm going to spend (no more than) 30 minutes looking into what is letting these get through https://github.com/risc0/risc0/actions/runs/7559670229/job/20583980072?pr=1317

nategraf:
> Somehow files are getting merged into `main` with incorrect copyright headers. I'm going to spend (no more than) 30 minutes looking into what is letting these get through https://github.com/risc0/risc0/actions/runs/7559670229/job/20583980072?pr=1317

I figured out what is going on. If a PR was opened in 2023, and then merged in 2024, any files that were not changed in 2024 on the branch will have a different most-recent-commit date after the squash and merge (i.e. the date of the squash+merge commit) than when it was open as a pull request. As a result it will pass when in CI, then fail on `main` after the merge.

PR #1319, opened 2024-01-11
Systematically fix all the typos discovered with the help of JetBrains

weikengchen:
We are going to see more "small" typo PRs in the coming months due to airdrop expectation, though to my belief, RISC Zero never mentions having a token. It is good though to have early participants. Nevertheless, it sometimes becomes a headache when each of the PRs is too small, and it takes resources to do CI and many manual reviews.

A solution to handle this issue is to fix **ALL** the obvious typos systematically in a single attempt, so that even if in the future people discover new typos, these typos must be hidden deep, and the efforts can be appreciated. 

An approach that Alessandro Chiesa taught us in the grad school is to use a spellchecker and manually go through every reported typo. Today we can just use JetBrains. There is a secret tool in Code -> Analyze Code -> Run inspection by name... => "Typo". 

JetBrains proposes about 4000 typos---this is pedantic because JetBrains would consider revm, zkvm, etc to be typos. Human inspection is now needed. After a manual walkthrough, this PR contains all the typos that survive the human inspection. 

Typos are not avoidable. Compared to other projects, RISC Zero does write a lot of documents in the code. 

In addition, just to leave a last sentence: if you are using Mac, my experience in the grad school is that you might be using a keyboard that has been recalled. All the typos appear to be a result of sticky keyboard. https://support.apple.com/keyboard-service-program-for-mac-notebooks  And by "a small percentage" in Mac's announcement, it means all. The problem is about the butterfly keyboard design used throughout those years.


bonsai/examples/governance/script/Deploy.s.sol
```diff
@@ -75,7 +75,7 @@ contract Deploy is Script, BonsaiCheats {
         } else {
             // Deploy an IRiscZeroVerifier contract instance. Relay is stateless and owner-less.
             IRiscZeroVerifier verifier;
-            address verifierAddr = vm.envOr("DEPLOY_VERFIER_ADDRESS", address(0));
+            address verifierAddr = vm.envOr("DEPLOY_VERIFIER_ADDRESS", address(0));
```
weikengchen, line 78: It is important to call out that there are typo fixes that could lead to compatibility issues.

Here is the Bonsai governance Solidity script. It uses an environment variable called "DEPLOY_VERFIER_ADDRESS". Fixing this typo may cause issues with the upstream.

risc0/zkp/src/taps.rs
```diff
@@ -253,7 +253,7 @@ impl<'a> Debug for TapRef<'a> {
             .field("group", &self.group())
             .field("offset", &self.offset())
             .field("back", &self.back())
-            .field("conmbo_id", &self.combo_id())
+            .field("combo_id", &self.combo_id())
```
weikengchen, line 256: This would change how debug information is reported for TapRef.

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -12,7 +12,7 @@ message ServerRequest {
     ProveSegmentRequest prove_segment = 3;
     LiftRequest lift = 4;
     JoinRequest join = 5;
-    IdentityP254Request identiy_p254 = 6;
+    IdentityP254Request identity_p254 = 6;
```
weikengchen, line 15: My experience with "proto" is that changing the name does not break the compatibility, and the identity-p254-request is not in the public release. Please check if this is the case.

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -572,8 +572,8 @@ mod sys_verify {
             .is_err());
 
         // TODO(#982) With conditional receipts, implement the following cases.
-        // verify with proven corraboration in verifier success.
-        // verify with unresolved corraboration in verifier success.
+        // verify with proven corroboration in verifier success.
+        // verify with unresolved corroboration in verifier success.
```
weikengchen, line 576: I am aware that RISC Zero has specifically picked the word "corroboration" to mean evidence.

weikengchen:
And just a note that some pending typo PRs are still fairly novel as they found non-spelling issues. 

If people are interested, the current repository mentions SHA-256 a few times in the hashing HAL and others, but the hashfn is indeed being parameterized now, and the SHA-256 that the comment said may be actually Poseidon (and by default, Poseidon).

risc0/zkvm/src/host/protos/api.proto
flaub: Yes and no. It breaks if the encoding is JSON, but at this time we don't use any JSON encodings, so this should be fine.

flaub:
This looks really interesting: https://github.com/privacy-scaling-explorations/folding-schemes/pull/76

weikengchen:
oh my god we got headlined in PSE.

flaub:
Superseded by #1683

PR #1321, opened 2024-01-11
Change "corroborating receipt" to "assumption receipt"

nategraf:
Although very clever, the term "corroborating receipt" was introducing uncessesary confusion.
This removes the use of the word "corroborating" with reference to receipts that attest the the validity of assumptions.


rlukata (approved):
love it!

weikengchen:
Just to squeeze a quick conversation. If we want to simplify the terminology further, isn't "fact" a good idea?

Facts have receipts.
sys_verify is "require fact"
"resolve assumptions" is basically "supply facts"

PR #1322, opened 2024-01-12
Use 5th degree extension field to increase bits of security

flaub:


risc0/zkp/src/core/hash/sha/mod.rs
```diff
@@ -442,12 +442,13 @@ pub mod testutil {
     fn hash_extelems<S: Sha256>(len: usize) -> Digest {
         let items: Vec<BabyBearExtElem> = (0..len as u32)
             .map(|x| {
-                BabyBearExtElem::new(
-                    BabyBearElem::new(x * 4),
+                BabyBearExtElem::new([
+                    BabyBearElem::new(x * 4 + 0),
```
pdg744, line 446: what is role of the `4` here?

risc0/zkp/src/verify/mod.rs
```diff
@@ -320,10 +320,10 @@ where
         // tracing::debug!("Result = {result:?}");
 
         // Now generate the check polynomial
-        // TODO: This currently treats the extension degree as hardcoded at 4, with
+        // TODO: This currently treats the expansion degree as hardcoded at 4, with
```
pdg744, line 323: "expansion degree" isn't a term i think we should be using; `degree` can refer to a feature of a polynomial or a feature of an extension field, but if we're talking about the reed-solomon expansion, `rate` or `blow-up factor` is more appropriate. IIUC, what we're referencing here is called `invRate` in the codebase (for inverse of rate)

risc0/core/src/field/baby_bear.rs
```diff
@@ -362,7 +374,7 @@ const fn decode(a: u32) -> u32 {
 }
 
 /// The size of the extension field in elements, 4 in this case.
-const EXT_SIZE: usize = 4;
+const EXT_SIZE: usize = 5;
 
 /// Instances of `ExtElem` are elements of a finite field `F_p^4`. They are
```
pdg744, line 379: fp4 specific comments

risc0/zkp/src/verify/mod.rs
flaub: This was recommended by @jbruestle. Feel free to rewrite any of these comments.

BTW, this is a draft and it's not ready yet for review. I only pushed it so I could share what I have to get help getting inverse worked out.

risc0/core/src/inverse.rs
```diff
@@ -0,0 +1,143 @@
...
+            c[i] = a * b[i];
+        }
+        c
+    }
+
+    fn inv(&self, a: &[T]) -> Vec<T> {
+        let f = ExtensionField::new(self.degree + 1, self.irreducible.clone());
+        let mut t0 = f.zero();
+        let mut t1 = f.one();
+        let mut r0 = self.irreducible.clone();
```
jacobdweightman, line 80: There's a small discrepancy here: the C++ inv implementation in the compiler uses `r0 = [p - 11, 0, 0, 0, p - 1]`. I don't remember why that ended up that way, but I do recall testing intermediate states of that algorithm implementation against at least one non-trivial example I worked out by hand so I'm sure there was a good reason.

risc0/core/src/inverse.rs
```diff
@@ -0,0 +1,143 @@
...
+        let mut r1 = a.to_vec();
+
+        while Self::degree(&r1) > 0 {
+            let quot = self.div(&r0, &r1);
+            let r2 = f.sub(&r0, &f.mul(&quot, &r1));
+            r0 = r1;
+            r1 = r2;
+            let t2 = f.sub(&t0, &f.mul(&quot, &t1));
```
jacobdweightman, line 88: Oh, this is actually likely to be the issue: these should be naive multiplies

risc0/core/src/inverse.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn mul_elem(&self, a: T, b: &[T]) -> Vec<T> {
+        let mut c = self.zero();
+        for i in 0..self.degree {
+            c[i] = a * b[i];
+        }
+        c
+    }
+
+    fn inv(&self, a: &[T]) -> Vec<T> {
+        let f = ExtensionField::new(self.degree + 1, self.irreducible.clone());
```
jacobdweightman, line 77: This extension field is primarily used as a "trick" to get ordinary polynomial operations (`add`, `sub`, `naive_mul`) on higher degree polynomials than `self` works with. In particular, this irreducible polynomial isn't a valid choice for the larger field, and `mul` should be avoided.

risc0/core/src/inverse.rs
flaub: thanks, fixed this, still broken however

risc0/core/src/field/baby_bear.rs
```diff
@@ -681,14 +695,11 @@ impl ops::MulAssign for ExtElem {
...
-        let upper = 2 * EXT_SIZE - 2;
-        for i_rev in (0..EXT_SIZE - 1).rev() {
-            // We need to iterate from high degree to low, otherwise we might add to coefficients we've already zeroed out
-            let i = upper - i_rev;
-            for j in 0..(IRREDUCIBLE.len() - 1) {
-                // Perhaps there's a minor perf gain to be had by storing NEG_IRREDUCIBLE instead of IRREDUCIBLE itself
-                c[i - EXT_SIZE + j] += c[i] * -IRREDUCIBLE[j];
+        assert_eq!(IRREDUCIBLE[EXT_SIZE], Elem::ONE);
+        // We need to iterate from high degree to low, otherwise we might add to coefficients we've already zeroed out.
+        for i in EXT_SIZE..2 * EXT_SIZE - 1 {
```
tzerrell, line 700: This comment and this loop index contradict each other. For our choice of irreducible polynomial (and in general any irreducible polynomial of the form `x^EXT_SIZE + constant`) it's ok to iterate from low degree to high as coded here, but we should update the comment to match the code.
```suggestion
        // Since our irred. poly is `x^EXT_SIZE + c` it is ok to iterate from low degree (small `i`) to high
        for i in EXT_SIZE..2 * EXT_SIZE - 1 {
```

risc0/core/src/field/baby_bear.rs
```diff
@@ -681,14 +695,11 @@ impl ops::MulAssign for ExtElem {
...
+            for j in 0..EXT_SIZE {
+                c[i - EXT_SIZE + j] += c[i] * NEG_IRREDUCIBLE[j];
             }
```
tzerrell, line 703: Thinking about my previous comment, iterating from low degree to high is legal in our case because in the inner `j` loop, `NEG_IRREDUCIBLE[j] = 0` for `j != 0`; thus,  because of the size of the `i` loop, the `c[i] = Elem::ZERO` of the outer loop never gets reset by the inner loop.

This _also_ means that under the same assumptions we're already making, we don't need the inner loop at all; I propose we either remove it, or revert the order of the `i` loop.
```suggestion
            // No loop is needed over the irred. poly. since it only has leading and constant terms
            c[i - EXT_SIZE] += c[i] * NEG_IRREDUCIBLE[0];
```

flaub: But we are writing to the output from low to high, which is why I still consider it to be in reverse order. It's just that the index variable is going up.

flaub: What happens if we select a different irreducible polynomial in the future? I'd like to keep the code generic so we don't have to change two things.

tzerrell: > But we are writing to the output from high to low

One of us is missing something here, because I don't see this code writing the output from high to low: Noting that `j = 0` is the only iteration of the inner loop that writes anything (for our current irreducible polynomial), we write first to `c[0]`, then to `c[1]`, etc, and finally write to `c[EXT_SIZE - 1]` when `i == 2 * EXT_SIZE - 1`. Even if we were to use a more complicated polynomial, we would still be writing to increasing indices of `c` as `i` increased.

For keeping the code generic, I have two thoughts:
* It's a pretty simple change to keep it generic, especially with the inner loop already written.
* It's a kind of generic we're unlikely to ever use -- I'm pretty sure you can always find an irreducible polynomial of the form `x^EXT_SIZE + c`, so there's no need to use more complicated polynomials; and more complicated irreducible polynomials have minor performance disadvantages (right here, in fact: they require this inner `j` loop and `x^EXT_SIZE + c` polynomials don't).

Between those two considerations I'm ambivalent about whether to code for generic irreducible polynomials here. I do think we should design the full algorithm to be generic _or_ take advantage of every simplification available if we decide to not write code that's compatible with generic irreducible polynomials.

flaub: Is there a good resource you can recommend on where to learn about polynomial reductions?

flaub: Thanks, I'm going to copy paste this and remove from here because it doesn't seem like a good long term place for it and will hurt our ability to review the changes in the future. Generally we should treat PRs as a historical record for future developers.

tzerrell: Good point, it definitely makes more sense somewhere long-term. Can you link it here for reference? (Or if you haven't made the Notion page yet I can do the porting using Git's history feature.)

pdg744: Hmm... mb I can do a study club on this; I can't think of any references on this topic that would be particularly helpful. 

For developing intuition on how familiar arithmetic algorithms translate into polynomial algorithms, I think James Tanton's course on exploding dots is phenomenal, but it's very non-traditional: https://gdaymath.com/lessons/explodingdots/6-1-welcome/

pdg744: > What happens if we select a different irreducible polynomial in the future? I'd like to keep the code generic so we don't have to change two things.

There are lots of algorithmic tricks that are specific to different choices of extension size and minimal polynomial. For example, using a minimal polynomial such as `x^2+x+1` or `x^6+x^3+1` will yield field extensions that have a "golden-ratio structure" that may offer a performance advantage over something like `x^5+B`. That said, I think for the foreseeable future we'll probably only be using extensions of the form `x^n+B`.

pdg744: > Can you link it here for reference?

Ya I'd love to take a look at this!

flaub: https://www.notion.so/risczero/Extension-Fields-5e42931293894944ad164b3dc738e394?pvs=4

risc0/zkp/src/verify/mod.rs
jbruestle: Agree it's probably 'expansion factor', but also, this is a very 'draft' PR at this point, so I wouldn't sweat it.

pdg744, line 323:
```suggestion
        // TODO: This currently treats the expansion factor as hardcoded at 4, with
```

risc0/zkp/src/verify/mod.rs
```diff
@@ -320,10 +320,10 @@ where
...
         // the structure of the code and the value of `remap` (and how it is
-        // accessed) only working in the extension degree = 4 case.
-        // However, for generic fields the extension degree may be different
+        // accessed) only working in the expansion degree = 4 case.
```
pdg744, line 325:
```suggestion
        // accessed) only working in the expansion factor = 4 case.
```

risc0/zkp/src/verify/mod.rs
```diff
@@ -320,10 +320,10 @@ where
...
+        // However, for generic fields the expansion degree may be different
```
pdg744, line 326:
```suggestion
        // However, for generic fields the expansion factor may be different
```

PR #1324, opened 2024-01-12
Pin cargo-platform to 0.1.5 in zkvm-methods-std

nategraf:
An increased minimum required Rust version in the `cargo-platform` dependency has broken builds of the `release-0.18` branch. As a result, PRs changing the website (e.g. https://github.com/risc0/risc0/pull/1314) as stuck since CI requires doc tests to be run against the `release-0.18` branch.

This PR addresses this issue by updating the `Cargo.lock` file for `risc0-zkvm-methods-std`, and checking in that file. This matches how `Cargo.lock` is handled for `risc0-zkvm-methods-guest`, which is the `no-std` sibling of this package.


nategraf:
It looks like a number of unrelated issues exist in trying to run CI against `release-0.18` at this point. Do we want to fix these issues, or do we want to merge this without CI checks?

PR #1325, opened 2024-01-13
Add ProverServer::compress and use it to make receipts succinct by default

nategraf:
This PR adds a `ProverServer::compress` function that uses the `lift`, `join`, and `resolve` recursion programs to compress a `CompositeReceipt` into a `SuccinctReceipt`. This PR then uses this function to compress receipts when `prove_session` is called, and therefore compress receipts by default.

This PR also:
* Adds an `ExitCode::is_ok` function to clean up some code I saw along the way to implementing this PR


risc0/zkvm/examples/loop.rs
nategraf, line 1: Note that this _does_ change what is benchmarked here and in the `fib` example. In particular, it now includes any `lift` and `join` operations required to get a single `SuccinctReceipt`. The upside of this is constant size proofs. The downsize of this is extra compute spent compressing said proofs.

If folks would like, I can change this to avoid including compression.

flaub: I have used these examples as a way to do profiling. But perhaps we should make an example/tool specifically for that purpose.

risc0/zkvm/examples/loop.rs
```diff
@@ -77,13 +77,7 @@ fn main() {
...
         let (cycles, _) = session.get_cycles().unwrap();
 
-        let seal = receipt
-            .inner
-            .composite()
-            .unwrap()
-            .segments
-            .iter()
-            .fold(0, |acc, segment| acc + segment.get_seal_bytes().len());
+        let seal = receipt.inner.succinct().unwrap().seal.len() * 4;
```
flaub, line 80: What is `4` in this context? I'm guessing `sizeof(uint32_t)`.

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -87,8 +87,52 @@ pub trait ProverServer {
...
-    /// Convert a [SuccinctReceipt] with a poseidon hash function that uses a 254-bit field
+    /// Convert a [SuccinctReceipt] with a Poseidon hash function that uses a 254-bit field
     fn identity_p254(&self, a: &SuccinctReceipt) -> Result<SuccinctReceipt>;
+
+    /// Compress a [CompositeReceipt] into a single [SuccinctReceipt].
+    ///
+    /// A [CompositeReceipt] may contain an arbitrary number of receipts assembled into
+    /// continuations and compositions. Together, these receipts collectively prove a top-level
+    /// [ReceiptClaim](crate::ReceiptClaim). This function compresses all of the constituent receipts of a
+    /// [CompositeReceipt]crate::ReceiptClaim into a single [SuccinctReceipt] that proves the same top-level claim. It
```
flaub, line 98:
```suggestion
    /// [CompositeReceipt] into a single [SuccinctReceipt] that proves the same top-level claim. It
```

flaub (approved):

PR #1327, opened 2024-01-13
Fix FpExt Multiply and Inversion

tzerrell:
Fix bugs in multiply, degree, and inversion functions in the FpExt refactor. Tests now pass.

Note that the test from `galois` needed to be changed, as we use the opposite order for our polynomials (we use increasing degree, they use decreasing degree) and there was a typo in one of the outputs.


flaub (approved):
Thanks!!!

PR #1332, opened 2024-01-16
Duplicate border Property in CSS

avocadodefi:
Duplicate border Property in CSS: In your CSS, you have defined the border property twice for both .box and .qrcode-box classes. The second definition (i.e., border: none;) will override the first one (i.e., border: 1px solid black;). I removed the redundant definition. Updated the CSS for .box and .qrcode-box classes to have border: none;. Adjusted the figcaption elements to use the style attribute for inline styling. Updated the alt text for QR code images for better accessibility.


capossele (approved):

flaub (changes requested):
Holding off on this for now

flaub:
@nahoc Would you mind taking a look at this?

nahoc (commented):
@avocadodefi Changes look good, but why get rid of all the CSS comments? Is this just whatever tool/IDE you use that wipes them out automatically?

flaub (approved):

PR #1335, opened 2024-01-17
Update docs to include resolve step

rlukata:


mothran (approved):
Thanks for catching that!

PR #1338, opened 2024-01-17
Fix missing foundry toolchain

flaub:


austinabell (approved):
Yeah, it seems like those monthly releases are kept around https://github.com/foundry-rs/foundry/releases and should be more than fine until ethers-rs fix comes in

PR #1340, opened 2024-01-17
Suppress the "skipping duplicate package `xgboost`" warning

weikengchen:
If one uses a git path for dependencies of risc0 in Cargo.toml, chances are that the compiler would complain:

```
warning: skipping duplicate package `xgboost` found at `/Users/cusgadmin/.cargo/git/checkouts/risc0-b4649977e2e81438/341a014/examples/xgboost/methods/guest`
```

This is due to xgboost's example repo has the name `xgboost` and its methods/guest also has the name `xgboost`. Since this is a very boring warning but appears to show up frequently, it makes sense to suppress it, by just naming the xgboost example repo differently.


weikengchen:
could not reproduce the following error locally.
"Run python license-check.py
examples/groth[16](https://github.com/risc0/risc0/actions/runs/7560161227/job/20597320952#step:19:17)-verifier/groth16/tests/groth16.rs: invalid header!"

flaub:
Could you update the copyright header's year on each of the effected files?

weikengchen:
Let me dig out what happens... It is weird. The problem is that the header is indeed correct on the source repo:
https://github.com/l2iterative/risc0/blob/remove-xgboost-warning/examples/groth16-verifier/groth16/tests/groth16.rs#L1

```rust 
// Copyright 2024 RISC Zero, Inc.
//
// Licensed under the Apache License, Version 2.0 (the "License");
// you may not use this file except in compliance with the License.
// You may obtain a copy of the License at
//
//     http://www.apache.org/licenses/LICENSE-2.0
//
```

weikengchen:
Ok. @flaub I think it should work now. I realized that my previous commit here that merges the upstream should have fixed the issue.
https://github.com/l2iterative/risc0/commit/44709c0afdc336ecab7612267fa449e4cf1719ef

It is related to https://github.com/risc0/risc0/pull/1317#issuecomment-1896301867.

flaub (approved):

weikengchen:
It appears that Cargo.lock of examples needs to be updated because a new entry would be added:
```
[[package]]
name = "xgboost-example"
version = "0.1.0"
dependencies = [
 "forust-ml",
 "risc0-zkvm",
 "rmp-serde",
 "serde",
 "serde_json",
 "xgboost-methods",
]
```

A commit is pushed trying to fix the remaining CI errors.

weikengchen:
An automatic run of `cargo update` will create a lot of changes... if it is not desirable, please let me know so I can try to minimize the change to Cargo.lock to be name-only.

flaub:
The only thing that's needed is to run `cargo check` from the `examples` directory. This will update the `Cargo.lock` file to match what's in the `Cargo.toml`.

weikengchen:
The requested change has been made. `cargo check` instead of `cargo update` fixes the issue. Now there is only the following change in this PR.

```diff
[[package]]
- name = "xgboost"
+ name = "xgboost-example"
```

weikengchen:
Just an update that a new error has appeared.
warning: skipping duplicate package `methods` found at `/Users/cusgadmin/.cargo/git/checkouts/risc0-b4649977e2e81438/39a6fd8/examples/groth16-verifier/methods`

let me take a look as well since we clearly do not want to spend one more PR for this...

weikengchen:
A new commit has fixed the same issue about `groth16-verify` as well. Ready for review.

flaub (approved):

flaub (approved):

SchmErik:
@weikengchen we can't merge because this is outdated. Could you update your branch so we can merge?

weikengchen:
Updated. Realized that one problem with making a PR from organization-owned repo is that I couldn't auto-grant permission to modify to the RISC Zero team...

Issue #1323, opened 2024-01-12
Drop `slice_io`


nategraf:
We currently have two I/O mechanisms between the host and guest. One is `slice_io` and the other is POSIX I/O. POSIX I/O was created after `slice_io` and should be able to replace all uses of `slice_io`. We would like to drop `slice_io`, and ensure that we do so before zkVM 1.0.

flaub:
This didn't make it for 1.0

PR #1342, opened 2024-01-18
zkVM: remove Slice I/O

SchmErik:
We currently have slice_io and POSIX I/O to do I/O between guest and host. POSIX I/O was created after slice_io can replace all uses of slice_io.

This change removes slice io from the guest and host and makes necessary adjustments to protobuf definitions by reserving fields that are no longer in use.

Fixes: #1323


risc0/zkvm/src/host/client/env.rs
```diff
@@ -302,28 +296,6 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
-    pub fn slice_io(&mut self, channel: &str, handler: impl SliceIo + 'a) -> &mut Self {
-        self.inner
-            .slice_io
-            .borrow_mut()
-            .with_handler(channel, handler);
-        self
-    }
-
-    /// Add a handler for simple I/O handling.
-    pub fn io_callback<C: AsRef<str>>(
```
flaub, line 315: It might be nice to still provide this interface on-top of the PosixIo (if that even still makes sense)

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -127,9 +127,9 @@ message IdentityP254Result {
 }
 
 message ExecutorEnv {
+  reserved 3;
```
flaub, line 130:
```suggestion
  reserved 3; // was: repeated string slice_ios
```

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -127,9 +127,9 @@ message IdentityP254Result {
...
   Asset binary = 1;
   map<string, string> env_vars = 2;
-  repeated string slice_ios = 3;
```
flaub, line 132: I don't think we can actually do this since the idea is to have newer clients interact with older servers. I think we need to leave the protobuf alone and adjust the code that uses it.

SchmErik:
I'm going to close this and re-open after I'm done porting the waldo example to reduce the amount of open PR's.

Issue #1344, opened 2024-01-18
[BUG] Do we need LFS for recursion_zkr.zip?


weikengchen:
## Bug Report

There have been a few issues in this repository about LFS about recursion_zkr.zip. In December, the situation is as follows.

> This file is over 50MB currently and contains firmware code that is executed by the recursion circuit. I haven't yet found a way to make this code available as normal source since GitHub has a limit on the size that files can be that's checked in. So right now it's using LFS and then the build script downloads it at build time. I realize this is less than ideal and I'm looking for alternatives.

https://github.com/risc0/risc0/issues/1142
https://github.com/risc0/risc0/issues/1180

However, the file nowadays is no longer 50MB. It is 24.3MB.
https://github.com/risc0/risc0/blob/main/risc0/circuit/recursion/src/recursion_zkr.zip

Should we revisit the decision of incorporating the LFS?

--- btw, if we encounter the upload limit about crates.io, about 8 years ago crates.rs has decided to handle large-file requests in a case-by-case manner (https://github.com/rust-lang/crates.io/issues/40#issuecomment-157919165). It remains, however, a problem whether crates.io could handle such traffic and whether it is definitely a bad idea to include it in the uploaded crate.

flaub:
This issue is only relevant for users that want to build from source. If users use the binary release, then this can be avoided. I don't know of any good alternatives at this time.

flaub:
For using crates, the build script should download the ZKR from an S3 bucket, so crates from crates.io should work. If this isn't working as intended we can open a ticket for that issue.

weikengchen:
Actually that is what I want to follow up next. So, **for people who build from sources.**

Would it help to let the build script ignore a dummy LFS?
https://github.com/risc0/risc0/blob/main/risc0/circuit/recursion/build.rs#L54C12-L54C12

Currently, the build script would copy-paste the zip file if it **_exists_**.

```rust
if std::fs::metadata(&src_path).is_ok() {
    let tgt_path = out_dir.join(FILENAME);
    std::fs::copy(&src_path, &tgt_path).unwrap();
}
```

But the one that exists could be the 133-byte LFS metadata file. 
```
version https://git-lfs.github.com/spec/v1
oid sha256:ae5736a42189aec2f04936c3aee4b5441e48b26b4fa1fae28657cf50cdf3cae4
size 25527822
```

This would lead to an error:
```
[...]

Issue #1142, opened 2023-11-13
`cargo install cargo-risczero --git ...` fails


nategraf:
It should be possible to install `cargo-risczero` from a `git` ref, but it currently gives an error. Here is an example:

```
$ cargo install cargo-risczero --git https://github.com/risc0/risc0 --rev b9a9302f1ca5454b62d214a14cb9503bfbfdbc83
    Updating git repository `https://github.com/risc0/risc0`
warning: skipping duplicate package `xgboost` found at `/Users/victorgraf/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/b9a9302/examples/xgboost/methods/guest`
error: package `/Users/victorgraf/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/b9a9302/bonsai/sdk/Cargo.toml` is a member of the wrong workspace
expected: /Users/victorgraf/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/b9a9302/Cargo.toml
actual:   /Users/victorgraf/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/b9a9302/bonsai/Cargo.toml
```

Note: The warning about `xgboost` is unrelated.

flaub:
The bigger issue with this is that LFS assets are not properly handled by cargo when using git dependency references.

flaub:
Related: https://github.com/rust-lang/cargo/issues/9692

PR #1347, opened 2024-01-18
Only use local recursion_zkr.zip if it matches the hash

weikengchen:
This resolves the remaining problem in #1344 

The problem is that, if one uses git path as dependency (see https://github.com/risc0/risc0/issues/1142), cargo does not do git-lfs, and therefore the `recursion_zkr.zip` would remain as a lfs metadata file instead of the actual file. This causes some errors in the upstream, where the recursion prover would complain that the zip archive for the recursion circuits could not be decompressed.

```
thread '<unnamed>' panicked at /Users/cusgadmin/.cargo/git/checkouts/risc0-b4649977e2e81438/341a014/risc0/circuit/recursion/src/zkr.rs:29:71:
called `Result::unwrap()` on an `Err` value: InvalidArchive("Could not find central directory end")
Traceback (most recent call last):
  File "/Users/cusgadmin/RustroverProjects/r0prover-python/test/test.py", line 33, in <module>
    receipt_1_lifted = l2_r0prover.lift_segment_receipt(receipt_1)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: InvalidArchive("Could not find central directory end")
```

This PR aims to solve this remaining issue systematically. There are a few steps.

**First.** Now, if the build script finds that the local recursion_zkr.zip does not match the expected hash value (which can be due to many reasons, including the lfs metadata issue mentioned above, or the zip file is obsolete), then the build script would choose to download the new zip from S3 bucket.

```diff
+       if sha2::Sha256::digest(data).to_vec() == decode_hex(SHA256_HASH).unwrap() {
            let tgt_path = out_dir.join(FILENAME);
            std::fs::copy(&src_path, &tgt_path).unwrap();
+           return;
+       }
```

**Second.** One thing about the `downloader` crate is that it would refuse to download a file that overwrites the existing one, as it always uses "create_new". This, however, may happen if the user runs the build script again.
https://github.com/hunger/downloader/blob/main/src/backend.rs#L91

To handle this situation, we also check the output file. If a valid downloaded version of `recursion_zkr.zip` exists, then we do not need to download again. If it exists but it does not match, the file must be deleted, in order to allow `downloader` to download the new one at the same path.

Local test shows that this can fix the issue, enabling the upstream to correctly run the recursion prover.

My own **_negative_** test (through println and panic) shows that this would not unnecessarily download the zip from S3 bucket if the local file actually matches the hash---this may need someone from the RISC Zero team to have an independent check.

### DIFF
--- a/risc0/circuit/recursion/build.rs
+++ b/risc0/circuit/recursion/build.rs
@@ -52,11 +52,21 @@ fn download_zkr() {
     let src_path = PathBuf::from_str(src_path.as_str()).unwrap();
     let out_dir = env::var("OUT_DIR").unwrap();
     let out_dir = Path::new(&out_dir);
+    let out_path = out_dir.join(FILENAME);
+
+    if std::fs::metadata(&out_path).is_ok() {
+        let data = std::fs::read(&out_path).unwrap();
+        if sha2::Sha256::digest(data).to_vec() == decode_hex(SHA256_HASH).unwrap() {
+            return;
+        }
+    } else {
+        std::fs::remove_file(out_path).unwrap();
+    }
+
     if std::fs::metadata(&src_path).is_ok() {
         let data = std::fs::read(&src_path).unwrap();
         if sha2::Sha256::digest(data).to_vec() == decode_hex(SHA256_HASH).unwrap() {
-            let tgt_path = out_dir.join(FILENAME);
-            std::fs::copy(&src_path, &tgt_path).unwrap();
+            std::fs::copy(&src_path, &out_path).unwrap();
             return;
         }
     }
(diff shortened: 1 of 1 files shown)


weikengchen:
@flaub it seems that although the `release-0.20` has the correct ZKR hash, it is not the case for the main?
https://github.com/risc0/risc0/blob/main/risc0/circuit/recursion/build.rs

In other words, #1348 is still there.

flaub:
Prebuilt binaries are only offered for releases right now, so my process currently is to update that hash once we upload the new ZKR just ahead of a release.

PR #1372, opened 2024-01-23
Validate Poseidon `hash_pair` input

tzerrell:
Our Poseidon/Poseidon2 `hash_pair` functionality is intended to work with Digests generated by Poseidon hashing. However, prior to this PR, it does not perform checking to ensure that its input Digests are in this format. This means a developer unaware of this nuance might accidentally write code that expects these `hash_pair` functions to work as hash functions on general Digests with values outside those possible from Poseidon hashing, which could create a vulnerability depending on the use case.

This PR adds an `assert` that validates all inputs are in the required format by adding the ability to test that words in a Digest are elements of the underlying field in reduced form. Thus, any developer misusing `hash_pair` in the above way would hit this `assert` instead of running to completion and producing a "hash" that does not have all the security properties such a developer might expect.


nategraf (approved):
This is very helpful, and looks good to me 👍

nategraf:
> This is very helpful, and looks good to me 👍

Removed my approval just pending any bench-marking we can do. I know you ran it locally, but we should think a bit on what more we might want to do. For example, we should confirm the same lack of performance diff occurs on a CUDA machine.

risc0/core/src/field/mod.rs
```diff
@@ -122,6 +128,12 @@ pub trait Elem:
         self
     }
 
+    /// Returns this element, but checks to make sure it's in reduced form.
+    fn ensure_reduced(&self) -> &Self {
+        assert!(self.is_reduced());
```
flaub, line 133: I wonder if we should use `debug_assert!`?

flaub:
I've been thinking about this and wondering if we could change the structure of a Digest so that it prevents misuse by leveraging the type system more.

tzerrell:
Summarizing what I understood from our in-person discussion:
* We have not seen perf impacts in my CPU testing, and we expect to not see perf impacts (because running 16 `u32` `<` ops is small compared to the whole of Poseidon)
* This doesn't change CUDA/Metal code because this is on the verification side, which isn't GPU accelerated
* The proof system should not be using this `hash_pair` on values `>= P`. Also, the `hash_pair` code is not a very sensible thing to use outside the proof system, although at the moment it is public in a way that outside developers in principle could use & misuse it without warnings. Adding this assert has negligible performance impact and prevents misuse by us or others from causing security defects.
* There may be deeper changes that are unnecessary for this specific problem but make it harder to create similar sharp edges in the future. E.g. @flaub's comment "change the structure of a Digest so that it prevents misuse by leveraging the type system more," and I know @nategraf has been thinking about this as well. These can be handled in other PRs.

risc0/core/src/field/mod.rs
nategraf: It would depend on how how this is used, but I would air on the side of having it be a full assert. However, now that I look, I don't see this function used anywhere, so I would actually suggest we simply remove it.

tzerrell: Ah, good call, this is remnants of an earlier approach. Removing this also reduces potential confusion between `ensure_reduced` (`assert` if >= prime) and `ensure_valid` (`debug_assert` if `0xffffffff`).

tzerrell:
Checking in on this PR, because GitHub is flagging this as "approved", but the only approval (from @nategraf) has been removed (https://github.com/risc0/risc0/pull/1372#issuecomment-1906998538) -- are we ready to land this?

nategraf:
In [aba887d](https://github.com/risc0/risc0/pull/1372/commits/aba887df11989b8755101cbd3a0a7de88d4f6437) I pushed a change to fix the benchmarks in the `benchmarks` directory. These were broken because they expected to a `CompositeReceipt` back from the `prove` method, but this was changed in https://github.com/risc0/risc0/pull/1325 and updating this location was missed.

tzerrell:
> In [aba887d](https://github.com/risc0/risc0/pull/1372/commits/aba887df11989b8755101cbd3a0a7de88d4f6437) I pushed a change to fix the benchmarks in the `benchmarks` directory. These were broken because they expected to a `CompositeReceipt` back from the `prove` method, but this was changed in #1325 and updating this location was missed.

These changes look fine to me; were you then able to run benchmarks?

nategraf (commented):
So I now have a run of the benchmarks against this branch, and against it's base commit in `main`. Results are available in the `ghpages` repo as commits:
* e7d3495ef6ad0c20c0454826f7fd23ca547426ea for `main` (with fixes for the benchmarks themselves)
* e2a5f3a20192eafc8e250d7634f4a3b6fd5407fe for this branch

On GPU only, it appears there _is_ a persistent and significant performance difference, at least from this batch of benchmarks. I'm not entirely sure of the explanation for this, or if these benchmarks are stable, but this at least warrants looking into it some more. I decided to kick off the workflow again `main` to check self-consitency. https://github.com/risc0/risc0/actions/runs/7704641894

If you want to see the reports, you can check out the `ghpages` repo, and one of those two commits listed above. You can locally serve the static HTML with `python -m http.server`. Comparing `http://localhost:8000/dev/benchmarks/` at each commit will give you the comparison. Do note that the top level table on that page with the averages across different jobs is broken because that the `average` bin was removed at some point.

tzerrell:
Fascinating! Yay for empiricism, because I didn't think these `assert`s were in code used by the GPUs -- but with this data I double-checked, and they are indeed run even on GPU.

Given that and the noticeable performance slowdown, I think we should pause this until we see what comes of your type cleanup work.

flaub:
We now have better benchmark comparisons and I'm not seeing any perf regression here: https://github.com/risc0/risc0/actions/runs/8811852210?pr=1372

PR #1361, opened 2024-01-22
Updates to Proof System Docs

pdg744:
This PR started in response to some broken links around receipt types on the proof system overview page. 

While fixing those links, I went ahead and made some revisions to the docs to reflect the three layer proving system. This includes adding some information about our Groth16Receipts, and revising verbiage for our STARK protocol to reflect the fact that it's one piece in the picture. 

This PR also adds the following terms to the terminology page:
  - deterministic builds
  - Groth16
  - Groth16 Receipt
  - Groth16 Circuit
  - Recursion Circuit
  - RISC-V Circuit

Note: There's currently a [ReceiptClaim] link on the proof system page which is broken and which this PR does not address. I think this will work as soon as 0.20 lands; folks should be aware when editing docs that our proof system docs aren't currently part of the versioned docs system 

TODO - images aren't displaying in vercel preview.

PR #1377, opened 2024-01-25
Clean-up on terminology.md & proof-system.md

pdg744:
I was a bit hasty on #1361; this PR cleans up a bit.

### DIFF
--- a/website/docs/proof-system/proof-system.md
+++ b/website/docs/proof-system/proof-system.md
@@ -55,7 +55,7 @@ In addition to the links in the sidebar, we recommend the following resources:
 [ReceiptClaim]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.ReceiptClaim.html
 [SegmentReceipts]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.SegmentReceipts.html
 [SegmentReceipt]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.SegmentReceipt.html
-[flat receipt]: https://docs.rs/risc0-zkvm/0.19.1/risc0_zkvm/struct.FlatReceipt.html
+[flat receipt]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/enum.InnerReceipt.html#variant.Flat
 [succinct receipt]: https://docs.rs/risc0-zkvm/0.19.1/risc0_zkvm/struct.SuccinctReceipt.html
 [Groth16]: https://dev.risczero.com/terminology#groth16
 [Groth16 receipt]: https://dev.risczero.com/terminology#groth16-receipt
--- a/website/docs/proof-system/proof-system.md
+++ b/website/docs/proof-system/proof-system.md
@@ -20,7 +20,7 @@ These claims are summarized in the [ReceiptClaim].
 
 A [Receipt] can take four main forms.
 
-1. A **[flat receipt]** is a vector of [ZK-STARK]s, one for each [segment]. Segments & segment proofs are constructed by the [RISC-V Circuit].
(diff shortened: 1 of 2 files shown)


website/docs/proof-system/proof-system.md
```diff
@@ -55,7 +55,7 @@ In addition to the links in the sidebar, we recommend the following resources:
 [ReceiptClaim]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.ReceiptClaim.html
 [SegmentReceipts]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.SegmentReceipts.html
 [SegmentReceipt]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.SegmentReceipt.html
-[flat receipt]: https://docs.rs/risc0-zkvm/0.19.1/risc0_zkvm/struct.FlatReceipt.html
+[flat receipt]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/enum.InnerReceipt.html#variant.Flat
```
flaub, line 58:
```suggestion
[flat receipt]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/enum.InnerReceipt.html#variant.Flat
```

BTW this is going to break very soon, we just renamed this in `0.20`. I think we should consider trying to avoid pointing to code from these sorts of docs.

website/docs/terminology.md
```diff
@@ -230,6 +235,8 @@ By repeating this compression, an arbitrary number of related [receipts] (e.g. a
 
 ### Recursion Circuit
 
+[Recursion Circuit]: #recursion-circuit
```
flaub, line 238: Would you mind putting all references at the bottom of the file? This makes it a lot easier to maintain over time since we can find all the links in a common place in all files.

website/docs/proof-system/proof-system.md
pdg744: Hmm...  I know the naming is unstable but not sure how this is solved by avoiding the pointer.

Can you clarify what you mean by "these sorts of docs" and what you have in mind as an alternative?  We point to the crate docs on a large number of pages, and providing a route for users to dig deeper into the details seems like a primary purpose of these docs.

pdg744: Like, I guess on most pages we could point to `terminology.md` rather than the crate docs, but it strikes me as a big mistake to avoid navigation paths from dev docs website to crate docs.

pdg744: will revise from `latest` to `0.19.1` for now, as I believe our convention is to avoid `latest` in order to make sure that we have an effective link audit as we push new releases

flaub: Consistency, it's shorter.

But I'm coming around to the idea that for `/docs` we should always point to the latest stable. But now we need to have a process where when we cut a new release we have to comb thru the `/docs` to update all the versions and do a pass on naming changes.

website/docs/terminology.md
pdg744: If it's ok with you, let's land this as is and do that on a follow-up PR. There are a ton of other links on this page and I think implementing that change on this PR will distract from the more substantive changes being implemented here.

website/docs/proof-system/proof-system.md
pdg744: Hmm...I'm happy to defer to your judgment on this. 

The site right now includes use of `/0.18/`, `/0.19/` more than `/latest/`. I'm having trouble figuring out if/where we use `/*/` because ctrl-f isn't place nice with `*`.

flaub (approved):

tzerrell (approved):

PR #1383, opened 2024-01-26
Website Previews for Contributors (risc0 side)

RDaneelOlivaw0:
Re-creating to test Amplify preview from fork.

Modifying website-preview.yml to use risc0-website-previews bucket to host website previews for contributors.


RDaneelOlivaw0:
Reopening to test Amplify preview with app created with Terraform.

PR #1391, opened 2024-01-30
examples/digital-signature: verify identity

matthiasgeihs:
Proposes a change to the `digital-signature` example:
Previously, the signer identity was not verified. However, this is an essential part of signature verification, so this PR proposes to verify the identity of the signer (in addition to verifying the message and the receipt).


flaub:
It seems like the extra check is unnecessary, because the passphrase is hashed inside the zkvm, which means we can be sure that it was computed properly.

matthiasgeihs:
> It seems like the extra check is unnecessary, because the passphrase is hashed inside the zkvm, which means we can be sure that it was computed properly.

I understand that. Nevertheless, from the standpoint of the verifier, it is essential to check that the identity output by the proof matches the expected identity, right?

Typically, signature verification has the following function signature: `(pk, m, sig) -> b`
In the example, we implicitly have `pk = Hash(passphrase)` and `sig = transcript`. The verification test checks that `transcript.verify()` and `H(m) == transcript.message`, but it doesn't check that `transcript.identity == pk`.

It's not really something about the ZK part, just about how this example is presented. It was confusing me while I was going through the examples and thought I put up this little PR.

PR #1396, opened 2024-01-30
Refactor private outputs

### LINKED ISSUE #1038: Improve devex for private outputs
There are two issues with private outputs:
* Private outputs written via `env::write` in the guest end up in the stdout stream. This will collide with any `println!` calls within the guest. 
* The host-side API is cumbersome. To retrieve private outputs one must hook up a stdout handler and then manually deserialize the results.

Proposal:
* Create a new dedicated file descriptor for private outputs.
* Add a 'private journal' to the `Session` which will allow users to retrieve private outputs similar to the new `Journal` type.

Cardosaum:
- Write outputs to `private_journal`, using a separate file descriptor (instead of using `stdout`).
- Refactor `Prover::prove` to return a `ProveResult`

Closes #1038


Cardosaum:
In [api.proto], should we deprecate `journal` and create a new `public_journal`? 

```proto
message SessionInfo {
  uint32 segments = 1;
  // Should we deprecated this?
  bytes journal = 2 [deprecated = true];
  protos.base.ExitCode exit_code = 3;
  bytes public_journal = 4;
  bytes private_journal = 5;
}
```

Alternatively, we could keep using `journal` as public journal and simply add `private_journal`, which would be a backwards compatible change from what I understand.

```proto
message SessionInfo {
  uint32 segments = 1;
  bytes journal = 2;
  protos.base.ExitCode exit_code = 3;
  bytes public_journal = 4;
}
```

[api.proto]: https://github.com/risc0/risc0/blob/daf10e8b44114a76e9336b04f572cff5a3fb804a/risc0/zkvm/src/host/protos/api.proto#L158

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -199,3 +199,29 @@ pub(crate) fn get_r0vm_path() -> PathBuf {
         .unwrap_or("r0vm".to_string())
         .into()
 }
+
+/// TODO: Write documentation.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[cfg_attr(test, derive(PartialEq))]
```
flaub, line 205: Why not make this always have a `PartialEq`?

Cardosaum: Oh, I was following the pattern on what was defined on other structs, but can totally make it always have it!

SchmErik:
> I wonder if we can treat the private journal as just another file descriptor so that if the user wants to have multiple private journals they can do that.

I think that might be a good idea. I was thinking about how this would work.. It means that the guest would need to designate which private journal to write to. So `env::write(...)` would need to take a parameter that represents some sort of file descriptor `fd`, like so: `env::write(fd, ...)`. We can ensure that the fd is not overlapping by checking that the fd is not already defined (i.e. stderr, stdout, journal, stdin). after execution we create a vector of pairs `(fd:u32, private_journal:Vec<u8>)`.

I also thought of a scenario: let's say that a user decides to take fd number 5 as their private journal and in a future release, we end up using that fd for something else. This situation will result in a collision that will break the user's code... I don't think we need to consider this situation because if we add new fd's to a list of known fd's, it means that we should declare a new version and users should expect some incompatibility.

SchmErik:
We're going to take a different approach and document what we already have implemented. Closing

PR #1401, opened 2024-02-01
reproducible builds: update dockerfile to use new toolchain

SchmErik:
This change downloads a newer version of cargo-risczero and to build using a newer toolchain version. This solution expects us to publish new docker images on every toolchain update. While this may create more burden on us, I think this is much more performant for users because they would not need to install the toolchain when running docker builds.

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -128,7 +128,7 @@ fn create_dockerfile(
     .join(" ");
 
     let build = DockerFile::new()
-        .from_alias("build", "risczero/risc0-guest-builder:v0.17")
+        .from_alias("build", "risczero/risc0-guest-builder:v2024-01-31.1")
         .workdir("/src")
         .copy(".", ".")
         .env(manifest_env)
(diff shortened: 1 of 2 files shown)


risc0/build/src/docker.rs
```diff
@@ -128,7 +128,7 @@ fn create_dockerfile(
     .join(" ");
 
     let build = DockerFile::new()
-        .from_alias("build", "risczero/risc0-guest-builder:v0.17")
+        .from_alias("build", "risczero/risc0-guest-builder:v2024-01-31.1")
```
capossele, line 131: This would still require to push to docker-hub any image with a different toolchain version.
We could add the `latest` tag to make this line constant, e.g.,  `.from_alias("build", "risczero/risc0-guest-builder:latest")` however I personally like being explicit about what version of the toolchain we are running.
A different approach would instead be to remove the installation of the toolchain from this image and install it in the next step, the one we use to build the guest. The advantage would be that we don't need to push a new image at every update, the disadvantage that the guest build would take more time.

flaub (approved):

PR #1407, opened 2024-02-01
Add a changelog

flaub:


SchmErik (approved):
Looks good to me, so is the idea that we would be contributing to this document in the "upcoming release" section as a part of future PR's? Nevermind, please disregard this question

flaub:
I think so, I think there's 2 ways to use this:
1. As fixes come in, we update the Fixes section
2. During release planning meetings, we can update this file with our ideas of features we want to introduce.

It's a living document, but can be frozen once we release branch is cut. When a release happens, we promote "Next" to the released version and push all the features that didn't make it to the new "Next".

PR #1414, opened 2024-02-05
Use Poseidon2 for recursion hashing

tzerrell:
This switches our recursion system to Poseidon2 hashing.

[Marking as a draft while I resolve the stark2snark issues]

Note that there is not forwards or backwards compatibility between this Poseidon2-based system and the existing Poseidon system. I.e. this verifier won't verify old proofs, and new proofs won't be verified by the old verifier. So once this lands on `main` proofs & verifications will be out of sync w/ v0.20 and will not be in sync with a released version again until v0.21. ~For this reason, I have marked this PR as a draft until we have a plan we're happy with for rolling out this change; the PR is otherwise ready for review, there is no outstanding work I know of.~ [We have decided the right rollout process is to land this on `main`, and then do any following rollout (e.g. to Bonsai) as part of the standard release process with v0.21.]


risc0/zkvm/src/host/recursion/prove/exec.rs
```diff
@@ -394,6 +394,61 @@ impl<'a> ParallelHandler<'a> {
         }
         self
     }
+
+    fn log(&mut self, msg: &str, args: &[BabyBearElem]) {
```
tzerrell, line 398: This code is copied from the `MachineContext` version of the logging code -- I needed logging for `ParallelHandler` and this is working for me locally. That said, I haven't convinced myself that this is a good solution globally, so I'm interested in review feedback here.

jbruestle (approved):

PR #1417, opened 2024-02-07
fix: delete `TempDir` when Session goes out of scope

SchmErik:
This solves the problem where the temporary directory used for storing segments are not deleted after proving. One solution is to represent the directory as `Rc<TempDir>` instead of a `PathBuf` and have `Session` or `SegmentRef` hold a reference to the `TempDir`. Once the session goes out of scope, the `TempDir` gets deleted. I've discovered that adding `TempDir` as a field of `Session` doesn't work because `TempDir` does not implement Serialize or Deserialize while Session and `SegmentRef` implement these traits.

This solution adds the `PathBuf` of the `TempDir` to `Session` and implements a `Drop` trait that removes this directory when Session goes out of scope. I've also noticed that in the case where the user defines the segment path, we're using that path to store the segments. This changes this execution path by creating a `TempDir` inside the user- defined directory so that it does not accidentally delete the user's directory upon termination.

fixes: #1223

### DIFF
--- a/risc0/zkvm/Cargo.toml
+++ b/risc0/zkvm/Cargo.toml
@@ -105,6 +105,7 @@ client = [
   "dep:prost-build",
   "dep:protobuf-src",
   "std",
+  "dep:tempfile",
 ]
 cuda = [
   "prove",
--- a/risc0/zkvm/Cargo.toml
+++ b/risc0/zkvm/Cargo.toml
@@ -104,8 +104,8 @@ client = [
   "dep:prost",
   "dep:prost-build",
   "dep:protobuf-src",
-  "std",
   "dep:tempfile",
+  "std",
 ]
 cuda = [
   "prove",
--- a/risc0/zkvm/src/host/client/env.rs
+++ b/risc0/zkvm/src/host/client/env.rs
@@ -28,6 +28,7 @@ use bytemuck::Pod;
 use bytes::Bytes;
 use risc0_zkvm_platform::{self, fileno};
 use serde::Serialize;
+use tempfile::TempDir;
 
 use crate::serde::to_vec;
 use crate::{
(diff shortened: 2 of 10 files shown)


risc0/zkvm/Cargo.toml
```diff
@@ -105,6 +105,7 @@ client = [
   "dep:prost-build",
   "dep:protobuf-src",
   "std",
+  "dep:tempfile",
```
flaub, line 108: Can you sort this?

flaub:
I think what we wanted to do is use an `Rc<TempDir>` that would be on each `FileSegmentRef`. That way when all the refs get dropped the temp dir automatically gets removed.

risc0/zkvm/src/host/client/env.rs
```diff
@@ -351,9 +352,9 @@ impl<'a> ExecutorEnvBuilder<'a> {
     }
 
     /// Set the path where segments will be stored.
-    pub fn segment_path<P: AsRef<Path>>(&mut self, path: P) -> &mut Self {
-        self.inner.segment_path = Some(path.as_ref().to_path_buf());
-        self
+    pub fn segment_path<P: AsRef<Path>>(&mut self, path: P) -> Result<&mut Self> {
+        self.inner.segment_path = Some(TempDir::new_in(path.as_ref())?.into_path());
```
flaub, line 356: I think this breaks the user's intent. If the user wants to store segments in a particular directory, that feels to me like they're going to manage the lifetime of the segments.

risc0/zkvm/src/host/server/session.rs
```diff
@@ -80,6 +80,22 @@ pub struct Session {
 
     /// The system state of the final [MemoryImage] at the end of execution.
     pub post_state: SystemState,
+
+    /// Path of temporary segment storage used internally.
+    temp_segment_dir: Option<PathBuf>,
```
flaub, line 85: I think what we discussed was to have something like `Rc<TempDir>` on the `FileSegmentRef`.

SchmErik:
Superseded by https://github.com/risc0/risc0/pull/1421

PR #1418, opened 2024-02-07
Tune ECDSA example by adding codegen-units=1 to guest Cargo.toml

nategraf:
I happened to be looking at the ECDSA example today, and tried out adding `codegen-units=1` to the guest `Cargo.toml`.
Result was reducing 1.17M cycles to a little less than 1M. As a result, the proof is one segment and is about 20% faster.

### DIFF
--- a/examples/ecdsa/methods/guest/Cargo.lock
+++ b/examples/ecdsa/methods/guest/Cargo.lock
@@ -316,9 +316,9 @@ dependencies = [
 
 [[package]]
 name = "digest"
-version = "0.10.6"
+version = "0.10.7"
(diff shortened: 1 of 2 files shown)


flaub:
Is it possible you were running different versions of the toolchain? We recently released a new 1.70 guest toolchain where we've seen similar improvements in cycle count.

nategraf:
> Is it possible you were running different versions of the toolchain? We recently released a new 1.70 guest toolchain where we've seen similar improvements in cycle count.

In this comparison, the 1.17M vs <1M numbers come from just changing the `codegen-units=1` line, keeping all else constant.

flaub (approved):

PR #1420, opened 2024-02-07
Refactor groth16

capossele:
The following changes are suggested to make easier the integration of `Groth16` with `bonsai-alpha`.

- update Bonsai SDK to use the new `risc0_groth16::Seal`
- drop the build.Dockerfile to reduce the number of dockerfiles; circom gets called with `--r1cs` in the ceremony and with `--c` in the prover
- re-exporting the bare minimum from groth16 via `risc0-zkvm` to avoid the need to import the `risc0-groth16` crate

Solves https://github.com/risc0/risc0/issues/1524


bonsai/ethereum-relay/src/uploader/completed_proofs/snark.rs
```diff
@@ -69,40 +70,38 @@ pub(crate) async fn get_snark_receipt(
     Ok(proof)
 }
 
-pub fn tokenize_snark_receipt(proof: &Groth16Seal) -> anyhow::Result<Token> {
-    if proof.b.len() != 2 {
-        anyhow::bail!("hex-strings encoded proof is not well formed");
+pub fn tokenize_snark_receipt(seal: &Groth16Seal) -> anyhow::Result<Token> {
```
mothran, line 73: Are we planning on dropping the `bonsai/ethereum-relay` in favor of the code in `risc0-etheruem`?

bonsai/sdk/Cargo.toml
```diff
@@ -13,6 +13,7 @@ reqwest = { version = "0.11", default-features = false, features = [
   "json",
   "rustls-tls",
 ] }
+risc0-groth16 = { workspace = true }
```
mothran, line 16: We depend on `bonsai-sdk` in the `risc0-zkvm` crate as a option prover backend, this wont create a circular dep right?

compact_proof/docker/prover.Dockerfile
```diff
@@ -30,19 +38,35 @@ RUN git submodule init && \
...
+COPY groth16/stark_verify.circom ./groth16/stark_verify.circom
+
+# Build the witness generation
+RUN (cd groth16; circom --c stark_verify.circom) && \
+  sed -i 's/g++/clang++/' groth16/stark_verify_cpp/Makefile && \
+  sed -i 's/O3/O0/' groth16/stark_verify_cpp/Makefile && \
+  (cd groth16/stark_verify_cpp; make)
+
+# Download the proving key
+RUN wget https://risc0-artifacts.s3.us-west-2.amazonaws.com/zkey/2024-02-07/stark_verify_final.zkey.gz -O groth16/stark_verify_final.zkey.gz && \
```
mothran, line 55: This is a really nice clean up from downloading outside the docker build! Nice!

compact_proof/scripts/install_prover.sh
```diff
@@ -2,9 +2,4 @@
 
 set -eoux
 
-docker build -f docker/build.Dockerfile . -t snark-build --output=groth16 --target outputs
-
-wget https://risc0-artifacts.s3.us-west-2.amazonaws.com/zkey/2024-02-07/stark_verify_final.zkey.gz -O groth16/stark_verify_final.zkey.gz
-(cd groth16; gzip -df stark_verify_final.zkey.gz)
-
 docker build -f docker/prover.Dockerfile . -t risc0-groth16-prover
```
mothran, line 5: I would love if we could publish this docker image into our CI s3 bucket or ECR so we don't have to git clone the whole `risc0` repo to access the source of truth build of this docker image.

mothran (approved):
Looks good on the bonsai side, left a few comments in line

bonsai/ethereum-relay/src/uploader/completed_proofs/snark.rs
capossele: Yes that is the plan, I just have not removed that on this PR. Should be done by a different one together with some new CI to keep in check the other repo

bonsai/sdk/Cargo.toml
capossele: Yes that is exactly why I left the sdk to depend directly on the `risc0-groth16` crate rather then using the new re-export from the `risc0-zkvm`. Basically both risc0-zkvm and bonsai-sdk depend on `risc0-groth16`, so we brake the circularity.

compact_proof/scripts/install_prover.sh
capossele: Yes I agree with that! 
I feel like we have a few options:
1. publish on ECR
2. publish on GHCR
3. publish on Docker Hub

My favourite is the 3. but I might be the minority here. The the build should only be done every time we update the ceremony (i.e., we re-run the xtsask bootstrap-groth16), so not that often.  And providing the docker image publicly would also make CI faster during tests as well as faster for users to potentially run locally.

Overall though I'm open to any of the above. Docker save is also nice but not as clean imo

mothran: I would say, option 1, 3 and 4 (docker save + s3) are a best for us. We can probably make any of those options work really easily.

mothran: Option 3 and 4 are nice because it is easiest for external folks to pull the image as well and extend it.

bonsai/sdk/src/alpha.rs
```diff
@@ -49,6 +49,7 @@ pub enum SdkErr {
 
 /// Collection of serialization object for the REST api
 pub mod responses {
+    use risc0_groth16::Seal;
```
SchmErik, line 52: I saw that you exported this here: https://github.com/risc0/risc0/pull/1420/files#diff-17e9437e9950f794a22d8fcae4e2e99e315c137e1eb2950612bf0ea774188f4dR136 so IIUC, you can use it here by `use risc0-zkvm::Groth16Seal as Seal` and `risc0-groth16` from this crate's Cargo.toml

capossele: I avoided doing that as risc0-zkvm depends on bonsai-sdk, so I've imported the risc0-groth16 crate to brake the circular dependency

compact_proof/docker/ceremony.Dockerfile
```diff
@@ -1,4 +1,34 @@
 # syntax=docker/dockerfile:1
+FROM rust:1.76.0 AS dependencies
```
flaub, line 2: Let's stick to using the same version that is in use in `risc0/risc0/rust-toolchain.toml`

SchmErik (approved):

flaub (approved):

PR #1421, opened 2024-02-08
Clean up segment path by using `Rc<TempDir>`

SchmErik:
At this time, the zkVM fails to clean up segment information. Fix this by reference-counting each segment file if it uses a tempdir. We will let the user manage the lifetime of the segment if they designate a specific path to store segments. Some of this is based off of Austin's work. In this implementation, I used `Arc` because `SegmentRef` implements the `send` trait.

### DIFF
--- a/risc0/zkvm/src/host/client/env.rs
+++ b/risc0/zkvm/src/host/client/env.rs
@@ -73,7 +73,7 @@ pub(crate) struct Assumptions {
 
 #[allow(dead_code)]
 #[derive(Clone)]
-pub enum DirectoryPath {
+pub enum SegmentPath {
     TempDir(Arc<TempDir>),
     Path(PathBuf),
 }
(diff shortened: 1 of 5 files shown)


risc0/zkvm/src/host/client/env.rs
```diff
@@ -69,6 +71,13 @@ pub(crate) struct Assumptions {
     pub(crate) accessed: Vec<Assumption>,
 }
 
+#[allow(dead_code)]
+#[derive(Clone)]
+pub enum DirectoryPath {
+    TempDir(Arc<TempDir>),
```
flaub, line 77:
```suggestion
    TempDir(Rc<TempDir>),
```

Seems like we should uniformly use `Rc` or `Arc` but seems odd to have both.

risc0/zkvm/src/host/server/session.rs
```diff
@@ -336,13 +332,25 @@ impl SegmentRef for FileSegmentRef {
     }
 }
 
+impl DirectoryPath {
```
flaub, line 335:
```suggestion
impl SegmentPath {
```

flaub (approved):

PR #1424, opened 2024-02-09
`risc0_build` returning a list of `GuestListEntry`

capossele:
In the context of knowing the imageID of each method being built during the build.rs script, it would be useful to return a list of `Risc0Methods` object.

This could be then used to do something like:
```rust
let use_docker = env::var("RISC0_USE_DOCKER").ok().map(|_| DockerOptions {
        root_dir: Some("../".into()),
    });

    let methods = embed_methods_with_options(HashMap::from([(
        "bonsai-starter-methods-guest",
        GuestOptions {
            features: Vec::new(),
            use_docker,
        },
    )]));

    let image_ids = methods
        .into_iter()
        .map(|method| {
            let name = method.name.to_uppercase().replace('-', "_");
            let image_id = hex::encode(method.make_image_id());
            format!("bytes32 public constant {name}_ID = bytes32(0x{image_id});")
        })
        .collect::<Vec<_>>()
        .join("\n");

    // Building the final file content.
    let file_content = format!("{SOL_HEADER}{LIB_HEADER}\n{image_ids}\n}}");
    fs::write(SOLIDITY_IMAGE_ID_PATH, &file_content).unwrap_or_else(|err| {
        panic!(
            "failed to save changes to {}: {}",
            SOLIDITY_IMAGE_ID_PATH, err
        );
    });
```

In this case we are populating a Solidity library file with the ImageID values of each method. It also integrates nicely with the reproducible build via Docker


flaub (approved):

risc0/build/src/lib.rs
```diff
@@ -63,14 +63,18 @@ pub struct GuestListEntry<'a> {
     pub path: &'a str,
 }
 
+/// Represents a RISC Zero method
 #[derive(Debug)]
-struct Risc0Method {
-    name: String,
-    elf_path: PathBuf,
+pub struct Risc0Method {
```
nategraf, line 68: Why make both `Risc0Method` and `GuestListEntry` public? Don't they contain the same information? `GuestListEntry` looks more complete in this regard.

nategraf (commented):
Looking at this, it might make sense remove the `guest-list` feature flag and just always provide the guest list, both embedded in the methods `lib.rs` and returned from `embed_methods_with_options`. I don't really recall any downsides to providing `guest-list`.

flaub:
Can we describe what exactly we need the GuestList, like what problem are we trying to solve?

nategraf:
> Can we describe what exactly we need the GuestList, like what problem are we trying to solve?

We introduced it in https://github.com/risc0/risc0/pull/585, with the main motivating use case being in the Foundry template. We has cases where, either in the CLI or from Solidity (which uses the CLI) we needed to specify which guest to run by name, and then fetch the correct (image ID, ELF). We use it right now in the template to template by enabling the `guest-list` feature, and then having the `cli` crate (which used to be the `relay`) depend on the `methods` crate. By importing the methods crate, and scanning the guest list, we can find the specified ELF and image ID.

The motivation for this PR is to refactor this usage such that instead of having the `cli` or app crate depend on the `methods` crate, we can produce, from `build.rs` on the methods crate, a Solidity file that we can then input into the guest with image IDs and paths for the ELFs (paths only being for local testing and not included in deployed contracts).

flaub:
LGTM, thanks for talking it thru

nategraf:
I've pushed a few commits to this branch, with [8d918e0](https://github.com/risc0/risc0/pull/1424/commits/8d918e018089d1a537ee8b9544bfb49da810456a) being the main one. I refactored `GuestListEntry` to use `Cow<'static, [u8]>` and the respective type for `str`. This allows the value to be populated by a `Vec<u8>`, within the build script, or with `'static [u8]` in the methods crate.

I tried to remove the `guest-list` feature flag, and simply always codegen the list. Unfortunately I realized generating code referencing `GuestListEntry` means that the methods crate needs to depend on `risc0-build` or the build will fail. Requiring all methods crates to depend on `risc0-build`, in the main `[dependencies]` rather than only at build-time, seems less than ideal. So although a `Vec<GuestListEntry>` is always returned, the guest list is only emitted in codegen when the `guest-list` feature is enabled.

I also went ahead and deleted `Risc0Method`, since `GuestListEntry` could fill it's place, and refactored the code accordingly.

PR #1434, opened 2024-02-13
Add docs on website for I/O in the zkVM

Cardosaum:


SchmErik:
@Cardosaum looks like it failed some website tests

Cardosaum:
Thanks for pointing out, @SchmErik !

I didn't know we actually executed rust examples in the docs.

Will fix it ✅

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+use risc0_zkvm::guest::env;
+risc0_zkvm::guest::entry!(main);
+
+fn main() {
+  // Read a single value from the standard input
+  let input: u64 = env::read()?;
+
+  // Read a slice from the standard input
+  let mut buffer = [0u8; 32];
+  env::read_slice(&mut buffer)?;
```
nategraf, line 113: If this code is used with the host code above, it would fail because it tries to read more data than is written. Since we are presenting both together, I think the implication is that they should work together.

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+}
+```
+
+## Handling Private Outputs
+
+The [`stdout`][stdout] file descriptor is used to send private output from the
+guest to the host. The zkVM provides some convenience methods to write data to
+this file descriptor.
+
+- [`env::write`][write_method]: Writes a value to the `stdout` file descriptor.
```
nategraf, line 129: Sam comment as above about serialization

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+  let env = ExecutorEnv::builder()
+              // Step 2: Instruct the guest to write to the desired buffer
+              .stdout(&mut output)
+              .build();
+  let prover = default_prover();
+  let receipt = prover.prove(env, EXAMPLE_ELF).unwrap();
+
+  // Step 3: Deserialize the output
+  let example_struct: ExampleStruct = from_slice(&output).unwrap();
+  let example_slice: [u8; 3] = from_slice(&output).unwrap();
```
nategraf, line 167: This will not do what's intended, because `from_slice` is a method for deserialization, and the data written with `write_slice` is not serialized.

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+For example, in the host program, we could instantiate the `ExecutorEnv` as usual, and then get the journal data from the receipt.
+
+`/src/main.rs`
+
+```rust
+use risc0_zkvm::{default_prover, ExecutorEnv};
+
+fn main() {
+  let env = ExecutorEnv::builder()
+              .build();
```
nategraf, line 219: When we want the default, we can use the `Default` trait, which provides nice semantic clarity.
```suggestion
  let env = ExecutorEnv::default()
```

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+  let prover = default_prover();
+  let receipt = prover.prove(env, EXAMPLE_ELF).unwrap();
+
+  // Get the journal data from the receipt
+  let journal = receipt.journal;
+
+  // Do something with the journal data...
```
nategraf, line 226: Like the examples you have above, it would be good to show decoding the journal data. In particular, the `decode()` function on the `Journal` type is a nice convenience to show devs.

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+
+And in the guest program, we would write to the journal as follows:
+
+`/methods/guest/src/main.rs`
+
+```rust
+#![no_main]
+
+use risc0_zkvm::guest::env;
+risc0_zkvm::guest::entry!(main);
```
nategraf, line 238: I think in our examples, we should go ahead and assume they are using `std`. It makes the code cleaner, and lowers the initial conceptual load. I'm not 100% others agree on this, so it may be worth asking the Slack.
```suggestion
use risc0_zkvm::guest::env;
```

nategraf (commented):
I like this document a lot! I think it's structure is clear, and will help developers. I have a number of suggestions, and some issues I see in the provided examples. In particular, the examples contain bugs, which I think we should work hard to avoid. We need to make sure to test the code snippets we give in our documentation.

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
+# Introduction to Input/Output in the zkVM
+
+This document provides an overview of how to perform I/O operations in the zkVM,
+going over the different file descriptors and methods used to handle inputs and
+outputs, the distinction between public and private data and additional
+considerations when working with I/O in the zkVM.
```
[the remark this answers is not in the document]
Cardosaum: Thanks! reads better this way. :)

Qgoni:
**work work work**

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+[journal][journal_terminology] as a result of a zkVM application's execution.
+Since this data is the _output_ of the application, public data is often also
+referred to as _public outputs_.
+
+On the other hand, private data is everything that is not included in the proof.
+In other words, private data is only accessible by the host and guest. This kind
+of data can be both in the form of _inputs_ from the host to the guest and
+_outputs_ from the guest to the host.
+
+While private data is not included in the proof, it is still important to take
```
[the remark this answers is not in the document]
SchmErik: What if instead of session, we describe this situationally. We could say that "the private output is accessible to party that is generating the proof".

website/api/zkvm/input-output.md
```diff
@@ -0,0 +1,274 @@
...
+going over the different file descriptors and methods used to handle inputs and
+outputs, the distinction between public and private data and additional
+considerations when working with I/O in the zkVM.
+
+As briefly mentioned in [Guest Code 101][guest-code-101-io], we can perform I/O
+between the guest and host by using some methods under the [`env`
+module][env-module-functions]. These methods allow sending private data between
+them. In the zkVM, there is a distinction between public and private data.
+Public data is included in the journal and is part of the proof, while private
+data is sent between the guest and host and is not included in the journal nor
```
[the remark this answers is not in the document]
pdg744: we need to both be careful about creating barriers by assuming familiarity with jargon *and* be careful of explaining everything on every page. my preference is to keep things succinct and link out to more info as appropriate. 

i..e., here, i'd suggest we just link to the terminology pages for guest and host rather than explaining the terms in-line

Cardosaum:
Thanks for the feedback, @pdg744! 
Agree with your points and will work on implementing it.

website/api/zkvm/tutorials/io.md
```diff
@@ -0,0 +1,327 @@
...
+[example-voting-machine-host-init]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L77-L83
+[example-voting-machine-host-submit-output-buffer]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L88
+[example-voting-machine-host-submit-output-from_slice]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L95
+[example-voting-machine-host-submit-output-stdout]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L91
+[example-voting-machine-host-submit]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L85-L97
+[example-voting-machine-host]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs
+[example-voting-machine]: https://github.com/risc0/risc0/tree/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine
```
Cardosaum, line 303: I chose to use permalinks here because I'll be pointing to specific lines.

risc0/zkvm/src/guest/env.rs
```diff
@@ -519,7 +704,8 @@ pub struct FdWriter<F: Fn(&[u8])> {
 }
 
 impl<F: Fn(&[u8])> FdWriter<F> {
-    fn new(fd: u32, hook: F) -> Self {
+    /// Creates a new FdWriter writing to the given file descriptor.
+    pub fn new(fd: u32, hook: F) -> Self {
```
Cardosaum, line 708: I thought of making this public so users could instantiate their own FdWriters if they want.

website/api/zkvm/tutorials/io.md
```diff
@@ -0,0 +1,329 @@
...
+There are more examples available in the [examples page] that you can use as
+reference if you wish.
+
+Happy coding!
+
+[article-composition]: https://www.risczero.com/blog/proof-composition
+[docs-guest-code-optmization]: ../optimization.md#when-reading-data-as-raw-bytes-use-envread_slice
+[docs-hello-world]: ./hello-world.md
+[docs-remote-proving]: ../../generating-proofs/remote-proving.md
+[example-chess-core]: https://github.com/risc0/risc0/blob/main/examples/chess/core/src/lib.rs
```
pdg744, line 283: I think our best practice is to point to the latest release rather than `main`, although looking around at other files, it looks like we point to `main` more often than I had thought...
```suggestion
[example-chess-core]: https://github.com/risc0/risc0/blob/v0.20.1/examples/chess/core/src/lib.rs
```

flaub: If it's in next, we point to main, if it's a specific version, part of the promotion process is to go thru each main link and point it to the appropriate stable link.

risc0/zkvm/src/guest/env.rs
flaub: I think we should have `env::write_fd` and `env::read_fd` to avoid having to make the `FdWriter` public.

Cardosaum: I think `FdWriter` has [always been public](https://github.com/risc0/risc0/commit/e16e860e303b2a750cfdff3ff00f83a1c431f2d0#diff-e1117fda42459c7cce9f894e70e03f3041b2405449de7be9cee9cc966a1db30dR337)?

website/api/zkvm/tutorials/io.md
```diff
@@ -280,29 +280,29 @@ Happy coding!
...
-[example-voting-machine-host-submit-output-buffer]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L88
-[example-voting-machine-host-submit-output-from_slice]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L95
-[example-voting-machine-host-submit-output-stdout]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L91
-[example-voting-machine-host-submit]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs#L85-L97
-[example-voting-machine-host]: https://github.com/risc0/risc0/blob/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine/src/lib.rs
-[example-voting-machine]: https://github.com/risc0/risc0/tree/2deb83a15d8bd438bdc594a86438fc816d02b480/examples/voting-machine
+[example-chess-core]: https://github.com/risc0/risc0/blob/release-0.21/examples/chess/core/src/lib.rs
```
SchmErik, line 283: We actually want to point this to `main`

Cardosaum: Thanks!

website/api/zkvm/tutorials/io.md
```diff
@@ -0,0 +1,329 @@
+# Understanding I/O in the zkVM
+
+In the [Hello World Tutorial][docs-hello-world], we had a brief introduction to
+how to perform I/O operations in the zkVM. Now we'll dive deeper into the subject.
+By following the steps in this guide, you will learn:
```
pdg744, line 5: This phrase sets the reader up to expect a step-by-step tutorial or how-to guide, when this is actually an [explanation(https://documentation.divio.com/explanation.html). Suggesting a specific revision, but mostly I just want to remove the word "steps." 

```suggestion
Keep reading to learn more about:
```

pdg744: (or if we want to keep the word "steps," let's revise the text to be clear about in what sense this is a step-by-step guide.

pdg744 (approved):
Thanks, Matheus! 

This is a major improvement to clarify a common source of confusion for users around: 
- public/private inputs & outputs
- std, no-std, etc. 

As I mentioned earlier, I have some concern that this doc will be a relatively high maintenance burden for us, but we can make revisions to address that if/when it becomes an issue. 

Left one suggestion around the word "steps" in the intro, which feels misleading to me.

website/api/zkvm/tutorials/io.md
Cardosaum: Removing "steps" seems good to me, thanks for the feeback!

Cardosaum:
> As I mentioned earlier, I have some concern that this doc will be a relatively high maintenance burden for us, but we can make revisions to address that if/when it becomes an issue.

@pdg744  I agree it can be a high maintenance one, we can make iteration on this to make more manageable and identify where're the pain points and fix over time. 

I appreciate your thoughtfulness and all feedback provided! 🙇‍♂️

PR #1445, opened 2024-02-15
links to bottom

pdg744:
moving links to bottom of terminology page


morgangibbs87 (approved):
Just markdown file changes. All tests passed. LGTM.

website/docs/terminology.md
```diff
@@ -293,24 +237,74 @@ A program on the [host] that [verifies] a [receipt].
...
+[host]: #host
+[imageID]: #image-id
 [ImageID documentation]: https://docs.rs/risc0-zkvm/*/risc0_zkvm/struct.SystemState.html
 [ImageID excerpt from Study Club]: https://www.youtube.com/watch?v=QwzrBHHkzFE&list=PLcPzhUaCxlCirUkJY0ltpjdtzWcz5U_6y&index=4
+[journal]: #journal
+[method]: #method
+[prove]: #prove
+[proves]: #prove
+[proven]: #prover
+[prover]: #prover
```
flaub, line 288: This (and other examples) can be collapsed. In markdown you can do things like `[proven][prover]` and `[proven][prover]`, so that we only need a single entry in the list of links.

pdg744: tried to implement this suggestion but it seems to break everything. 
landing this as is because i have a follow-up PR i want to open

tzerrell (commented):
These definitions sound good to me

PR #1447, opened 2024-02-15
Remove optimizations from build scripts

nategraf:
Throughout our repo we apply optimizations to Cargo build scripts (i.e. binaries built to run as part of the build process).
This was introduced to make calculation of control IDs faster, as this was the bottleneck for the build process.

Now that control IDs are not calculated on a per-guest basis, it is unclear if optimizing the build scripts is still beneficial to build time.
Opening this small PR to see what the build times are in CI across the various examples and tests without this setting.


SchmErik:
So I'm looking at this run and https://github.com/risc0/risc0/pull/1483 and I'm not seeing a huge difference in execution time but it makes the Cargo.toml's look cleaner, which is nice

flaub (approved):
Since this is confined to just the `build-override` it LGTM

nategraf:
I'm going to do one more test, which is to build the examples from a clean clone and check build times with `--timings`. (Not sure why I didn't think of this flag before)

nategraf:
I've confirmed that build+run times are reduced with this PR. The effect on overall build times appears to be lower than the noise floor, but some items on the critical path (e.g. `ring`) build+run their build script significantly faster.

PR #1449, opened 2024-02-16
Update SessionStatusRes

rlukata:
This PR adds the SessionStats struct, as well as it update the SessionStatusRes to include such struct in it.


mothran (approved):
Looks good to me, can you double check it works as expected by import this branch into the integration suite on the bonsai side and double checking the API still works?

PR #1461, opened 2024-02-20
DevMode: discard segments during execution

SchmErik:
Previously, DevMode kept segments in memory. However, this is not practical for very large workloads because the segments can take up a lot of space. This change drops segments from memory during devmode by introducing a new SegmentRef impl. I tried conditionally assigning different closures to the callback variable but I quickly ran into adding starting to complicate type signatures by adding `Box<dyn ...>`. This solution wraps the callback in another closure that checks for dev mode each time segments are resolved.


mothran (approved):
This looks good on my end, I lightly tested it in my project and hit other issues we talked about that I think would make sense in a follow up. But this would solve my issue of running out of tmpfs on bigger guests while doing dev.

PR #1477, opened 2024-02-26
Change runners to self-hosted so website preview jobs can use sccache in PRs from external contributors

morgangibbs87:
The `main` and `version-0_20` jobs need to be changed to using self-hosted runners so PRs from external contributors can use the `sccache` like other workflows.

Until this is pushed, jobs like this will fail:
https://github.com/risc0/risc0/actions/runs/7972603437/job/21999427789?pr=1459#step:6:49

### DIFF
--- a/.github/workflows/website-test.yml
+++ b/.github/workflows/website-test.yml
@@ -65,7 +65,7 @@ jobs:
   main:
     if: needs.changes.outputs.main == 'true'
     needs: changes
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


rlukata:
My only caveat is that using self-hosted could mean that there are config settings (in this case credentials) that aren't visible but are required. Would it make sense to pass in AWS credentials by leveraging [aws-config](https://github.com/aws-actions/configure-aws-credentials)? This way it's less opaque what's going on.

rlukata (approved):

PR #1478, opened 2024-02-26
CI: run all jobs when Cargo.lock is changed

SchmErik:


morgangibbs87 (approved):
LGTM. Adding the root `Cargo.lock`to `base` will inherit and trigger all of the other jobs that reference `*base` in the paths filter.

PR #1482, opened 2024-02-27
'Enforce in CI that code does not contain a "DO NOT MERGE" comment'

nategraf:
Adding comments with "DO NOT MERGE" is a personal practice that is supposed to remind myself and reviewers that something of importance needs to taken care of before merging the PR. I realized when I ran `grep` in my clone of this repo that a comment had slipped through and neither I nor the reviewer ensured I had removed it. The comment was addressed, so no harm done, but I would like to have a rule in CI to check for this.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -144,7 +144,7 @@ jobs:
           python-version: "3.10"
       - run: python license-check.py
       - name: check for "D0 NOT MERGE" comments
-        run: grep -re 'DO[_ ]\?NOT[_ ]\?MERGE' $(git ls-tree --full-tree --name-only -r HEAD)
+        run: grep -re 'DO[_ ]\?NOT[_ ]\?MERGE' $(git ls-tree --full-tree --name-only -r HEAD); [ $? -eq 1 ]
 
   test:
     if: needs.changes.outputs.test == 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -143,6 +143,9 @@ jobs:
         with:
           python-version: "3.10"
       - run: python license-check.py
+      - name: check for "D0 NOT MERGE" comments
```
SchmErik, line 146: haha, i see what you did here

flaub (approved):
I guess this is fine, it feels more like a pre-commit hook to me. 

Later, it might make sense to have a python script to do all this sorts of little things, I've seen workflow files become the dumping ground of little ad-hoc bash scripts that quickly become unmaintainable.

nategraf:
> I guess this is fine, it feels more like a pre-commit hook to me.
> 
> Later, it might make sense to have a python script to do all this sorts of little things, I've seen workflow files become the dumping ground of little ad-hoc bash scripts that quickly become unmaintainable.

Ok, ya I don't want to degrade the health of our CI code 😕 . I hope this will provide enough value to make it worth the bit of extra cruft. I'll also set a reminder for 3 months to reconsider keeping this rule.

I would make it a pre-commit hook, but the intention is that it is allowed to be committed, as long as it does not make it's way to `main` or `release`.

Issue #1479, opened 2024-02-26
[Feature] Executor: configure segment storage


SchmErik:
## Feature
At the moment, the default behavior is to save the segments in memory, disk or not saving any segments. We would like the default behavior to store in memory until a certain threshold has been reached.

## Motivation
this will improve devex by switching the storage only for larger proofs.

## Implementation
change the default implementation for when `Executor.run()` function is called.

PR #1483, opened 2024-02-27
zkVM: change segment storage behavior

SchmErik:
This PR changes the segment storage behavior. Before this change, the default behavior is to store segments on disk. Change the behavior so that the first 64 segments are stored in memory. Beyond this limit, we will store segments on disk. This limit is configurable through the `ExecutorEnv`.

Here's my train of thought as I implemented this:
There’s two paths to using the zkVM: using `client/server` and using `-F prove`. For both client/server mode and using the -F prove flags, I wanted the default behavior to be “save x amount of segments to memory and save any segment beyond this limit to files”. This means that we need to initialize the segment path in either case so that in case we go over the default limit, we can save them to the `tempdir`.

`-F prove` mode:

gives the user the ability to use their own callbacks. So my thought was that calling `Executor.run()` should trigger the aforementioned default behavior. If they want to customize the behavior and provide their own call backs, they should call `Executor.run_with_callback()`.

client/server mode:

With this mode, we don't support adding custom segment callbacks. So the only behavior that’s supported is the default that I’ve described above. I’ve allowed a way for the developer to configure `ExecutorEnv` to state how many segments they want stored in RAM. If they want to store only to file, they can set this value to `Some(0)`.

Fixes: https://github.com/risc0/risc0/issues/1479


SchmErik:
converting to draft - need to write tests

flaub:
Perhaps we can revisit this later, but doesn't seem like an urgent need at the moment.

PR #1495, opened 2024-03-01
bump version to 0.22.0-alpha.1

SchmErik:


tzerrell (commented):
Wait, why are we back-to-back bumping to 0.21.0-rc.1 and 0.22.0-alpha.1 on the same branch? Should one or both of these be pointed to a release branch instead of `main`?

SchmErik:
> Wait, why are we back-to-back bumping to 0.21.0-rc.1 and 0.22.0-alpha.1 on the same branch? Should one or both of these be pointed to a release branch instead of `main`?
> 
> (Referencing #1496)

@tzerrell this PR is targeting the main branch and #1496 is targeting the release-0.21 branch

flaub (approved):

risc0/zkvm/Cargo.toml
```diff
@@ -95,6 +95,7 @@ serial_test = "3.0"
 tar = "0.4"
 tempfile = "3"
 test-log = { version = "0.2", default-features = false, features = ["trace"] }
+risc0-groth16 = { workspace = true, features = ["docker"] }
```
flaub, line 98: We should not se the `features = ["docker"]` here, instead we need to forward the `docker` flag from `risc0-zkvm` to `risc0-groth16.`

flaub (changes requested):

flaub (approved):

PR #1496, opened 2024-03-01
bump version to 0.21.0-rc.1

SchmErik:


SchmErik:
CI passes but it's technically broken. Will mark as ready after I fix the test

.gitignore
```diff
@@ -15,3 +15,4 @@ target/
 tmp/
 # Ignore .pb files generated by profiling
 *.pb
+tools/smoke-test/Cargo.lock
```
tzerrell, line 18: Is the idea of gitignorning this lock file that we want to see this test fail if dependencies update in a way that breaks us?

SchmErik:
> CI passes but it's technically broken. Will mark as ready after I fix the test

tests are fixed!

tzerrell (approved):

risc0/zkvm/Cargo.toml
```diff
@@ -95,6 +95,7 @@ serial_test = "3.0"
 tar = "0.4"
 tempfile = "3"
 test-log = { version = "0.2", default-features = false, features = ["trace"] }
+risc0-groth16 = { workspace = true, features = ["docker"] }
```
flaub, line 98: I think we need to drop the `features = ` here, and then forward the `docker` feature flag from this crate to `risc0-groth16`

PR #1498, opened 2024-03-01
Docker push CI for risc0-groth16-prover

capossele:
This PR adds a GitHub workflow to build and push the `risc0-groth16-prover` docker image on Docker Hub.
Currently it only works **on dispatch**, but we can add more automation if needed.
Versioning is managed by the env `RISC0_GROTH16_PROVER` set to the same value as the proving key on S3, but we can use the `RISC0_VERSION` if we prefer.

For this workflow to work we need to create a CI/admin user on the [risczero](https://hub.docker.com/u/risczero) Docker Hub account add these secrets to our repo:
- `DOCKER_USER`
- `DOCKER_PASSWORD`


capossele:
If we are ok with this, we can then update the `stark_to_snark` test to re-use this image rather than building it from scratch.

PR #1505, opened 2024-03-01
changing `resources` to `education hub`

pdg744:
In the status quo, it feels awkward/jarring that when you click on "resources," you land directly on the study club page. More broadly, our "resources" section feels a bit weak, and I think we have a lot of educational assets to show off.

This PR rebrands the "resources" section as an "Education Hub" and adds a landing page to orient users to the contents of that section. Previously, clicking "resources" pointed users to the study club page in a way that felt rather jarring to me.

On a follow-up PR, I think we should probably pop the proof system docs out of this section.


tzerrell (commented):
I like this approach!

website/docs/education.md
```diff
@@ -0,0 +1,32 @@
+---
+displayed_sidebar: docs
```
flaub, line 2: What does this do?

pdg744: The name of the sidebar here is confusing (I'll change it), but the function here is to make this page display a sidebar called `docs`, which is defined in the sidebars file and looks like this: 

image

image

PR #1511, opened 2024-03-04
Docs: add instructions for updating crates

SchmErik:
This PR adds clarification that one must update the `risc0` dependency in their projects to use new versions (in addition to updating `cargo-risczero` and updating the rust toolchain).


pdg744 (approved):
Thank you!

PR #1513, opened 2024-03-05
Refactor client/server API to use Receipt directly instead of Asset

nategraf:
A proposed refactor for the client/server API to use the receipt types directly instead of wrapping as `Asset`


nategraf:
NOTE: CI test is currently broken because I didn't yet update the client API tests and made a breaking change to the `Client`. https://github.com/risc0/risc0/blob/victor/receipt-asset-refactor/risc0/zkvm/src/host/api/tests.rs

risc0/zkvm/src/host/api/tests.rs
```diff
@@ -202,29 +201,29 @@ fn lift_join_identity() {
...
             .unwrap();
     }
-    let rollup_p254 = client.identity_p254(opts, rollup);
 
-    let rollup_receipt = Receipt::new(
-        InnerReceipt::Succinct(rollup_p254),
-        session.journal.bytes.into(),
-    );
+    // NOTE: identiy_p254 is generated, but never checked in this test.
+    // As long as the prover does not return an error, this test will pass.
```
flaub, line 223: AFAIK, we don't have a way to run `identity_p254` verify in rust, because we don't have a circuit that supports this configuration.

risc0/zkvm/src/host/api/tests.rs
```diff
@@ -202,29 +201,29 @@ fn lift_join_identity() {
...
             .verify_integrity_with_context(&VerifierContext::default())
             .unwrap();
     }
-    let rollup_p254 = client.identity_p254(opts, rollup);
 
-    let rollup_receipt = Receipt::new(
-        InnerReceipt::Succinct(rollup_p254),
-        session.journal.bytes.into(),
-    );
+    // NOTE: identiy_p254 is generated, but never checked in this test.
```
flaub, line 222:
```suggestion
    // NOTE: identity_p254 is generated, but never checked in this test.
```

flaub:
I still think we want to wrap with Asset, to treat the Receipt as opaque from the wire/protocol level. I thought we were going to encode using protobuf (the interior of the Asset).

Also curious what is spawning this change? Since it's a breaking change, is it worth it?

nategraf:
> I still think we want to wrap with Asset, to treat the Receipt as opaque from the wire/protocol level. I thought we were going to encode using protobuf (the interior of the Asset).
> 
> Also curious what is spawning this change? Since it's a breaking change, is it worth it?

The minimal and non-breaking version of this change is indeed to change the inner encoding of `Asset` when used for receipts. I can certainly do that. What prompted this was our discussion yesterday. Either this PR, or a new version that changes the inner `Asset` encoding to protobuf is a prereq for https://github.com/risc0/risc0/pull/1518. This implementation gives cleaner, but breaks the `Client` interface.

Happy to rebase https://github.com/risc0/risc0/pull/1518 on a more minimal version of this PR, with only the changes to the inner encoding. Either way, we can have the `compress` available.

nategraf:
Based further discussion, we are going to go in the other direction and make further use of `Asset` instead of less. Marking this PR as draft until I can come back to it.

SchmErik:
superseded by #1680. closing. we may want to open a tracking issue to use asset in more places to avoid the asset->receipt-> asset conversion...

PR #1516, opened 2024-03-05
Updated bonsai-sdk elapsed_time to f64 to match bonsai-api

mothran:
This PR updates the type on the bonsai-sdk to match the API types. We migrated to f64 for the seconds.

@flaub @SchmErik Could this be tagged for backporting into 0.21?


SchmErik (approved):

bonsai/sdk/src/alpha.rs
```diff
@@ -129,8 +129,8 @@ pub mod responses {
         pub state: Option<String>,
```
flaub, line 129: Can we make this an enum?

bonsai/sdk/src/alpha.rs
```diff
@@ -129,8 +129,8 @@ pub mod responses {
...
         /// Elapsed Time
         ///
-        /// Elapsed time for a given session, in seconds
-        pub elapsed_time: Option<u64>,
+        /// Elapsed time for a given session, in seconds (f64 floating point, string encoded)
+        pub elapsed_time: Option<String>,
```
flaub, line 133: I would think using a stronger data type than a `String` is usually preferred, seems less error prone and we can rely more on the type system. Is this just some serde-json compat issue?

bonsai/sdk/src/alpha.rs
mothran: Currently it cannot, it has some dynamically generated components.

bonsai/sdk/src/alpha.rs
mothran: Will migrate to f64 on the bonsai side, fixed in: f594fecf612cce1407baec31ce638469ed62587f

flaub (approved):

PR #1518, opened 2024-03-06
Add compress method to the Prover trait

nategraf:
This PR adds the `compress` method to the `Prover` trait, which was previously only available on the `ProverServer`

### DIFF
--- a/risc0/zkvm/src/host/client/prove/mod.rs
+++ b/risc0/zkvm/src/host/client/prove/mod.rs
@@ -79,6 +79,19 @@ pub trait Prover {
         elf: &[u8],
         opts: &ProverOpts,
     ) -> Result<Receipt>;
+
+    /// Compress a [Receipt], guaranteeing that the resulting receipt is of constant size.
+    ///
+    /// Proving will, by default, will produce a [CompositeReceipt](crate::CompositeReceipt), which
+    /// may contain an arbitrary number of receipts assembled into continuations and compositions.
+    /// Together, these receipts collectively prove a top-level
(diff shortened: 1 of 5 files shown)


nategraf:
This PR is on top of https://github.com/risc0/risc0/pull/1513. It will only be merged after that PR.

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -124,12 +124,12 @@ pub trait ProverServer {
                 InnerReceipt::Composite(assumption) => {
                     self.resolve(&conditional, &self.compress(assumption)?)
                 }
-                InnerReceipt::Fake { .. } => bail!(
+                InnerReceipt::Fake { .. } => Err(anyhow!(
```
flaub, line 127: `bail!` seems more concise, are there any actual differences between these two?

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -79,6 +79,19 @@ pub trait Prover {
         elf: &[u8],
         opts: &ProverOpts,
     ) -> Result<Receipt>;
+
+    /// Compress a [Receipt], guaranteeing that the resulting receipt is of constant size.
+    ///
+    /// Proving will, by default, will produce a [CompositeReceipt](crate::CompositeReceipt), which
```
nazar-pc, line 85:
```suggestion
    /// Proving will, by default, produce a [CompositeReceipt](crate::CompositeReceipt), which
```

SchmErik:
superseded by #1680. closing

PR #1521, opened 2024-03-06
Revert "risc0-build: move the env var check for RISC0_SKIP_BUILD (#1419)"

SchmErik:
This reverts commit c68575c3bea0f2bcf275dbc60d99fc59dd73dbdd and moves the logic to skip build inside docker and non-docker build functions.

Moving the env var to the embed_methods results in skipping the embedding of methods in addition to skipping the build. By doing this, RISC0_SKIP_BUILD will only skip the build and continue to embed the methods.rs file containing the ELF and image ID.

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -24,6 +24,8 @@ use risc0_zkvm_platform::{
 };
 use tempfile::tempdir;
 
+use crate::get_env_var;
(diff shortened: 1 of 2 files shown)


mothran (approved):
Works in my testing!

capossele (approved):
Awesome!

PR #1523, opened 2024-03-06
Specify `sccache` idle timeout to fix build issues

morgangibbs87:
Specifying sccache idle timeout to 20 minutes (1200 sec), up from 10 minutes (600 sec is the default)

### DIFF
--- a/.github/actions/sccache/action.yml
+++ b/.github/actions/sccache/action.yml
@@ -8,6 +8,8 @@ inputs:
 
 runs:
   using: composite
+  env:
+    SCCACHE_IDLE_TIMEOUT: 1200
   steps:
     - if: runner.os == 'Linux' && runner.arch == 'X64'
       run: |


.github/actions/sccache/action.yml
```diff
@@ -8,6 +8,8 @@ inputs:
 
 runs:
   using: composite
+  env:
```
mothran, line 11: I don't think this applies the env var to when sccache is actually running in the `run` steps does it?

morgangibbs87:
@flaub - I added `SCCACHE_IDLE_TIMEOUT: 1200` for all  jobs using `sccache` that also had a `cargo` command.

This may have been overkill, so let me know if it can be removed from some of these jobs.

.github/workflows/bench_datasheet.yml
```diff
@@ -55,6 +55,7 @@ jobs:
 
     env:
       FEATURE: ${{ matrix.feature }}
+      SCCACHE_IDLE_TIMEOUT: 0
```
flaub, line 58: Why don't we add this to `.github/actions/sccache` so we don't need to duplicate it everywhere

mothran: I suggested that would not apply the env var globally, but only for the duration of the run of the composite. I might have been wrong though.

flaub: Env vars can be 'exported' from the composite via: https://github.com/risc0/risc0/blob/main/.github/actions/sccache/action.yml#L45

morgangibbs87: This was added. Build times good, checks all passed.

flaub (approved):

PR #1525, opened 2024-03-07
Integrate `risc0-ethereum` on CI

capossele:


capossele:
The risc0-ethereum CI tests are now failing as we updated the `ALLOWED_IDS_ROOT`, and that is expected.

.github/cargo_local_patch.py
```diff
@@ -0,0 +1,75 @@
...
+import toml
+import os
+
+def modify_dependency(details, new_path):
+    if 'path' not in details:
+        details['path'] = new_path
+        details.pop('version', None)
+        details.pop('git', None)
+        details.pop('rev', None)
+        details.pop('branch', None)
```
nategraf, line 11: It took me a minute to understand this one, and was unsure what it should do it there is already a path.
```suggestion
# Modify the dict details in-place, replacing version and git references with the given path.
def modify_dependency(details, new_path):
    if 'path' not in details:
        details['path'] = new_path
        details.pop('version', None)
        details.pop('git', None)
        details.pop('rev', None)
        details.pop('branch', None)
    else:
        raise ValueError(f"dependency has existing path: {details['path']}")
```

.github/cargo_local_patch.py
```diff
@@ -0,0 +1,75 @@
...
+
+def process_dependencies(dependencies, base_path, dep_path_mapping):
```
nategraf, line 13:
```suggestion
# Modify the list of dependencies in-place, setting a path reference for those in the mapping.
def process_dependencies(dependencies, base_path, dep_path_mapping):
```

.github/cargo_local_patch.py
```diff
@@ -0,0 +1,75 @@
...
+        return os.path.join(*[".."] * depth)
+
+def main(cargo_toml_path, dep_path_mapping):
+    base_path = calculate_base_path(cargo_toml_path)
+
+    with open(cargo_toml_path, 'r') as file:
+        data = toml.load(file)
+
+    if 'dependencies' in data:
+        dependencies = data['dependencies']
```
nategraf, line 36: Some of the entries in 'dependencies' will be a string instead of a dict (e.g. `anyhow = "1.0"`), which will end up causing this to fail. Maybe that's ok though 🤷

.github/cargo_local_patch.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+        return os.path.join(*[".."] * depth)
+
+def main(cargo_toml_path, dep_path_mapping):
+    base_path = calculate_base_path(cargo_toml_path)
+
+    with open(cargo_toml_path, 'r') as file:
+        data = toml.load(file)
+
+    if 'dependencies' in data:
```
nategraf, line 35: We likely want to process the build dependencies as well.

.github/cargo_local_patch.py
```diff
@@ -0,0 +1,75 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Modify dependencies in Cargo.toml")
+    parser.add_argument("cargo_toml_path", help="Path to the Cargo.toml file to modify")
+    args = parser.parse_args()
+
+    # Define risc0, risc0-ethereum and bonsai dependencies and their paths here
+    # Add more dependencies and paths as needed
+    dep_path_mapping = {
+        "bonsai-rest-api-mock": "../risc0/bonsai/rest-api-mock",
```
nategraf, line 53: In order to remove a dependency on the current working directory when this script is called, it would be better to make it relative to the script itself. Using something like: `os.path.dirname(os.path.realpath(__file__)`

nategraf: We should also go ahead and check ahead of running the rest of this script that all the specified directories actually exist. It would be much harder to debug later.

.github/workflows/bonsai.yml
```diff
@@ -53,56 +53,21 @@ jobs:
...
           lfs: true
-          submodules: recursive
 
       # Install dependencies.
       - uses: ./.github/actions/rustup
       - uses: ./.github/actions/sccache
         with:
           key: Linux-default
 
-      - name: Install Foundry
```
nategraf, line 64: Happy to see this 🎉

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Checkout risc0-ethereum
+        uses: actions/checkout@v4
+        with:
+          repository: 'risc0/risc0-ethereum'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          submodules: recursive
+          path: 'risc0-ethereum'
+          ref: ${{ github.event.inputs.risc0-ethereum-branch || 'main' }}
+      - name: Local patch on risc0-ethereum
+        run: python ../risc0/.github/cargo_local_patch.py Cargo.toml
```
nategraf, line 107: Can this be made relative to a "repo root" variable or similar instead of relying on the current working directory?

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,146 @@
...
+          cargo risczero install --version $RISC0_TOOLCHAIN_VERSION
+        working-directory: risc0
+      - uses: risc0/foundry-toolchain@2fe7e70b520f62368a0e3c464f997df07ede420f
+        with:
+          version: nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
+      - name: Checkout risc0-ethereum
+        uses: actions/checkout@v4
+        with:
+          repository: 'risc0/risc0-ethereum'
+          token: ${{ secrets.GITHUB_TOKEN }}
```
nategraf, line 102: Does this need a token? If it works without it, it would be nice to remove this. Otherwise, no worries.

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,146 @@
...
+        run: python -m pip install --upgrade toml
+      - name: Install risczero toolchain
+        run: |
+          cargo install --force --path risc0/cargo-risczero
+          cargo risczero install --version $RISC0_TOOLCHAIN_VERSION
+        working-directory: risc0
+      - uses: risc0/foundry-toolchain@2fe7e70b520f62368a0e3c464f997df07ede420f
+        with:
+          version: nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
+      - name: Checkout risc0-ethereum
```
nategraf, line 98: Can we move this step to just after checking out `risc0`. It's not a functional issue, I just found it just took me a while to find this step in the flow.

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,146 @@
...
+          repository: 'risc0/risc0-ethereum'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          submodules: recursive
+          path: 'risc0-ethereum'
+          ref: ${{ github.event.inputs.risc0-ethereum-branch || 'main' }}
+      - name: Local patch on risc0-ethereum
+        run: python ../risc0/.github/cargo_local_patch.py Cargo.toml
+        working-directory: risc0-ethereum
+      - name: Bootstrap-groth16 test receipt
+        run: cargo xtask bootstrap-groth16 --test-receipt-only
```
nategraf, line 110: Does this assume `risc0-ethereum` is checked out at a particular relative path?

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,146 @@
...
+          repository: 'risc0/bonsai-foundry-template'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          submodules: recursive
+          path: 'bonsai-foundry-template'
+          ref: ${{ github.event.inputs.bonsai-foundry-template-branch || 'main' }}
+      - name: Local patch on bonsai-foundry-template
+        run: |
+          python ../risc0/.github/cargo_local_patch.py Cargo.toml
+          python ../risc0/.github/cargo_local_patch.py methods/guest/Cargo.toml
+          python ../risc0/.github/cargo_local_patch.py lib/risc0-ethereum/Cargo.toml
```
nategraf, line 133: Maybe we could have the script search recursively for `Cargo.toml` files instead of maintaining these three lines. It just seems easy to miss a place where a substitute should have been made. Also, there is another workspace in `risc0-ethereum`, which is the guest used to test the relay, so I guess that's an example of missing a `Cargo.toml` file.
```suggestion
          python ../risc0/.github/cargo_local_patch.py Cargo.toml
          python ../risc0/.github/cargo_local_patch.py methods/guest/Cargo.toml
          python ../risc0/.github/cargo_local_patch.py lib/risc0-ethereum/Cargo.toml
```

xtask/src/bootstrap_groth16.rs
```diff
@@ -26,8 +30,15 @@ use risc0_zkvm::{
 };
 use risc0_zkvm_methods::{multi_test::MultiTestSpec, MULTI_TEST_ELF, MULTI_TEST_ID};
 
-#[derive(Parser)]
-pub struct BootstrapGroth16;
+#[derive(Debug, Parser)]
+pub struct BootstrapGroth16 {
+    /// ris0-ethereum repository path
+    #[arg(long, env, default_value = "../risc0-ethereum")]
```
nategraf, line 36: I feel like this should have a default. This util doesn't need to be super simple to use, and it seems better to me to be explicit where it is being invoked. That and avoiding assumption about the existence of paths outside the repo whenever possible.
```suggestion
    #[arg(long, env)]
```

xtask/src/bootstrap_groth16.rs
```diff
@@ -158,19 +178,22 @@ fn bootstrap_test_receipt() {
...
-    ));
+    fs::write(&solidity_test_receipt_path, content).unwrap_or_else(|_| {
+        panic!(
+            "failed to save changes to {}",
+            solidity_test_receipt_path.display()
+        )
+    });
 
     // Use forge fmt to format the file.
     Command::new("forge")
```
nategraf, line 192: By the way, I had issues with this on my `lxc` machine because on Linux, `sudo` is required to run docker commands, and I didn't install `forge` in the `PATH` variable for `root`. It's a bit of a catch 22, and I ended up getting around it by commenting out all the `format` calls. Not really sure how to resolve that, and it doesn't need to be resolved in this PR as long as it works in CI.

nategraf (approved):
Love to see a red PR! 🟥 

Overall, this looks good to me. I have a number of comments and suggestions on ways to make it potentially more robust.

As I was reading this I started to think about the scenario that we forget to make a substitution, and how that would lead to some confusing to debug issues. Worst case, it would lead to use no detecting a breaking change because the patch didn't actually get applied. If it seems easy to do, it would be use to have a script to run at the end that combs all the `Cargo.lock` files for dependencies starting with `risc0-*` and confirm they are using a path reference, and not a version reference.

.github/cargo_local_patch.py
capossele: I don't think this is what we should do, as typically the external repo could have intra-workspace dependencies, i.e.:
```
risc0-build-ethereum = { version = "0.8.0-alpha.1", default-features = false, path = "build" }
risc0-ethereum-contracts = { version = "0.8.0-alpha.1", default-features = false, path = "contracts" }
risc0-ethereum-relay = { version = "0.8.0-alpha.1", default-features = false, path = "relay" }
risc0-forge-ffi = { version = "0.8.0-alpha.1", default-features = false, path = "ffi" }
```

and we want to keep those

.github/cargo_local_patch.py
capossele: I'll add a functionality to address that case

xtask/src/bootstrap_groth16.rs
capossele: I see, the easiest would be to add your user permission to run docker:
```console
sudo groupadd docker
sudo usermod -aG docker $USER
```
https://docs.docker.com/engine/install/linux-postinstall/

.github/cargo_local_patch.py
capossele: well actually we could overwrite those as well, and they will just point to the same directory practically speaking

.github/workflows/risc0-ethereum.yml
capossele: I've added the env `RISC0-ETHEREUM-PATH`

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,154 @@
...
+        run: |
+          cargo install --force --path risc0/cargo-risczero
+          cargo risczero install --version $RISC0_TOOLCHAIN_VERSION
+        working-directory: risc0
+      - uses: risc0/foundry-toolchain@2fe7e70b520f62368a0e3c464f997df07ede420f
+        with:
+          version: nightly-09fe3e041369a816365a020f715ad6f94dbce9f2
+      # Apply local patch
+      - name: Local patch on risc0-ethereum
+        run: python ../risc0/.github/cargo_local_patch.py
```
capossele, line 120:
```suggestion
        run: python ../risc0/.github/cargo_local_patch.py .
```

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,154 @@
...
+        working-directory: risc0-ethereum
+      - name: Local patch on bonsai-foundry-template
+        run: python ../risc0/.github/cargo_local_patch.py
```
capossele, line 123:
```suggestion
        run: python ../risc0/.github/cargo_local_patch.py .
```

capossele:
Converting to draft until we fix one small issue after some introduced changes

.github/cargo_local_patch.py
capossele: Changing that was creating issues between risc0-ethereum and bonsai-foundry-template. I ended up checking if the path is a subdirectory of the specific repo, and if so it does not need to be patched

risc0/groth16/src/verifier.rs
```diff
@@ -22,8 +22,9 @@ use crate::{
     from_u256, g1_from_bytes, g2_from_bytes, ProofJson, PublicInputsJson, Seal, VerifyingKeyJson,
 };
 
-// Constants from: bonsai/ethereum/contracts/groth16/Groth16Verifier.sol
-// When running a new ceremony, update them by running cargo xtask bootstrap-groth16.
+// Constants from: risc0-ethereum/contracts/src/groth16/Groth16Verifier.sol
+// When running a new ceremony, update them by running cargo xtask bootstrap-groth16
+// after udpating the new Groth16Verifier.sol on the risc0-ethereum repo.
```
SchmErik, line 27: Maybe a future improvement of this will be to save each ID to a file so we can update them more easily by running the bootstrap command rather than copy/paste. we could do `include_str!(file);` where file contains the id's themselves that get updated. I know a few places where we could do something like this like the reproducible build workflow so we could pile them up into a single gh issue.

SchmErik: btw - no action needed on this, I'm just thinking out loud

.github/workflows/risc0-ethereum.yml
```diff
@@ -0,0 +1,157 @@
...
+        run: exit 1
+
+  test:
+    if: ${{ needs.changes.outputs.test == 'true' || github.event_name == 'workflow_dispatch'}}
+    needs: changes
+    runs-on: [self-hosted, prod, Linux, cpu, docker]
+    env:
+      RUST_BACKTRACE: full
+    steps:
+      # This is a workaround from: https://github.com/actions/checkout/issues/590#issuecomment-970586842
```
SchmErik, line 78: This is another comment, no action needed at this time: so now that we don't have any sub-modules in risc0, we could consider dropping this workaround. We can still keep it around though

SchmErik (approved):

.github/cargo_local_patch.py
```diff
@@ -1,75 +1,109 @@
...
         details.pop('rev', None)
         details.pop('branch', None)
+    return details
 
-def process_dependencies(dependencies, base_path, dep_path_mapping):
-    for dep, details in dependencies.items():
-        # Check if the current dependency is in our mapping
+# Modify the list of dependencies in-place, setting a path reference for those in the mapping.
+def process_dependencies(dependencies, base_path, dep_path_mapping, start_directory):
+    for dep, details in list(dependencies.items()):
```
nategraf, line 39: Why create a `list` here rather than simply iterating over the iterator returned by `items()`?
```suggestion
    for dep, details in dependencies.items():
```

nategraf (approved):
Just gave this another look. Thanks for making those changes! I think this looks good 👌

xtask/src/bootstrap_groth16.rs
nategraf: Oh nice, that's a good tip... kinda scary too since talking to Docker gives trivial access to `root` 😅

Issue #1541, opened 2024-03-11
[BUG] Right-hand side of navbar isn't viewable on mobile


pdg744:
## Bug Report
On safari on iphone, I can't figure out how to change from v0.20 to other versions of the docs, and I can't find the links that are listed under `community`. 

## Steps to Reproduce
1. Navigate to dev.risczero.com on an iphone.

## Expected behavior
The entire navbar would be visible

flaub:
This is hidden, but I think it's possible:

1. click the menu/hamburger icon in the upper left corner
2. click "<- Back to main menu"
3. select the version you'd like (including "Next") from the "Versions" drop down

PR #1549, opened 2024-03-13
fix: always show version picker, even on mobile

nahoc:
Resolves: https://github.com/risc0/risc0/issues/1541

Context: previously, on mobile you needed to click on
* 1. mobile menu
* 2. "Back to main menu"

in order to see the version picker:

## Before:

image
image

---

This PR makes it so the version picker shows up at all times on medium/small screens:

## After:

image
image

Sadly, this isn't something that is officially supported by Docusaurus, so we need to roll out our own CSS rules for it.

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -64,12 +64,12 @@
 }
 
 @media (max-width: 996px) {
-  .navbarSearchContainer_node_modules-\@docusaurus-theme-classic-lib-theme-Navbar-Search-styles-module  {
-    position: relative !important;
-    right: 0 !important;
+  .navbar__items--right > div:has(.DocSearch-Button)  {
+    position: relative;
+    right: 0;
   }
 
-  .navbar__items--right > .navbar__item:first-child {
+  .navbar__item:has(.docsVersionDropdown) {
     display: block; /* so the version picker always shows up, even on small screens */
   }
 }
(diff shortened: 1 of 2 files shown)


website/src/css/custom.css
```diff
@@ -62,3 +62,14 @@
 .anchor {
   scroll-margin-top: 50pt;
 }
+
+@media (max-width: 996px) {
+  .navbarSearchContainer_node_modules-\@docusaurus-theme-classic-lib-theme-Navbar-Search-styles-module  {
+    position: relative !important;
+    right: 0 !important;
```
nahoc, line 69: have to use `!important` since the CSS for this element are lower down the CSS cascade and can't be overwritten otherwise

website/src/css/custom.css
```diff
@@ -62,3 +62,14 @@
...
+  }
+
+  .navbar__items--right > .navbar__item:first-child {
```
nahoc, line 72: This isn't ideal either, but I don't see a clean way to inject custom classes/ID to those navbar elements via docusaurus.

flaub:
Will this change be effected by upgrades to docusaurus?

nahoc:
> Will this change be effected by upgrades to docusaurus?

Can't guarantee anything, but IMO it'll be fine unless it's a major version change.

flaub (approved):

Issue #1517, opened 2024-03-05
Investigate tools for checking semver violations


SchmErik:

SchmErik:
Here's an excellent discussion: https://github.com/rust-lang/cargo/issues/374#issuecomment-1194163503

SchmErik:
From reading the thread `rust-semverver` was deprecated because there weren't any active maintainers https://github.com/rust-lang/cargo/issues/374#issuecomment-1194069386

SchmErik:
[cargo-breaking](https://github.com/iomentum/cargo-breaking) is another one but it's also deprecated and recommends cargo-semver-checks.

PR #1556, opened 2024-03-14
CI: add cargo-semver-checks

SchmErik:
This adds an action to check an in-coming PR against the target to point out any changes in the crates we publish that might break semver. This implementation lists a few packages that we exclude from the check. Among the excluded are packages that do not need to be published, will be moved, or ones that cargo semverchecks cannot locate correctly due to a possible bug (`risc0-zkvm-platform`).

This change is designed for us to get used to seeing what sort of changes break semver and is intended to not block PR's from getting at this time.

Closes: #1517


SchmErik:
marking as draft - need to make sure this is working as expected on the CI before marking as ready for review

SchmErik:
@flaub this is ready for review. A few things: this will fail because the runners need `cmake` installed. I'd like to make this check non-blocking for PR's meaning that if it fails on a future PR that make breaking changes, we should still be able to merge the PR

morgangibbs87 (approved):

PR #1557, opened 2024-03-15
fix: bug with unaligned short read

austinabell:
So what's currently happening is when requesting an unaligned amount of bytes and the read is short, such that it crosses a word boundary, the guest will overwrite the last word with the lastword return from the host. This is a workaround to avoid changing the host.

Right now what this does is:

```
|----------request 15 bytes---------------|
|-----3x SyscallRecord words------|
00 01 02 03 04 05 06 07 08 09 0a 00
```

and then the last word writes over the end of the buffer, zeroing the bytes written in the last word until aligned:

```
                        |lastword-|
00 01 02 03 04 05 06 07 00 00 00 00
```

Alternative being the host doesn't statically set the guest memory based on `#requested.floor_div(4)`, but instead only create a `SyscallRecord` with the full words written, and re-use this fill from last word logic.

```
|----------request 15 bytes---------------|
|--2x SyscallRecord---| |lastword|
00 01 02 03 04 05 06 07 08 09 0a
```

~~where currently what it does in this case (from changes in this PR) is just removing writing the lastword if a short read past a word boundary:~~ Edit: this functionality was changed to write zero bytes in the lastword slot, such that zeroes are filled until the request length
```
|----------request 15 bytes---------------|
|-----3x SyscallRecord words------|
00 01 02 03 04 05 06 07 08 09 0a 00
```

Note that with this proposed implementation, as well as the current implementation, if a lot more bytes are requested than available, there are more words in the `SyscallRecord` created than necessary, setting all of the buffer bytes to 0. Unclear to me how much of an impact this has on proving.

```
|---------------request 19 bytes------------------------|
|----------4x SyscallRecord words-------------|
00 01 02 03 04 05 06 07 08 09 0a 00 00 00 00 00
```

I don't know what is more ideal, but hope this workaround or at least one of these tests is helpful. Those tests were just created to help narrow down the issue, happy to change them if this PR actually comes through :)


risc0/zkvm/methods/guest/src/bin/multi_test.rs
```diff
@@ -169,7 +169,7 @@ fn main() {
             for (pos, len) in pos_and_len {
                 let num_read =
                     unsafe { sys_read(fd, buf.as_mut_ptr().add(pos as usize), len as usize) };
-                assert_eq!(num_read, len as usize);
+                debug_assert!(num_read <= len as usize);
```
nategraf, line 172: This `debug_assert` will get compiled into the guest. We should instead make the assert test the more accurate condition we need to check.

austinabell: yeah this was just changed because one of the tests doesn't uphold this (since it does a valid short read). Not sure why I switched it to a debug assert though, that was a mistype

Happy to refactor this if that test specifically is kept.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -476,16 +476,20 @@ pub unsafe extern "C" fn sys_read(fd: u32, recv_ptr: *mut u8, nread: usize) -> u
...
-    );
+    if nread_main > main_requested - unaligned_at_end {
+        // If read more than the expected max aligned, write the extra bytes to the end
+        // of the buffer.
+        fill_from_word(
+            main_ptr.add(read_words * WORD_SIZE),
+            lastword,
+            unaligned_at_end,
+        );
+    }
```
nategraf, line 492:
```suggestion
    // The last 0-3 bytes are returned in lastword. Write those to complete the _requested_ read amount.
    fill_from_word(
        main_ptr.add(main_words * WORD_SIZE),
        lastword,
        unaligned_at_end,
    );
```

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -348,6 +348,34 @@ fn posix_style_read() {
     }
 }
 
+#[test]
+fn unaligned_short_read_bug() {
```
nategraf, line 352: When we look at this in 6 months, it might mean odd that this test is labeled "bug".
```suggestion
fn unaligned_short_read() {
```

risc0/zkvm/methods/guest/src/bin/multi_test.rs
```diff
@@ -169,7 +169,7 @@ fn main() {
             for (pos, len) in pos_and_len {
                 let num_read =
                     unsafe { sys_read(fd, buf.as_mut_ptr().add(pos as usize), len as usize) };
-                assert_eq!(num_read, len as usize);
+                assert!(num_read <= len as usize);
```
nategraf, line 172: It would be good to commit `num_read` here, and check it on the host in the test.

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -738,6 +766,23 @@ fn args() {
     }
 }
 
+#[test]
+fn buf_read() {
+    let input = b"1234567";
+    let env = ExecutorEnv::builder()
+        .env_var("TEST_MODE", "BUF_READ")
+        // Previously failed on anything > buf and not % 4 == 0
```
nategraf, line 774: It's worth referencing this PR for more context.
```suggestion
        // Previously failed on anything > buf and not % 4 == 0
        // https://github.com/risc0/risc0/pull/1557
```

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -738,6 +766,23 @@ fn args() {
     }
 }
 
+#[test]
+fn buf_read() {
+    let input = b"1234567";
```
nategraf, line 771:
```suggestion
    // Host-provided input is 7 bytes, while the guest requests to read 9.
    let input = b"1234567";
```

nategraf (approved):

nategraf:
After discussion, we confirmed this issue but clarified the expected behavior. When the guest requests more data than the host provides, the circuit constrains the copy-in will overwrite exactly as many words as the number of bytes requested, rounded down to the nearest multiple of `WORD_SIZE=4`. In the code adjusted with https://github.com/risc0/risc0/pull/1557/commits/2e41d84692579a48bff9de61ecca515bf266c8b9, the guest upholds the simple invariant that it will overwrite the entire region pointed to in the `sys_read` request. When the host has less data to send, any data in the tail past what the host provides will be filled with zeroes, up to and including any trailing unaligned writes. (If we are being very precise, the host always provides exactly as much data as the guest requests, but it will set this tail to zeroes and return the length of meaningful data in it's return registers from the ecall)

risc0/zkvm/methods/std/src/bin/standard_lib.rs
```diff
@@ -44,6 +44,14 @@ fn main() {
             let args: Vec<String> = std::env::args().collect();
             risc0_zkvm::guest::env::commit(&args);
         }
+        "BUF_READ" => {
+            use std::io::BufRead;
```
flaub, line 48: This should go to the top of file

risc0/zkvm/methods/std/src/bin/standard_lib.rs
```diff
@@ -44,6 +44,14 @@ fn main() {
...
+            let capacity: usize = risc0_zkvm::guest::env::read();
+            let mut reader =
+                std::io::BufReader::with_capacity(capacity, risc0_zkvm::guest::env::stdin());
```
flaub, line 51: A use would be good here

flaub (approved):

nategraf:
Looks like the Ethereum CI is failing with an issue on recursive checkout.
https://github.com/risc0/risc0/actions/runs/8514140789/job/23319263768?pr=1557#step:5:168

I am going to look into that issue. Also, I feel quite confident that this PR is not at risk of breaking the `risc0-ethereum`, so someone wants to hit the (red) merge button, feel free. I don't want to make that call by myself.

PR #1558, opened 2024-03-15
bandaid: switch back to cloudflare

SchmErik:
This is a temporary solution to get CI to pass again


SchmErik:
Another thought: we could just save the eth transaction locally rather than hammering on the rpc end point..

PR #1563, opened 2024-03-15
Restore the getrandom feature in rust-runtime.a static lib

nategraf:
Restore the getrandom feature in `rust-runtime.a` static lib. This static library build primarily supports the `cargo risczero test` tool, which we use in our workflows to test patched cryptography crates and potentially other crates developed for the zkVM. `getrandom` is intended to be supported, but was removed with the change of default behavior in https://github.com/risc0/risc0/pull/1162.

### DIFF
--- a/risc0/build/src/lib.rs
+++ b/risc0/build/src/lib.rs
@@ -317,12 +317,7 @@ pub fn cargo_command(subcmd: &str, rust_flags: &[&str]) -> Command {
 pub fn build_rust_runtime() -> String {
     build_staticlib(
         "risc0-zkvm-platform",
-        &[
-            "rust-runtime",
-            "panic-handler",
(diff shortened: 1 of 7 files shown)


SchmErik:
@nategraf it looks like you have some other changes in addition to the getrandom feature

nategraf:
Sorry, yes. I was working on getting `cargo risczero test` to work, which is where the null segment ref changes come in. I'll refactor/clean this up tomorrow. I do want to land all the included changes, but the description needs to reflect the contents, and possibly two PRs are in order.

flaub:
I think we need to get some CI around this staticlib. Ideally we'd run `cargo risczero test` somewhere in CI

SchmErik:
superseded by #1750. closing

PR #1567, opened 2024-03-18
Benchmarks: Update `write_slice` and `session.segments.len`

morgangibbs87:
1. `write_slice` set to `10_000` so segments in all three CI environments will be `3`

2. `session.segments.len` set to `> 2` instead of equal to 2

<br /><br />
The above changes are required for Benchmarks to work and avoid failures like these where CI panics when the segment does not equal `2`:
https://github.com/risc0/risc0/actions/runs/8332180237/job/22800741662

<br /><br />
These changes need to be committed before the rest of the `bench_pr.yml` changes can be tested properly, as that workflow requires checking out the `main` branch with these changes in them so the initial baseline works. After that,  a reference branch can be checked out and compared.

@flaub @nategraf @ssh-nkar @SchmErik


SchmErik:
@morgangibbs87 could you edit this https://github.com/risc0/risc0/pull/1567#issue-2193634588 so that it doesn't include images, when we squash and merge the PR, the image links could get in the commit messages which isn't very helpful when reading commit logs

SchmErik (approved):
Please make sure that we don't put image links in the commit messages

SchmErik (changes requested):
we'll implement a different solution for benchmarks

PR #1571, opened 2024-03-20
Update recursion circuit

tzerrell:
Since this includes a lot of generated code, I want to highlight the important and manually written bits:
* When making sequential Poseidon2 hash calls, instead of adding the new data into the existing state, overwrite the state with this new data. This involves both CPU and GPU changes.
* Adds a `checked_bytes` operation to the recursion circuit allowing the input / verification of non-det data constrained to be bytes.
* Add a "`bn254_control_id`" to constrain the stark2snark circom circuit to only allow the proper final recursion program.
* Pinning `dockerfile` version to handle a strange bug where it wanted to roll back to a pre-1.4 version on my system

The Poseidon2 changes only affect the state data that is written to when adding input data (i.e. the first 16 elements in our system); the other elements (the last 8 in our system) retain their state data.


xtask/src/templates/control_id_zkr.rs
```diff
@@ -16,3 +16,5 @@
 pub const ALLOWED_IDS_ROOT: &str = "{}";
 
 pub const RECURSION_CONTROL_IDS: [(&str, &str); {}] = [{}];
+
+pub const RECURSION_PROGRAM_ID: &str = "{}";
```
flaub, line 20: I think maybe this didn't get updated?

tzerrell: Thank you! Failed to save the file. Should be fixed now: I've re-run `cargo xtask bootstrap` (which is what this would break) and it changes no files.

flaub (approved):

PR #1574, opened 2024-03-22
link updates for verifier contract

pdg744:
This PR addresses two issues with link pointers on the dev docs page for the verifier contract:

- The link on line 43 was pointing to the verifier on `main` in `risc0-ethereum`, but the verbiage there seems to be discussing the deployed verifier contract. To address this, this PR replaces the pointer to `github` with a pointer to `etherscan`. 
- The entry in the table indicates that the the verifier on `main` in `risc0-ethereum` is the one deployed at `0x83C2e9CD64B2A16D3908E94C7654f3864212E2F8`. This means that if `main` changes, our docs will become misleading. To address this, this PR updates the link to point to a specific commit rather than `main`.

PR #1576, opened 2024-03-22
Update verifier contract address on docs

Cardosaum:
After changing the deployed contract on Sepolia, we have to update the docs so users can find it.


capossele:
I think we already have #1574 for this

Cardosaum:
Closing in favor of #1574

PR #1579, opened 2024-03-22
improving findability of control root info

pdg744:
The control root is a key part of our version control story, but at the moment it's not super easy to find an appropriate source of truth for what this actually is. 

This PR makes it easier to navigate from the terminology entry to the relevant section in the codebase. 

On a follow-up PR, would be great to align the naming here: it seems that what we call `control root` in the terminology page is called `ALLOWED_IDS_ROOT` in the recursion repo.


pdg744:
I haven't really thought about the naming question, but I think `ALLOWED_IDS_ROOT` is perhaps a better choice than `control root`, since it's less likely to be confused with `control ID`.

pdg744:
Addressed by #1588

PR #1588, opened 2024-03-25
adding pointer to awesome risc0

pdg744:


austinabell (approved):
Has the gh account inversebrah agreed to maintain this repo, and do we have permissions to maintain it? LGTM generally, though

website/docs/terminology.md
```diff
@@ -70,7 +70,8 @@ The control root is the Merkle hash of a set of [control ID]s.
 
 The control root plays a key role in enforcing zkVM version control. Each [control ID] identifies one [recursion program]; the control root identifies the list of allowed recursion programs. When the RISC-V circuit or recursion programs change, the control root is updated to reflect the new set of allowed functions.
 
-This design allows for updates to the [RISC-V circuit] without necessitating a new trusted setup for the [Groth16 circuit].
+This design allows for updates to the [RISC-V circuit] without necessitating a new trusted setup for the [Groth16 circuit]. <br/>
+See also: [Code reference for control root]
```
SchmErik, line 74: Looks like you have changes from groth16 related content here. Was this intentional?

website/sidebarsApi.js
```diff
@@ -173,6 +173,11 @@ export default {
           label: "zkVM technical specification",
           id: "zkvm/zkvm-specification",
         },
+        {
+          type: "link",
+          label: "Awesome risc0",
+          href: "https://github.com/inversebrah/awesome-risc0",
```
SchmErik, line 179: I feel like this is a little bit of a strange territory... If we create this link, we're reliant on this individual maintaining this list on our behalf. Why not have a page on our own page listing these projects and others? I think it would be better for maintainability if we have a page that features projects that we maintain

website/docs/terminology.md
pdg744: hmm i can't recall whether I intentionally put these into the same PR or not; in retrospect I should have put them into separate PRs but these are sufficiently minor changes that I'm going to go ahead and land them.

pdg744:
Agreed that it's a little bit of a strange territory, but we have a findability problem around "What interesting projects have external people built using RISC Zero?"

I think this is the clear best option given what currently exists. I'm open-minded about what we do in the future, but I'm going to go ahead and land this as a good-enough-for-now option. 

-- 
As to the question of whether to continue using this moving forward... The repo describes itself as community-maintained; I think it makes sense for us to support the community maintenance of this existing resource rather than take on the maintenance burden of managing this ourselves. If we do want to take on the burden of managing this ourselves, we need to identify an owner.

pdg744:
> Has the gh account inversebrah agreed to maintain this repo, and do we have permissions to maintain it? LGTM generally, though

The repo describes itself as community maintained.

PR #1592, opened 2024-03-25
Add cargo risczero deploy

### LINKED ISSUE #1363: [Feature] `cargo risczero deploy` to upload ELF binaries to Bonsai
## Feature
The ability to upload guest images to Bonsai from the command line.

This could be a flag to the `build` command (ala `cargo risczero build --deploy`), or could be offered as a separate command (`cargo risczero deploy`).

## Motivation
The `cargo risczero build` tool performs a reproducible build of zkVM guest images. In many use cases, the next step (after the build) is to upload the image(s) to Bonsai. It would be convenient for the upload to be performed by the `cargo risczero` cool.

## Implementation
`cargo risczero` would need to add either (1) a flag to the `build` command and/or (2) a new command that creates a Bonsai client and performs the upload(s).

Cardosaum:
This PR adds a new command to `cargo risczero`, building the desired guest code using reproducible builds and uploading it to Bonsai. 

The image ids of all uploaded ELFs are printed upon completion.

Usage: 
```
Uploads the guest code to Bonsai

Usage: cargo risczero deploy [OPTIONS] --manifest-path <MANIFEST_PATH>

Options:
      --manifest-path <MANIFEST_PATH>
          Location of the Cargo.toml for the guest code.
          
          This path is relative to the current directory.

      --features <FEATURES>
          Feature flags passed to cargo

      --api-url <API_URL>
          API URL for Bonsai

      --api-key <API_KEY>
          API key for Bonsai

  -h, --help
          Print help (see a summary with '-h')
```

<details>
<summary>
Example Run
</summary>

```
$ cargo risczero deploy --manifest-path examples/prorata/methods/guest/Cargo.toml --api-url='<url>' --api-key='<key>'

cargo:rerun-if-env-changed=RISC0_SKIP_BUILD
Docker context: <path>
Building ELF binaries in prorata-guest for riscv32im-risc0-zkvm-elf target...
Docker version 25.0.3, build 4debf411d1
[+] Building 17.3s (10/10) FINISHED                                                                                                                                                                               docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                        0.0s
 => => transferring dockerfile: 768B                                                                                                                                                                                        0.0s
 => [internal] load metadata for docker.io/risczero/risc0-guest-builder:v2024-02-08.1                                                                                                                                       0.6s
 => [build 1/5] FROM docker.io/risczero/risc0-guest-builder:v2024-02-08.1@sha256:575eed761f3c84ee1ae6fedb36e0794ae9c191d540437688b3fd6d7038cf5222                                                                           0.0s
 => [internal] load build context                                                                                                                                                                                           0.1s
 => => transferring context: 113.15kB                                                                                                                                                                                       0.1s
 => CACHED [build 2/5] WORKDIR /src                                                                                                                                                                                         0.0s
 => [build 3/5] COPY . .                                                                                                                                                                                                    1.6s
 => [build 4/5] RUN cargo +risc0 fetch --locked --target riscv32im-risc0-zkvm-elf --manifest-path examples/prorata/methods/guest/Cargo.toml                                                                                 4.0s
 => [build 5/5] RUN cargo +risc0 build --release --locked --target riscv32im-risc0-zkvm-elf --manifest-path examples/prorata/methods/guest/Cargo.toml                                                                      10.6s 
 => CACHED [export 1/1] COPY --from=build /src/target/riscv32im-risc0-zkvm-elf/release /prorata_guest                                                                                                                       0.0s 
 => exporting to client directory                                                                                                                                                                                           0.1s 
 => => copying files 51.31MB                                                                                                                                                                                                0.1s 
ELFs ready at:                                                                                                                                                                                                                   
ImageID: 1d476edb66bebb98b49af9bc22740f3a039aaac706554fc8fbb72591ebd33ad3 - "target/riscv-guest/riscv32im-risc0-zkvm-elf/docker/prorata_guest/prorata-guest"                                                                     
Docker context: <path>                                                                                                                                                                           
Uploaded ELF `prorata-guest` with image ID `1d476edb66bebb98b49af9bc22740f3a039aaac706554fc8fbb72591ebd33ad3`.
```

</details>

---
Closes #1363


risc0/cargo-risczero/src/commands/deploy.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        Ok(())
+    }
+
+    fn deploy(&self, client: Client) -> Result<()> {
+        // Ensure we have an up to date artifact before deploying
+        if let BuildStatus::Skipped = build_guest::build(&self.manifest_path, &self.features)? {
+            eprintln!("Build skipped, nothing to deploy.");
+            return Ok(());
+        }
```
Cardosaum, line 65: I opted to always compile the guest using reproducible builds as it seemed the right thing to do. Please let me know if it would be desirable to upload non-reproducible builds too.

risc0/cargo-risczero/src/commands/deploy.rs
```diff
@@ -0,0 +1,115 @@
...
+                    "Uploaded ELF `{}` with image ID `{}`.",
+                    target.name, image_id_hex
+                );
+            }
+        }
+
+        Ok(())
+    }
+
+    fn get_client(&self) -> Result<Client> {
```
SchmErik, line 88: i can foresee this being moved somewhere more general if we create more commands that use bonsai API. Maybe we can move this to this file https://github.com/risc0/risc0/blob/40e38a7ddbdc6d3a479a5efde999c4c2bc3c1e6a/risc0/cargo-risczero/src/utils.rs?

Cardosaum: Thanks for the feedback!

I moved the `get_client` and `get_client_envs` functions to `utils.rs` - Also learned that you can `flatten` clap arguments so we're able to now leverage `ClientEnvs` in other places if we need `get_client` and it will be consistent. :)

SchmErik (dismissed):
Looks like you added bonsai/ethereum/lib submodules to this. please drop them. I believe some of the lock file changes are rooted in this as well..

Cardosaum:
> Looks like you added bonsai/ethereum/lib submodules to this. please drop them. I believe some of the lock file changes are rooted in this as well..

Oh Thanks for spotting that!

Removed these submodules, but I think the lock file changes are due to the addition of the following dependencies: 

```
which = "6.0.1"
backoff = "0.4.0"
```

compact_proof/seal_to_json/Cargo.lock
flaub, line 1: This file should be dropped

risc0/build/Cargo.toml
```diff
@@ -19,6 +19,8 @@ risc0-zkvm-platform = { workspace = true }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde_json = "1.0"
 tempfile = "3"
+which = "6.0.1"
+backoff = "0.4.0"
```
flaub, line 23:
```suggestion
which = "6.0"
backoff = "0.4"
```

flaub: Please sort deps

risc0/build/src/docker.rs
```diff
@@ -76,21 +95,62 @@ pub fn docker_build(manifest_path: &Path, src_dir: &Path, features: &[String]) -
...
+
+fn canonicalize_path(path: &Path) -> Result<PathBuf> {
+    path.canonicalize()
+        .context(format!("Failed to canonicalize path: {path:?}"))
+}
+
+/// Get the path to the ELF binary.
+pub fn get_elf_path(
+    src_dir: impl AsRef<Path>,
+    pkg_name: impl AsRef<Path> + ToString,
```
flaub, line 122:
```suggestion
    pkg_name: impl ToString,
```

risc0/build/src/docker.rs
```diff
@@ -36,23 +48,30 @@ const DOCKER_IGNORE: &str = r#"
...
-    eprintln!("Docker context: {src_dir:?}");
-
-    let meta = MetadataCommand::new()
-        .manifest_path(&manifest_path)
-        .exec()
-        .context("Manifest not found")?;
-    let root_pkg = meta.root_package().context("failed to parse Cargo.toml")?;
+    let manifest_path = canonicalize_path(manifest_path)?;
+    let src_dir = canonicalize_path(src_dir)?;
+    let root_pkg = get_root_pkg(&manifest_path, &src_dir)?;
```
flaub, line 74: Why was this changed?

risc0/build/src/docker.rs
```diff
@@ -76,21 +95,62 @@ pub fn docker_build(manifest_path: &Path, src_dir: &Path, features: &[String]) -
...
         let rel_manifest_path = manifest_path.strip_prefix(&src_dir)?;
         create_dockerfile(rel_manifest_path, temp_path, pkg_name.as_str(), features)?;
-        build(&src_dir, temp_path)?;
+        build(src_dir.as_path(), temp_path)?;
     }
     println!("ELFs ready at:");
 
-    let target_dir = src_dir.join(TARGET_DIR);
-    for target in root_pkg.targets.iter() {
+    for target in get_targets(&root_pkg) {
```
flaub, line 102: Why did this change?

risc0/cargo-risczero/Cargo.toml
```diff
@@ -25,12 +26,14 @@ downloader = { version = "0.2", default-features = false, features = [
 ] }
 flate2 = "1"
 fs2 = "0.4"
+hex = { version = "0.4.3", default-features = false }
```
flaub, line 29:
```suggestion
hex = { version = "0.4", default-features = false }
```

risc0/cargo-risczero/src/utils.rs
```diff
@@ -22,8 +22,13 @@ use std::{
 };
 
 use anyhow::{anyhow, Context, Result};
+use bonsai_sdk::{alpha::Client, API_KEY_ENVVAR, API_URL_ENVVAR};
+use clap::Args;
 use fs2::FileExt;
 
+/// Reports the current version of this crate.
+pub const VERSION: &str = env!("CARGO_PKG_VERSION");
```
flaub, line 30: The version of `cargo-risczero` shouldn't be used for this. We should use the version of `risc0-zkvm`.

risc0/build/src/docker.rs
```diff
@@ -205,6 +265,39 @@ fn compute_image_id(elf_path: &Path) -> Result<String> {
     Ok(image.compute_id().to_string())
 }
 
+/// Check if docker is running.
+fn check_docker(docker: impl AsRef<std::ffi::OsStr>, max_elapsed_time: Duration) -> Result<()> {
+    eprintln!("Checking if Docker is running...");
+    let backoff = backoff::ExponentialBackoffBuilder::new()
```
flaub, line 271: I don't think we need to do backoff for checking if docker is running. What's the thinking behind this?

risc0/build/src/docker.rs
```diff
@@ -205,6 +265,39 @@ fn compute_image_id(elf_path: &Path) -> Result<String> {
...
+    match result {
+        Ok(true) => Ok(()),
+        _ => Err(anyhow!("Docker engine is not running")),
+    }
+}
+
+/// Ensures that docker is running.
+fn ensure_docker_is_running() -> Result<()> {
+    let docker = which("docker").context(DOCKER_MSG)?;
+    check_docker(docker, Duration::from_secs(5)).context(DOCKER_MSG)?;
```
flaub, line 297: This doesn't feel very robust, why `5` seconds? My bigger worry is that we might mask underlying issues by trying to detect if docker is around. If we let a docker command fail on-demand, at least then the user can see the full message (and it might not be related to whether docker is running or not).

risc0/build/src/docker.rs
```diff
@@ -76,21 +95,62 @@ pub fn docker_build(manifest_path: &Path, src_dir: &Path, features: &[String]) -
...
         }
     }
 
-    Ok(())
+    Ok(BuildStatus::Success)
+}
+
+fn canonicalize_path(path: &Path) -> Result<PathBuf> {
+    path.canonicalize()
+        .context(format!("Failed to canonicalize path: {path:?}"))
```
flaub, line 116: Does this error show up often? It doesn't feel like it's worth having a special helper function just for this case.

flaub (changes requested):
There are some required changes, and then a bunch of nits mostly.

compact_proof/seal_to_json/Cargo.lock
Cardosaum: Oh okay, will delete it ✅

risc0/build/Cargo.toml
Cardosaum: Dependencies sorted now.

And given the feedback about checking if docker is running, I ended up deleting these dependencies as I removed the entire docker check.

risc0/build/src/docker.rs
Cardosaum: Thanks!

risc0/build/src/docker.rs
Cardosaum: I thought it would be interesting to create a helper function that could be used both in `docker.rs` and in other places if we need to get the root package. 

This was the thinking behind adding some new public functions in the file - I decided to remove them and just use the same logic in `deploy.rs` now, so the only new export that `docker.rs` has is `TARGET_DIR` and `BuildStatus`.

risc0/build/src/docker.rs
Cardosaum: Same as [here](https://github.com/risc0/risc0/pull/1592#discussion_r1544908955) - To avoid code duplication. 

I realize now it's better to just use `root_pkg` as it was before - so the changes were reverted ✅

risc0/cargo-risczero/Cargo.toml
Cardosaum: thanks!

risc0/cargo-risczero/src/utils.rs
Cardosaum: Oh, I thought that by sharing the same workspace and version definition there wouldn't be a problem doing the same thing as in `risc0-zkvm`. But I changed no to use `risc0_zkvm::VERSION` instead ✅

Thanks for pointing out!

risc0/build/src/docker.rs
Cardosaum: The backoff was originally added because we first had an initial check to see if docker is running, and if it's not then we'd try to start the engine, which could take a while hence the backoff when doing the checks. 

I realize this was a bad idea and the whole docker check is already reverted.

risc0/build/src/docker.rs
Cardosaum: Makes sense! 

I removed the docker checks ✅

risc0/build/src/docker.rs
Cardosaum: It does not, I reverted the code to simply use `path.canonicalize()?` ✅

Cardosaum:
Thank you for your time reviewing the PR @flaub !

I made some changes to address your comments.

.vscode/settings.json
```diff
@@ -1,75 +1,92 @@
...
-		"cstddef": "cpp",
-		"*.bazel": "starlark",
-		"fstream": "cpp",
-		"ios": "cpp",
-		"iosfwd": "cpp",
-		"vector": "cpp",
-		"charconv": "cpp",
-		"stdexcept": "cpp",
-		"array": "cpp",
+  "[rust]": {
```
SchmErik, line 2: @Cardosaum were these changes intentional?

Cardosaum: Oh.. My vscode probably auto-formatted the file, sorry for not catching that.

The only changes I spotted there are two new entries to `cSpell.words`. Would it be okay to leave as is (would we have a default formatter for json?)

flaub (approved):

PR #1608, opened 2024-03-28
Update p0tion metadata

tzerrell:
Update metadata for p0tion to new file locations, improved descriptions, etc. Dates are still placeholders but at least now they're placeholders in the future rather than the past.

### DIFF
--- a/compact_proof/groth16/p0tionConfig.json
+++ b/compact_proof/groth16/p0tionConfig.json
@@ -21,8 +21,8 @@
         "cfOrVm": "CF"
       },
       "artifacts": {
-        "r1csStoragePath": "http://risc0-artifacts.s3.us-west-2.amazonaws.com/tsc/stark_verify.r1cs",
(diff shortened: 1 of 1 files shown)


compact_proof/groth16/p0tionConfig.json
```diff
@@ -1,30 +1,30 @@
...
         "paramsConfiguration": []
       },
       "verification": {
         "cfOrVm": "CF"
       },
       "artifacts": {
-        "r1csStoragePath": "http://risc0-p0tion-ceremony.s3.us-east-2.amazonaws.com/verify.r1cs",
-        "wasmStoragePath": "http://risc0-p0tion-ceremony.s3.us-east-2.amazonaws.com/verify.wasm"
+        "r1csStoragePath": "http://risc0-artifacts.s3.us-east-2.amazonaws.com/stark_verify.r1cs",
+        "wasmStoragePath": "http://risc0-artifacts.s3.us-east-2.amazonaws.com/stark_verify.wasm"
```
flaub, line 25: We might want to have this be versioned (perhaps via timestamp), just in case we need to turn another ceremony down the road.

flaub (approved):

PR #1614, opened 2024-03-29
Make list of allowed ZKRs explicit in bootstrap

nategraf:
Currently, the `test_recursion_circuit` ZKR is included in the allowed list even though it is only for us in tests. The allowed set of ZKRs should only include operations that are meant to be applied to succinct and segment receipts. This PR adds an explicit list of the ZKRs to be included in the allowed set. This means that the test ZKR(s), and accelerators in the future, can be included in the zip, without being automatically included in the allowed set.

### DIFF
--- a/compact_proof/README.md
+++ b/compact_proof/README.md
@@ -12,6 +12,7 @@ which is suitable for publishing on-chain.
 ## Quickstart
 
 To install the necessary dependencies, run:
+
 ```bash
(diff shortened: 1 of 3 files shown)


flaub (approved):

flaub:
For reasons I don't yet understand, the `stark2snark` test fails with these changes.

PR #1615, opened 2024-03-29
Drop unneeded Poseidon254 dependencies

tzerrell:
This PR removes an unused import of `Poseidon254HashSuite` in GPU recursion host code (we deleted its last use); it was causing a build failure when building this with `-F cuda` (or `-F metal`, presumably).


SchmErik:
> This PR removes an unused import of `Poseidon254HashSuite` in GPU recursion host code (we deleted its last use); it was causing a build failure when building this with `-F cuda` (or `-F metal`, presumably).

@tzerrell looks like you're removing an unused export? Also, could you tell me which build is failing? more specifically which command is exhibiting this failure?

tzerrell:
> > This PR removes an unused import of `Poseidon254HashSuite` in GPU recursion host code (we deleted its last use); it was causing a build failure when building this with `-F cuda` (or `-F metal`, presumably).
> 
> @tzerrell looks like you're removing an unused export? Also, could you tell me which build is failing? more specifically which command is exhibiting this failure?

This is a _very_ good question -- I'm going to have to try to remember when in the last month I ran a build with `-F cuda`, because I definitely don't remember doing that and once I do it'll remind me of what I was doing when I made this PR. I'm going to have to let this sit for a bit until my memory kicks back in.

flaub:
It's odd: CI should catch if there's a build issue. I build with CUDA quite a bit and haven't run into it, but perhaps you're trying to build a specific crate (which sometimes doesn't work in isolation).

flaub (approved):
I guess if CI is happy with this, we can drop the extra uses. We can always bring them back if needed later.

tzerrell:
> I guess if CI is happy with this, we can drop the extra uses. We can always bring them back if needed later.

Yeah. I'll comment again here if I remember the where I encountered this.

PR #1620, opened 2024-04-02
Fix send_recv_slice when U != u8

nategraf:
`send_recv_slice`, which is used to allow the host to define "custom syscalls" has a bug which means that it only works when the generic type `U` is `u8`. This is because the host returns to the guest a length for how many _bytes_ it will send to the guest on the next call. The guest code currently interprets this number of bytes to be a number of elements of type `U`. So when `mem::size_of::<U> != 1`, the length will be incorrect and the second read will fail.


flaub:
@hashcashier did this fix your issue?

flaub:
Is there a test we can write which shows the issue?

SchmErik:
> Is there a test we can write which shows the issue?

friendly ping! @hashcashier could you answer this question?

SchmErik:
superseded by: #1874

PR #1656, opened 2024-04-10
DRAFT: Implement syscall for receipt verification

preston-evans98:
# Description
This commit implements a new syscall allowing the guest to verify a specific receipt, rather than simply verifying that *some* receipt is known which satisfies the claim.

This commit implements the requisite syscall but omits the crucial step of recursively invoking the Rust verifier when the host claims that the provided receipt is invalid. This omission is intentional, since the Risc0 verifier currently does not run on Risc0.

### DIFF
--- a/risc0/zkvm/platform/src/syscall.rs
+++ b/risc0/zkvm/platform/src/syscall.rs
@@ -137,6 +137,7 @@ pub mod nr {
     declare_syscall!(pub SYS_WRITE);
     declare_syscall!(pub SYS_VERIFY);
     declare_syscall!(pub SYS_VERIFY_INTEGRITY);
+    declare_syscall!(pub SYS_VERIFY_RECEIPT);
 }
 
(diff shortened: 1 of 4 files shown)


SchmErik:
@preston-evans98 the code to perform receipt verification for composite, succinct, and compact receipts has landed in main just now from this PR: https://github.com/risc0/risc0/pull/1776. I've added a test that you can use as a starting place:
Host Code:
https://github.com/risc0/risc0/blob/53ada25b03fd22c9bb9c13065bb5022c10574bb6/risc0/zkvm/src/host/server/prove/tests.rs#L602
Guest code:
https://github.com/risc0/risc0/blob/53ada25b03fd22c9bb9c13065bb5022c10574bb6/risc0/zkvm/methods/std/src/bin/verify.rs#L17
Seeing as this is now implemented, I'll close this PR. Feel free to reach out if you have any questions!

PR #1662, opened 2024-04-12
fix: create result bug

nahoc:


Cardosaum (approved):
Looks good to me, thanks for the bug fix

PR #1664, opened 2024-04-12
[security model doc] translate from cost estimate to time estimate

pdg744:
This doc updates the description for "How secure is the STARK protocol?" from a cost-based estimate to a time-based estimate. 

The primary motivation to update this is that the current estimate  is dependent on a few assumptions from 2020. Switching from cost to time removes the dependency on both bitcoin hashrate and price of bitcoin.

The draft here was written by Kevin; I'm hesitating on the specifics because it seems weird to be referencing `sha1`. Would Poseidon2 be the appropriate benchmark for this?


website/api/security-model.md
```diff
@@ -77,13 +77,15 @@ The RISC-V Prover and the Recursion Prover both use STARK-based protocols, which
...
 ### How secure are the STARK Provers?
 
-We estimate that it would cost over 400 billion USD of computational resources to construct a fake proof for one of our STARK circuits.
+We estimate that it would take approximately 400,000 years to brute force a collision against our STARK protocol using GPUs in a parallel attack, using 1 million high-end GPUs.
 
-These estimates are based on the approach used in [this article by Justin Thaler].
-The napkin math here is as follows:
+- Currently, we look to the RTX 4090 as a high-end reference GPU.
+- RTX 4090 [benchmarks] provide a reference SHA2-256 hashrate at less than 25,000 million hashes per second.
+- We have seen several [examples] of large purchases of GPUs on the order of 350,000 so we conservatively suggest that a distributed attacker may be able to operate 1 million RTX 4090s in parallel.
```
tzerrell, line 84: Hmm, slightly awkward to cite the 4090 and then link to a purchase of H100s. People often think of H100s as higher end than 4090s (certainly H100s are way more expensive), so this might look like we are unfairly weakening our case -- but actually the H100 benchmarks only just over 12,000 megahashes/second for SHA2-256 (https://gist.github.com/Chick3nman/e1417339accfbb0b040bcd0a0a9c6d54).

That said, 4090s are more recognizable to the not-immersed-in-GPU-stats-all-day audience, and I can't think of a concise way to explain this complexity, and this isn't really the core focus here. So my vote is: think about it for a couple minutes to see if you see any opportunities for revision I'm missing, and then probably just leave it as is.

tzerrell: Possible alternative metric: Nvidia is expecting to put out "up to 2 million" H100s this year, up from around 550k last year. So we could perhaps cite using "every H100 ever made or projected to be made by the end of 2024" instead.

tzerrell (approved):
Looks good, I suggested a possible adjustment to how we do this computation but also I'm not sure whether it's an improvement.

website/api/security-model.md
pdg744: Chatted with Tim over DMs about this point. 
We decided to leave it as is for now; we can always re-visit this later.

PR #1665, opened 2024-04-12
fix: remove redundant zeroing of memory allocated by bump alloc

austinabell:
TLDR memory is zeroed by default, so no need to manually zero the memory allocated https://doc.rust-lang.org/src/core/alloc/global.rs.html#201.

Ideally there would be a way to also test that this doesn't require paging in the memory allocated until it's actually written, but that seems like a hard variant to test. Can revisit if we want to be sure of this.


flaub (approved):

flaub:
I recently added a test to `MultiTest` and now the `AllocZeroed` test case is failing. Leading me to think that zeroing of memory isn't actually gauranteed. I'll try to be more conclusive about this, but we may need to revert this.

flaub:
Follow up, looks like this was actually related to using `inout` vs `inlateout` in our assembly instructions when invoking a syscall. When using `inout` there were situations where the compiler would emit a `lui` instruction (which loads the upper bits of an immediate value into a register) but would sometimes elide the corresponding `add` to ensure the lower 12-bits are also loaded into the register. It's still not clear to me why `inlateout` fixed it or what the issue with `inout` is.

PR #1661, opened 2024-04-12
Updating calculator

pdg744:
This PR: 
- updates the calculator to report conjectured security based on the ethSTARK toy problem analysis, rather than the conjectured soundness analysis based on the conjectures from ProxGaps/DEEP-FRI. 
- updates the arithmetic for computing the plonk_plookup_error
- adds plonk_plookup_error to the `toy_model_security()` function
- clarifies the definition of the parameter which was previously called `d`

- [x] Note how this test uses DEFAULT_SEGMENT_LIMIT_PO2 where the tests next to it in the same file use 1 << 20; I [Tim] kinda like the version in this test, so I left it, but I think we should make all 3 consistent.

PR #1666, opened 2024-04-12
WIP pass parameters to security calculation

tzerrell:
A suggested solution for #1661

Note that this doesn't actually pass (it doesn't quite hit 100 bits of security) -- I think that's correct and the test should be adjusted, but you would know better than me.

Also note how this test uses `DEFAULT_SEGMENT_LIMIT_PO2` where the tests next to it in the same file use `1 << 20`; I kinda like the version in this test, so I left it, but I think we should make all 3 consistent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5542,6 +5542,7 @@ dependencies = [
  "rand",
  "rand_core",
  "rayon",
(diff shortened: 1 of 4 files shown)


pdg744 (approved):
Thank you!

PR #1675, opened 2024-04-15
Require Stable Circom

tzerrell:
Makes CI test that the hashes of the two circom files secured by a trusted setup ceremony are unchanged from what we used for the ceremony (i.e., that these circom files are the files secured by the ceremony).

Also adds a test that the Control Root is unchanged.


.github/scripts/circom_stable.sh
```diff
@@ -0,0 +1,13 @@
...
+then echo "Unexpected SHA-256 of stark_verify.circom: $STARK_VERIFY_HASH"; \
+exit 1; \
+fi
+
+RISC0_LIB_HASH=$(shasum -a 256 compact_proof/groth16/risc0.circom | awk '{ print $1 }')
+RISC0_LIB_EXPECTED_HASH="38d3937558dae8f9de3d0bdc534c76c9adc6c7ffe37cc43935073324e1e53bf3"
+if [ $RISC0_LIB_HASH != $RISC0_LIB_EXPECTED_HASH ]; \
+then echo "Unexpected SHA-256 of risc0.circom: $RISC0_LIB_HASH"; \
+exit 1; \
+fi
```
flaub, line 13:
```suggestion
STARK_VERIFY_HASH=$(shasum -a 256 compact_proof/groth16/stark_verify.circom | awk '{ print $1 }')
STARK_VERIFY_EXPECTED_HASH="a3789471909ba1a13cca783dc5269b25b6d41295abe60234a8075f750017518c"
if [ $STARK_VERIFY_HASH != $STARK_VERIFY_EXPECTED_HASH ]; then
    echo "Unexpected SHA-256 of stark_verify.circom: $STARK_VERIFY_HASH"
    exit 1
fi

RISC0_LIB_HASH=$(shasum -a 256 compact_proof/groth16/risc0.circom | awk '{ print $1 }')
RISC0_LIB_EXPECTED_HASH="38d3937558dae8f9de3d0bdc534c76c9adc6c7ffe37cc43935073324e1e53bf3"
if [ $RISC0_LIB_HASH != $RISC0_LIB_EXPECTED_HASH ]; then
    echo "Unexpected SHA-256 of risc0.circom: $RISC0_LIB_HASH"
    exit 1
fi
```

.github/workflows/main.yml
```diff
@@ -416,6 +421,20 @@ jobs:
...
+    if: needs.changes.outputs.stable-ceremony == 'true'
+    needs: changes
+    runs-on: [self-hosted, prod, Linux, cpu]
+    steps:
+      # This is a workaround from: https://github.com/actions/checkout/issues/590#issuecomment-970586842
+      - run: "git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :"
+      - uses: actions/checkout@v4
+      - name: Compare circom hashes to golden values
+        run: bash ./circom_stable.sh
+        working-directory: ./.github/scripts
```
flaub, line 435: I recommend we move this to the `check` job, which always runs (doesn't depend on the `changes` job)

flaub (changes requested):
This is currently failing silently: https://github.com/risc0/risc0/actions/runs/8697583918/job/23853086335?pr=1675#step:4:8

.github/workflows/main.yml
```diff
@@ -155,6 +160,12 @@ jobs:
       - name: check for "D0 NOT MERGE" comments
         run: |
           [ "$(grep -re 'DO[_ ]\?NOT[_ ]\?MERGE' $(git ls-tree --full-tree --name-only -r HEAD) | tee /dev/fd/2 | wc -l)" -eq "0" ]
+      - name: Load circom check script
+      # This is a workaround from: https://github.com/actions/checkout/issues/590#issuecomment-970586842
+        run: "git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :"
```
flaub, line 165: This can be dropped, we already checkout the repo at the top of this job

.github/workflows/main.yml
```diff
@@ -416,6 +427,20 @@ jobs:
...
+    if: needs.changes.outputs.stable-ceremony == 'true'
+    needs: changes
+    runs-on: [self-hosted, prod, Linux, cpu]
+    steps:
+      # This is a workaround from: https://github.com/actions/checkout/issues/590#issuecomment-970586842
+      - run: "git checkout -f $(git -c user.name=x -c user.email=x@x commit-tree $(git hash-object -t tree /dev/null) < /dev/null) || :"
+      - uses: actions/checkout@v4
+      - name: Compare circom hashes to golden values
+        run: bash ./circom_stable.sh
+        working-directory: ./.github/scripts
```
flaub, line 441:
```suggestion
      - name: Compare circom hashes to golden values
        run: bash .github/scripts/circom_stable.sh
```

tzerrell:
The test is now failing for a substantive reason: CI is hashing the LFS pointer to `stark_verify.circom` instead of the contents of that file. (I have manually confirmed that the hash value the test is reporting in CI matches the hash I get from hashing `git cat-file -p :compact_proof/groth16/stark_verify.circom`. I.e. it really is hashing the pointer.)

I see two possible approaches:
* Download the LFS file in CI
* Accept both the hash of the file and the hash of the pointer to the file in this new test.

@morgangibbs87 are you someone who knows about the costs/practicality of pulling down an LFS file in CI? It's roughly 56MB.

morgangibbs87:
@tzerrell -  In terms of cost, this is from [Amazon's EC2 Pricing Documentation](https://aws.amazon.com/ec2/pricing/on-demand/):

> Pricing is **_per instance-hour consumed for each instance_**, from the time an instance is launched until it is terminated or stopped. Each partial instance-hour consumed will be billed per-second for Linux, Windows, Windows with SQL Enterprise, Windows with SQL Standard, and Windows with SQL Web Instances, and as a full hour for all other instance types.

Thus, downloading a ~56MB LFS file in CI should only add a few seconds of uptime to our runners and should not affect our costs significantly. Most runners will be up anyway until they are idle for some time and spin down.

tzerrell:
Thank you; probably the bigger issue then is whether I have to also pull down other, bigger LFS files or can just grab the one I need. I'm investigating that now.

tzerrell:
Commit [523a486](https://github.com/risc0/risc0/pull/1675/commits/523a486db66da39c680fb3352f87b051ad5e6878) (which added `lfs: true`) ran `check` in 2m 41s; previous run was 2m 42s. I think this is the way to go.

xtask/src/templates/control_id_zkr.rs
```diff
@@ -18,3 +18,11 @@ pub const ALLOWED_IDS_ROOT: &str = "{}";
...
 
 pub const BN254_CONTROL_ID: &str = "{}";
+
+#[test]
+fn stable_root() {{
+    // This tests that none of the control IDs have changed unexpectedly
+    // If you have _intentionally_ changed control IDs, update this hash and bootstrap.
+
+    assert_eq!(ALLOWED_IDS_ROOT, "88c1f749250aba181168c33839d7a351671e7a5b7f3e746dde91ef6c6e9ef344");
+}}
```
flaub, line 28: I don't think a test should go into a template. Could we move the test into a place that isn't generated by an xtask?

flaub:
Pulling down LFS is totally fine, but we could probably just check the git object hash: https://git-scm.com/docs/git-hash-object

xtask/src/templates/control_id_zkr.rs
tzerrell: That makes sense, I moved it to `risc0/zkvm/src/host/recursion/tests.rs`

tzerrell:
> Pulling down LFS is totally fine, but we could probably just check the git object hash: https://git-scm.com/docs/git-hash-object

This seems to be working as you expected, thank you!

flaub (approved):

PR #1681, opened 2024-04-17
Return ProveInfo from prover

SchmErik:
This PR changes `prover.prove()` to return a new struct called `ProveInfo` that contains the receipt, number of segments, total cycles, and total user cycles. This is helpful because it allows users to access more information about the proof generation through the prover API.


SchmErik:
looks like I have merge conflicts, I'm going to convert as draft and mark it as ready after I resolve them tomorrow

mothran (commented):
YAY! Just wanted to say this looks like exactly what I have needed in the past, thanks!

flaub (approved):
The CHANGELOG needs to be updated with this breaking change, but otherwise looks good!

SchmErik:
> The CHANGELOG needs to be updated with this breaking change, but otherwise looks good!

Thanks for the heads up, I really appreciate it! I'll merge after updating the change log this afternoon

PR #1686, opened 2024-04-18
risc0-build: allow overriding the Cpp toolchain

SchmErik:
The previous commit adding the Cpp toolchain disabled overriding the toolchain from the command line via the `CC` and
`CFLAGS_riscv32im_risc0_zkvm_elf` environment variable. This change allows developers to override their settings by supplying the default values only if the user does not specify these environment variables as a part of their `cargo` command.


SchmErik:
@austinabell could you test this? I was able to get it to invoke the clang compiler on my local testing but it seems like my clang configuration somehow got screwed up...

austinabell:
> @austinabell could you test this? I was able to get it to invoke the clang compiler on my local testing but it seems like my clang configuration somehow got screwed up... The cc override should happen using the `CC_riscv32im_risc0_zkvm_elf` env var

So I'm still getting the following errors:
```
error: invalid inline table
expected `}`
 --> ../../../../.cargo/git/checkouts/risc0-7d6ab9951628e5b1/084c523/templates/rust-starter/methods/Cargo.toml:7:17
  |
7 | risc0-build = { {{ risc0_build }} }
  |                 ^
  |
error: invalid inline table
expected `}`
 --> ../../../../.cargo/git/checkouts/risc0-7d6ab9951628e5b1/084c523/templates/rust-starter/methods/guest/Cargo.toml:2:9
  |
2 | name = {{ guest_package_name }}
  |         ^
  |
error: invalid inline table
expected `}`
 --> ../../../../.cargo/git/checkouts/risc0-7d6ab9951628e5b1/084c523/templates/rust-starter/host/Cargo.toml:8:16
  |
8 | risc0-zkvm = { {{ risc0_zkvm }} }
  |                ^
  |

```

but not the other errors I was previously getting, so I think this is better. I have no idea if that other part is related, as it happened sometime inbetween 0.21 and what's on main. (switching back to 0.21 doesn't have this issue)

SchmErik:
@flaub this is ready to go. It addresses an issue that Austin was seeing which is that my recent cpp change disabled the old way of overriding the CC_riscv32im_risc0_zkvm_elf flag. This adds the logic to check before overriding the CC env var. I'm not sure what's going on with the other error in the template that austin is seeing but that's likely some other issue...

flaub (approved):

PR #1688, opened 2024-04-19
Prover: Switch receipt kind based on `ProverOpts`

SchmErik:
This change adds a way to switch receipt formats using the `ProverOpts`. There's a new field called `ReceiptFormat` that is used to specify the desired receipt format generated from the prove function. Fake receipts were not included as a part of this because `DevMode` has a separate `ProverServer` implementation. This implements a way to configure `Prover::prove` to return either composite or succinct receipts. In the near future, an implementation for `CompactReceipts` will be implemented for x86-linux as well as a way to configure the receipt types for `Prover::Compress`.

Closes: #1689


SchmErik:
forgot about r0vm and client/server. Putting in draft and I'll mark as ready once I get that done.

flaub:
I'm not sure that `ReceiptFormat` is the right name for this.

* `ReceiptStage`?
* `ProofStage`?
* `ReceiptKind`?

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -209,6 +210,7 @@ impl From<ProverOpts> for pb::api::ProverOpts {
         Self {
             hashfn: opts.hashfn,
             prove_guest_errors: opts.prove_guest_errors,
+            receipt_format: opts.receipt_format as i32,
```
flaub, line 213:
```suggestion
            receipt_format: opts.receipt_format as u32,
```

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -120,6 +121,32 @@ pub struct ProverOpts {
...
+/// This enum represents the various receipt formats that can be generated.
+#[derive(Clone, Serialize, Deserialize, PartialEq, ValueEnum)]
+pub enum ReceiptFormat {
+    /// Composite Receipts resulting from proving a session
+    Composite = 0,
+    /// Succinct Receipts resulting from applying the lift and join programgs so Composite receipts
+    Succinct = 1,
+    /// Compact Receipts resulting from stark2snark on the Succinct receipts
+    Compact = 2,
+}
```
flaub, line 137:
```suggestion
pub enum ReceiptFormat {
    /// Composite Receipts resulting from proving a session
    Composite,
    /// Succinct Receipts resulting from applying the lift and join programgs so Composite receipts
    Succinct,
    /// Compact Receipts resulting from stark2snark on the Succinct receipts
    Compact,
}
```

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -120,6 +121,32 @@ pub struct ProverOpts {
...
+
+    fn try_from(value: i32) -> Result<Self> {
+        match value {
+            0 => Ok(ReceiptFormat::Composite),
+            1 => Ok(ReceiptFormat::Succinct),
+            2 => Ok(ReceiptFormat::Compact),
+            _ => bail!("Unknown receipt format number: {value}"),
+        }
+    }
 }
```
flaub, line 150: I think we should remove this and only have conversions for protobuf over in the `api/convert.rs`

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -139,8 +167,34 @@ impl ProverOpts {
...
+            prove_guest_errors: false,
+            receipt_format: ReceiptFormat::Compact,
+        }
+    }
+
+    /// Choose the prover that enables compact, snark receipts, only supported for x86_64 linux
+    #[cfg(not(all(target_arch = "x86_64", target_os = "linux")))]
+    pub fn compact() -> Self {
+        unimplemented!("compact receipts are only supported on x86_64 linux")
+    }
```
flaub, line 197: Not sure about this, I'm thinking it's OK to fail later when it tries to run the docker image.

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -61,6 +95,9 @@ pub trait ProverServer {
     /// Prove the specified [Segment].
     fn prove_segment(&self, ctx: &VerifierContext, segment: &Segment) -> Result<SegmentReceipt>;
 
+    /// Preturn the receipt format to be gerneated by proving.
```
flaub, line 98:
```suggestion
    /// Return the receipt format to be generated by proving.
```

SchmErik:
> I'm not sure that `ReceiptFormat` is the right name for this.
Yeah, I agree it's not the greatest name... I was thinking `ReceiptType` but that felt awkward to as a rust enum
> 
>     * `ReceiptStage`?
> 
>     * `ProofStage`?
> 
>     * `ReceiptKind`?

I like `ReceiptKind`. I'll switch it to that for now

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -52,7 +53,40 @@ pub trait ProverServer {
     ) -> Result<ProveInfo> {
         let mut exec = ExecutorImpl::from_elf(env, elf)?;
         let session = exec.run()?;
-        self.prove_session(ctx, &session)
+        let prove_info = self.prove_session(ctx, &session)?;
```
flaub, line 56: I would expect that if I call `prover.prove_session` that it would also follow the `ReceiptFormat` (aka `ReceiptKind`). So it seems like `prove_session` should take the `ProverOpts` as a parameter.

risc0/zkvm/src/host/client/prove/mod.rs
SchmErik: Another thought: how about we catch it during proof generation so if this ProverOpts gets passed to ProveSession, I can do a check that bails earlier all of the receipts are generated?

SchmErik: oh, the compact stuff will be in a later PR so I'll just drop the `unimplemented` for this PR

risc0/zkvm/src/host/api/convert.rs
SchmErik: So this one is weird - i tried u32 at first but the compiler complained. It seems like protobuf enums are expressed as signed integers...

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -120,6 +121,19 @@ pub struct ProverOpts {
     // moment if there is a better place. At some point before 1.0, this option should be moved or
     // dropped.
     pub prove_guest_errors: bool,
+    /// The kind of receipt to be generated.
+    pub receipt_format: ReceiptKind,
```
flaub, line 125:
```suggestion
    pub receipt_kind: ReceiptKind,
```

SchmErik: thanks for catching this. I did a search/replace locally

risc0/zkvm/Cargo.toml
```diff
@@ -63,6 +63,7 @@ addr2line = { version = "0.21", features = ["rustc-demangle"], optional = true }
 bincode = { version = "1.3", optional = true }
 bonsai-sdk = { workspace = true, optional = true }
 bytes = { version = "1.4", features = ["serde"], optional = true }
+clap = { version = "4.4", features = ["derive"] }
```
flaub, line 66: Is this needed?

SchmErik: This is needed for r0vm because r0vm needs this (and other fields of prover Opts) as a command line argument. I could create a separate struct within r0vm that's meant to capture this information and translate that to `ReceiptKind` which would allow us to drop this dependency. Would that be preferred?

flaub: Why wouldn't this go on the `risc0-r0vm` crate? I don't see why `risc0-zkvm` needs it.

SchmErik: It was for derive `ValueEnum` that I had placed. I pushed a change that removes this derive, drops clap and created a new struct in r0vm. i feel like it's probably a better approach to keep the clap stuff away from risc0-zkvm

risc0/r0vm/src/lib.rs
```diff
@@ -63,6 +63,10 @@ struct Cli {
     /// to read it.
     #[arg(long, env = "RISC0_PPROF_OUT")]
     pprof_out: Option<PathBuf>,
+
+    /// The receipt format produced by the r0vm prover
```
flaub, line 67:
```suggestion
    /// The receipt kind produced by the r0vm prover
```

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -61,6 +62,9 @@ pub trait ProverServer {
     /// Prove the specified [Segment].
     fn prove_segment(&self, ctx: &VerifierContext, segment: &Segment) -> Result<SegmentReceipt>;
 
+    /// Return the receipt format to be generated by proving.
+    fn get_receipt_kind(&self) -> ReceiptKind;
```
flaub, line 66: It feels wrong for this to be defined on the trait. Can we drop it?

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -120,6 +120,19 @@ pub struct ProverOpts {
     // moment if there is a better place. At some point before 1.0, this option should be moved or
     // dropped.
     pub prove_guest_errors: bool,
+    /// The kind of receipt to be generated.
+    pub receipt_kind: ReceiptKind,
+}
+
+/// This enum represents the various receipt formats that can be generated.
```
flaub, line 127:
```suggestion
/// This enum represents the various receipt kinds that can be generated.
```

flaub (approved):
One more format -> kind, but LGTM

PR #1690, opened 2024-04-20
rv32im circuit updates

flaub:
* Replace `NVCC_PREPEND_FLAGS` and `NVCC_APPEND_FLAGS` with `RISC0_NVCC_FLAGS`. If `RISC0_NVCC_FLAGS` is not set, then `-arch=native` is used as a default.
* Add `sys_input` which is a syscall that guests can use to access the input digest that is optionally specified for each proof request.
* Adjust rv32im circuit to avoid need to do PC adjustments for claims.
* The post state digest is now always `Digest::ZERO` when the exit code is `Halted`.
* Add warmup phase to benchmarks to account for kernel JIT compilation


risc0/build_kernel/src/lib.rs
```diff
@@ -136,8 +136,7 @@ impl KernelBuild {
 
     fn compile_cuda(&mut self, output: &str) {
         println!("cargo:rerun-if-env-changed=RISC0_CUDA_OPT");
-        println!("cargo:rerun-if-env-changed=NVCC_PREPEND_FLAGS");
-        println!("cargo:rerun-if-env-changed=NVCC_APPEND_FLAGS");
+        println!("cargo:rerun-if-env-changed=RISC0_NVCC_FLAGS");
```
flaub, line 139: @mothran @rlukata Note: this was changed so that by default, `-arch=native` is used. In order to build for a specific architecture, use `RISC0_NVCC_FLAGS` instead of `NVCC_PREPEND_FLAGS` or `NVCC_APPEND_FLAGS`.

jbruestle (approved):

tzerrell (approved):

risc0/zkvm/src/guest/env.rs
```diff
@@ -752,3 +752,17 @@ impl<F: Fn(&[u8])> std::io::Write for FdWriter<F> {
         Ok(())
     }
 }
+
+/// Read the input digest from the input commitment.
+pub fn input() -> Digest {
```
nategraf, line 757: In order to support a future version of this API that returns the preimage, we might want to call this `input_digest`. (If we don't rename this more generally)

risc0/zkvm/src/host/receipt.rs
```diff
@@ -464,22 +463,14 @@ impl CompositeReceipt {
...
                     .post
                     .as_value()
                     .map_err(|_| VerificationError::ReceiptFormatError)?
-                    .clone();
-                post.pc = post
-                    .pc
-                    .checked_sub(WORD_SIZE as u32)
-                    .ok_or(VerificationError::ReceiptFormatError)?;
-                post.digest()
-            });
```
nategraf, line 482: Love to see it 🎉

nategraf (approved):
Looks good to me. Some comments provided as suggestions, or considerations.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -304,6 +304,27 @@ pub unsafe extern "C" fn sys_pause(user_exit: u8, out_state: *const [u32; DIGEST
     );
 }
 
+#[cfg_attr(feature = "export-syscalls", no_mangle)]
+pub extern "C" fn sys_input(index: u32) -> u32 {
+    let t0 = ecall::INPUT;
+    let index = index & 0x07;
```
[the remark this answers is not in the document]
flaub: Panic in this codepath would cause a lot of bloat. Ideally this function could be inlined. Generally I don't like the idea of the low-level `sys-` functions doing a panic. We should maybe have them return error codes (I'd say `Result` but it'd be nice for these functions to be callable from C). I also don't know how C will handle panics, so another reason to avoid panic from these functions.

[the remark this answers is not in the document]
nategraf: Ok, that's helpful context

PR #1692, opened 2024-04-20
Update zkvm-specification.md

richlukas:


flaub (changes requested):
These changes are incorrect.

PR #1696, opened 2024-04-22
Fix broken image path

AdventureSeeker987:
image


flaub:
This page is designed to be hosted by docusaurus, and can be seen here: https://dev.risc0.com/api/zkvm/receipts

PR #1703, opened 2024-04-22
updating link to calculator

pdg744:
Since the calculator is not included in a stable release yet, the links in the security model doc are pointing to the most recent PR. 
Note that this PR removes a pointer to `main` that apparently slipped through CI at some point in the past week or two.


pdg744:
I'm a bit unsure of our current best practice on how to manage these sorts of links -- it seems important to be able to link to things that aren't yet included in a stable release. Linking to the PR is an OK solution but seems awkward.

flaub:
Next points to things that aren't yet released. You can have a 'next' version of the security model doc that points to things in `main`. The `release-0.21` doc should only point to things in the `release-0.21` branch.

flaub (approved):

PR #1709, opened 2024-04-23
more conservative numbers

pdg744:
This PR slightly revises the numerical estimate of soundness for our STARK-to-SNARK translator.

This change was prompted by a comment from Daniel Lubarov pointing out the fact that [https://eprint.iacr.org/2017/334] claims BN254 can be attacked with 99.69 bits of work. 

I think that estimate is unreasonably pessimistic, in light of [Ariel's observation here](https://github.com/zcash/zcash/issues/714#issuecomment-295813446), but given that it's not the bounding term in our proof system, it seems simplest to just revise to use the more conservative number.


tzerrell (approved):
Seems like a good approach to me.

Issue #1329, opened 2024-01-15
[BUG] Do not import `risc0-zkvm-platform/rust-runtime` for the host


weikengchen:
## Bug Report

We are looking to see if one can deploy RISC Zero prover as a [Ray remote function](https://docs.ray.io/en/latest/ray-core/tips-for-first-time.html) so that one can run distributed proof generation directly over io.net.

All it takes is to add a Python interface for RISC Zero prover, through https://github.com/PyO3/pyo3, by compiling RISC Zero into a Python module. An obstacle is that, when, in Python, we want to load this module, the system would complain that:

```
>>> import l2_r0prover
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/cusgadmin/.pyenv/versions/pyo3/lib/python3.11/site-packages/l2_r0prover/__init__.py", line 1, in <module>
    from .l2_r0prover import *
ImportError: dlopen(/Users/cusgadmin/.pyenv/versions/pyo3/lib/python3.11/site-packages/l2_r0prover/l2_r0prover.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '__end'
```

People who are familiar with RISC Zero can immediately think about what this `__end` is. It is from the allocator.
```rust
#[cfg(feature = "export-syscalls")]
#[no_mangle]
pub unsafe extern "C" fn sys_alloc_aligned(bytes: usize, align: usize) -> *mut u8 {
    extern "C" {
        // This symbol is defined by the loader and marks the end
        // of all elf sections, so this is where we start our
        // heap.
        //
        // This is generated automatically by the linker; see
        // https://lld.llvm.org/ELF/linker_script.html#sections-command
        static _end: u8;
    }
    ......
}
```
https://github.com/risc0/risc0/blob/main/risc0/zkvm/platform/src/syscall.rs#L657

The problem is that `__end` is not always available, and additionally, it is not really necessary for the host. Today, however, `Cargo.toml` of zkvm always import risc0-zkvm-platform and has rust-runtime feature on.
```toml
risc0-zkvm-platform = { workspace = true, features = [
  "rust-runtime",
  "export-getrandom",
] }
```
https://github.com/risc0/risc0/blob/main/risc0/zkvm/Cargo.toml#L35C1-L47C4

## Solution

It seems necessary to make sure that `rust-runtime` and `export-getrandom` are only used when it is in zkvm. 

A proposed change is to replace the toml entry above to:
```toml
[...]

PR #1715, opened 2024-04-24
zkVM: fix build issues arising from undefined `_end`

SchmErik:
When building zkvm as a host-side staticlib to link with platforms such as go, the build can fail due to an undefined `_end` symbol. Hide this when building for the host.

Fixes: #1329


risc0/zkvm/src/lib.rs
```diff
@@ -71,6 +71,7 @@
 
 extern crate alloc;
 
+#[cfg(target_os = "zkvm")]
```
flaub, line 74: I think this breaks the ability to use rust-analyzer

SchmErik: Thinking about this some more, the original bug report and issues we've seen internally point to the undefined `_end` symbol used in `sys_alloc_aligned`. I'm thinking that it's probably better to revert all of these changes and put cfg's around where _end appears... Let me know if you have thoughts on this. I'm going to make this change and change the title/description of this PR.

Wollac:
It seems strange to me that the `_end` is only hidden on non-zkvm targets. This means that `sys_alloc_aligned` is still compiled and exported, but is probably incorrect because parts of it are changed.

I actually preferred the previous approach where no syscalls were exported or even compiled on the host as they are not needed here. Can we maybe keep this and just make sure that the `guest` mod is not behind a target flag? 
If not, I think it would still be better to disable the entire `sys_alloc_aligned` function on non-zkvm targets (but nothing else). Rather than have it compile a somehow modified version.

flaub:
It's nice to have the IDE do analysis and references. Thus the guest code is still meant to be able to be compiled on a target that it won't run on.

flaub (approved):

PR #1718, opened 2024-04-25
zkvm: use `u64` to keep track of cycles instead of `u32` or `usize`

SchmErik:
closes: #1673


SchmErik:
not ready yet. closing

PR #1720, opened 2024-04-25
Check control root during succinct receipt verification

nategraf:
Prior to this PR, the control ID of the succinct receipt is checked during verification, but the control root encoded in the output is not. Both must be checked as the control ID on the receipt binds the most receipt program that was run, and the control root in the output binds the closure of recursion programs that were run in all previous steps of recursion. Without the latter, a prover can run an out-of-set program, then pass it through an in-set program, such as identity, without detection.


risc0/zkvm/src/host/recursion/receipt.rs
```diff
@@ -97,8 +101,22 @@ impl SuccinctReceipt {
             seal_claim.push_back(elem.as_u32())
         }
 
-        // TODO: Read root hash
-        seal_claim.drain(0..16);
+        // NOTE: read_sha_halfs is a misnomer here, because the digest is Poseidon2, but we don't
+        // attempt to do anything with the hash that would make this a problem.
+        let control_root =
+            read_sha_halfs(&mut seal_claim).map_err(|_| VerificationError::ReceiptFormatError)?;
```
jbruestle, line 107: Is the Poseidon2 hash used actually encoded in the globals as halfs?  I thought that Poseidon hashes get written to the globals as 16 elements with every other one being zero or something weird like that.

jbruestle (commented):
100% we need to check this, that said, I think the specific are not quite right, also tests clearly don't like it.

risc0/zkvm/src/host/recursion/receipt.rs
```diff
@@ -97,8 +101,22 @@ impl SuccinctReceipt {
...
+        let allowed_root = Digest::from_hex(ALLOWED_IDS_ROOT)
+            .map_err(|_| VerificationError::ImplementationError)?;
```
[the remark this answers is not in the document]
nategraf: Changed to unwrap in https://github.com/risc0/risc0/pull/1720/commits/b125ac8d931aa6e68cf4e0ba9ac99a1f84e0e432

risc0/zkvm/src/host/recursion/receipt.rs
nategraf: @jbruestle, I've fixed this in https://github.com/risc0/risc0/pull/1720/commits/7ccba07c7063208d9ff35736df5a871c06430f92

I thought about creating a new "decode poseidon" function, but I'm not sure it would actually be used anywhere else, and it's kind of an odd encoding. I decided to forgo the function, include the logic inline, and comment on what it is doing.

flaub (approved):

jbruestle (approved):

PR #1721, opened 2024-04-25
allow configuring rust toolchain branch

austinabell:
Not sure if this is correct, just opening for visibility to help unblock someone needing to build the toolchain given the updated Rust version (1.70 too low).

~~Also changes the clone to minimize data fetched.~~ Removed this for simplicity, it was causing weird edge case issues

Ignore this PR, unless you also want this.

Tested with:
```
cargo install --force --path risc0/cargo-risczero
cargo risczero build-toolchain --git-tag v2024-04-22.0
```

### DIFF
--- a/risc0/cargo-risczero/src/commands/build_toolchain.rs
+++ b/risc0/cargo-risczero/src/commands/build_toolchain.rs
@@ -85,7 +85,7 @@ impl BuildToolchain {
 
         if !path.join(".git").is_dir() {
             Command::new("git")
-                .args(["clone", "--branch", tag, source])
+                .args(["clone", source])
                 .arg(path)
                 .run_verbose()?;
         } else {
--- a/risc0/cargo-risczero/src/commands/build_toolchain.rs
+++ b/risc0/cargo-risczero/src/commands/build_toolchain.rs
@@ -31,7 +31,10 @@ const CONFIG_TOML: &'static str = include_str!("config.toml");
 
 /// `cargo risczero build-toolchain`
 #[derive(Parser)]
-pub struct BuildToolchain {}
+pub struct BuildToolchain {
+    #[arg(long)]
+    git_tag: Option<String>,
+}
 
 /// Output info of a successful rust toolchain build.
 pub struct RustBuildOutput {
@@ -54,7 +57,8 @@ impl BuildToolchain {
 
         let is_ci = std::env::var("CI").is_ok();
         if !is_ci {
-            self.prepare_git_repo(ToolchainRepo::Rust.url(), RUST_BRANCH, &rust_dir)?;
+            let git_tag = self.git_tag.as_deref().unwrap_or(RUST_BRANCH);
+            self.prepare_git_repo(ToolchainRepo::Rust.url(), git_tag, &rust_dir)?;
         }
 
         let out = self.build_toolchain(&rust_dir)?;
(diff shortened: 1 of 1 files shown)


risc0/cargo-risczero/src/commands/build_toolchain.rs
```diff
@@ -31,7 +31,10 @@ const CONFIG_TOML: &'static str = include_str!("config.toml");
 
 /// `cargo risczero build-toolchain`
 #[derive(Parser)]
-pub struct BuildToolchain {}
+pub struct BuildToolchain {
+    #[arg(long)]
+    git_tag: Option<String>,
```
flaub, line 36:
```suggestion
    #[arg(long, default_value = RUST_BRANCH)]
    git_ref: String,
```

risc0/cargo-risczero/src/commands/build_toolchain.rs
```diff
@@ -54,7 +57,8 @@ impl BuildToolchain {
 
         let is_ci = std::env::var("CI").is_ok();
         if !is_ci {
-            self.prepare_git_repo(ToolchainRepo::Rust.url(), RUST_BRANCH, &rust_dir)?;
+            let git_tag = self.git_tag.as_deref().unwrap_or(RUST_BRANCH);
+            self.prepare_git_repo(ToolchainRepo::Rust.url(), git_tag, &rust_dir)?;
```
flaub, line 61:
```suggestion
            self.prepare_git_repo(ToolchainRepo::Rust.url(), self.git_ref, &rust_dir)?;
```

flaub:
I think we should try to use a git ref rather than a branch name so that it's more generic. Some adjustments might need to be made to allow for this.

risc0/cargo-risczero/src/commands/build_toolchain.rs
```diff
@@ -82,9 +86,14 @@ impl BuildToolchain {
...
             Command::new("git")
-                .args(["clone", source])
+                .args(["clone", "--branch", tag, source])
                 .arg(path)
                 .run_verbose()?;
+        } else {
+            Command::new("git")
+                .args(["fetch", "origin", tag])
+                .current_dir(path)
+                .run_verbose()?;
```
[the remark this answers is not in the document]
austinabell: Needed because if the branch/tag is created after you clone, it won't resolve? Maybe I'm overlooking something, though

austinabell:
> I think we should try to use a git ref rather than a branch name so that it's more generic. Some adjustments might need to be made to allow for this.

This does work with tags and refs, just tested refs also to double check, that's why I had the naming as tags (seemed most generic).

flaub:
My understanding is that `git clone --branch` only takes branches and tags, but not a commit hash: https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---branchltnamegt

> Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.

So my thinking is to drop the `--branch` on clone and let the following `checkout` command do the switch.

austinabell:
> My understanding is that `git clone --branch` only takes branches and tags, but not a commit hash: [git-scm.com/docs/git-clone#Documentation/git-clone.txt---branchltnamegt](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---branchltnamegt)
> 
> > Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to  branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.
> 
> So my thinking is to drop the `--branch` on clone and let the following `checkout` command do the switch.

You're right! I missed testing that case without a cloned branch. That was more of an artifact from trying to pull only what was needed.

I'll run the cases in the background and make that change now, good catch

SchmErik:
Implemented in #1851. closing...

PR #1725, opened 2024-04-25
adding mention of TSC to security model doc

pdg744:
Daniel suggested that the assumption for our STARK-to-SNARK prover could be improved. Specifically, he suggested that: 
1. The two existing assumptions are actually equivalent. 
2. We should add an assumption about the integrity of the trusted setup ceremony.

This PR adds mention of the TSC to address point (2). 
As to point (1) here, I think we're OK being a bit redundant on this issue. 

We'll want to follow-up with information about how to verify the correctness of the TSC, as per [this notion issue](https://www.notion.so/risczero/Publish-and-duplicate-ceremony-contributions-6b42b06309c14162bf95dadddaad483b) and [kobi's twitter thread](https://twitter.com/kobigurk/status/1782502969453494530).


tzerrell (approved):
Yes, good to include this here

PR #1726, opened 2024-04-25
Change `ProverServer::compress` to work with `Receipt`

SchmErik:
This change re-implements the compress function. We want the ability for a user to generate any kind of receipt using the Prover and ProverServer. In other words, we want to be able to translate composite receipts to succinct receipts, and succinct receipts into compact receipts to move up through the various receipt stages like so:
```
Composite -> Succinct -> Compact
```
In a previous PR, I've added a `ReceiptKind` field with in `ProverOpts` so that `ProverImpl::prove_session` can generate any kind of receipt specified by `ProverOpts`. In this PR, I changed the semantics of the Compress function to mean "for a given ProverOpts and Receipt, compress the receipt to the desired receipt kind". The following table illustrates the semantics of the new compress functionality:

| InnerReceipt | ReceiptFormat | Result                           |
|--------------|---------------|----------------------------------|
| Fake         | any           | no-op                            |
| Composite    | Composite     | no-op                            |
| Composite    | Succinct      | Composite -> Succinct            |
| Composite    | Compact       | Composite -> Succinct -> Compact |
| Succinct     | Composite     | Error                            |
| Succinct     | Succinct      | no-op                            |
| Succinct     | Compact       | Succinct -> Compact              |
| Compact      | Composite     | Error                            |
| Compact      | Succinct      | Error                            |
| Compact      | Compact       | no-op                            |

In order to implement this, I realized that the existing compress function took a composite receipt and compressed it to a succinct receipt. I've renamed this to `composite_to_succinct` and created a similar `succinct_to_compact` function so that I can use the compress function to move up the various receipt types like so:
```
composite_to_succinct
          |   succinct_to_compact
          |           |
          V           V
Composite -> Succinct -> Compact
    ^                      ^
    |______________________|
          compress
```


flaub:
I feel like this is missing a bunch of tests that basically covers the table above.

SchmErik:
@flaub tests have been written

flaub (approved):

PR #1735, opened 2024-04-27
Cleanup datasheet and benchmarks

flaub:


flaub:
@nahoc A new column is introduced to the datasheet CSV, the first column is now a `name` field.

PR #1738, opened 2024-04-29
feat: add datasheet new columns

nahoc:
new table columns on `next`:
image


flaub:
There should be more rows now as well. Could we make the columns be in this order?

```
Name │ Hash │ Cycles | Duration │ RAM │ Seal | Speed
```

flaub (approved):

PR #1750, opened 2024-04-30
cargo-risczero test: fix getrandom and add to CI

SchmErik:
There was a regression where the getrandom feature got dropped from the static library. This PR fixes that and adds a `cargo risczero test` command to CI to prevent future regressions


SchmErik:
converting to draft, will reopen after tests are passing

.github/workflows/main.yml
```diff
@@ -279,6 +284,36 @@ jobs:
...
+      - uses: ./risc0/.github/actions/sccache
+        with:
+          key: ${{ matrix.os }}-default
+      - name: Install risczero toolchain
+        run: |
+          cargo run --bin cargo-risczero --no-default-features -- risczero install --version $RISC0_TOOLCHAIN_VERSION
+          cargo install --force --path risc0/cargo-risczero -F experimental
+        working-directory: risc0
+      - name: run cargo risczero test
+        working-directory: ${{ github.workspace }}/RustCrypto-crypto-bigint
```
flaub, line 313: Can/should we add this to an existing job rather than having to spin up a new machine? How long does it take to run this test? Do we need to run it on all targets or is Linux sufficient?

Basically we need to be mindful of the number of CI machines we have available and the impact this has on build times.

SchmErik: @flaub sorry for the late response - I somehow missed this message. This test doesn't take a very long time and we can include it as a part of another workflow. I think linux is sufficient but doesn't hurt to run it on mac. We don't care about running this on metal/cuda.

How about we rename the "check-template" job to "test cargo-risczero commands" and have that workflow run tests involving cargo risczero?

SchmErik:
@flaub i've reverted this back to a different job

flaub (approved):

PR #1752, opened 2024-04-30
Better Error Message

rlukata:
This PR updates the `receipt_download` function to filter for 404 errors, and return an appropriate response.

### DIFF
--- a/bonsai/sdk/src/alpha.rs
+++ b/bonsai/sdk/src/alpha.rs
@@ -45,6 +45,9 @@ pub enum SdkErr {
     /// Missing file
     #[error("failed to find file on disk: {0:?}")]
     FileNotFound(#[from] std::io::Error),
+    /// Receipt not found
+    #[error("Receipt not found")]
+    ReceiptNotFound,
 }
 
 /// Collection of serialization object for the REST api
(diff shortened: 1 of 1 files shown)


bonsai/sdk/src/alpha.rs
```diff
@@ -482,11 +485,14 @@ impl Client {
             .send()?;
 
         if !res.status().is_success() {
+            if res.status().as_u16() == 404 {
```
mothran, line 488: Instead of the as_u16 we could use: `if res.status() == reqwest::StatusCode::NOT_FOUND`
Not there is a `if res.status() == 204 {` in `get_image_upload_url` we should also change.

mothran (approved):

bonsai/sdk/src/alpha.rs
```diff
@@ -482,11 +485,14 @@ impl Client {
...
+                return Err(SdkErr::ReceiptNotFound);
+            }
             let body = res.text()?;
             return Err(SdkErr::InternalServerErr(body));
```
Cardosaum, line 492: I wonder if it could be the case to add a check here to ensure the body is not empty, otherwise we could still present an empty error message to users.

rlukata: I'm not sure this is an issue, sometimes we want to be opaque so as to not reveal what is going on in the backend.

bonsai/sdk/src/alpha.rs
rlukata: [119475c](https://github.com/risc0/risc0/pull/1752/commits/119475cdb56ee15b18ca569a2a29c61d3b5467c2)

PR #1762, opened 2024-04-30
Support producing `CompactReceipt` through `BonsaiProver` and improve `Prover::compress`

nategraf:
This PR adds support for producing a `CompactReceipt` through the `BonsaiProver` impl of the `Prover` trait. This unblocks moving proving in `risc0-ethereum` and the Foundry template to the `risc0-zkvm` API (https://github.com/risc0/risc0-foundry-template/issues/92), and helps simplify the implementation of receipt metadata (coming in a PR soon).

This PR additionally fixes up some missing logic in the `Prover::compress` implementations, and changes the semantics of `opts.receipt_kind`. With this PR, the receipt kind field is interpreted a _minimum_ compression level, rather than exact. As a result, it is no longer an error to call  `compress` on a `Compact` receipt, with `ReceiptKind::Succinct`. Instead, it is a no-op. I believe this better aligns with the expectation of a caller, and prevents needing to check the receipt type before calling compress.


risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -120,18 +133,27 @@ pub struct ProverOpts {
     // moment if there is a better place. At some point before 1.0, this option should be moved or
     // dropped.
     pub prove_guest_errors: bool,
-    /// The kind of receipt to be generated.
+    /// Kind of receipt that to be generated by the prover.
```
SchmErik, line 136:
```suggestion
    /// Kind of receipt to be generated by the prover.
```

SchmErik:
> This PR additionally fixes up some missing logic in the `Prover::compress` implementations,

Thanks so much for doing this btw!

> and changes the semantics of `opts.receipt_kind`. With this PR, the receipt kind field is interpreted a _minimum_ compression level, rather than exact. As a result, it is no longer an error to call `compress` on a `Compact` receipt, with `ReceiptKind::Succinct`. Instead, it is a no-op. I believe this better aligns with the expectation of a caller, and prevents needing to check the receipt type before calling compress.

After looking at the code, I understand what you mean...
`ProverOpts::Default()` doesn't make much sense when you're running bonsai. The question is "what should we do when we ask bonsai for `ProverOpts::Default()`?" I think there's two answers A) return an error or B) return a succinct receipt and this solution goes with option B. So if we go with B it's easy to change the receipt format semantics to mean "this is the minimum compression level to return".

I'm entertaining the thought "what if we go with A and be strict about the type that the impls return"? If we do that, the user wouldn't really need to check the receipt type they get from prove (they are certainly welcome to do this if they want to program defensively).

I feel that going with these new semantics makes things easier if you're interested in compact receipts. but I'm not sure what the devex would be like if you specifically want succinct receipts. Then this semantics makes it more difficult for users to trust that they will get the right kind of receipt back. @flaub do you have any thoughts?

risc0/r0vm/src/lib.rs
```diff
@@ -174,14 +174,13 @@ impl Cli {
...
-                ReceiptKind::Succinct => risc0_zkvm::ReceiptKind::Succinct,
-                ReceiptKind::Compact => risc0_zkvm::ReceiptKind::Compact,
-            },
+        let mut opts = ProverOpts::default();
+        opts.hashfn = hashfn.to_string();
+        opts.prove_guest_errors = self.prove_guest_errors;
+        opts.receipt_kind = match self.receipt_kind {
+            ReceiptKind::Composite => risc0_zkvm::ReceiptKind::Composite,
+            ReceiptKind::Succinct => risc0_zkvm::ReceiptKind::Succinct,
+            ReceiptKind::Compact => risc0_zkvm::ReceiptKind::Compact,
```
flaub, line 183: I prefer to avoid using `mut` as much as possible to prevent future bugs.

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -120,18 +133,27 @@ pub struct ProverOpts {
...
 }
 
-/// This enum represents the various receipt kinds that can be generated.
-#[derive(Clone, Serialize, Deserialize, PartialEq)]
+/// An enumeration of receipt kinds that can be requested to be generated.
+#[derive(Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum ReceiptKind {
-    /// Composite Receipts resulting from proving a session
+    /// Request that a [CompositeReceipr][crate::CompactReceipt] be generated.
```
flaub, line 144:
```suggestion
    /// Request that a [CompositeReceipt][crate::CompositeReceipt] be generated.
```

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -498,96 +498,81 @@ fn sys_input() {
...
+        host::server::prove::{DevModeProver, ProverServer},
+        ExecutorEnv, InnerReceipt, ProverOpts, Receipt, ReceiptKind,
     };
-    use anyhow::{bail, Result};
-    use risc0_groth16::docker::stark_to_snark;
     use risc0_zkvm_methods::{multi_test::MultiTestSpec, MULTI_TEST_ELF, MULTI_TEST_ID};
 
     #[test]
-    fn stark2snark() {
+    fn prove_nothing_compact() {
```
flaub, line 508: I'd much prefer to keep the old name, it's what i use for filtering tests

flaub:
I'm not sure about all these changes, I need to think on them more. I'm hoping we can first get all the changes needed for circuit composition and then maybe go back and see if we need to adjust the API afterwards.

risc0/zkvm/src/host/server/prove/tests.rs
nategraf: Changed back in 9af4758c18320d63188a75d9063c85a391b9666e

nategraf:
> I'm not sure about all these changes, I need to think on them more. I'm hoping we can first get all the changes needed for circuit composition and then maybe go back and see if we need to adjust the API afterwards.

Sounds good. I made these changes in support of getting receipt metadata plumbed through to the smart contracts. (In particular, it allows me to move the Ethereum code to use the `Prover` trait and prevents me having to implement the other changes twice)

flaub (approved):

risc0/r0vm/src/lib.rs
nategraf: In 4fdda5fb03057db0228d3d9d367307c793f9732c I addressed this by adding a kind of builder pattern to the `ProverOpts`. Note that the reason for this change is that I added `#[non_exhaustive]` to `ProverOpts`. Without `#[non_exhaustive]`, because all the fields are public, it is a breaking change to add new fields.
https://github.com/risc0/risc0/blob/4fdda5fb03057db0228d3d9d367307c793f9732c/risc0/zkvm/src/host/client/prove/mod.rs#L122-L123

PR #1763, opened 2024-04-30
Added the session/stop method to the Bonsai SDK

mothran:
This PR adds support for the new stop session API in Bonsai.


bonsai/sdk/src/alpha.rs
```diff
@@ -258,6 +258,17 @@ impl SessionId {
...
         Ok(res.text()?)
     }
+
+    /// Stops a running proving session
+    pub fn stop(&self, client: &Client) -> Result<(), SdkErr> {
+        let url = format!("{}/sessions/stop/{}", client.url, self.uuid);
+        let res = client.client.get(url).send()?;
+        if !res.status().is_success() {
+            let body = res.text()?;
+            return Err(SdkErr::InternalServerErr(body));
```
rlukata, line 268: would it make sense to filter for errors stemming from `SessionStopErr` enum?

mothran: Yeah that would be nice, unfortunately the backend does not route errors to status codes well enough to unify that error handling. Lets chat about overhauling the backend error handling system. I have a design I would like to roll out that would make this much easier client side.

rlukata: Sounds good, I still think we should clarify the error, I suggest we copy the enum for now:

```
#[derive(Error, Debug)]
pub enum SessionStopErr {
    #[error("You are unauthorized to stop session id: {0}")]
    Unauthorized(String),
    #[error("Unable to find session id")]
    NotFound,
    #[error("Workflow not currently Running")]
    NotRunning,
}
```

mothran: Yeah but the API side does not return enough to remap those on the client side yet, hence the refactor server side.

rlukata: I see what youre saying, fair enough!

rlukata (approved):

PR #1764, opened 2024-04-30
Add ReceiptMetadata with verifier_info field for checking for verifier parameter mismatch

nategraf:
This PR adds a `ReceiptMetadata` struct and attaches it to the `Receipt` type.

The first field added to the `ReceiptMetadata` is a `verifier_info: Digest` value, which is constructed from a hash of the parameters for a given a given verifier.
This value can be used to tell when a `Receipt` and a verifier have different parameters (e.g. the control root may have changed).
It is intented additionally to serve as a routing key for the Ethereum on-chain verifier contracts, ensuring we send a given receipt to the right place.

Additionally, this PR:

* Adds an `Input` type to the `ReceiptClaim` to serve as a placeholder for a future value type.
  It is implemented as an uninhabited type, which means it cannot be constructed.
  This ensures that adding fields to this type is not a breaking change to our APIs, and all code paths will deal with the digest only.
* Refactors the `risc0_groth16` public API to introduce a new `VerifyingKey` type.
  This type implements the `Digestible` trait and is serializable with `serde`, which eases implementation of `ReceiptMetadata`.
  Additionally, it works to hide the `arkworks` types such that we have the option to swap arkworks (which is primarly a library for R&D of proof systems) out for a different Groth16 verifier implementation.

This PR is in draft state while I am still working on:

* [x] Adding tests for receipt metadata, including that it remains stable between releases.
* [x] Refacoring to remove the `metadata.rs` file and simply move the types along side the receipt implementation.
* [x] Rebase onto a non-functional change PR that just splits apart `receipt.rs`


flaub:
I wonder if we can first make a NFC that splits out the receipt types so that it's a lot easier to review.

nategraf:
> I wonder if we can first make a NFC PR that splits out the receipt types so that it's a lot easier to review.

I'll go ahead and do that.

nategraf:
> I wonder if we can first make a NFC PR that splits out the receipt types so that it's a lot easier to review.

This is done now

nategraf:
> > I wonder if we can first make a NFC PR that splits out the receipt types so that it's a lot easier to review.
> 
> I'll go ahead and do that.

This is done now

risc0/zkvm/methods/rand/Cargo.lock
```diff
@@ -354,6 +354,12 @@ dependencies = [
  "either",
 ]
 
+[[package]]
+name = "lazy_static"
+version = "1.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
```
flaub, line 361: I wonder why this is showing up in all the guest lock files?

nategraf: I added `lazy_static`as a dependency of `risc0_groth16`, which is a dependency of `risc0_zkvm` (used to verify the Groth16 receipts).

flaub: Is it possible to use `OnceCell` now that this is in stable rust? I think I recently went thru and replaced all lazy static with that.

nategraf:
Ok, all tests are passing now expect the Ethereum tests (https://github.com/risc0/risc0-ethereum/pull/93 should help keep this test green) and semver checks (this PR contains breaking changes)

nategraf:
In [0a37650](https://github.com/risc0/risc0/pull/1764/commits/0a3765055625a7d13ca80bfad0a82984c5817582), I fixed support for `no-std` in `risc0-groth16`. It appears is was half-implemented and the `cfg` to actually enable `no-std` in `lib.rs` was never added. I added that and fixed a handful of errors that popped up as a result. No `risc0-groth16` can be used in `no-std` and there is no need for the `cfg` statements that were added in https://github.com/risc0/risc0/pull/1776 to conditionally compile `CompactReceipt`. This avoids the need for any hackyness in getting `CompositeReceiptVerifierInfo` to be available and consistent between zkVM and host.

nategraf:
@flaub, when you get a chance, this is ready for review

nategraf:
It would be helpful to get your review here @capossele, in particular on the changes to the interface of `risc0_groth16`. I did some work to:
* Remove arkworks from the public interface, as it feels quite likely we will want to use a different basis for verification in the future and it would be good to have forwards compatibility there. One reason in particular for this is that we want to have acceleration for Groth16 verification in the near-ish future.
* Change from have the prepared verifier key be what people pass around to just being the verifier key. It's not clear to me what the security implications are of receiving a prepared verifying key and using it. It also was odd to have the "prepared" version be the only public option, and we could reintroduce precomputed preparation of the key under the surface.

examples/groth16-verifier/methods/guest/src/bin/groth16_verifier.rs
```diff
@@ -12,22 +12,16 @@
...
+use risc0_zkvm::{guest::env, sha::Digestible};
 
-use risc0_groth16::Verifier;
-use risc0_zkvm::{guest::env, sha::Digest};
-use sha2::{Digest as _, Sha256};
+pub fn main() {
+    let (seal, public_inputs, verifying_key): (Seal, Vec<Fr>, VerifyingKey) = env::read();
 
-risc0_zkvm::guest::entry!(main);
+    Verifier::new(&seal, &public_inputs, &verifying_key)
```
capossele, line 21: Now, with this change we are almost doubling the required cycles to verify a groth16 proof in the guest, specifically from ~490 on main VS 980 on this branch. Are we aware of this?

capossele (approved):
Just a note on the performance hit while verifying a groth16, but rest looks good and is also reasonable

examples/groth16-verifier/methods/guest/src/bin/groth16_verifier.rs
nategraf: I saw that, and I should have called it out in particular. My thought here is that whether it's 1B vs 500M cycles, it's kind of both in the realm of "too expensive to run on a regular basis". My suggestion here is that we make this refactor now, and revisit this later, especially when we have an accelerator for pairings. My goal here is forwards compatibility with those kinds of changes, and we could reintroduce a extra info like the key preparation host-side when we revisit this to improve performance.

risc0/zkvm/src/receipt/segment.rs
```diff
@@ -107,6 +120,33 @@ impl SegmentReceipt {
     }
 }
 
+/// Verifier parameters used to verify a [SegmentReceipt].
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct SegmentReceiptVerifierParameters {
+    /// Set of control ID with which the receipt is expected to verify.
+    pub control_ids: BTreeSet<Digest>,
```
flaub, line 128: I wonder if this will be good for future deserializers. `BTreeSet` is a nice type, but `Vec` might be more future compatible. Not sure what the right thing here is.

risc0/zkvm/src/receipt/composite.rs
```diff
@@ -49,6 +57,16 @@ pub struct CompositeReceipt {
 }
 
 impl CompositeReceipt {
+    /// Information about the parameters used to verify the receipt. Includes parameters that are
+    /// useful in deciding whether the verifier is compatible with a given receipt.
+    pub fn verifier_parameters() -> CompositeReceiptVerifierParameters {
+        CompositeReceiptVerifierParameters {
+            segment: SegmentReceipt::verifier_parameters(),
+            succinct: SuccinctReceipt::verifier_parameters(),
+            compact: CompactReceipt::verifier_parameters(),
```
flaub, line 66: It seems kind of redundant with this structuring, IIUC, the prover info and circuit info will be embedded here multiple times.

risc0/zkvm/src/receipt/composite.rs
```diff
@@ -221,3 +242,54 @@ impl CompositeReceipt {
...
+pub struct CompositeReceiptVerifierParameters {
+    /// Verifier parameters related to [SegmentReceipt].
+    pub segment: SegmentReceiptVerifierParameters,
+    /// Verifier parameters related to [SuccinctReceipt].
+    pub succinct: SuccinctReceiptVerifierParameters,
+    /// Verifier parameters related to [CompactReceipt].
+    pub compact: CompactReceiptVerifierParameters,
+}
+
+impl Digestible for CompositeReceiptVerifierParameters {
```
flaub, line 262: I'm kind of unsure why we'd need to worry about making the metadata digestible if it can't impact the claim or attestations.

risc0/zkvm/src/receipt/composite.rs
nategraf: This is partially true. Prove system identifier (that 16 byte string we added to hash into Fiat-Shamir) will appear in segment and succinct. Other than that, the info is mostly distinct. With the slightly weird exception that the recursion control tree includes control IDs for rv32im programs with Poseidon2. This is kind of an implementation quirk though.

In practice, I don't think this should matter too much. In all the encoded forms, only the digest of this struct is passed around.

risc0/zkvm/src/receipt/composite.rs
nategraf: And to be clear, the `ReceiptMetadata` as a whole does no define a digest method, and it wouldn't make much sense to do so.

https://github.com/risc0/risc0/pull/1764/files/7e5c06f843847db1928b41b5e195eca45b738a47#diff-63cd47df26551e5cfb26448b6449ba92432b6f5a277788d6000168d5e0569189R412

risc0/zkvm/src/receipt/segment.rs
nategraf: Good question. I'm not sure what is best here either. On serialization, at least `serde` treats `BTreeSet` as a generic sequence [1]. I was curious, so I tried decoding an encoded `BTreeSet` into a `Vec` and that also worked [2]. Doesn't fully answer the question though.

[1] https://docs.rs/serde/latest/src/serde/ser/impls.rs.html#239-243
[2] https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=437b6972c4d482f7f024e76c5d12a69c

risc0/zkp/src/verify/mod.rs
```diff
@@ -57,21 +58,24 @@ impl fmt::Debug for VerificationError {
 impl fmt::Display for VerificationError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            VerificationError::VerifierParametersMismatch { expected, received } => {
+                write!(f, "receipt was produced for a version of the verifier with info digest {received}; expected {expected}")
```
nategraf, line 62:
```suggestion
                write!(f, "receipt was produced for a version of the verifier with parameters digest {received}; expected {expected}")
```

flaub (approved):

nategraf:
Looks like this is failing now due to the checked-in receipt for `cargo risczero verify`.

flaub:
I ran into a similar problem so my PR drops the risczero verify test until we can generate a receipt automatically at build time: https://github.com/risc0/risc0/pull/1807

PR #1765, opened 2024-04-30
build_kernel: fix nvcc failure by splitting RISC0_NVCC_FLAGS

SchmErik:


risc0/build_kernel/src/lib.rs
```diff
@@ -141,7 +141,9 @@ impl KernelBuild {
 
         let mut flags = vec![];
         if let Ok(nvcc_flags) = env::var("RISC0_NVCC_FLAGS") {
-            flags.push(nvcc_flags);
+            for flag in nvcc_flags.split(" ") {
+                flags.push(flag.to_string());
+            }
```
flaub, line 146: This won't work with things like `RISC0_NVCC_FLAGS=--foo='x y'`. This issue might not be worth worrying about for this use case, but this could solve it: https://docs.rs/shlex/latest/shlex/fn.split.html

SchmErik: thanks for catching this!

flaub (approved):
Looks like `shlex` doesn't work with other deps, probably not worrying about it we can just assume that we won't use quotes with spaces for flags.

mothran (approved):
Looks good from my perspective as long as it works with the bonsai NVCC sample string args. I can give it a test if you haven't already.

flaub (approved):

SchmErik:
> Looks good from my perspective as long as it works with the bonsai NVCC sample string args. I can give it a test if you haven't already.

I tested your command and it no longer gets stuck on the nvcc error that you saw!

PR #1766, opened 2024-05-01
fix: remove caching for datasheet fetches, lower other caches to 15mins

nahoc:


web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/datasheet/_actions/fetch-datasheet-commit-hash.ts
```diff
@@ -1,16 +1,18 @@
 "use server";
 
+import { unstable_noStore as noStore } from "next/cache";
 import env from "~/env";
 
 export async function fetchDatasheetCommitHash({ version }: { version: string }) {
+  noStore();
```
Comment (nahoc) [inline web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/datasheet/_actions/fetch-datasheet-commit-hash.ts:7]: `noStore` means no caching on Vercel's side

flaub (approved):

PR #1773, opened 2024-05-01
docs: add note for versioned install

austinabell:
Very opinionated how this is documented, but is a common question of how to handle version mismatches. I'm sure this wording can be improved, but just getting a pulse on if this would be helpful first.

- Should we include this line in previous versions?
- Should we make a note that this installed version must match the version used in the host/guest of the app? Probably?


website/api/zkvm/install.md
```diff
@@ -25,6 +25,8 @@ cargo binstall cargo-risczero
 cargo risczero install
 ```
 
+> Note: To install a specific version instead of using the latest stable version, use `cargo binstall cargo-risczero@<version>` where the version is the crates.io release tag.
```
flaub, line 28:
```suggestion
> Note: To install a specific version instead of using the latest stable version, use `cargo binstall cargo-risczero@<version>` where the version is a [release tag](https://github.com/risc0/risc0/releases). 
```

SchmErik (approved):

PR #1774, opened 2024-05-02
Add `rzup` installation script

hmrtn:
This introduces two shell scripts for easy installation and version management of the r0 toolchain with `rzup`.

To install `rzup`, download and execute the install script with this one-liner:

```sh
curl -L https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/install | bash
```

This automates the installation by downloading the `rzup` and adding it to the user's `PATH`.

Once installed, use `rzup`:

```sh
rzup
```

Running `rzup` automates the installation of `cargo-binstall`, `cargo-risczero`, and the toolchain, making it very easy for users get started. 

We should consider setting up a redirect to a more user-friendly URL, such as:

```
curl -L r0.xyz/install | bash
```

Prior to merging, we should edit the url paths to main, rather than the installer branch.
https://github.com/risc0/risc0/blob/hmrtn/rzup-installer/rzup/install#L14
https://github.com/risc0/risc0/blob/hmrtn/rzup-installer/rzup/README.md?plain=1#L9

I have tested this on MacOS and Linux. It would be great to have a windows user test it.


rzup/rzup
```diff
@@ -0,0 +1,157 @@
...
+    else
+      echo "Your Rust version ($current_version) is below the minimum requirement ($MIN_RUST_VERSION)."
+      echo "Please update Rust by running 'rustup update' and rerun this script."
+      exit 1
+    fi
+  fi
+}
+
+install_cargo_binstall() {
+  execute_with_feedback "cargo install cargo-binstall --quiet" "Installing cargo-binstall"
```
austinabell, line 91: Small note that this will build cargo-binstall from scratch if you have an older version, but this shouldn't be necessary to have the most up to date version. Feel free to ignore this comment, just figure I'd mention!

flaub:
Would it be possible to test this in CI?

flaub:
I should note that windows isn't a supported platform right now (unless you consider WSL/WSL2, but that's not really windows).

rzup/install
```diff
@@ -0,0 +1,49 @@
...
+set -eo pipefail
+
+echo "Installing rzup..."
+
+BASE_DIR=${XDG_CONFIG_HOME:-$HOME}
+RISC0_DIR=${RISC0_DIR-"$BASE_DIR/.risc0"}
+RISC0_BIN_DIR="$RISC0_DIR/bin"
+
+# TODO: Change this path to main
+BIN_URL="https://raw.githubusercontent.com/risc0/risc0/hmrtn/rzup-installer/rzup/rzup"
```
flaub, line 14: I suggest we host this on an S3 bucket to avoid github rate limiting.

rzup/rzup
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
+
+MIN_RUST_VERSION="1.74.1" # Minimum required Rust version
```
flaub, line 3: Is this correct? We should note this in our docs as the MSRV (and in the main README)

rzup/README.md
```diff
@@ -0,0 +1,41 @@
+# `rzup`
+
+Install, update, or revert to a specific RISC Zero version.
+
+## Install
+
+<!-- TODO: Replace this with the correct url path once merged. -->
+```sh
+curl -L https://raw.githubusercontent.com/risc0/risc0/hmrtn/rzup-installer/rzup/install | bash
```
flaub, line 9: @morgangibbs87 There's an existing bucket we can use for this. DM for details.

flaub (approved):

rzup/rzup
hmrtn: After asking @SchmErik about this I found out we do not have a specified MSRV for `cargo-risczero` host installation. I tested our install msrv with a script for every rust version `nightly`-`1.1.0` and found that it fails at versions <`1.60.0` due to `cargo-binstall` support (in hindsight I should have just checked crates.io). So I recommend we change this to variable `1.60.0`, if the check is worth keeping at all. Also okay to modify it simply to check if rust is installed, and let the user find out they are very out-of-date on their own.

hmrtn:
> Would it be possible to test this in CI?

I am not very familiar with our CI workflow, but messing around locally with a new job in similar fashion to the others, I think it is possible.

My local workflow testing with the `act` tool fails, but something like this is probably what we want:

```diff
diff --git a/.github/workflows/main.yml b/.github/workflows/main.yml
index ebdc9e20..988984dd 100644
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml

@@ -38,6 +38,7 @@ jobs:
       examples: ${{ steps.filter.outputs.examples }}
       test: ${{ steps.filter.outputs.test }}
       web: ${{ steps.filter.outputs.web }}
+      rzup: ${{ steps.filter.outputs.rzup }}
     steps:
       - uses: actions/checkout@v4
       - uses: risc0/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36
@@ -83,6 +84,9 @@ jobs:
               - examples/browser-verify/**
               - risc0/**
               - xtask/**
+            rzup:
+              - *base
+              - rzup/**

   # see: https://github.com/orgs/community/discussions/26822
   main-status-check:
@@ -431,3 +435,28 @@ jobs:
         env:
           RISC0_USE_DOCKER: "1"
       - run: sccache --show-stats
+
+  rzup:
+    if: needs.changes.outputs.rzup == 'true'
+    needs: changes
+    runs-on: [self-hosted, prod, "${{ matrix.os }}", cpu]
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [Linux, macOS]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/rustup
+      - uses: ./.github/actions/sccache
+        with:
+          key: ${{ matrix.os }}-default
+      - name: Install rzup
+        run: |
+          chmod +x rzup/install
+          rzup/install
+          source ~/.bashrc
+      - name: Run rzup
+        run: |
+          chmod +x rzup/rzup
+          rzup/rzup --verbose
+      - run: sccache --show-stats
```

rzup/rzup
hmrtn: I went with the simpler option, and now just check the rust installation.

rzup/README.md
hmrtn: replaced with: https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/install

rzup/install
hmrtn: replaced with https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/rzup

PR #1775, opened 2024-05-02
Add cargo risczero verify

### LINKED ISSUE #1619: [Feature] cargo-risczero: add a `verify` command to verifiy receipts locally or from bonsai
## Feature
Add a command in cargo-risczero that would verify a receipt stored locally or from bonsai. The command will fetch the receipt from bonsai or disk and run receipt.verify on it.

## Motivation
The primary motivation is improved devex. This feature allows one to easily verify existing receipts with out 1) going through the prover or writing code in order to verify receipts.

## Implementation
Create a new command `verify` in cargo-risczero. There should be two modes of operation: 1) download a receipt from bonsai and 2) fetch a receipt from local storage. In the bonsai case, we need to use the bonsai-sdk to download a receipt. Parameters to this should be the bonsai url/api key, and an ID that refers to a receipt. In the local case, we need the filepath of the receipt. After downloading the receipt, it should be de-serialized to a receipt type and we should be able to call receipt.verify() and return the result.

Cardosaum:
This PR adds the command `verify` to `cargo risczero`. 

The user should provide the `image_id` of the guest as a positional parameter, and specify as an argument either the path to a local receipt saved in binary format using `bincode` or specify the Bonsai ID of the receipt to download and verify. 

The usage of the new command looks like the following: 

```
$ cargo-risczero risczero verify --help

Verifies if a receipt is valid

Usage: cargo risczero verify [OPTIONS] <--path <PATH>|--id <ID>> <IMAGE_ID>

Arguments:
  <IMAGE_ID>  The image ID to verify the receipt against

Options:
      --path <PATH>        The path to the receipt file
      --id <ID>            The Bonsai ID of the receipt to download and verify
      --api-url <API_URL>  API URL for Bonsai
      --api-key <API_KEY>  API key for Bonsai
  -h, --help               Print help
```

--- 

Closes #1619


risc0/cargo-risczero/src/lib.rs
```diff
@@ -48,6 +48,7 @@ pub struct Risczero {
     pub command: RisczeroCmd,
 }
 
+#[non_exhaustive]
```
Cardosaum, line 51: I'm making this non-exhaustive so future changes are not considered major breaking changes in the semver check.

SchmErik (commented):
Could you add a test that exercises this command? We've had some cargo-risczero commands that broke recently because they weren't under test. I think it can be as simple as reading a receipt using the `--path` option.

Cardosaum:
> Could you add a test that exercises this command? We've had some cargo-risczero commands that broke recently because they weren't under test. I think it can be as simple as reading a receipt using the `--path` option.

Thanks for the feedback, @SchmErik !

I added 3 tests for `cargo risczero verify`, one that checks for the output of `--help`, one for a verification of a valid receipt, and one checking for a receipt with wrong image id. 

I have not added tests for Bonsai verification as this would require an API key and I'm not sure about what would be the appropriate way to integrate it with CI. 

Please let me know if this suffice!

Cardosaum:
I'm not sure why [CI tests](https://github.com/risc0/risc0/actions/runs/8968374177/job/24627667775?pr=1775) are failing for `tests/cmd/verify_receipt_path_invalid_image_id.toml`, it works fine locally. Maybe some `sccache` caching problem in CI?

SchmErik:
> I'm not sure why [CI tests](https://github.com/risc0/risc0/actions/runs/8968374177/job/24627667775?pr=1775) are failing for `tests/cmd/verify_receipt_path_invalid_image_id.toml`, it works fine locally. Maybe some `sccache` caching problem in CI?

Looks like CI has RUST_BACKTRACE=full and your results don't reflect this. Maybe try running with backtrace = full and capture expected tests?

Cardosaum:
> > I'm not sure why [CI tests](https://github.com/risc0/risc0/actions/runs/8968374177/job/24627667775?pr=1775) are failing for `tests/cmd/verify_receipt_path_invalid_image_id.toml`, it works fine locally. Maybe some `sccache` caching problem in CI?
> 
> Looks like CI has RUST_BACKTRACE=full and your results don't reflect this. Maybe try running with backtrace = full and capture expected tests?

Good call, thanks for that - will solve this.

risc0/cargo-risczero/tests/verify.rs
```diff
@@ -0,0 +1,68 @@
...
+        .arg("risczero")
+        .arg("verify")
+        .arg(image_id)
+        .arg("--path")
+        .arg(receipt)
+        .assert()
+        .code(code)
+        .stdout(stdout)
+        .stderr(stderr);
+}
```
Cardosaum, line 68: @SchmErik , I refactored the code, removing `trycmd` and using `assert_cmd` to check CLI output instead. 
Thanks for pointing out that we were using `RUST_BACKTRACE=full` in tests, this was the reason it was failing indeed. 

I had to drop `trycmd` in favor of `assert_cmd` because in the former we could only match on exact strings whereas in the latter we have the flexibility to use `predicates` to match on substrings, regex, empty strings etc. The tests are way simpler now. If you like the way we do here, I could open a new PR to test all other commands in `cargo risczero` in this way. Let me know your thoughts :)

SchmErik (approved):

flaub:
This new test breaks future receipts, we need to have it generate a new receipt for each build. I'm going to drop the test for now until this can be automated.

risc0/cargo-risczero/src/lib.rs
flaub: This doesn't make sense for `cargo-risczero`, it doesn't need to maintain backwards compabilitity (at the API-level, only at the command-line level).

PR #1776, opened 2024-05-03
zkvm: run verify in guest

SchmErik:
This PR enables receipt verification in the guest. Note, this is running the risc0-zkvm verifier as guest code, rather than running the recursion circuit or using composition. This allows verification of composite, succinct and compact receipts. One major restructuring was to move `risc0_zkvm::host::receipt` to `risc0_zkvm::receipt`.


SchmErik:
I'm marking this as a draft for now. Note: as of opening this PR, the verification for composite receipts have not been accelerated. I'll address this and the merge conflict tomorrow but for those that are interested, feel free to give me feedback. Not sure how to exactly handle the tracing calls. I might surround them in a macro that will get compiled away in the zkvm...

SchmErik:
@flaub this is basically ready - I left the composite receipt un-accelerated but let me know if you want me to take care of that for this PR

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -580,6 +583,38 @@ mod docker {
...
+            .unwrap()
+            .compress(&ProverOpts::succinct(), &composite_receipt)
+            .unwrap();
+        exec_verify(&succinct_receipt);
+        let compact_receipt = get_prover_server(&ProverOpts::compact())
+            .unwrap()
+            .compress(&ProverOpts::compact(), &succinct_receipt)
+            .unwrap();
+        exec_verify(&compact_receipt);
+        compact_receipt.inner.compact().unwrap();
```
flaub, line 615: I'm surprised this works, I would have guessed that compact receipts can't be verified in the guest right now.

flaub (approved):
Nice!

risc0/zkvm/src/host/server/prove/tests.rs
SchmErik: We have the groth16 verify as an example so I thought that it was feasible but I didn't know it would be this easy. Thanks @capossele!

PR #1780, opened 2024-05-06
Accelerated witness generation

flaub:


risc0/zkp/src/hal/cpu.rs
```diff
@@ -519,6 +523,12 @@ impl<F: Field> Hal for CpuHal<F> {
             });
     }
 
+    fn eltwise_zeroize_elem(&self, elems: &Self::Buffer<Self::Elem>) {
+        elems.as_slice_mut().par_iter_mut().for_each(|elem| {
+            *elem = elem.valid_or_zero();
+        });
+    }
```
nategraf, line 530: I see that this function isn't new. It is an off name though. At first I assumed it would set the elements to zero, but it only does so if the element is `INVALID`

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -87,20 +73,20 @@ fn hashfn_poseidon2() {
...
-    };
-    let env = ExecutorEnv::builder()
-        .write(&MultiTestSpec::DoNothing)
-        .unwrap()
-        .build()
-        .unwrap();
-    let prover = ProverImpl::new("cpu:blake2b", hal_pair, ReceiptKind::Composite);
-    prover.prove(env, MULTI_TEST_ELF).unwrap();
-}
+// #[test]
```
nategraf, line 76: Should we add an issue to track this getting re-enabled?

.github/workflows/main.yml
```diff
@@ -284,33 +284,33 @@ jobs:
...
-      - name: Install risczero toolchain
-        run: |
-          cargo run --bin cargo-risczero --no-default-features -- risczero install --version $RISC0_TOOLCHAIN_VERSION
-          cargo install --force --path risc0/cargo-risczero -F experimental
-        working-directory: risc0
-      - name: run cargo risczero test
-        working-directory: ${{ github.workspace }}/RustCrypto-crypto-bigint
-        run: cargo risczero test
-      - run: sccache --show-stats
+  # test-crates:
```
nategraf, line 287: Should we add an issue to track this getting re-enabled?

nategraf (approved):

risc0/circuit/rv32im-sys/cxx/ffi.cpp
```diff
@@ -14,66 +14,303 @@
...
-  });
+void inject_backs_ram(MachineContext* ctx, size_t steps, size_t cycle, Fp* data) {
+  uint8_t kind = ctx->isParSafeVerifyMem(cycle);
+  if (cycle > 2 && kind) {
+    size_t idx = ctx->ramIndex[cycle];
+    if (idx == 0) {
+      throw std::runtime_error("inject_backs_ram: idx == 0");
+    }
+
+    const RamArgumentRow& back1 = ctx->ramRows[idx - 1];
```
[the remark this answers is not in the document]
flaub: Actually @shkoo has a follow up PR that should be more self-documented by using the layout structures emitted by the compiler.

risc0/zkp/src/hal/cpu.rs
flaub: Any better name?

risc0/zkvm/src/host/server/prove/tests.rs
flaub: I plan to address this soon. I'm trying to get this landed so we can do an rc, then I'll go right back to fixing all the loose ends.

PR #1795, opened 2024-05-08
add mention of steel to dev docs site

pdg744:
The dev docs site doesn't mention Steel yet. This PR adds a brief section to the Ethereum Examples page, to improve findability of Steel. 

Questions worth thinking about:
- is the risc0-steel crate is the appropriate place to point users
- do we want to add mentions/paths to Steel elsewhere (mb in the sidebar?)

### DIFF
--- a/website/api/blockchain-integration/eth-examples.md
+++ b/website/api/blockchain-integration/eth-examples.md
@@ -6,6 +6,10 @@ While all of the [zkVM examples][zkvm-examples] can be run on Bonsai by [configu
 
 The [RISC Zero Foundry Template][foundry-template] provides a minimal application that can act as a template for developing your application.
 
+### STEEL
+
+[STEEL] is a library for [view call proofs], which enables users to easily integrate claims about Ethereum state into zkVM applications.
+
 ### Zeth
 
 [Zeth][zeth-repo] produces ZK validity proofs for Ethereum Virtual Machine blocks (EVM). This is accomplished by running [revm], a Rust implementation of the EVM in the zkVM. The end result is an open-source zkEVM with high code reuse and minute, not hour, proving times on Bonsai. For more info check out [Announcing Zeth: The first Type Zero zkEVM][zeth-article].
@@ -23,3 +27,5 @@ This [example app][governance-example] uses Bonsai as an Ethereum coprocessor. T
 [Governor smart contract standard]: https://docs.openzeppelin.com/contracts/4.x/api/governance
 [signature-aggregation]: https://github.com/risc0/risc0/blob/release-0.20/bonsai/examples/governance/methods/guest/src/bin/finalize_votes.rs
 [foundry-template]: https://github.com/risc0/bonsai-foundry-template
+[STEEL]: https://crates.io/crates/risc0-steel
+[view call proofs]: https://www.risczero.com/blog/introducing-steel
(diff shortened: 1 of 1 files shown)


NukeManDan:
Weird that can see the preview, but I cannot see the changes on https://docs-website-git-paulg-mention-steel-risczero.vercel.app/api/blockchain-integration/eth-examples 

Also a 404 for access the `inspect` url https://vercel.com/risczero/docs-website/5weXuQhj5RLTJ7ykoxWN4oKT2n5X - maybe I need to be grated access @pdg744 ?

NukeManDan:
Idea: Would it be appropriate to expand a bit on each example listed here here on the "why" of Them? The steel blog linked for view cals helps give me a bit more understanding:

> How is this different from storage proofs?
> With traditional storage proofs, developers must manually select which storage slots their smart contract uses, and re-implement smart contract logic. With steel, all storage slots are discovered and fetched automatically based on the view call execution. This saves developers a significant amount of time and decreases the room of implementation errors, reducing the chance of security vulnerabilities.

Overall I think this page could be a good way to succinctly duscuss not just pointers to what exists, but why each one of these is unique and important for EVM related integrations. :thinking:

NukeManDan (approved):
For getting minimal visibility this LGTM :+1:

website/api/blockchain-integration/eth-examples.md
```diff
@@ -6,6 +6,10 @@ While all of the [zkVM examples][zkvm-examples] can be run on Bonsai by [configu
 
 The [RISC Zero Foundry Template][foundry-template] provides a minimal application that can act as a template for developing your application.
 
+### Steel
+
+[Steel] is a library for [view call proofs], which enables users to easily integrate claims about Ethereum state into zkVM applications.
```
nategraf, line 11: I've been finding it helpful to use tags like this for links. It helps understand where the link goes when reading the Markdown text... and [my linter](https://github.com/DavidAnson/markdownlint) yells at me when the links break with this format, but not without the tag.
```suggestion
[Steel][steel-repo] is a library for [view call proofs][steel-blog], which enables users to easily integrate claims about Ethereum state into zkVM applications.
```

website/api/blockchain-integration/eth-examples.md
```diff
@@ -23,3 +27,5 @@ This [example app][governance-example] uses Bonsai as an Ethereum coprocessor. T
 [Governor smart contract standard]: https://docs.openzeppelin.com/contracts/4.x/api/governance
 [signature-aggregation]: https://github.com/risc0/risc0/blob/release-0.20/bonsai/examples/governance/methods/guest/src/bin/finalize_votes.rs
 [foundry-template]: https://github.com/risc0/bonsai-foundry-template
+[Steel]: https://crates.io/crates/risc0-steel
+[view call proofs]: https://www.risczero.com/blog/introducing-steel
```
nategraf, line 31:
```suggestion
[steel-repo]: https://crates.io/crates/risc0-steel
[steel-blog]: https://www.risczero.com/blog/introducing-steel
```

nategraf (approved):

pdg744:
> Idea: Would it be appropriate to expand a bit on each example listed here here on the "why" of Them? 

thanks for raising this idea! i think we should aim for keeping these pretty concise especially considering that this list is likely to grow. that said, this page, and our blockchain docs in general, could use some improvement. would be great for the docs to offer a clearer answer to the question of "where should i start?" and "where should i go next?" when it comes to blockchain development.

Issue #1676, opened 2024-04-16
[Documentation] write documentation for C/C++ rust crate guests


SchmErik:
## User's Objective
Since we now have initial support for rust crates that call into C/C++, we should provide documentation on how this works.

## User's Obstacle
There are no obstacles that the user faces but if a crate doesn't compile, it's nice to have docs that outline how our c/c++ support works so that users have an understanding of what's going on and troubleshoot on their own

## User's Request
No users are involved but it would be nice to have C/C++ crate documentation

PR #1797, opened 2024-05-08
DevDocs: Add docs on rust crates that bind to C++

SchmErik:
Closes: #1676


flaub:
A section on how to install the toolchain would be good and maybe a bit on how the versions work. Maybe this should also go into the existing install page.

SchmErik:
> A section on how to install the toolchain would be good and maybe a bit on how the versions work. Maybe this should also go into the existing install page.

the toolchain is installed by `cargo risczero install` so I don't think we really need to add much to the existing install page - I could clarify that this command will also install the C++ toolchain.

Versions: right now we have only one version and we only support installing. I'll write that in the docs.

> A section on how to install the toolchain would be good and maybe a bit on how the versions work. Maybe this should also go into the existing install page.

feedback addressed!

flaub (approved):

PR #1798, opened 2024-05-08
updating name of foundry template

pdg744:
We currently use "RISC Zero Foundry Template" rather than the previous "Bonsai Foundry Template." 

Fixing a couple spots on the dev docs site where this was out of date.


nategraf (approved):
Glad you found these! I missed those

PR #1799, opened 2024-05-08
Change ffi.cu to use layout

shkoo:


flaub (changes requested):
Blocking this until we have a change to integrate it

flaub (approved):
Integrated!

flaub (approved):

PR #1804, opened 2024-05-09
Document Trusted Setup Ceremony security

tzerrell:
Add documentation for how users can confirm the security properties of our trusted setup ceremony.


website/api/trusted-setup-ceremony.md
```diff
@@ -0,0 +1,70 @@
+# Trusted Setup Security
+
+RISC Zero has run a trusted setup ceremony for our Groth16 prover/verifier. This ceremony secures our STARK Verify circuit so we can publish compact receipts for our general purpose zkVM to limited-memory environments like blockchains.
+
+You don't need to take our word for it that this is secure, though! This document will walk you through how to verify the security of our ceremony for yourself. This can also be used to aid in the detection of attacks related to our ceremony, e.g. it may help detect fraudlent actors publishing something they claim to be the RISC Zero STARK Verifier but which is actually malicious code.
```
pdg744, line 5:
```suggestion
You don't need to take our word for it that this is secure, though! This document will walk you through how to verify the security of our ceremony for yourself. This can also be used to aid in the detection of attacks related to our ceremony, e.g. it may help detect fraudulent actors publishing something they claim to be the RISC Zero STARK Verifier but which is actually malicious code.
```

website/api/trusted-setup-ceremony.md
```diff
@@ -0,0 +1,70 @@
...
+[p0tion]: https://github.com/privacy-scaling-explorations/p0tion
+[p0tion-config]: https://github.com/risc0/risc0/blob/d4e427283027c28b38b8eda1562e8e0e68d1b0e2/compact_proof/groth16/p0tionConfig.json
+[powers-of-tau-hez-23]: https://storage.googleapis.com/zkevm/ptau/powersOfTau28_hez_final_23.ptau
+[powers-of-tau-hez-23-our-mirror]: https://risc0-artifacts.s3.us-west-2.amazonaws.com/tsc/2024-04-04/powersOfTau28_hez_final_23.ptau
+[PSE]: https://pse.dev
+[pse-risc0-ceremony]: https://ceremony.pse.dev/projects/RISC%20Zero%20STARK-to-SNARK%20Prover
+[r1cs-file]: https://risc0-artifacts.s3.us-west-2.amazonaws.com/tsc/2024-04-04/stark_verify.r1cs
+[risc0-circom-library]: https://github.com/risc0/risc0/blob/d4e427283027c28b38b8eda1562e8e0e68d1b0e2/compact_proof/groth16/risc0.circom
+[snarkjs]: https://github.com/iden3/snarkjs
+[stark-verify-circom]: https://github.com/risc0/risc0/blob/d4e427283027c28b38b8eda1562e8e0e68d1b0e2/compact_proof/groth16/stark_verify.circom
```
pdg744, line 70: hmm... raising the question of whether this is the right place to point for this. 

do we have a more enshrined source of truth than `blob/d4e427283027c28b38b8eda1562e8e0e68d1b0e2`? 

if we have put this circom file into the amazon bucket, we might point there instead.

website/api/trusted-setup-ceremony.md
```diff
@@ -0,0 +1,70 @@
...
+1. Verifying the circuit we are securing is what we intend to secure and does not include security holes.
+2. Verifying the transcript corresponds to this circuit
+3. Verifying contributors' attestations match the transcript
+4. Verifying the setup ceremony does not include security holes
+
+## The Circuit
+
+The circuit we are securing is the RISC Zero STARK Verify circuit, which is open source and [available on GitHub][stark-verify-circom] (along with [a short library][risc0-circom-library] it depends on).
+
+To ensure that the circuit itself does not have security holes, we have used a mixture of good software engineering practices, internal security reviews, and external audits. Stay tuned for detailed reports of our audit results!
```
pdg744, line 20: > Stay tuned

mb we should add a TODO comment so we remember to add a pointer to the audit results?

website/api/trusted-setup-ceremony.md
```diff
@@ -0,0 +1,70 @@
...
+
+## The Transcript Matches the Circuit
+
+Our ceremony transcript is included in the `zkey` published on ceremony.pse.dev in the "Download ZKey" tab of the [RISC Zero STARK-to-SNARK Prover page][pse-risc0-ceremony]. (We mirror this file [here][TODO].) You can verify it matches the circuit using Circom and snarkjs:
```
pdg744, line 24: Is your thinking here to let this PR sit until we get this TODO resolved?

tzerrell: Hmm, good call out, we _can't_ resolve this until the ceremony is fully finalized and published to the PSE website and then mirrored by us. That would put a fairly long hold on this.

I think instead we should drop "(We mirror this file here.)" for now and restore the line once it's actually mirrored; I'll make that change

website/api/trusted-setup-ceremony.md
tzerrell: No, in that this is from after the v0.21 release and so is not enshrined any release. It's also not on AWS (since it's checked in to GitHub). I originally pointed at `main`, but that isn't guaranteed to be stable; I can forward you a thread discussing this more.

website/api/trusted-setup-ceremony.md
tzerrell: Hmm, on further thought comments in Markdown are a bit awkward (see https://stackoverflow.com/questions/4823468/comments-in-markdown) -- so instead I opened an issue: https://github.com/risc0/risc0/issues/1805

website/api/trusted-setup-ceremony.md
tzerrell: https://github.com/risc0/risc0/issues/1806

tzerrell:
I added a copy of this into v0.21, as I think it would be nice for this to be navigable in the default view.

pdg744 (approved):
Looks good!

PR #1810, opened 2024-05-10
feat(benchmarks): mobile fixes

nahoc:
customized footer ascii art

image
image

new 404 page

image


web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/applications-benchmarks/[slug]/_utils/constants.tsx
```diff
@@ -1,6 +1,8 @@
+import Link from "next/link";
+
 export const FILENAMES_TO_TITLES = {
   "macOS-apple_m2_pro.csv": "Metal on Apple M2 Pro",
-  "Linux-nvidia_rtx_a5000.csv": "CUDA on NVIDIA RTX A5000",
+  "Linux-nvidia_rtx_a5000.csv": "CUDA on NVIDIA RTX 3090 Ti",
```
Comment (flaub) [inline web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/applications-benchmarks/[slug]/_utils/constants.tsx:3]: I think this can't be changed, but we need a new entry in this table for the new `.csv` file that next is producing.

Comment (nahoc) [reply web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/applications-benchmarks/[slug]/_utils/constants.tsx:3]: I agree, but I currently don't see such file on `main` here:
https://github.com/risc0/ghpages/tree/main/dev/benchmarks

image

Comment (flaub) [reply web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/applications-benchmarks/[slug]/_utils/constants.tsx:3]: OK, this makes me think there's some issue with the results not getting published. This is where it gets run: https://github.com/risc0/risc0/blob/main/.github/workflows/bench_reports.yml#L65

Which should yield a file named `Linux-nvidia_rtx_3090_ti.csv`. Need to do a bit more digging but won't be able to do that until perhaps next week.

Comment (flaub) [reply web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/[version]/applications-benchmarks/[slug]/_utils/constants.tsx:3]: Oops, this might be the bug: https://github.com/risc0/risc0/blob/main/.github/workflows/bench_reports.yml#L74

Should be `.csv` instead of `.json`.

flaub (approved):

PR #1813, opened 2024-05-10
Polish UX for `rzup`

nuke-web3:
Suggestion for better output in `rzup`.


rzup/README.md
```diff
@@ -4,9 +4,8 @@ Install, update, or revert to a specific RISC Zero version.
 
 ## Install
 
-<!-- TODO: Replace this friendly redirect URL once set up -->
 ```sh
-curl -L https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/install | bash
+curl -L https://raw.githubusercontent.com/risc0/risc0/main/rzup/install | bash
```
nuke-web3, line 8: Not sure if there was a plan for some other link, but IMHO github should be used over aws.

flaub:
The issue here is that GitHub has IP rate limiting. The chance of this goes up a lot due to for example CI builds coming from a single IP (like in a cluster in the cloud).

flaub:
Ooops, just saw that there are more changes than just the URL, re-opening

rzup/README.md
flaub: We will likely setup a more permanent link, but it will be hosted in a bucket rather than on GitHub.

nuke-web3:
Ah so we do want to use some other URL, but need devops to configure a risczero.com or other redirect?

I am happy to include only the suggested UX polish in this PR if desired or close it.

rzup/install
```diff
@@ -1,16 +1,19 @@
...
+
+echo "💾 Installing rzup"
 
 BASE_DIR=${XDG_CONFIG_HOME:-$HOME}
 RISC0_DIR=${RISC0_DIR-"$BASE_DIR/.risc0"}
 RISC0_BIN_DIR="$RISC0_DIR/bin"
 
-BIN_URL="https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/rzup"
+# TODO: Replace this friendly redirect URL once set up
+BIN_URL="https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/install"
```
flaub, line 16: Note, this is the wrong URL
```suggestion
BIN_URL="https://risc0-artifacts.s3.us-west-2.amazonaws.com/rzup/rzup"
```

nuke-web3: Doh! Copy pasta fail

flaub (approved):

PR #1814, opened 2024-05-10
Warn that the ceremony is incomplete

tzerrell:
Our trusted setup ceremony is not finalized yet; add a warning so users know to expect that.


pdg744 (approved):
Good call

PR #1830, opened 2024-05-14
[dev docs] updating security model doc to reflect version management design

pdg744:


website/api/security-model.md
```diff
@@ -19,10 +19,11 @@ RISC Zero offers the following components, each of which is ready for use on tes
    Each recursion program is identified by a [control ID], and the full list of allowed programs is identified by the [control root].
 
 4. The **STARK-to-SNARK Prover**, which verifies proofs from the RISC Zero Recursion Prover, compressing the STARK into a Groth16 SNARK.
-   The [control root] is passed to as an input, allowing for updates to our RISC-V Prover without requiring a new trusted setup ceremony.
+   The [control root] is passed to as a public input, allowing for updates to our RISC-V Prover without requiring a new trusted setup ceremony.
 
 5. The **on-chain verifier contract**, which verifies proofs from the RISC Zero STARK-to-SNARK Prover.
-   Version information for the on-chain verifier is available in our [verifier contract] documentation.
+   The control root is hard-coded into the on-chain verifier contract.
+   Addresses for the on-chain verifier contracts we have deployed are available in our [verifier contract] documentation, and a detailed description of the designs for governance, upgrades, and deprecation are available in our [Version Management Design][VersionManagement@main] doc.
```
tzerrell, line 26: What do you think about changing "designs" to "options" to make it clearer that using our governance/upgrade/deprecation architecture is opt-in?

(I think the argument against would be that people might think we have more optionality than we actually do -- my understanding is the options are more or less just "let us manage which upgrades etc, or do it yourself")

```suggestion
   Addresses for the on-chain verifier contracts we have deployed are available in our [verifier contract] documentation, and a detailed description of the options for governance, upgrades, and deprecation are available in our [Version Management Design][VersionManagement@main] doc.
```

pdg744: no objections on my end -- will leave the comment open for now in case @nategraf has an opinion on this question.

tzerrell (approved):

PR #1832, opened 2024-05-14
Test PR - External Contributors Testing

RDaneelOlivaw0:


SchmErik:
Hi @RDaneelOlivaw0, what are you trying to test specifically?

flaub:
That 3rd-party contributors can run builds

.github/scripts/iam_risc0_cache.sh
```diff
@@ -1,2 +1,2 @@
-cp /test/certificate.pem $HOME/.local/aws/certificate.pem
-cp /test/runner_key.pem $HOME/.local/aws/runner_key.pem
+cp ../../../_aws/certificate.pem $HOME/.local/aws/certificate.pem
+cp ../../../_aws/runner_key.pem $HOME/.local/aws/runner_key.pem
```
flaub: I suggest using absolute paths

.github/actions/sccache/action.yml
```diff
@@ -89,31 +89,17 @@ runs:
...
-      shell: bash
-
     - name: (External PR) Install AWS Signer (macOS)
       if: >
         runner.os == 'macOS' &&
         github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name &&
         runner.environment == 'self-hosted'
       env:
         DARWIN_SIGNER_URL: https://rolesanywhere.amazonaws.com/releases/1.1.1/X86_64/Darwin/aws_signing_helper
+        IAM_RA_PATH: ../../_aws
```
flaub, line 99: I think we should avoid using relative paths. Let's just assume this tool is installed in a well known location.

PR #1841, opened 2024-05-16
Enable composition of rv32im programs and generalized recursion programs

nategraf:
As a major step towards composition with recursion programs and circuits implementing accelerators, this PR adds a `control_root` value to the receipt claim encoding of assumptions, and enables proving and of recursion programs that are not in the main set of rv32im transforms.

In the course of this PR, there are also a number of changes to make it possible to pass alternate parameters to proving and verification. In particular, `ProverOpts` now holds the set of allowed control roots and can be configured to use an alternate set, and `VerifierContext` holds the verifier parameters for each receipt type.

`SuccinctReceipt` has been made more generic, allowing it to represent proof of any recursion program that follows the convention of placing a control root in the first output slot, and a SHA-256 claim digest in the second output slot. This claim digest can be over a generic claim, which allows for claims representing the execution of e.g. accelerators which are not `ReceiptClaim` and do not represent zkVM execution.

Verification routines on `Receipt` and `CompositeReceipt` are simplified, partially taking advantage of recent changes to the rv32im circuit to zero-out the post-state digest, and partially by dropping `ExitCode::Paused(0)` as a default accepted exit code.


risc0/zkvm/src/receipt/merkle.rs
nategraf, line 1: This file was moved from `risc0/zkvm/src/host/recursion/merkle.rs` in or to make it possible to add `MerkleProof` as a (`pusb(crate)`) field of `SuccinctReceipt`.

risc0/zkp/src/core/hash/mod.rs
```diff
@@ -86,3 +86,16 @@ impl<F: Field> Clone for HashSuite<F> {
...
+pub fn hash_suite_from_name(name: impl AsRef<str>) -> Option<HashSuite<BabyBear>> {
+    match name.as_ref() {
+        "sha-256" => Some(sha::Sha256HashSuite::new_suite()),
+        "poseidon2" => Some(poseidon2::Poseidon2HashSuite::new_suite()),
+        "blake2b" => Some(blake2b::Blake2bCpuHashSuite::new_suite()),
+        #[cfg(feature = "prove")]
+        "poseidon_254" => Some(poseidon_254::Poseidon254HashSuite::new_suite()),
+        _ => None,
+    }
+}
```
flaub, line 101: Nice. I was also thinking of making the `hashfn` a `non_exhaustive` enum.

risc0/zkp/src/core/hash/mod.rs
```diff
@@ -86,3 +86,16 @@ impl<F: Field> Clone for HashSuite<F> {
         }
     }
 }
+
+/// Construct a supported hash function given it's name. Returns None is the name does not
```
flaub, line 90:
```suggestion
/// Construct a supported hash function given its name. Returns None is the name does not
```

nategraf:
Some build failures in `risc0-ethereum` job. These are expected, and I think its fine in this case to ignore them for the purposes of this PR and follow up with a PR there.
https://github.com/risc0/risc0/actions/runs/9186038065/job/25261017020?pr=1841

nategraf:
Some failures in `risc0-ethereum` tests. These are expected due to breaking API changes, and I think it is fine to ignore these for the purpose of merging this PR, the follow up with a PR to that repo. Once we hit 1.0 (very soon now!), that test should be more reliable.
https://github.com/risc0/risc0/actions/runs/9186038065/job/25261017020?pr=1841

SchmErik:
> Some failures in `risc0-ethereum` tests. These are expected due to breaking API changes, and I think it is fine to ignore these for the purpose of merging this PR, the follow up with a PR to that repo. Once we hit 1.0 (very soon now!), that test should be more reliable. https://github.com/risc0/risc0/actions/runs/9186038065/job/25261017020?pr=1841

Agreed! risc0-ethereum should pass after making the fix and should be more stable after 1.0

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -481,19 +557,87 @@ impl TryFrom<pb::core::InnerReceipt> for InnerReceipt {
...
-            },
+            pb::core::inner_receipt::Kind::Fake(inner) => Self::Fake(inner.try_into()?),
+        })
+    }
+}
+
+// NOTE: InnerReceipt and InnerAssumptionReceipt are the same type in protobuf.
+// In Rust, they are distinct types becaue Rust needs to size everything on the
+// stack and e.g. SuccinctReceipt<ReceiptClaim> and SuccinctReceipt<Unknown>
+// have different sizes. Protobuf handles this without issue.
```
flaub, line 568: I wonder if this will bite us in the end when we want to start making changes to one or the other. What's nice about having separate types in protobuf is that they can vary independently in a backwards-compatible way.

nategraf: Part of my assumption here is that you can "fork" the types later as needed without introducing a backwards incompatibility at the protobuf layer. I _think_ this is true, but I'm not 100% sure.

I'd be happy to split this out now

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -674,11 +670,31 @@ mod sys_verify {
...
+        assert_eq!(
+            halt_claim.as_value().unwrap().exit_code,
+            ExitCode::Halted(exit_code as u32)
+        );
+
         halt_receipt
     }
 
+    // The test_recursion_circuit is a program for the recursion VM that does some very simple
+    // operations, just to make sure the
```
flaub, line 686: Leaving me hanging :)

risc0/zkvm/src/receipt/succinct.rs
```diff
@@ -54,21 +51,30 @@ pub struct SuccinctReceipt {
...
-            proof_system_info: PROOF_SYSTEM_INFO,
-            circuit_info: risc0_circuit_recursion::CircuitImpl::CIRCUIT_INFO,
-        }
-    }
+    /// Name of the hash function used to create this receipt.
+    pub hashfn: String,
 
+    /// A digest of the verifier parameters that can be used to verify this receipt.
+    ///
+    /// Acts as a fingerprint to identity differing proof system or circuit versions between a
```
flaub, line 65:
```suggestion
    /// Acts as a fingerprint to identify differing proof system or circuit versions between a
```

risc0/zkvm/src/receipt/succinct.rs
```diff
@@ -54,21 +51,30 @@ pub struct SuccinctReceipt {
...
+    /// prover and a verifier. Is not intended to contain the full verifier parameters, which must
```
flaub, line 66:
```suggestion
    /// prover and a verifier. It ss not intended to contain the full verifier parameters, which must
```

risc0/zkvm/src/host/server/prove/tests.rs
nategraf: Oops 😆 
53bd1a3d6e1adbe966d96e6ff610e1ab7f7f1741

risc0/zkvm/src/host/api/convert.rs
flaub: Ah, if we can fork later in a compatible way, then this seems fine to me.

flaub (approved):
Fantastic!

PR #1842, opened 2024-05-16
Add New Benchmarks

morgangibbs87:
- [ ] g4dn.xlarge
- [ ] L4 (g6.xlarge like Bonsai)
- [ ] c7i.48xlarge to `nightly`?

### DIFF
--- a/.github/workflows/bench_pr.yml
+++ b/.github/workflows/bench_pr.yml
@@ -36,7 +36,10 @@ jobs:
         include:
           - os: Linux
             feature: cuda
-            device: [nvidia_rtx_3090_ti,g4dn.xlarge]
+            device: nvidia_rtx_3090_ti
+          - os: Linux
+            feature: cuda
+            device: g4dn.xlarge
           - os: macOS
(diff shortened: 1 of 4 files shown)


.github/workflows/bench_pr.yml
```diff
@@ -30,16 +30,23 @@ jobs:
...
           - os: Linux
             feature: cuda
-            device: nvidia_rtx_3090_ti
-          - os: macOS
-            feature: metal
-            device: apple_m2_pro
+            device: g4dn.xlarge
+          - os: Linux
+            feature: cuda
+            device: nvidia_rtx_l4
```
flaub, line 46: this should be named the AWS instance type

flaub:
I don't think we need `c7i.48xlarge`. Also we should rename L4 to the AWS instance type = `g6.xlarge`

flaub (changes requested):
Holding off for now

morgangibbs87:
Closing this PR to split into separate tests.

PR #1843, opened 2024-05-16
Update security model diagram

tzerrell:
Moves ZKVM to where Bonsai was (and drops its old box), changes `stark2snark` to `stark2snark Prover`.


pdg744 (approved):
Thank you!

PR #1844, opened 2024-05-16
Use recent trusted setup ceremony for stark2snark

tzerrell:
Update our stark2snark groth16 code to be based on the recent trusted setup ceremony.


risc0/groth16/src/docker.rs
```diff
@@ -51,7 +51,7 @@ pub fn stark_to_snark(identity_p254_seal_bytes: &[u8]) -> Result<Seal> {
         .arg("--rm")
         .arg("-v")
         .arg(&format!("{}:/mnt", work_dir.to_string_lossy()))
-        .arg("risczero/risc0-groth16-prover:v2024-04-03.2")
+        .arg("risczero/risc0-groth16-prover:v2024-05-16")
```
flaub, line 54: Generally we should add a suffix in case we need to create another image in the same day. (and so it's consistent)

tzerrell:
The test failures on the most recent CI round are all risc0-ethereum errors, which we expect since this is a breaking change that will require a risc0-ethereum update. See also https://github.com/risc0/risc0-ethereum/pull/118

flaub (approved):

PR #1851, opened 2024-05-17
cargo-risczero build-toolchain: add optional `--version` flag

SchmErik:
This new --version flag behaves similar to `cargo risczero install --version`.


flaub (approved):

risc0/cargo-risczero/src/commands/build_toolchain.rs
```diff
@@ -52,9 +56,13 @@ impl BuildToolchain {
         };
         let rust_dir = root_dir.join("rust");
 
+        let tag = match &self.version {
+            Some(tag) => tag,
+            None => RUST_BRANCH,
+        };
```
flaub, line 62: BTW, you can set a default value in the `#[arg]` attribute above.

Issue #1840, opened 2024-05-15
[BUG] cargo version > 1.76 results in build errors when risc0 deps are specified as git dependencies in `Cargo.toml`


SchmErik:
## Bug Report
When using `risc0` crates as git dependencies (meaning `git =` is used in the Cargo.toml instead of `version =`), compiling `risc0-zkvm` crate results in the following error when running `cargo build` on a risc zero project:
```
error: invalid inline table
expected `}`
 --> ../../../../../../../home/ubuntu/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/ecb7a53/templates/rust-starter/host/Cargo.toml:8:16
  |
8 | risc0-zkvm = { {{ risc0_zkvm }} }
  |                ^
  |
error: invalid inline table
expected `}`
 --> ../../../../../../../home/ubuntu/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/ecb7a53/templates/rust-starter/methods/Cargo.toml:7:17
  |
7 | risc0-build = { {{ risc0_build }} }
  |                 ^
  |
error: invalid inline table
expected `}`
 --> ../../../../../../../home/ubuntu/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/ecb7a53/templates/rust-starter/methods/guest/Cargo.toml:2:9
  |
2 | name = {{ guest_package_name }}
  |         ^
  |
```
What appears to be happening is that the risc0 rust-starter template is getting compiled. I found that removing the templates directory from the dependency makes this error message disappear.

A few solutions off the top of my head:
1) get rid of cargo-generate
2) move the template to a separate repo
3) figure out a way to suppress compilation of the template directory. I tried to exclude this in the workspace's Cargo.toml but this had no effect.

## Steps to Reproduce
1. create a project, use `git =` for risc0 dependencies
2. run cargo build. Be sure to run `cargo update` to ensure that the deps are updated (helpful in certain situations where cargo does not pick up the newest changes pushed to a repo).
3. error is displayed

## Expected behavior
Error should not be displayed.

## Your Environment
- risc0-zkvm version: 1.0.0-rc.4
- Rust version: host side = 1.77, 1.78
- Platform/OS: mac and linux

PR #1867, opened 2024-05-21
cargo-risczero: drop cargo-generate from new command

SchmErik:
The `cargo-generate` crate is designed with the assumption that templates to reside in a small repository. Our template is a part of a large repository and takes anywhere between 3-10 seconds to run `cargo risczero new` because it has to clone the repo. This change drops cargo-generate and instead uses regex patterns to search and replace the template during generation. Locally, this takes about 0.5 seconds to run. I've dropped several options that seemed irrelevant. Another notable change is the renaming of the template's `Cargo.toml` to `Cargo-toml`. By doing so, I have fixed https://github.com/risc0/risc0/issues/1840.

Fixes: #1840


flaub (approved):
Yay!

PR #1869, opened 2024-05-21
Add Working External PR Changes + Test

morgangibbs87:


.github/scripts/iam_risc0_cache.sh
```diff
@@ -0,0 +1,2 @@
+cp ../../../../_aws/certificate.pem $HOME/.local/aws/certificate.pem
+cp ../../../../_aws/runner_key.pem $HOME/.local/aws/runner_key.pem
```
flaub, line 2: Can we move this to ansible/host-config rather than being in a github action?

.github/actions/sccache/action.yml
```diff
@@ -87,3 +88,50 @@ runs:
...
+      run: |
+        mkdir -p $HOME/.local/aws
+        curl $DARWIN_SIGNER_URL -o $HOME/.local/aws/aws_signing_helper
+        chmod +x $HOME/.local/aws/aws_signing_helper
+      shell: bash
+
+    - name: (External PR) Install IAM Anywhere Resources for risc0 Public Cache
+      if: >
+        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name &&
+        (contains(runner.name, 'puget') || contains(runner.name, 'mini-'))
```
flaub, line 121: Hmm, I wonder if we could use a different tag/label for this detection rather than the runner name

morgangibbs87: I made an improvement, but on-prem Linux runners will still require a reference to `runner.name`, at least for now.

- `Benchmark Check` jobs need a way to distinguish between:
  - `self-hosted` runners from AWS (names begin with `i-`, IAM Anywhere setup steps not required) 
  - `self-hosted` on-prem runners (names vary, IAM Anywhere setup steps required)
- If this is not done, jobs running on AWS runners look for the IAM Anywhere stuff, which isn't loaded on there and is not required for the jobs to work.

-----

Fix so on-prem Linux self-hosted runners will pick up the IAM Anywhere jobs, but not AWS ones:
```
if: >
  runner.os == 'Linux' &&
  github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name &&
  (runner.environment == 'self-hosted' && (contains(runner.name, 'i-') == 'false'))
```

Fix for on-prem macOS self-hosted runners, which do not need `contains(runner.name,'foo')` since we do not use macOS runners in AWS that could get picked up:
```
if: >
  runner.os == 'macOS' &&
  github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name &&
  runner.environment == 'self-hosted'
```

-----

**More info:**
- GitHub [Actions Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context) do not currently offer a `runner.label` property to extract and process this through conditions.
- Others have this issue ([example](https://github.com/actions/runner/issues/821)) and have proposed adding this, but nothing yet. 
- There is a [third-party action](https://github.com/marketplace/actions/runner-label-resolver) that will do this, but it requires a Token with permissions at the Org level to pull the labels.

.github/scripts/iam_risc0_cache.sh
morgangibbs87: Absolutely. The files are on the runners already, so this script is actually no longer necessary and was removed.

Instead, the files are now just referenced in relative a path when used to set credentials:
https://github.com/risc0/risc0/blob/morgan/external-prs/.github/actions/sccache/action.yml#L127-129

```
env:
  ...
  IAM_RA_PATH: ../../_aws
run: |
  aws configure set credential_process "$HOME/.local/aws/aws_signing_helper credential-process --trust-anchor-arn $TRUST_ANCHOR_ARN --profile-arn $PROFILE_ARN --role-arn $ROLE_ARN --certificate $IAM_RA_PATH/certificate.pem --private-key $IAM_RA_PATH/runner_key.pem" --profile default
```

morgangibbs87:
@flaub - For reference, here is the external PR showing the recent fixes worked:
https://github.com/risc0/risc0/pull/1832

.github/scripts/iam_risc0_cache.sh
morgangibbs87: https://github.com/risc0/devops/pull/225

.github/actions/sccache/action.yml
```diff
@@ -87,3 +88,45 @@ runs:
...
         use-fallback: false
+
+    - name: (External PR) Install AWS Signer (Linux)
+      if: >
+        runner.os == 'Linux' &&
+        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name &&
+        runner.environment == 'self-hosted' &&
+        startsWith(runner.name, 'i-') != true
+      env:
+        LINUX_SIGNER_URL: https://rolesanywhere.amazonaws.com/releases/1.1.1/X86_64/Linux/aws_signing_helper
```
flaub, line 99: If this is happening on machines managed by ansible, why not install it there?

.github/actions/sccache/action.yml
```diff
@@ -87,3 +88,45 @@ runs:
...
+        chmod +x $HOME/.local/aws/aws_signing_helper
+      shell: bash
+
+    - name: (External PR) Install AWS Signer (macOS)
+      if: >
+        runner.os == 'macOS' &&
+        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name &&
+        runner.environment == 'self-hosted'
+      env:
+        DARWIN_SIGNER_URL: https://rolesanywhere.amazonaws.com/releases/1.1.1/X86_64/Darwin/aws_signing_helper
```
flaub, line 112: Do we really want the x86_64 binary?

PR #1878, opened 2024-05-23
CI: re-enable cargo risczero test

SchmErik:
There seems to be an intermittent error for this job. Run it in CI but do not require it for pull requests to land.


SchmErik:
Looks like it's failing now... I'm going to close this so that I can get to the bottom of this without triggering ci for each small change...

PR #1880, opened 2024-05-23
Updates to documentation from a review pass on May 23

nategraf:
A number of smaller docs updates from a quick pass over the documentation.

Largest change is to include the `risc-groth16` README in the crate documentation, thus bringing it under test in CI, and to drop the verbose example of STARK to SNARK proving, which can now be accomplished through the main APIs.

### DIFF
--- a/risc0/zkvm/src/lib.rs
+++ b/risc0/zkvm/src/lib.rs
@@ -54,7 +54,6 @@
 //! | client           | all except rv32im | std        | Enables the client API.                                                                                                                                      |
 //! | cuda             |                   | prove, std | Enables CUDA GPU acceleration for the prover. Requires CUDA toolkit to be installed.                                                                         |
 //! | disable-dev-mode | all except rv32im |            | Disables dev mode so that proving and verifying may not be faked. Used to prevent a misplaced `RISC0_DEV_MODE` from breaking security in production systems. |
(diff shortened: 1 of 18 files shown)


risc0/zkvm/src/lib.rs
```diff
@@ -47,13 +47,14 @@
...
 //! | ---------------- | ----------------- | ---------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
 //! | client           | all except rv32im | std        | Enables the client API.                                                                                                                                      |
 //! | cuda             |                   | prove, std | Enables CUDA GPU acceleration for the prover. Requires CUDA toolkit to be installed.                                                                         |
 //! | disable-dev-mode | all except rv32im |            | Disables dev mode so that proving and verifying may not be faked. Used to prevent a misplaced `RISC0_DEV_MODE` from breaking security in production systems. |
+//! | docker           | all except rv32im |            | Enables features that require Docker. In particular, this enables Groth16 proving.                                                                           |
```
flaub, line 57: I think we should not require the docker flag to enable Groth16 proving, we should make this is a runtime issue, rather than a compile time one. Note that we should still use the docker flag for tests.

nategraf: Ah, and this is indeed how it is implemented. I've removed this line in cdcab5d62d79aec9f78a4294c8e53ccafaff988c

flaub (approved):

PR #1891, opened 2024-05-27
fix: remove repetitive words in comments

polymaer:
removed the duplicated words in the comments to improve readability. no production code was touched.


risc0/zkvm/src/host/recursion/prove/preflight.rs
```diff
@@ -194,7 +194,7 @@ impl<'a, Ext: Externs> Preflight<'a, Ext> {
 
             let result;
             if u32arg[2] != 0 {
-                // AND and combine [a, b, 0, 0] & [c, d, 0, 0] -> [(a&c) + ((b&d) << 16), 0, 0,
+                // AND combine [a, b, 0, 0] & [c, d, 0, 0] -> [(a&c) + ((b&d) << 16), 0, 0,
```
SchmErik, line 197: I think this is intended to mean "do a bitwise AND operation and combine" so I would revert this hunk

polymaer: yeah, that makes sense. i must have misunderstood the meaning. the change has been reverted. kindly give another review.

flaub (approved):

PR #1898, opened 2024-05-28
C guest example

austinabell:
Basic template, adapted slightly to fit into the examples workspace


examples/c-guest/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+# Regenerate the header file. Assumes cbindgen installed locally.
+.PHONY: headers
+headers:
+	cbindgen --crate zkvm-platform -c ./platform/cbindgen.toml -o ./guest/platform.h
+
+.PHONY: gcc
+gcc:
+	@if ! [ -d "./riscv32im-${MACHINE}" ]; then \
+        curl -L https://github.com/risc0/toolchain/releases/download/2024.01.05/riscv32im-${MACHINE}.tar.xz | tar xvJ -C ./; \
```
austinabell, line 37: @SchmErik I'd love to use the installed c toolchain added recently, but unclear if that path is stable enough and this logic (https://github.com/risc0/risc0/blob/618fdc2747f05ae8d1d7c3db178a6e184812a152/risc0/build/src/lib.rs#L44) is a little hacky to translate to a makefile.

What do you recommend here? I could also have the compilation done in a build script, so that I can just call this fn before building, but it seems a bit wasteful.

examples/c-guest/platform/src/lib.rs
flaub, line 1: Seems like this should go into `risc0-zkvm-platform` itself. Can do that later of course.

examples/c-guest/Makefile
SchmErik: Maybe we could have `cargo risczero` emit information about where the c-toolchain is installed?

examples/c-guest/.gitignore
```diff
@@ -0,0 +1,4 @@
+.DS_Store
+target/
+riscv32im*/
+Cargo.lock
```
flaub, line 4: Should we include the `out` directory here?

examples/c-guest/guest/.gitignore
```diff
@@ -0,0 +1 @@
+out/
```
flaub, line 1: Oh, surprised to see this in another `.gitignore `file. I guess either will work

examples/c-guest/guest/main.c
```diff
@@ -0,0 +1,34 @@
...
+	// See: https://gnu-mcu-eclipse.github.io/arch/riscv/programmer/#the-gp-global-pointer-register
+	// Linker relaxations must be disabled to avoid the initialization beign
+	// relaxed with an uninitialized global pointer: mv gp, gp
+	".option push;\n\t"
+	".option norelax\n\t"
+	"la gp,      __global_pointer$\n\t"
+	".option pop\n\t"
+
+	// Jump to main function
+	"call main\n\t");
```
flaub, line 21: Is there anyway to put this in `platform`?

examples/c-guest/host/src/main.rs
```diff
@@ -0,0 +1,26 @@
+use risc0_zkvm::{compute_image_id, default_prover, ExecutorEnv};
+use std::fs;
+
+fn main() -> anyhow::Result<()> {
+    // Load built gcc program and compute it's image ID.
+    let consensus_elf = fs::read("./guest/out/main")?;
+    let consensus_id = compute_image_id(&consensus_elf)?;
```
flaub, line 7: This seems like a dangerous example to set. Ideally the image_id is computed at build time.

examples/c-guest/Makefile
```diff
@@ -0,0 +1,52 @@
...
+    ifneq ($(filter arm%,$(UNAME_P)),)
+        MACHINE := $(MACHINE)-arm64
+    endif
+endif
+
+default: execute
+
+.PHONY: platform
+platform:
+	cargo +risc0 rustc -p zkvm-platform --target riscv32im-risc0-zkvm-elf --lib --crate-type staticlib --release --target-dir ./guest/out/platform
```
SchmErik, line 27: I'm mostly writing this so I remember this in the future but I think this could be something we distribute on our end and it's something that can be downloaded by a command like `cargo risczero install`

examples/c-guest/Makefile
flaub: I feel like we should arrange for the env vars to be configured via some sort of `rzup` command. That way paths don't need to be specified in the makefile, we make it an environmental issue. I think we should probably drop the `curl` and just rely on the `cargo risczero install` command to get the g++ toolchain installed somewhere.

examples/c-guest/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+> Warning: This example is experimental, and has not been extensively tested
+
+This example is a template of a basic program written in C.
+
+## Quick Start
+
+First, make sure [rustup] is installed.
+
+Then, install the [Risc Zero toolchain](https://dev.risczero.com/api/zkvm/install#install).
```
flaub, line 11:
```suggestion
Then, install the [Risc Zero toolchain](https://dev.risczero.com/api/zkvm/install).
```

examples/c-guest/platform/src/lib.rs
austinabell: yeah, I'm just not sure of what this API would look like, and the types are primarily based around C compatibility. 👍 to do later though

examples/c-guest/guest/.gitignore
austinabell: yeah maybe just a habit, but prefer to have dir specific ignores to be more resilient to refactors 🤷 Happy to switch to just using one though, if you prefer!

examples/c-guest/guest/main.c
austinabell: hmm, good thought. Let me try, I assumed not, but not for a good reason

examples/c-guest/host/src/main.rs
austinabell: agree, but very bloated to do at compile time. Would a lazy static be sufficient here? Or should I resolve it through a build script/proc macro?

examples/c-guest/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+# Regenerate the header file. Assumes cbindgen installed locally.
+.PHONY: headers
+headers:
+    # Note: this is currently broken on main, as there is a FFI incompatible type
```
austinabell, line 32: Noticed on main that this no longer works (something in this crate is not FFI compatible since 1.0.0-rc.5.

Not a big deal and this obviously isn't a constraint that previously mattered, just pointing out :)

examples/c-guest/host/src/main.rs
austinabell: But I guess practically for this example, should this happen in a build script that builds some minimal crate with those types or in a proc macro?

lazy_static now that I think about it wouldn't alleviate your concern, as it's still being retrieved at runtime (once).

flaub: We can iterate on this, it doesn't need to block landing it. But I do think we need to work on how do we generate the image_id at build time generally. I think running it from the Makefile would make sense to me.

austinabell: Added it as a TODO here, to at least mark it if anyone building off it before we update 👍

austinabell:
Bit of a tangent, but was curious about someone asking about cpp and these hacky changes _seem_ to suggest that works fine with this setup also:

```diff
diff --git a/examples/Cargo.lock b/examples/Cargo.lock
index 46e7e491..6951aa0b 100644
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -488,7 +488,7 @@ version = "0.1.0"
 dependencies = [
  "bevy-core",
  "bevy-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -612,7 +612,19 @@ name = "bonsai-sdk"
 version = "0.8.0-rc.2"
 dependencies = [
  "reqwest 0.12.4",
- "risc0-groth16",
+ "risc0-groth16 1.0.0-rc.5",
+ "serde",
+ "thiserror",
+]
+
+[[package]]
+name = "bonsai-sdk"
+version = "0.8.0-rc.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d8f46b40f7039b0bd2f568e18fad3f63d3388d343c4877147c35af610d67c337"
+dependencies = [
+ "reqwest 0.12.4",
+ "risc0-groth16 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde",
  "thiserror",
 ]
@@ -646,7 +658,7 @@ name = "browser-verify"
 version = "0.1.0"
 dependencies = [
  "bincode",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "risc0-zkvm-receipts",
  "wasm-bindgen-test",
 ]
@@ -734,7 +746,7 @@ name = "c-guest-host"
 version = "0.1.0"
 dependencies = [
  "anyhow",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde",
  "tracing-subscriber 0.3.18",
 ]
@@ -825,7 +837,7 @@ dependencies = [
  "chess-core",
  "chess-methods",
  "clap",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "shakmaty",
 ]
 
@@ -951,7 +963,7 @@ version = "0.1.0"
 dependencies = [
  "composition-example-methods",
  "hello-world",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -1260,7 +1272,7 @@ dependencies = [
  "prorata-methods",
  "rand",
  "rand_core",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "rmp-serde",
  "rust_decimal",
  "serde",
@@ -1366,7 +1378,7 @@ dependencies = [
  "clap",
  "digital-signature-core",
  "digital-signature-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "sha2",
  "test-log",
  "tracing",
@@ -1377,7 +1389,7 @@ dependencies = [
 name = "digital-signature-core"
 version = "0.1.0"
 dependencies = [
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "serde",
 ]
 
@@ -1483,7 +1495,7 @@ dependencies = [
  "ecdsa-methods",
  "k256",
  "rand_core",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -2176,7 +2188,7 @@ name = "hello-world"
 version = "0.1.0"
 dependencies = [
  "hello-world-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -2568,7 +2580,7 @@ dependencies = [
 name = "json-core"
 version = "0.1.0"
 dependencies = [
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "serde",
 ]
 
@@ -2578,7 +2590,7 @@ version = "0.1.0"
 dependencies = [
  "json-core",
  "json-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -2646,7 +2658,7 @@ version = "0.1.0"
 dependencies = [
  "jwt-core",
  "jwt-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -3208,7 +3220,7 @@ dependencies = [
  "password-checker-core",
  "password-checker-methods",
  "rand",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
 ]
 
 [[package]]
@@ -3411,7 +3423,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "fibonacci-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "tracing-subscriber 0.3.18",
 ]
 
@@ -3439,7 +3451,7 @@ dependencies = [
  "clap",
  "prorata-core",
  "prorata-methods",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "rust_decimal",
 ]
 
@@ -3879,8 +3891,22 @@ version = "1.0.0-rc.5"
 dependencies = [
  "anyhow",
  "elf",
- "risc0-zkp",
- "risc0-zkvm-platform",
+ "risc0-zkp 1.0.0-rc.5",
+ "risc0-zkvm-platform 1.0.0-rc.5",
+ "serde",
+ "tracing",
+]
+
+[[package]]
+name = "risc0-binfmt"
+version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "174598af56aad42537c50d05c226cc7e346f8fda5f1fbe39ded52836fed3753e"
+dependencies = [
+ "anyhow",
+ "elf",
+ "risc0-zkp 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkvm-platform 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde",
  "tracing",
 ]
@@ -3893,9 +3919,9 @@ dependencies = [
  "cargo_metadata",
  "dirs",
  "docker-generate",
- "risc0-binfmt",
- "risc0-zkp",
- "risc0-zkvm-platform",
+ "risc0-binfmt 1.0.0-rc.5",
+ "risc0-zkp 1.0.0-rc.5",
+ "risc0-zkvm-platform 1.0.0-rc.5",
  "serde",
  "serde_json",
  "tempfile",
@@ -3929,21 +3955,35 @@ dependencies = [
  "rand",
  "rayon",
  "risc0-circuit-recursion-sys",
- "risc0-core",
+ "risc0-core 1.0.0-rc.5",
  "risc0-sys",
- "risc0-zkp",
+ "risc0-zkp 1.0.0-rc.5",
  "sha2",
  "tracing",
  "zip",
 ]
 
+[[package]]
+name = "risc0-circuit-recursion"
+version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4bf5ca899bb3650a9eed60450795a749925b5eb09a29e8a7fb8ef3122034ac6"
+dependencies = [
+ "anyhow",
+ "bytemuck",
+ "hex",
+ "risc0-core 1.0.0-rc.6",
+ "risc0-zkp 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tracing",
+]
+
 [[package]]
 name = "risc0-circuit-recursion-sys"
 version = "1.0.0-rc.5"
 dependencies = [
  "glob",
  "risc0-build-kernel",
- "risc0-core",
+ "risc0-core 1.0.0-rc.5",
  "risc0-sys",
  "sppark",
 ]
@@ -3964,24 +4004,39 @@ dependencies = [
  "nvtx",
  "rand",
  "rayon",
- "risc0-binfmt",
+ "risc0-binfmt 1.0.0-rc.5",
  "risc0-circuit-rv32im-sys",
- "risc0-core",
+ "risc0-core 1.0.0-rc.5",
  "risc0-sys",
- "risc0-zkp",
- "risc0-zkvm-platform",
+ "risc0-zkp 1.0.0-rc.5",
+ "risc0-zkvm-platform 1.0.0-rc.5",
  "serde",
  "sha2",
  "tracing",
 ]
 
+[[package]]
+name = "risc0-circuit-rv32im"
+version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ec21b15f8dcccf0826e0ea5a6efbb572b1c324d24c7c011d5b1ebf7201b47dbb"
+dependencies = [
+ "anyhow",
+ "risc0-binfmt 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-core 1.0.0-rc.6",
+ "risc0-zkp 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkvm-platform 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde",
+ "tracing",
+]
+
 [[package]]
 name = "risc0-circuit-rv32im-sys"
 version = "1.0.0-rc.5"
 dependencies = [
  "glob",
  "risc0-build-kernel",
- "risc0-core",
+ "risc0-core 1.0.0-rc.5",
  "risc0-sys",
  "sppark",
 ]
@@ -3994,6 +4049,16 @@ dependencies = [
  "rand_core",
 ]
 
+[[package]]
+name = "risc0-core"
+version = "1.0.0-rc.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c920fcc8ac6afec4a85310f9da74e681af885433fc19ae92e2737fd98b5cebac"
+dependencies = [
+ "bytemuck",
+ "rand_core",
+]
+
 [[package]]
 name = "risc0-groth16"
 version = "1.0.0-rc.5"
@@ -4007,22 +4072,42 @@ dependencies = [
  "hex",
  "num-bigint 0.4.5",
  "num-traits",
- "risc0-binfmt",
- "risc0-core",
- "risc0-zkp",
+ "risc0-binfmt 1.0.0-rc.5",
+ "risc0-core 1.0.0-rc.5",
+ "risc0-zkp 1.0.0-rc.5",
  "serde",
  "serde_json",
  "tempfile",
  "tracing",
 ]
 
+[[package]]
+name = "risc0-groth16"
+version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "313483931948778b91ea105c115ccc833f4b058649fb13f1b36f99325e04f0e8"
+dependencies = [
+ "anyhow",
+ "ark-bn254",
+ "ark-ec",
+ "ark-groth16",
+ "ark-serialize",
+ "bytemuck",
+ "hex",
+ "num-bigint 0.4.5",
+ "num-traits",
+ "risc0-binfmt 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkp 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde",
+]
+
 [[package]]
 name = "risc0-groth16-verifier-example"
 version = "0.1.0"
 dependencies = [
  "groth16-verifier-methods",
- "risc0-groth16",
- "risc0-zkvm",
+ "risc0-groth16 1.0.0-rc.5",
+ "risc0-zkvm 1.0.0-rc.5",
  "serde_json",
 ]
 
@@ -4057,9 +4142,30 @@ dependencies = [
  "rand",
  "rand_core",
  "rayon",
- "risc0-core",
+ "risc0-core 1.0.0-rc.5",
  "risc0-sys",
- "risc0-zkvm-platform",
+ "risc0-zkvm-platform 1.0.0-rc.5",
+ "serde",
+ "sha2",
+ "tracing",
+]
+
+[[package]]
+name = "risc0-zkp"
+version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "488a3d73143affc2ccae0e2619790c6ab44f794591923b5f0e7df1ede09c8509"
+dependencies = [
+ "anyhow",
+ "blake2",
+ "bytemuck",
+ "cfg-if",
+ "digest",
+ "hex",
+ "paste",
+ "rand_core",
+ "risc0-core 1.0.0-rc.6",
+ "risc0-zkvm-platform 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde",
  "sha2",
  "tracing",
@@ -4072,7 +4178,7 @@ dependencies = [
  "addr2line 0.22.0",
  "anyhow",
  "bincode",
- "bonsai-sdk",
+ "bonsai-sdk 0.8.0-rc.2",
  "bytemuck",
  "bytes",
  "cfg-if",
@@ -4084,13 +4190,13 @@ dependencies = [
  "prost",
  "rand",
  "rayon",
- "risc0-binfmt",
- "risc0-circuit-recursion",
- "risc0-circuit-rv32im",
- "risc0-core",
- "risc0-groth16",
- "risc0-zkp",
- "risc0-zkvm-platform",
+ "risc0-binfmt 1.0.0-rc.5",
+ "risc0-circuit-recursion 1.0.0-rc.5",
+ "risc0-circuit-rv32im 1.0.0-rc.5",
+ "risc0-core 1.0.0-rc.5",
+ "risc0-groth16 1.0.0-rc.5",
+ "risc0-zkp 1.0.0-rc.5",
+ "risc0-zkvm-platform 1.0.0-rc.5",
  "rrs-lib",
  "rustc-demangle",
  "semver",
@@ -4101,9 +4207,50 @@ dependencies = [
  "typetag",
 ]
 
+[[package]]
+name = "risc0-zkvm"
+version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "749eaa5169256cabc910e2143d2fcaf25c465dc9e348fd7e01b696f02d19d957"
+dependencies = [
+ "anyhow",
+ "bincode",
+ "bonsai-sdk 0.8.0-rc.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "bytemuck",
+ "bytes",
+ "cfg-if",
+ "getrandom",
+ "hex",
+ "prost",
+ "risc0-binfmt 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-circuit-recursion 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-circuit-rv32im 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-core 1.0.0-rc.6",
+ "risc0-groth16 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkp 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkvm-platform 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rrs-lib",
+ "semver",
+ "serde",
+ "sha2",
+ "tempfile",
+ "tracing",
+]
+
+[[package]]
+name = "risc0-zkvm-platform"
+version = "1.0.0-rc.5"
+dependencies = [
+ "bytemuck",
+ "getrandom",
+ "libm",
+]
+
 [[package]]
 name = "risc0-zkvm-platform"
 version = "1.0.0-rc.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "20e006ec91eecbff9c0849a9028c5558495292d333d3a0841e35db0f77baf3fb"
 dependencies = [
  "bytemuck",
  "getrandom",
@@ -4605,7 +4752,7 @@ version = "0.12.0"
 dependencies = [
  "clap",
  "hex",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "sha-methods",
 ]
 
@@ -4786,7 +4933,7 @@ dependencies = [
 name = "smartcore-ml"
 version = "0.1.0"
 dependencies = [
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "rmp-serde",
  "smartcore 0.3.2 (git+https://github.com/risc0/smartcore.git?rev=4bd3cadd50ed988c45c239f5264c3e2c2af0a690)",
  "smartcore-ml-methods",
@@ -5549,7 +5696,7 @@ name = "voting-machine"
 version = "0.12.0"
 dependencies = [
  "hex",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "test-log",
  "tracing",
  "tracing-subscriber 0.3.18",
@@ -5561,7 +5708,7 @@ dependencies = [
 name = "voting-machine-core"
 version = "0.12.0"
 dependencies = [
- "risc0-zkp",
+ "risc0-zkp 1.0.0-rc.5",
  "serde",
 ]
 
@@ -5579,7 +5726,7 @@ dependencies = [
  "bincode",
  "clap",
  "image",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "tracing-subscriber 0.3.18",
  "viuer",
  "waldo-core",
@@ -5598,7 +5745,7 @@ dependencies = [
  "merkle_light",
  "merkle_light_derive",
  "rand",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "serde",
  "sha2",
 ]
@@ -5629,7 +5776,7 @@ checksum = "9c8d87e72b64a3b4db28d11ce29237c246188f4f51057d65a7eab63b7987e423"
 name = "wasm"
 version = "0.1.0"
 dependencies = [
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "wasm-methods",
  "wat",
 ]
@@ -6036,7 +6183,7 @@ name = "wordle"
 version = "0.1.0"
 dependencies = [
  "rand",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "serial_test",
  "wordle-core",
  "wordle-methods",
@@ -6046,7 +6193,7 @@ dependencies = [
 name = "wordle-core"
 version = "1.0.0"
 dependencies = [
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "serde",
 ]
 
@@ -6090,7 +6237,7 @@ name = "xgboost-example"
 version = "0.1.0"
 dependencies = [
  "forust-ml",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "rmp-serde",
  "serde_json",
  "xgboost-methods",
@@ -6178,7 +6325,7 @@ dependencies = [
  "clap",
  "ethers-core",
  "ethers-providers",
- "risc0-zkvm",
+ "risc0-zkvm 1.0.0-rc.5",
  "tokio",
  "tracing",
  "tracing-subscriber 0.3.18",
@@ -6197,9 +6344,9 @@ dependencies = [
 name = "zkvm-platform"
 version = "0.1.0"
 dependencies = [
- "risc0-binfmt",
- "risc0-zkp",
- "risc0-zkvm-platform",
+ "risc0-binfmt 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkp 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "risc0-zkvm-platform 1.0.0-rc.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
diff --git a/examples/c-guest/Makefile b/examples/c-guest/Makefile
index e99f0152..b1b7e4c6 100644
--- a/examples/c-guest/Makefile
+++ b/examples/c-guest/Makefile
@@ -42,7 +42,7 @@ gcc:
 	 
 .PHONY: guest
 guest: gcc platform
-	${ROOT_DIR}riscv32im-${MACHINE}/bin/riscv32-unknown-elf-gcc -nostartfiles ./guest/main.c -o ./guest/out/main -L${ROOT_DIR}guest/out/platform/riscv32im-risc0-zkvm-elf/release -lzkvm_platform -T ./guest/riscv32im-risc0-zkvm-elf.ld
+	${ROOT_DIR}riscv32im-${MACHINE}/bin/riscv32-unknown-elf-g++ -nostartfiles ./guest/main.cpp -o ./guest/out/main -L${ROOT_DIR}guest/out/platform/riscv32im-risc0-zkvm-elf/release -lzkvm_platform -T ./guest/riscv32im-risc0-zkvm-elf.ld
 
 .PHONY: execute
 execute: guest
diff --git a/examples/c-guest/guest/main.c b/examples/c-guest/guest/main.c
deleted file mode 100644
index c7155bd1..00000000
--- a/examples/c-guest/guest/main.c
+++ /dev/null
@@ -1,25 +0,0 @@
-#include "platform.h"
-#include <stdint.h>
-
-union u32_cast {
-  uint32_t value;
-  uint8_t buffer[4];
-};
-
-int main() {
-  // TODO introduce entropy into memory image (for zk)
-  sha256_state* hasher = init_sha256();
-
-  // Read two u32 values from the host, assuming LE byte order.
-  union u32_cast a;
-  union u32_cast b;
-  env_read(a.buffer, 4);
-  env_read(b.buffer, 4);
-
-  a.value *= b.value;
-
-  env_commit(hasher, a.buffer, sizeof(a.buffer));
-  env_exit(hasher, 0);
-
-  return 0;
-}
diff --git a/examples/c-guest/guest/platform.h b/examples/c-guest/guest/platform.h
deleted file mode 100644
index 3c1e6a8d..00000000
--- a/examples/c-guest/guest/platform.h
+++ /dev/null
@@ -1,731 +0,0 @@
-#include <stdarg.h>
-#include <stdbool.h>
-#include <stdint.h>
-#include <stdlib.h>
-
-/**
- * Maximum size of a BER/DER-encoded OID in bytes.
- */
-#define ObjectIdentifier_MAX_SIZE 39
-
-/**
- * Size of a zkVM memory page.
- */
-#define PAGE_SIZE 1024
-
-#define MEM_BITS 28
-
-#define MEM_SIZE (1 << MEM_BITS)
-
-#define GUEST_MIN_MEM 1024
-
-#define GUEST_MAX_MEM (SYSTEM).start
-
-/**
- * Top of stack; stack grows down from this location.
- */
-#define STACK_TOP 2098176
-
-/**
- * Program (text followed by data and then bss) gets loaded in
- * starting at this location.  HEAP begins right afterwards.
- */
-#define TEXT_START 2099200
-
-#define DIGEST_WORDS 8
-
-#define DIGEST_BYTES (WORD_SIZE * DIGEST_WORDS)
-
-/**
- * Number of words in each cycle received using the SOFTWARE ecall
- */
-#define IO_CHUNK_WORDS 4
-
-#define MAX_BUF_BYTES (4 * 1024)
-
-#define MAX_BUF_WORDS (MAX_BUF_BYTES / WORD_SIZE)
-
-#define MAX_SHA_COMPRESS_BLOCKS 1000
-
-#define HALT 0
-
-#define INPUT 1
-
-#define SOFTWARE 2
-
-#define SHA 3
-
-#define BIGINT 4
-
-#define TERMINATE 0
-
-#define PAUSE 1
-
-#define SPLIT 2
-
-#define REG_ZERO 0
-
-#define REG_RA 1
-
-#define REG_SP 2
-
-#define REG_GP 3
-
-#define REG_TP 4
-
-#define REG_T0 5
-
-#define REG_T1 6
-
-#define REG_T2 7
-
-#define REG_S0 8
-
-#define REG_FP 8
-
-#define REG_S1 9
-
-#define REG_A0 10
-
-#define REG_A1 11
-
-#define REG_A2 12
-
-#define REG_A3 13
-
-#define REG_A4 14
-
-#define REG_A5 15
-
-#define REG_A6 16
-
-#define REG_A7 17
-
-#define REG_S2 18
-
-#define REG_S3 19
-
-#define REG_S4 20
-
-#define REG_S5 21
-
-#define REG_S6 22
-
-#define REG_S7 23
-
-#define REG_S8 24
-
-#define REG_S9 25
-
-#define REG_S10 26
-
-#define REG_S11 27
-
-#define REG_T3 28
-
-#define REG_T4 29
-
-#define REG_T5 30
-
-#define REG_T6 31
-
-#define REG_MAX 32
-
-#define OP_MULTIPLY 0
-
-/**
- * BigInt width, in bits, handled by the BigInt accelerator circuit.
- */
-#define WIDTH_BITS 256
-
-/**
- * BigInt width, in bytes, handled by the BigInt accelerator circuit.
- */
-#define WIDTH_BYTES (WIDTH_BITS / 8)
-
-
-
-#define STDIN 0
-
-#define STDOUT 1
-
-#define STDERR 2
-
-#define JOURNAL 3
-
-/**
- * An enum representing the available verbosity levels of the logger.
- *
- * Typical usage includes: checking if a certain `Level` is enabled with
- * [`log_enabled!`](macro.log_enabled.html), specifying the `Level` of
- * [`log!`](macro.log.html), and comparing a `Level` directly to a
- * [`LevelFilter`](enum.LevelFilter.html).
- */
-enum Level {
-  /**
-   * The "error" level.
-   *
-   * Designates very serious errors.
-   */
-  Error = 1,
-  /**
-   * The "warn" level.
-   *
-   * Designates hazardous situations.
-   */
-  Warn,
-  /**
-   * The "info" level.
-   *
-   * Designates useful information.
-   */
-  Info,
-  /**
-   * The "debug" level.
-   *
-   * Designates lower priority information.
-   */
-  Debug,
-  /**
-   * The "trace" level.
-   *
-   * Designates very low priority, often extremely verbose, information.
-   */
-  Trace,
-};
-typedef uintptr_t Level;
-
-/**
- * Wrapper around [`BufferKindUser`].
- *
- * It handles data buffering and implements the slice-based traits.
- */
-typedef struct CoreWrapper_CtVariableCoreWrapper_Sha256VarCore_Impl_____U32__OidSha256 CoreWrapper_CtVariableCoreWrapper_Sha256VarCore_Impl_____U32__OidSha256;
-
-/**
- * SHA-256 implementation cross-compatible with `sha2::Sha256`.
- */
-typedef struct CoreWrapper_CtVariableCoreWrapper_Sha256VarCore_Impl_____U32__OidSha256 Sha256_Impl;
-
-/**
- * C wrapper for guest sha256 implementation.
- *
- * This is used in the guest to generate any sha256
- * hash, but also to accumulate the Sha256 state of all data written to journal through
- * [commit].
- *
- * Initialize with [init_sha256], and can retrieve the final hash through [sha256_finalize],
- * or pass it into [zkvm_exit] to exit the program.
- */
-typedef struct sha256_state {
-  Sha256_Impl *inner;
-} sha256_state;
-
-/**
- * Digest represents the results of a hashing function.  It is always 256 bits
- * of storage although depending on the hash it may have additional structure
- * (for example Poseidon's output is actually composed of field elements).  The
- * storage is in u32's in part to simplify alignment requirements, especially
- * in the zkVM.
- */
-typedef uint32_t Digest[DIGEST_WORDS];
-
-
-
-
-
-
-
-
-
-
-
-struct sha256_state *init_sha256(void);
-
-/**
- * Update the sha256 state with the bytes passed in.
- *
- * # Safety
- * This is safe assuming that pointers have not been manually modified, and len does not go past
- * the buffer of the data in memory.
- */
-void sha256_update(struct sha256_state *hasher, const uint8_t *data, uint32_t len);
-
-/**
- * Finalize the hasher, returning an allocated digest of the output hash.
- *
- * Note: This digest is leaked, as the zkvm currently uses a bump allocator so it does not need to
- * be freed. If switching allocator, should implement a free method for [Digest].
- *
- * # Safety
- * Assumes [sha256_state] has not been freed previously.
- */
-Digest *sha256_finalize(struct sha256_state *hasher);
-
-/**
- * Free allocations from [sha256_state].
- *
- * # Safety
- * This assumes the state has not already been freed or manually modified.
- */
-void sha256_free(struct sha256_state *hasher);
-
-/**
- * Exit the zkvm, using the [sha256_state].
- *
- * # Safety
- * Assumes hasher has not been freed.
- */
-void env_exit(struct sha256_state *hasher, uint8_t exit_code);
-
-/**
- * Write data to the journal, updating the sha256 state accumulation with that data.
- *
- * # Safety
- * This is safe assuming that pointers have not been manually modified, and len does not go past
- * the buffer of the data in memory.
- */
-void env_commit(struct sha256_state *hasher, const uint8_t *bytes_ptr, uint32_t len);
-
-/**
- * Reads `len` bytes into buffer from the host.
- *
- * # Safety
- * Assumes that the buffer has at least `len` bytes allocated.
- */
-void env_read(uint8_t *bytes_ptr, uint32_t len);
-
-#if defined(DEFINE_SYSCALLS)
-/**
- * # Safety
- *
- * This function should be safe to call, but clippy complains if it is not marked as `unsafe`.
- */
-uint8_t *sys_alloc_aligned(uintptr_t bytes, uintptr_t align);
-#endif
-
-#if defined(DEFINE_SYSCALLS)
-/**
- * Send a ReceiptClaim digest to the host to request verification.
- *
- * A cooperative prover will only return if there is a verifying proof
- * associated with that claim digest, and will always return a result code
- * of 0 to register a0. The caller must encode the claim_digest into a
- * public assumptions list for inclusion in the guest output.
- *
- * # Safety
- *
- * `claim_digest` must be aligned and dereferenceable.
- */
-void sys_verify_integrity(const uint32_t (*claim_digest)[DIGEST_WORDS]);
-#endif
-
-extern uint8_t *sys_alloc_aligned(uintptr_t nwords, uintptr_t align);
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float acosf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float acoshf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double acosh(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double acos(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float asinf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float asinhf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double asinh(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double asin(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float atan2f(float y, float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double atan2(double y, double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float atanf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float atanhf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double atanh(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double atan(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float cbrtf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double cbrt(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float ceilf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double ceil(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float copysignf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double copysign(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float cosf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float coshf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double cosh(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double cos(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float erfcf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double erfc(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float erff(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double erf(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float exp10f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double exp10(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float exp2f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double exp2(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float expf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float expm1f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double expm1(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double exp(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float fabsf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double fabs(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float fdimf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double fdim(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float floorf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double floor(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float fmaf(float x, float y, float z);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double fma(double x, double y, double z);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float fmaxf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double fmax(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float fminf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double fmin(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float fmodf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double fmod(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float hypotf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double hypot(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-int32_t ilogbf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-int32_t ilogb(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float j0f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double j0(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float j1f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double j1(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float jnf(int32_t n, float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double jn(int32_t n, double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float ldexpf(float x, int32_t n);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double ldexp(double x, int32_t n);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double lgamma(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float log10f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double log10(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float log1pf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double log1p(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float log2f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double log2(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float logf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double log(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float nextafterf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double nextafter(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float powf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double pow(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float remainderf(float x, float y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double remainder(double x, double y);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float roundf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double round(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float scalbnf(float x, int32_t n);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double scalbn(double x, int32_t n);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float sinf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float sinhf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double sinh(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double sin(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float sqrtf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double sqrt(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float tanf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float tanhf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double tanh(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double tan(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float tgammaf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double tgamma(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float truncf(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double trunc(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float y0f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double y0(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float y1f(float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double y1(double x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-float ynf(int32_t n, float x);
-#endif
-
-#if (defined(DEFINE_LIBM) && defined(DEFINE_ZKVM))
-double yn(int32_t n, double x);
-#endif
-
-#if (defined(DEFINE_ZKVM) && defined(DEFINE_ENTRYPOINT))
-void __start(void);
-#endif
diff --git a/examples/c-guest/host/Cargo.toml b/examples/c-guest/host/Cargo.toml
index 49d624a0..9ad081e9 100644
--- a/examples/c-guest/host/Cargo.toml
+++ b/examples/c-guest/host/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-risc0-zkvm = { path = "../../../risc0/zkvm" }
+risc0-zkvm = { version = "=1.0.0-rc.5" }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 serde = "1.0"
 anyhow = "1.0"
diff --git a/examples/c-guest/platform/Cargo.toml b/examples/c-guest/platform/Cargo.toml
index 86864405..99fd0a81 100644
--- a/examples/c-guest/platform/Cargo.toml
+++ b/examples/c-guest/platform/Cargo.toml
@@ -4,6 +4,6 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-risc0-zkvm-platform = { path = "../../../risc0/zkvm/platform", features = ["rust-runtime", "export-syscalls"] }
-risc0-binfmt = { path = "../../../risc0/binfmt", default-features = false }
-risc0-zkp = { path = "../../../risc0/zkp", default-features = false }
+risc0-zkvm-platform = { version = "=1.0.0-rc.5", features = ["rust-runtime", "export-syscalls"] }
+risc0-binfmt = { version = "=1.0.0-rc.5", default-features = false }
+risc0-zkp = { version = "=1.0.0-rc.5", default-features = false }
diff --git a/examples/c-guest/platform/cbindgen.toml b/examples/c-guest/platform/cbindgen.toml
index 144f548c..c65415b2 100644
--- a/examples/c-guest/platform/cbindgen.toml
+++ b/examples/c-guest/platform/cbindgen.toml
@@ -1,4 +1,4 @@
-language = "C"
+language = "C++"
 
 [defines]
 "target_os = zkvm" = "DEFINE_ZKVM"

```

Diff super unreadable, just pasting so I can reference it if needed. TLDR changes are: switch gcc for g++ in prebuilt toolchain, change main to cpp, update cbindgen config to generate cpp header file (and do some hacks to modify the code since it included a bunch of garbage).

examples/c-guest/guest/riscv32im-risc0-zkvm-elf.ld
```diff
@@ -0,0 +1,77 @@
...
+ENTRY(_start)
+
+/* Must match risc0/zkvm/platform/src/memory.rs */
+/* TODO double check these memory sections, probably not ideal */
+MEMORY {
+  stack        : ORIGIN = 0x00000400, LENGTH =  8M - 1K
+  data    (RW) : ORIGIN = 0x00080000, LENGTH =  24M
+  heap         : ORIGIN = 0x02000000, LENGTH =  80M
+  prog    (X)  : ORIGIN = 0x07000000, LENGTH =  80M
+}
```
austinabell, line 28: I'd love to have this be resolved before this comes in. Seems like the Rust memory is:
```
stack (0x400 - 0x0020_0400) | (start 0x0020_0800) prog/.text | .data and .bss shared (top 0x0C00_0000)
```

Would it be reasonble to match this instead, and have these sections be more dynamic rather than fixed as they are here? Ordering is also diff with prog above the heap, which may not have the same checks (seems like the allocator just checks if memory is above the system range, but this can allocate over the program?)

flaub: I'm not sure how to match the rust model with a linker script, because the heap starts right at the end of text. At least in C, it should be possible to rely on the `_end` symbol to exist (I think most linkers define it).

flaub (approved):

austinabell:
So the examples fail for a reason other than this, but it shouldn't work as is because I haven't set it up to use the makefile (previous runs it seems the examples aren't being run?).

What I could do for this is have the host have a build script that compiles the guest within it, to remove the need for the makefile and have the ELF be pulled into the host at compile time with `include_bytes!`. This would also make it so that the example can just be run with `cargo run`. Would this be preferred?

austinabell:
It's a bit cursed, having the build script build the static lib and using the preinstalled gcc to build all within the build.rs file, but it's also kind of nice to use the `risc0_path` to not have to redownload the toolchain (and this supports building from source).

Let me know if ideal to remove makefile. I could see some wanting to use it or reference the commands outside of Rust, but also might just be more annoying to maintain

flaub:
Wondering if this is landable? Even if it's still got issues we could land it so we can iterate.

austinabell:
> Wondering if this is landable? Even if it's still got issues we could land it so we can iterate.

I'm good with it landing as is if you are! Wasn't going to merge myself as I had made changes after your approval

PR #1902, opened 2024-05-29
Fix broken link in docs: Prepend "www" to ensure accessibility

0xBitzz:
## Description:

This pull request addresses a broken link "https://risczero.com" in the documentation that was causing a connection error.

## Changes made:

Prepended "www" to the URL in the documentation, making it https://www.risczero.com. This temporary fix ensures users can access the intended resource.

## Considerations:

This is a temporary solution. To permanently resolve the issue:
Consider configuring the web server to handle connections for both "https://risczero.com" and "https://www.risczero.com", directing them to the appropriate website content.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ exclude = ["tools/crates-validator"]
 version = "1.0.0-rc.5"
 edition = "2021"
 license = "Apache-2.0"
-homepage = "https://risczero.com/"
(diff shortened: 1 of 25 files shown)


nuke-web3:
Love the initiative here to get this done! 

IMHO though this is not as good a solution as inverting the redirect (www -> TLD) but need our ops to chime in on this :pray: (pinged them internally)

PR #1903, opened 2024-05-29
Remove relay references from risc0-ethereum CI job

nategraf:
Follow-up on https://github.com/risc0/risc0-ethereum/pull/122, this PR removes references to the relay from the cargo patch script in the `risc0-ethereum` tests.


nategraf:
`risc0-ethereum CI` is ✅ ! This means that `main` here does indeed work with `risc0-ethereum` and `risc0-foundry-template`, pending https://github.com/risc0/risc0-foundry-template/pull/106.

nategraf:
With https://github.com/risc0/risc0-foundry-template/pull/106 merged, this PR is now ready for review. Unfortunately, the `risc0-ethereum` job will be failing again because of control root changes in https://github.com/risc0/risc0/pull/1910. Once those changes, and any subsequent changes are landed, the job should now work again and should stay working as long as we don't make any breaking changes.

capossele (approved):

PR #1909, opened 2024-05-30
Improve `SyscallName`

nvzqz:
- `SyscallName::as_str` returns `&'static str`
- `SyscallName::from_bytes_with_nul` is `unsafe`
- Methods should take `self` because `SyscallName` is `Copy` and small

This adds an `unsafe` block into `declare_syscall!`, which may affect crates using `#[deny(unsafe)]`.


shkoo:
Since as of rust 1.69, `CStr::from_bytes_until_nul` is `const`, would it make sense to use that instead?  And it's not even `unsafe`...

nvzqz:
Superseded by #1912 due to change in base repo.

PR #1910, opened 2024-05-30
Increase ZK_CYCLES

tzerrell:
Increase the number of cycles used to ensure zero knowledge.


risc0/circuit/rv32im/src/prove/emu/testutil.rs
```diff
@@ -69,7 +69,7 @@ pub fn simple_loop() -> Program {
         entry: 0x4000,
         image: BTreeMap::from([
             (0x4000, 0x00000713), // li      a4,0
-            (0x4004, 0x3e800793), // li      a5,1000
+            (0x4004, 0x06400793), // li      a5,100
```
flaub, line 72: I'm guessing this was to accommodate the larger amount of ZK_CYCLES now causes small tests to go into 3 segments rather than 2?

tzerrell: Yes, or in some cases even more than 3

flaub (approved):

PR #1923, opened 2024-06-04
feat(benchmarks): react19/next15

nahoc:
Depends on https://github.com/risc0/risc0/pull/1924

This PR includes these changes:
* react compiler seems to break column visibility picker, so turned off for the time being -- can still safely migrate to react 19
* ppr now works properly


flaub:
I'd like to stay on stable versions for production. Is there a feature in canary that you'd really like to have?

nahoc:
> I'd like to stay on stable versions for production. Is there a feature in canary that you'd really like to have?

Thing is, Vercel really changed how caching works in NextJS v15 (for the better)

On 14 everything was cached by default, now on 15 it's the opposite, caching is now opt-in. So it makes for more predictable caching behavior.

PR #1928, opened 2024-06-05
Update docs to 1.0

flaub:


website/github-link-version-check.py
```diff
@@ -7,44 +7,63 @@
 import re
 
 EXTENSIONS = [
-    '.md',
+    ".md",
+]
+
+EXCEPTIONS = [
+    # allow refs to trusted setup ceremony assets
+    "d4e427283027c28b38b8eda1562e8e0e68d1b0e2"
```
pdg744, line 15: I guess this is the commit hash for the TSC assets?

PR #1933, opened 2024-06-05
Add v1.0.0 verifier Sepolia deployment to dev docs

nategraf:


website/api_versioned_docs/version-1.0/blockchain-integration/contracts/verifier.md
```diff
@@ -43,21 +43,22 @@ You can choose to use this contract or deploy your own.
...
-[article-proof-composition]: https://www.risczero.com/news/proof-composition
+[release-v1.0.0]: https://github.com/risc0/risc0-ethereum/releases/tag/v1.0.0
 [release-v0.10.0]: https://github.com/risc0/risc0-ethereum/releases/tag/v0.10.0
 [release-v0.9.0]: https://github.com/risc0/risc0-ethereum/releases/tag/v0.9.0
 [foundry-template]: https://github.com/risc0/bonsai-foundry-template
+[sepolia-verifier-v1.0.0]: https://sepolia.etherscan.io/address/0x36be51af39a2d430368ffee8c664c46d2298083d#code
 [sepolia-verifier-v0.10.0]: https://sepolia.etherscan.io/address/0x4e9559f1fdf82dd1babdd42b4550ffdb1a1c59de#code
 [sepolia-verifier-v0.9.0]: https://sepolia.etherscan.io/address/0xe57255C10291771A2E14f7eb257555AE82D78302#code
-[Groth16Receipt]: https://docs.rs/risc0-zkvm/1.0/risc0_zkvm/struct.Groth16Receipt.html
+[Groth16Receipt]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Groth16Receipt.html
```
flaub, line 61: This should point to 1.0

website/api/blockchain-integration/contracts/verifier.md
```diff
@@ -43,18 +43,19 @@ You can choose to use this contract or deploy your own.
 | ---------------------------------------------------------------- | -------------------------- | --------- | ---------------------------------------------------------------------- |
 | [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v0.9]  | [v0.9.0][release-v0.9.0]   | [Sepolia] | [0xe57255C10291771A2E14f7eb257555AE82D78302][sepolia-verifier-v0.9.0]  |
 | [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v0.10] | [v0.10.0][release-v0.10.0] | [Sepolia] | [0x4e9559f1fdf82dd1babdd42b4550ffdb1a1c59de][sepolia-verifier-v0.10.0] |
+| [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v1.0]  | [v1.0.0][release-v1.0.0]   | [Sepolia] | [0x36be51af39a2d430368ffee8c664c46d2298083d][sepolia-verifier-v1.0.0] |
 
 [EvenNumber.sol@v0.10]: https://github.com/risc0/bonsai-foundry-template/blob/release-0.10/contracts/EvenNumber.sol
 [RiscZeroVerifierRouter@v0.10]: https://github.com/risc0/risc0-ethereum/blob/release-0.10/contracts/src/RiscZeroVerifierRouter.sol
```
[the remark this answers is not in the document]
pdg744: or perhaps we can point to `latest`?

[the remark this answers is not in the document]
pdg744: ...or maybe there's a deployed version that makes more sense to point to?

website/api/blockchain-integration/contracts/verifier.md
```diff
@@ -43,18 +43,19 @@ You can choose to use this contract or deploy your own.
 | ---------------------------------------------------------------- | -------------------------- | --------- | ---------------------------------------------------------------------- |
 | [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v0.9]  | [v0.9.0][release-v0.9.0]   | [Sepolia] | [0xe57255C10291771A2E14f7eb257555AE82D78302][sepolia-verifier-v0.9.0]  |
 | [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v0.10] | [v0.10.0][release-v0.10.0] | [Sepolia] | [0x4e9559f1fdf82dd1babdd42b4550ffdb1a1c59de][sepolia-verifier-v0.10.0] |
+| [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v1.0]  | [v1.0.0][release-v1.0.0]   | [Sepolia] | [0x36be51af39a2d430368ffee8c664c46d2298083d][sepolia-verifier-v1.0.0] |
 
 [EvenNumber.sol@v0.10]: https://github.com/risc0/bonsai-foundry-template/blob/release-0.10/contracts/EvenNumber.sol
```
pdg744, line 48: This link to 0.10 could use a refresh -- not sure what the best way to manage maintenance is on these. 

I think when it comes to cutting new zkVM versions we have some systems in place so that links get updated during the release process, but I don't think we have those systems set up yet for risc0-ethereum versioning.

nategraf:
Should we go ahead and only display addresses for one version of the contracts per version of the docs?

pdg744:
> Should we go ahead and only display addresses for one version of the contracts per version of the docs? In the case of "next" this might mean only adding addresses as part of the release process, when the next version is fixed and the contracts are deployed.

I'm not sure I follow your question/suggestion, but I'm happy to defer to you for how best to manage this on `next` -- seems like a tricky issue that you have a better handle on than I do.

nategraf:
> > Should we go ahead and only display addresses for one version of the contracts per version of the docs? In the case of "next" this might mean only adding addresses as part of the release process, when the next version is fixed and the contracts are deployed.
> 
> I'm not sure I follow your question/suggestion, but I'm happy to defer to you for how best to manage this on `next` -- seems like a tricky issue that you have a better handle on than I do.

Essentially, the versioned pages would each have a table with one entry per contract, all at the current version (e.g. 1.0 or 1.1). If they go to the API docs for the "next" version, there would be no table, but a note about how we deploy contracts when the version is released.

website/api/blockchain-integration/contracts/verifier.md
```diff
@@ -43,18 +43,19 @@ You can choose to use this contract or deploy your own.
 | ---------------------------------------------------------------- | -------------------------- | --------- | ---------------------------------------------------------------------- |
 | [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v0.9]  | [v0.9.0][release-v0.9.0]   | [Sepolia] | [0xe57255C10291771A2E14f7eb257555AE82D78302][sepolia-verifier-v0.9.0]  |
 | [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v0.10] | [v0.10.0][release-v0.10.0] | [Sepolia] | [0x4e9559f1fdf82dd1babdd42b4550ffdb1a1c59de][sepolia-verifier-v0.10.0] |
+| [RiscZeroGroth16Verifier.sol][RiscZeroGroth16Verifier.sol@v1.0]  | [v1.0.0][release-v1.0.0]   | [Sepolia] | [0x36be51af39a2d430368ffee8c664c46d2298083d][sepolia-verifier-v1.0.0] |
```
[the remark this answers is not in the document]
nategraf: I'll start adding a version string constant to the contracts.

nategraf:
Changing the base to https://github.com/risc0/risc0/pull/1957, which should be run first to fix the CI check job and run `prettier`

website/api/blockchain-integration/contracts/verifier.md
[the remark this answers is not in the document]
nategraf: Check out the latest version of this PR and let me know what you think

nategraf:
@pdg744 this is ready for you to take a look at. I've simplified the table such that each versioned page only has the table of addresses for its own version.

pdg744 (approved):
Nice! This looks quite clean -- I like the way you handled `next` vs. `1.0` vs. older versions.

pdg744:
I've approved the changes re: the actual copy for the website... I'm not super well-informed on the tests that you've changed here (and not sure whether there's a separate codeowner whose approval is required for those tests)

nategraf:
Indeed, it looks like @flaub needs to approve this. I believe because I touched `.github/workflows/website.yml`

website/api/blockchain-integration/contracts/verifier.md
```diff
@@ -31,36 +28,43 @@ contract EvenNumber {
...
+[RiscZeroVerifierRouter]: https://github.com/risc0/risc0-ethereum/blob/main/contracts/src/RiscZeroVerifierRouter.sol
+[VersionManagement]: https://github.com/risc0/risc0-ethereum/blob/main/contracts/version-management-design.md
+[foundry-template]: https://github.com/risc0/risc0-foundry-template
 [term-image-id]: /terminology#image-id
 [term-journal]: /terminology#journal
 [term-receipt]: /terminology#receipt
 [term-verify]: /terminology#verify
 [term-zkvm-program]: /terminology#zkvm-program
-[VersionManagement@main]: https://github.com/risc0/risc0-ethereum/blob/main/contracts/version-management-design.md
+[doc-released-contracts]: /api/blockchain-integration/contracts/verifier#contract-addresses
```
flaub, line 70: This needs to be a relative path I think

flaub: Basically, once we run the docusaurus version command to promote the next version, this URL will break. But if we keep it relative, it will work across promotions.

flaub (approved):

website/api/blockchain-integration/contracts/verifier.md
nategraf: The intention here is to link to the latest release, rather than the "next" release that the reader is currently viewing. So that's why it's absolute. Does that seem right to you? I can open up a follow-up PR to fix that. (I enabled auto-merge before I saw this)

flaub: I wonder if we should be putting this kind of content that is 'versionless' under `docs` rather than the versioned `api`?

PR #1941, opened 2024-06-07
[docs] git checkout tag, not release for version

nuke-web3:
Might be missing other instances to replace


nategraf:
Checking out the release branch actually _should_ work here
https://github.com/risc0/risc0/blob/release-1.0/Cargo.toml#L30

flaub:
The reason for checking out a release branch is because this can be a smoother experience since users won't be in a detached HEAD state and also it should be compatible.

pdg744:
I'll defer to Frank and/or Erik on this topic.

sashaaldrick:
this came about because I had an issue with the release branch, but I think this was during a lot of migration. I just tried the hello world example again with `git checkout release-1.0` and it worked.

my only gripe is this line:

image

it's not super clear what the "corresponding branch" of the example is for beginners, for reference [hello world](https://github.com/risc0/risc0/tree/release-1.0/examples/hello-world) example. It does say in the top left corner, but I was expecting to see it in the README for hello world and didn't think to look there, as I'm not used to this versioning system. 

Not sure if this is true for everyone, but I feel we could make it clearer if we had something like this in the README:

```md
## Quick Start

First, follow the [examples guide] to install dependencies and check out the correct version of the example.

Then, run the example with:

`git checkout release-1.0`

and

```bash
cargo run --release
```

Congratulations! You just constructed a zero-knowledge proof that you know the
factors of 391.
```

**Comment by sashaaldrick** [comment]
@pdg744 I think we could start with the hello-world readme as a test and that will be the place with the most `beginner` eyes on it. I am wary of introducing extra maintenance for when the example is updated to the latest release. I've added a ticket to the backlog, and will have a think myself the best way to do this.

**Comment by pdg744** [comment]
Makes sense to me! 
Should we keep this open as a draft for now or close it in favor of #1979?

**Comment by flaub** [comment]
Could we say: `release-$MAJOR.$MINOR` and then describe what major and minor is and the list can be found from the releases page? Otherwise we'll need to bump the docs for every patch version it seems.

**Comment by sashaaldrick** [comment]
when you say list @flaub, you mean something like so:

hello-world: release-1.0
chess: release-0.20
etc.

kept somewhere for reference? I think that would help a lot and would mean the maintenance burden would only be on one file

**Comment by flaub** [comment]
Here's what I mean:

```
$ git branch -a | grep release-
  release-0.15
  release-0.16
  release-0.17
  release-0.18
  release-0.19
  release-0.20
  release-1.0
```

sashaaldrick:
@flaub now I understand, thanks for that. For context, I thought some examples lagged behind in being updated to 1.0 (for example), but I see that's not the case. 

So the suggestion is to replace the command in the READMEs with:

`git checkout release-$MAJOR.$MINOR` 

and add a link to the releases page so that they can check for themselves. 

WDYT @pdg744 and @nuke-web3?

pdg744:
> For context, I thought some examples lagged behind in being updated to 1.0 (for example), but I see that's not the case.

The content in `risc0/examples/` stays up-to-date. 
The content in other repos (`demos`, `risc0-ethereum`, `risc0-foundry-template`) may be out of date. 

> So the suggestion is to replace the command in the READMEs with:
> git checkout release-$MAJOR.$MINOR
> and add a link to the releases page so that they can check for themselves.

SGTM

sashaaldrick:
@pdg744 I have made the changes we have discussed, please could you review again?

pdg744 (approved):
Looks good!

PR #1944, opened 2024-06-07
Add version banner to help direct people to the tagged releases

nategraf:


nategraf:
Based on other discussion, maybe we want to be linking to the release _branches_ instead here?

flaub (approved):

PR #1946, opened 2024-06-07
Add test for po2 = 16 recursion program

flaub:


flaub:
This is targetted for main, so it shouldn't matter if it breaks compatiblity

nategraf:
So does that mean this will only be released with 2.0? Additionally, it does seem like, regardless of how we manage versioning and branches, the semver checks CI job should detect breaking changes with the prover API.

nategraf (approved):

nategraf:
Approving this PR as the changes themselves look good. I do think there is more discussion here on versioning, at the very least so I understand our approach, but that will happen in another context.

flaub:
I don't think anyone is using the Prover API directly, so I think the issue around versioning could be minor. I think we can make any changes we'd like to main and determine later if we need to backport a change to the stable branch. I don't think we need this change on stable AFAIK.

Issue #1786, opened 2024-05-06
clarify usability of stark2snark on various architectures


pdg744:
In theory and according to our docs, one can use our tooling locally or remotely. 
But in practice, the stark2snark prover [only works on some hardware](https://github.com/risc0/risc0/issues/1749). 
This is a frequent point of confusion for users -- would be good to clarify the current status and recommended workflow(s) in the docs. 

- [x] add admonition to local proving page & proving options page 
- [x] add admonition to bonsai foundry template

nuke-web3:
Related to #1749

nuke-web3:
Idea: could we insert a compile time error to mention this issue until it's resolved for those using metal and other non-supported GPUs? Would help users a ton while this is not resolved

sashaaldrick:
first PR opened and ready for review: #1952 @pdg744

sashaaldrick:
#1952 merged

sashaaldrick:
second PR opened for foundry template [#117](https://github.com/risc0/risc0-foundry-template/pull/117), waiting for review

pdg744:
Thanks @sashaaldrick!

PR #1952, opened 2024-06-10
added info callout about stark2snark issues on apple silicon

sashaaldrick:
Part of [#1786](https://github.com/risc0/risc0/issues/1786), adding as follows:

```
:::info

The `stark2snark` prover currently *only* works on x86 architecture, and so Apple Silicon is *currently unsupported* (even via Docker). 

You can find out more info in the relevant issues [here](https://github.com/risc0/risc0/issues/1520) and [here](https://github.com/risc0/risc0/issues/1749). 
 
:::
```

to `generating-proofs/local-proving` and `generating-proofs/proving-options`


pdg744 (approved):
Thanks Sasha!

pdg744:
@sashaaldrick you can fix the test failure by running `npx prettier . --write` from the `website` directory

PR #1955, opened 2024-06-10
HAL improvements

flaub:
Test OP#37


flaub:
[#37] Test OpenProject PR

https://risczero.openproject.com/work_packages/37

PR #1959, opened 2024-06-11
Adding doc & diagram for end-to-end recursive proving

pdg744:
Adding a doc & diagram to clarify how `lift`, `join`, `identity_p254`, and `compress` are related.

To streamline the review process, I've only included changes on `next`. Will duplicate the changes onto `1.0` after initial review process.


website/api/recursion.md
```diff
@@ -0,0 +1,122 @@
...
+RISC Zero's zkVM consists of three circuits.
+
+1. The RISC-V Circuit is a STARK circuit that proves correct execution of RISC-V programs.
+1. The Recursion Circuit is a separate STARK circuit, that's designed to efficiently generate proofs for the verification of STARK proofs and to support the integration of custom accelerator circuits into the zkVM.
+This circuit has a similar architecture to the RISC-V Circuit, but with fewer columns and a much simpler instruction set.
+The Recursion Circuit supports a number of programs, including [lift], [join], [resolve], and [identity_p254].
+The same [proof system] is used for both the RISC-V Circuit and the Recursion Circuit.
+1. The STARK-to-SNARK Circuit is an R1CS circuit that proves verification of proofs from the *identity_p254* program.
+
+*The size of a [Receipt] varies depending on the size of the ReceiptClaim and its ReceiptKind. For SuccinctReceipts and Groth16 Receipts, the size of the *seal* is constant.
```
nategraf: I don't think you need this caveat. The proof itself is indeed constant size, and this is true of any ZKP, even as an ideal block box.

website/api/recursion.md
```diff
@@ -0,0 +1,123 @@
...
+## Recursive Circuit Architecture
+
+RISC Zero's zkVM consists of three circuits.
+
+1. The RISC-V Circuit is a STARK circuit that proves correct execution of RISC-V programs.
+1. The Recursion Circuit is a separate STARK circuit, that's designed to efficiently generate proofs for the verification of STARK proofs and to support the integration of custom accelerator circuits into the zkVM.
+   This circuit has a similar architecture to the RISC-V Circuit, but with fewer columns and a much simpler instruction set.
+   The Recursion Circuit supports a number of programs, including [lift], [join], [resolve], and [identity_p254].
+   The same [proof system] is used for both the RISC-V Circuit and the Recursion Circuit.
+1. The STARK-to-SNARK Circuit is an R1CS circuit that proves verification of proofs from the _identity_p254_ program.
```
nategraf, line 113: It's somewhat more accurate to say that it proves verification of proofs from the _recursion circuit_. It's execution is independent of the program that the recursion circuit ran, which is committed to by the control ID passed in as a public input. (This is good, because it means we can change the program without breaking the trusted setup)

website/api/recursion.md
```diff
@@ -0,0 +1,123 @@
...
+
+*The size of a [Receipt] varies depending on the size of the ReceiptClaim and its ReceiptKind. For SuccinctReceipts and Groth16 Receipts, the size of the *seal\* is constant.
```
nategraf, line 115:
```suggestion
```

I don't think we need this caveat. This is true of every ZKP system, even idealized block box versions. The proof size itself is indeed constant.

website/api/recursion.md
```diff
@@ -0,0 +1,123 @@
+# Recursive Proving
+
+RISC Zero's zkVM uses recursive proving in order to achieve unbounded computation size, constant\* proof size, proof aggregation, and proof composition.
+
+The end-to-end process for proof generation is shown in the following diagram, including links to the associated crate docs:
```
nategraf, line 5: It's unclear to me what value the links add. Who is this doc aimed at an what are they interested in figuring out? One thing I don't want is to give developers the impression that they need to be using the methods and types linked here, which should be implementation details for most developers.

flaub: I'm also concerned that these links will go stale over time.

pdg744: this doc is housed under `reference` and not something that most users would encounter unless they're digging into questions about architecture, or we point them to it directly. 

at this stage, our `reports.risczero.com` page is constructed in a way that is going to invite all sorts of questions about these operations -- e.g. [here](https://discord.com/channels/953703904086994974/1152032592170844180/1247999471758086207). 

i'd like to have an easy-to-link-to & comprehensive answer to these questions so that we don't have to re-explain things a million times.

pdg744: i'll add an :::info block that clarifies that this doc is provided for informational purposes and that users don't need to use these functions directly

pdg744: re: stale links... 

not sure how this is any different than everywhere else on the dev docs site...?
note that there's only really like 5 links in the diagram -- would be pretty straight-forward to ctrl-f and update.

pdg744: i'm open to dropping the links if y'all think that's best but generally don't like having informational dead ends

pdg744: i guess another option would be to drop the item-by-item links and just add a single link to the ApiClient page? 

/shrug

flaub: I think most users won't be using the `ApiClient` directly, instead they should use the `Prover` trait.

pdg744: > I think most users won't be using the ApiClient directly, instead they should use the Prover trait.

Thanks Frank! That's helpful.

website/api/recursion.md
```diff
@@ -0,0 +1,148 @@
...
+1. The `identity_p254()` program verifies a proof from the Recursion Prover using the Poseidon254 hash function. The identity_p254 program is used as the last step in the prover pipeline before running the Groth16 prover.
+
+## STARK-to-SNARK Wrapping
+All of the recursion programs in the previous section output a [SuccinctReceipt], which is a STARK proof (~200kB)
+
+The final step in the recursion process is `compress()`, which outputs a [Groth16Receipt], which can be verified on-chain using the [RISC Zero Verifier Contract].
+
+[composite, succinct or groth16 receipts]: https://docs.rs/risc0-zkvm/1.0/risc0_zkvm/enum.ReceiptKind.html
+[Groth16Receipt]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Groth16Receipt.html
+[proof system]: ../docs/proof-system/proof-system-sequence-diagram.md
```
flaub: This will break when we promote the version from next to 1.0.

If you are making links from 'unversioned' to 'versioned' or vice-versa (e.g. /api -> /docs) then you need to use absolute links rather than relative ones.

This section tries to describe the reasoning: https://docusaurus.io/docs/versioning#recommended-practices

flaub:
```suggestion
[proof system]: /docs/proof-system/proof-system-sequence-diagram
```

This will break when we promote the version from next to 1.0.

If you are making links from 'unversioned' to 'versioned' or vice-versa (e.g. /api -> /docs) then you need to use absolute links rather than relative ones.

This section tries to describe the reasoning: https://docusaurus.io/docs/versioning#recommended-practices

website/api/recursion.md
```diff
@@ -0,0 +1,148 @@
...
+1. The `identity_p254()` program verifies a proof from the Recursion Prover using the Poseidon254 hash function. The identity_p254 program is used as the last step in the prover pipeline before running the Groth16 prover.
+
+## STARK-to-SNARK Wrapping
+All of the recursion programs in the previous section output a [SuccinctReceipt], which is a STARK proof (~200kB)
+
+The final step in the recursion process is `compress()`, which outputs a [Groth16Receipt], which can be verified on-chain using the [RISC Zero Verifier Contract].
+
+[composite, succinct or groth16 receipts]: https://docs.rs/risc0-zkvm/1.0/risc0_zkvm/enum.ReceiptKind.html
+[Groth16Receipt]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Groth16Receipt.html
+[proof system]: ../../docs/proof-system/proof-system-sequence-diagram.md
```
flaub, line 143:
```suggestion
[proof system]: /docs/proof-system/proof-system-sequence-diagram
```

This will break when we promote the version from next to 1.0.

If you are making links from 'unversioned' to 'versioned' or vice-versa (e.g. /api -> /docs) then you need to use absolute links rather than relative ones.

This section tries to describe the reasoning: https://docusaurus.io/docs/versioning#recommended-practices

pdg744:
OK I've revised this based on the first round of feedback... hopefully tests will pass this time but I'm signing off for the day.

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+
+## STARK-to-SNARK Wrapping
+
+All of the recursion programs in the previous section output a [SuccinctReceipt], which is a STARK proof (~200kB)
+
+The final step in the recursion process is `compress()`, which outputs a [Groth16Receipt], which can be verified on-chain using the [RISC Zero Verifier Contract].
+
+[composite, succinct or groth16 receipts]: https://docs.rs/risc0-zkvm/1.0/risc0_zkvm/enum.ReceiptKind.html
+[Groth16Receipt]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Groth16Receipt.html
+[proof system]: /docs/proof-system/proof-system-sequence-diagram
```
nuke-web3, line 145:
```suggestion
[proof system]: /docs/proof-system/proof-system-sequence-diagram.md
```

Remember #1972 :wink:

nuke-web3: @pdg744 you forgot to include changes from main that updated yarn - you must run `yarn` to install docusaurus 3.4.0 locally . Just merged main for you.

Strangely I cannot for the life of me resolve the right link :face_exhaling:  any ideas?

website/sidebarsApi.js
```diff
@@ -173,6 +173,11 @@ export default {
       label: "Reference",
       collapsed: true,
       items: [
+        {
+          type: "doc",
+          label: "Recursion",
+          id: "recursion",
+        },
```
nuke-web3, line 180: locally on the sidebar, I do not see this item at all (maybe an error from link issues?) please confirm

website/api/recursion.md
nuke-web3: with .md I CANNOT get anything to work. Without md, any path "works" but does not show up

image

website/sidebarsApi.js
pdg744: so far I'm only working in `/api/` and `sidebarsApi.js`, which means the changes only appear on `next`.

will duplicate to `/api_versioned_sidebars/version-1.0` and `api_versioned_sidebars/sidebar-version-1.0` once all the copy editing and content review are figured out.

pdg744:
> with .md I CANNOT get anything to work. Without md, any path "works" but does not show up

I also can't get it to work with .md. 
Without md, it shows up under `next`.

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
+# Recursive Proving
+
+RISC Zero's zkVM uses recursive proving in order to achieve unbounded computation size, constant proof size, proof aggregation, and proof composition.
+
+:::tip
+
+[prove_with_opts] allows users to choose between [composite, succinct or groth16 receipts].
```
nategraf, line 7: Just to add a little more context to where they can find the function.
```suggestion
[Prover::prove_with_opts][prove_with_opts] allows users to choose between [composite, succinct or groth16 receipts].
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+
+The _Groth16Receipt_ can now be posted on-chain and verified by the [RISC Zero Verifier Contract].
+
+## Recursive Circuit Architecture
+
+RISC Zero's zkVM consists of three circuits.
+
+1. The RISC-V Circuit is a STARK circuit that proves correct execution of RISC-V programs.
+1. The Recursion Circuit is a separate STARK circuit, that's designed to efficiently generate proofs for the verification of STARK proofs and to support the integration of custom accelerator circuits into the zkVM.
+   This circuit has a similar architecture to the RISC-V Circuit, but with fewer columns and a much simpler instruction set.
```
nategraf, line 120:
```suggestion
   This circuit has a similar architecture to the RISC-V Circuit, but with fewer columns and an instruction set optimized for cryptography.
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+   The same [proof system] is used for both the RISC-V Circuit and the Recursion Circuit.
+1. The STARK-to-SNARK Circuit is an R1CS circuit that proves verification of proofs from the Recursion Circuit.
```
nategraf, line 122:
```suggestion
1. The STARK-to-SNARK Circuit is an R1CS circuit that verifies proofs from the Recursion Circuit.
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Recursion Programs
+
+The Recursion Circuit supports a number of programs, including `lift()`, `join()`, `resolve()`, and `identity_p254()`.
```
nategraf, line 126: In this context, these names are references to named programs rather than functions, so the parenthesis don't really make sense.
```suggestion
The Recursion Circuit supports a number of programs, including `lift`, `join`, `resolve`, and `identity_p254`.
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+Instead, users should access recursion via the [Prover].
```
nategraf, line 127:
```suggestion
These are using internally to the [Prover] implementations to produce [SuccinctReceipt] and [Groth16Receipt].
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+
+To aid in conceptual understanding, we offer the following explanations of the recursion programs:
+
```
nategraf, line 130: I don't actually think this line adds much
```suggestion
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+1. The `lift()` program verifies a STARK proof from the RISC-V Prover, using the Recursion Prover. This recursion proof has a single constant-time verification procedure, with respect to the original segment length, and is then used as the input to all other recursion programs (e.g. join, resolve, and identity_p254).
```
nategraf, line 131:
```suggestion
1. The `lift` program verifies a STARK proof from the RISC-V Prover, using the Recursion Prover. This recursion proof has a single constant-time verification procedure, with respect to the original segment length, and is then used as the input to all other recursion programs (e.g. join, resolve, and identity_p254).
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+
+1. The `join()` program verifies two proofs from the Recursion Prover, using the Recursion Prover. By repeated application of `join()`, any number of receipts for execution spans within the same session can be compressed into a single receipt for the entire session.
```
nategraf, line 133:
```suggestion
1. The `join` program verifies two proofs from the Recursion Prover, using the Recursion Prover. By repeated application of `join`, any number of receipts for execution spans within the same session can be compressed into a single receipt for the entire session.
```

website/api/recursion.md
```diff
@@ -0,0 +1,150 @@
...
+
+1. The `identity_p254()` program verifies a proof from the Recursion Prover using the Poseidon254 hash function. The identity_p254 program is used as the last step in the prover pipeline before running the Groth16 prover.
```
nategraf, line 135:
```suggestion
1. The `identity_p254` program verifies a proof from the Recursion Prover using the Recursion Prover with the Poseidon254 hash function. The identity_p254 program is used as the last step in the prover pipeline before running the Groth16 prover.
```

nategraf (approved):

pdg744:
@sashaaldrick @nuke-web3 -- i think this is ready to land, pending your approval

nuke-web3:
Love the theme :star_struck:  - how was that managed?

image

One issue: the text is _waaaaaay_ to small to read, ideally I would like to see a github wrapper for our mermaid stuff to allow zooming/panning/fullscreen toggle, like:

```mermaid
flowchart TD
    A[Christmas] -->|Get money| B(Go shopping)
    B --> C{Let me think}
    C -->|One| D[Laptop]
    C -->|Two| E[iPhone]
    C -->|Three| F[fa:fa-car Car]
```

website/api/recursion.md
nuke-web3: This works locally for me now, without the .md - but not for the "right" reason with the link checker: 

image

pdg744:
> Love the theme 🤩 - how was that managed? 

I don't know what you mean. 

> One issue: the text is waaaaaay to small to read

Oh hmm good point. I've rotated the diagram which mitigates the fontsize issue. 
Open to other suggestions...

website/api/recursion.md
```diff
@@ -0,0 +1,148 @@
...
+
+```mermaid
+flowchart TB
+
+execute(".execute()")
+--> Session
+
+subgraph Session["<a href='https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Session.html'>Session</a>"]
+direction TB
+s1("<a href='https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Segment.html'>Segment</a>")
```
nuke-web3, line 25: do you want all these links to open in the same window? I typically add target to open in a new tab:


```suggestion
s1("<a target="_blank" href='https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Segment.html'>Segment</a>")
```

If you wanna, this should be a quick find and replace - I am looking too to see if we can declare labels instead of the full URL every time :crossed_fingers:

nuke-web3:
Theme is the colors on the diagrams, the default is on gh, we have something much nicer rendering IMHO on our site:

image

I think it's great as is for a start! We can do a latter PR getting mermaid styled and sized really nice.

nuke-web3 (approved):

website/api/recursion.md
pdg744: good call

pdg744:
I guess the theme is maybe coming from here? 
https://github.com/risc0/risc0/blob/89ebf7b971a3e26ed0f6b2a9433196c9bb59047a/website/docusaurus.config.js#L35

PR #1961, opened 2024-06-11
ZKVM-484: revamp `rzup` cli 

hmrtn:
This PR is a Rust revamp of the original `rzup` bash scripts. The goal of this revamp is to have `rzup` act similarly to `rusutp` for toolchain management, and to eventually supersede `cargo risczero install` and `cargo risczero build-toolchain`. 

The features and necessary requirements are likely to change with suggestions and updates. Below is running list the progress. 

To try it out: `cargo run --bin rzup` shows the info previewed below. 

To install the sane defaults (rust, cpp, cargo-risczero): `rzup install`

To install specifics: 
- cargo-risczero: `rzup extension install cargo-risczero` or `rzup install cargo-risczero`
- rust: `rzup toolchain install rust` or `rzup install rust`
- cpp: `rzup toolchain install cpp` or `rzup install cpp`

To check for updates of cargo-risczero and the active toolchains: `rzup check`

To show all installed things and meta: `rzup show`

To set new defaults (i.e. active) toolchains: `rzup default rust <TOOLCHAIN_NAME>`, `rzup default cpp <TOOLCHAIN_NAME>`

To use different cargo-risczero versions (and zkVM): `rzup extension use cargo-risczero <VERSION>`

Each of these have more subcommands that you can see with `--help`.

Todo: 
  - [x] quick start install `rzup install` (latest rust toolchain, cpp, & cargo risczero) 
  - [x] cargo-risczero: extension installation / checking / updating
  - [x] List installed toolchains
  - [x] Install specific toolchains (cpp/rust)
  - [x] Build rust toolchain 
  - [x] Show toolchain / extension / rzup info*
  - [x] rust toolchain update checks 
  - [x] cargo risczero update check
  - [x] pretty print 
  - [x] c++ toolchain update check 
  - [x] Self (rzup) ~~check~~, update, etc 
  - [x] Set default toolchain + `settings.toml` persistence (?)
  - [x] `rzup-init.sh` script for detecting the target and downloading correct `rzup`
  - [x] Github token support to avoid throttling
  - [x] true verbosity 
  - [x] help/discussion coverage 
  - [x] Improve error handling

Preview: 
```
rzup 0.2.1

    ██████  █▀▀▀██
    ██████  █ ▄▀ █        Prove anything.
    ██████  ██▄▄▄█       Verify everywhere.
    ██████
    ██████   RISC
    ██████   ZERO

Contribute      : https://github.com/risc0/
Documentation   : https://dev.risczero.com/
Chat & Support  : https://discord.com/invite/risczero/


Usage: rzup [OPTIONS] [COMMAND]

Commands:
  show       Show the active and installed toolchains and extensions
  install    Install toolchains or extensions
  update     Update toolchains or extensions
  check      Check for updates to toolchains and extensions
  default    Set the default toolchains
  toolchain  Modify or query the installed toolchains
  extension  Modify or query the installed extensions
  help       Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose
          Enable verbose output

  -h, --help
          Print help

  -V, --version
          Print version

Discussion:
    rzup installs RISC Zero toolchains and extensions from the official
    release channels, enabling you to easily switch between different
    versions and keep them updated.

    If you are new to RISC Zero, consider visitng our documentation
    to learn more.
```


rzup/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+serde_json = "1.0.117"
+tar = "0.4.40"
+tempfile = "3.10.1"
+termcolor = "1.4.1"
+tokio = { version = "1.37.0", features = ["full", "rt"] }
+xz = "0.1.0"
+
+[[bin]]
+name = "rzup"
+path = "src/bin/rzup.rs"
```
flaub, line 37: This section could be removed if `src/bin/rzup.rs` is renamed to `src/main.rs`

rzup/src/cmd/check.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn handle_check() -> Result<()> {
+    let client = Client::builder().user_agent("rzup").build()?;
+
+    let rust_repo = "https://api.github.com/repos/risc0/rust/releases/latest";
+    let cpp_repo = "https://api.github.com/repos/risc0/toolchain/releases/latest";
+
+    let rt = tokio::runtime::Runtime::new()?;
+
+    let rust_release_data = rt.block_on(get_latest_version(&client, rust_repo))?;
+    let cpp_release_data = rt.block_on(get_latest_version(&client, cpp_repo))?;
```
flaub, line 28: Alternatively, we could use the blocking API: https://docs.rs/reqwest/latest/reqwest/blocking/

rzup/src/common.rs
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub fn version() -> &'static str {
+    "0.2.0"
+}
```
flaub, line 17: We should be able to grab the current version without needing to hardcode and then maintain this over time (since it's already specified in `Cargo.toml`). 

See: https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/lib.rs#L155

rzup/src/utils.rs
```diff
@@ -0,0 +1,184 @@
...
+    return Some("x86_64-apple-darwin");
+
+    #[cfg(all(target_arch = "aarch64", target_os = "macos"))]
+    return Some("aarch64-apple-darwin");
+
+    #[cfg(all(target_arch = "x86_64", target_os = "windows"))]
+    return Some("x86_64-pc-windows-msvc");
+
+    None
+}
```
flaub, line 184: Using `cfg_if` might be good here, and maybe this could be a `const HOST_TARGET_TRIPLE: &str = ...`

flaub: I think I borrowed this code from some other codebase (which is why I'm to blame for this originally)

rzup/src/bin/rzup.rs
```diff
@@ -0,0 +1,114 @@
...
+        Some(RzupSubcmd::Default { .. }) => todo!(),
+        Some(RzupSubcmd::Uninstall { .. }) => todo!(),
+        None => todo!(),
+    }
+}
+
+#[derive(Debug, Parser)]
+#[command(
+    name = "rzup",
+    bin_name = "rustup[EXE]",
```
flaub, line 53:
```suggestion
    bin_name = "rzup[EXE]",
```
?

flaub:
I wonder if we could use `cargo-risczero` as a library for the transition period of having `cargo-risczero` and `rzup` having some duplicate functionality.

austinabell:
```
~/dev/ris/risc0 (hmrtn/rzup-revamp|✔) 🌲 rzup update
thread 'main' panicked at rzup/src/cmd/update.rs:22:49:
Error getting updates: Invalid directory name format
```

Do I need to clear my install dir for these commands to work or is there a secret to this for now?

rzup/src/cmd/toolchain.rs
```diff
@@ -0,0 +1,109 @@
...
+        /// Toolchain version (i.e. 'latest' or a specific version)
+        #[arg(required = false)]
+        toolchain: Option<String>,
+    },
+    /// Install C++ toolchain
+    Cpp {
+        /// Toolchain version
+        #[arg(required = false)]
+        toolchain: Option<String>,
+    },
```
austinabell, line 58: Opinionated: Noting that going to install this is a bit unintuitive and ideally there is just a way to install both (esp if both have prebuilt for the architecture?)

by comparison, the current `rzup` command by itself would install both to latest, where going this flow, you'd have to run `rzup toolchain install rust` and `rzup toolchain install cpp`, which seems like it could have a simple to use default for those who don't know what they need to get running.

Edit: I see now that this is matching the rustup interfaces, with some indirect comparisons (release channel swapped for rust/cpp toolchain), so this comment is very opinionated.

rzup/src/cmd/toolchain.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+#[derive(Debug, Subcommand)]
+pub enum ToolchainInstallSubcmd {
+    /// Install Rust toolchain
+    Rust {
+        /// Toolchain version (i.e. 'latest' or a specific version)
+        #[arg(required = false)]
+        toolchain: Option<String>,
```
austinabell, line 51: Perhaps the error message for this could indicate what this is looking for? I ran `rzup toolchain install rust 0.21.0` and the follow error message might not be intuitive that it's looking for a git ref/tag:
```
Error during Rust toolchain installation: Failed to fetch release info

Caused by:
    HTTP status client error (404 Not Found) for url (https://api.github.com/repos/risc0/rust/releases/tags/0.21.0)
```

Ideally this links to versions we recommend?

rzup/src/cmd/default.rs
```diff
@@ -0,0 +1,46 @@
...
+    let toolchain = &toolchains
+        .iter()
+        .find(|t| t.name == opts.name)
+        .expect("Error finding toolchain {name}");
+
+    match toolchain.repo {
+        RepoType::Toolchain(ToolchainRepo::Rust) => {
+            RustupToolchain::link(RUSTUP_TOOLCHAIN_NAME, &toolchain.path)
+                .expect("Error linking toolchain {name}")
+        }
```
austinabell, line 42: Noting my experience with this:
```
~/dev/ris/risc0 (hmrtn/rzup-revamp|✔) [2]🌲 rzup default 1.0.1
thread 'main' panicked at rzup/src/cmd/default.rs:36:10:
Error finding toolchain {name}
```

which is after installing with rzup. Not sure what it's looking for.

rzup/src/cmd/extension.rs
```diff
@@ -0,0 +1,51 @@
...
+};
+
+#[derive(Debug, Subcommand, Clone)]
+pub enum ExtensionSubcmd {
+    /// Use a specific cargo-risczero version
+    #[command(aliases = ["install", "add"])]
+    Use {
+        #[arg(help = help::INSTALL_HELP)]
+        version: String,
+    },
```
austinabell, line 29: Similar feedback to previous, handling actual versions would be nice:

```
rzup extension use 1.0.1
thread 'main' panicked at rzup/src/cmd/extension.rs:42:14:
Error during cargo-risczero installation: Failed to fetch release info

Caused by:
    HTTP status client error (404 Not Found) for url (https://api.github.com/repos/risc0/risc0/releases/tags/1.0.1)
```

rzup/src/cmd/extension.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// Update cargo-risczero extension
+    Update,
```
austinabell, line 31: Can these just be one command? Confusing that they do the same (as far as I can see) but one just handles installs a versioned?

rzup/src/cmd/show.rs
```diff
@@ -0,0 +1,118 @@
...
+        None,
+        rzup_home().unwrap().to_str().unwrap(),
+    )?;
+
+    let extension_version = get_installed_extension_version().unwrap();
+    pretty_print_message(&mut stdout, true, None, "cargo-risczero: ")?;
+    pretty_println_message(&mut stdout, false, None, &extension_version)?;
+
+    pretty_print_header(&mut stdout, "\ninstalled toolchains")?;
+    print_installed_toolchains(false)?;
```
austinabell, line 78: Opinionated: Would be nice to see the rustc versions of these on those lines maybe? Only shows the rust version on the active toolchain

rzup/src/bin/rzup.rs
```diff
@@ -0,0 +1,157 @@
...
+    Default {
+        #[command(flatten)]
+        opts: DefaultOpts,
+    },
+    /// Manage and install RISC Zero toolchains
+    Toolchain {
+        #[command(subcommand)]
+        subcmd: ToolchainSubcmd,
+    },
+    /// Manage cargo-risczero extensions
```
austinabell, line 138: This is a bit of a confusing hint message to me. Will there be other extensions supported? Or is this just planned to update the `cargo-risczero` version?

rzup/src/cmd/toolchain.rs
hmrtn: Good call out - @SchmErik and I were just talking about this. We are probably going to go with the default `install` installing both the c and rust toolchains. imo having the specific control for toolchains selection will still be useful for more lang support in the future, though.

rzup/src/cmd/toolchain.rs
hmrtn: This is where our naming semantics gets a little hairy. `0.21.0` is a specific version of the `cargo-risczero` extension, not necessarily the toolchain. Toolchain versions are tagged by date: https://github.com/risc0/rust/releases

So the correct toolchain install command would be something like `rzup toolchain install rust v2024-04-22.0` :/

rzup/src/cmd/default.rs
hmrtn: See above - again related to the naming semantics of lang toolchains v. cargo extension

rzup/src/cmd/extension.rs
hmrtn: `use` is for using specific versions of `cargo-risczero`. 
`update` checks for all uneatable items (tool chains and extension) and updates, if any are available. Definitely confusing thanks for pointing it out!

rzup/src/bin/rzup.rs
hmrtn: My goal was to make this future proof, and allow many versions of cargo-risczero, or even other lang "extensions". Though this might be better to kick this can down the road with less confusing wording.

rzup/src/cmd/toolchain.rs
austinabell: Right, guess I was assuming a superset of what this command did. Perhaps just making clear what this version is would be helpful

Cargo.toml
```diff
@@ -1,28 +1,29 @@
...
+    "risc0/tools",
+    "risc0/zkp",
+    "risc0/zkvm",
+    "risc0/zkvm/methods",
+    "risc0/zkvm/platform",
+    "risc0/zkvm/receipts",
+    "rzup",
+    "tools/hotbench",
+    "website/doc-test/main",
+    "xtask",
```
flaub, line 26: Not sure where these whitespace changes are coming from. I think my editor (vscode) follows the `.editorconfig` which defaults to `indent_size = 2`

rzup/Cargo.toml
flaub: If there's only a single binary, yep.

rzup/src/bin/rzup.rs
flaub: What is an extension?

rzup/src/toolchain/repo.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use crate::utils::{fetch_release_info, get_http_client, GithubReleaseData, Repo};
+use anyhow::Result;
+use clap::{Parser, ValueEnum};
+
+#[derive(Default, Debug, Clone, Parser, ValueEnum)]
+pub enum ToolchainRepo {
+    #[default]
+    Rust,
+    Cpp,
```
flaub, line 23: @SchmErik brought up a really good point. If I want rust support, I will also want the c++ toolchain since I'll want to allow C code that is called from rust code to work out of the box. Same goes for future golang support. So C is really like a 'base' feature.

rzup/src/toolchain/repo.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        let repo_name = self
+            .url()
+            .trim_start_matches("https://github.com/")
+            .trim_end_matches(".git");
+
+        let release_url = format!(
+            "https://api.github.com/repos/{}/releases/{}",
+            repo_name, version_tag
+        );
```
flaub, line 63:
```suggestion
        let release_url = format!(
            "https://api.github.com/repos/{repo_name}/releases/{version_tag}"
        );
```

Cargo.toml
hmrtn: I noticed that my editor was not following that - I'll change it back

rzup/src/toolchain/repo.rs
hmrtn: Sounds good, Erik and I landed on the same page about this and it will be installed by default as well.

rzup/src/bin/rzup.rs
hmrtn: cargo-risczero is an example of a cargo "extension", but perhaps its just not a very useful way of describing it here if its unclear.

rzup/rzup-init.sh
```diff
@@ -0,0 +1,216 @@
...
+set -euo pipefail
+
+# Example URLs:
+# for a 64-bit Linux system (x86_64):
+# BASE_URL/rzup-x86_64-linux
+# for a 64-bit macOS system (arm64):
+# BASE_URL/rzup-arm64-darwin
+
+# NOTE: Temporary release URL
+RZUP_BINARY_UPDATE_ROOT="${RZUP_BINARY_UPDATE_ROOT:-https://github.com/hmrtn/asset-test/releases/download/test1/}"
```
hmrtn, line 25: This is a temporary URL as a placeholder for testing purpose only. Will be changed once we have the CI pipeline in place

rzup/rzup-init.sh
```diff
@@ -0,0 +1,216 @@
...
+  get_architecture
+
+  _arch="$RETVAL"
+  assert_nz "$_arch" "arch"
+
+  _url="${RZUP_BINARY_UPDATE_ROOT}/rzup-${_arch}"
+  _dir="$(ensure mktemp -d)"
+  _file="${_dir}/rzup"
+
+  [ -f "$DEPRECATED_RISC0_DIR/bin/rzup" ] && rm -rf "$DEPRECATED_RISC0_DIR"
```
hmrtn, line 75: Note, the init script detects the deprecated installed rzup and removes it if it exists

rzup/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+use anyhow::{anyhow, Context, Result};
+
+pub struct Rzup;
+
+impl Rzup {
+    pub fn version() -> &'static str {
+        env!("CARGO_PKG_VERSION")
+    }
+    pub async fn update() -> Result<()> {
+        // TODO
```
hmrtn, line 33: Updating self (rzup) is an important feature but with s3 hosting, gathering metadata about builds will differ from the toolchain or extension update process. Perhaps this can be skipped entirely and just run the init script?

rzup/src/config.toml
SchmErik, line 1: I added more changes to this file recently. Could you port the new version from the HEAD of main?

rzup/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+lazy_static = "1.5.0"
+regex = "1.10.5"
+reqwest = { version = "0.12.5", features = ["json", "rustls-tls"] }
+serde = { version = "1.0.203", features = ["derive"] }
+serde_json = "1.0.117"
+tar = "0.4.41"
+tempfile = "3.10.1"
+termcolor = "1.4.1"
+thiserror = "1.0.61"
+tokio = { version = "1.38.0", features = ["full"] }
```
flaub, line 26: Do we actually need `full`? Seems like a heavy dep.

rzup/rzup-init.sh
flaub: Seems like we can decide on the official URL now and just manually update it to get this landed.

rzup/rzup-init.sh
```diff
@@ -0,0 +1,225 @@
...
+QUIET=no
+
+# Set XDG_DATA_HOME based on OS
+_ostype="$(uname -s)"
+case "$_ostype" in
+  Darwin)
+    XDG_DATA_HOME="${XDG_DATA_HOME:-${HOME}/Library/Application Support}"
+    ;;
+  *)
+    XDG_DATA_HOME="${XDG_DATA_HOME:-${HOME}/.local/share}"
```
flaub, line 35: Why here and not in home?

rzup/src/cli/extension.rs
flaub, line 1: Isn't this more like a 'tool' rather than an 'extension'?

rzup/rzup-init.sh
hmrtn: It can be. But I chose this because it is what the spec calls for to avoid polluting the user home. 

> If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

https://xdgbasedirectoryspecification.com/

rzup/Cargo.toml
hmrtn: Good call out - we can reduce this to just `macros` and `rt-multi-thread`.

rzup/src/cli/extension.rs
hmrtn: We can rename it to tool. The binary *symlink lives in `.cargo/bin` and is ran with `cargo` prepended - it seems "extension" is naming convention in the docs: https://doc.rust-lang.org/book/ch14-05-extending-cargo.html

flaub (approved):

PR #1962, opened 2024-06-12
skip cargo-binstall if installed

austinabell:
Small update as a stop-gap until #1961, I keep wanting to use rzup, but given it reinstalls cargo-binstall even if just a new patch version to update, I have avoided it. I'd use this if this change was made :D


hmrtn (approved):
Works as expected - lgtm!

PR #1974, opened 2024-06-14
Add `g6.xlarge` Benchmarks + CI Check for Existing Benchmark Data

morgangibbs87:


.github/workflows/bench_pr.yml
```diff
@@ -80,14 +84,27 @@ jobs:
...
+              echo "Previous Benchmark data for ${{ matrix.os }}-${{ matrix.device }} exists... Continuing..."
+          fi
+
+      - name: (New Benchmarks) Upload Initial Benchmark Data
+        if: env.NEW_BENCHMARK == 'true'
+        run: aws s3 cp target/hotbench/fib/benchmark.json $S3_BENCH_PATH/external.json
+
+      - name: Download previous benchmark data
+        run: aws s3 cp $S3_BENCH_PATH/external.json ./cache/external.json
+
```
morgangibbs87, line 107: @flaub - Added so all `bench_pr.yml` steps will work with new Benchmarks by uploading a baseline file to S3 if none exists. If a file does not exist in AWS, then the `Analyze benchmark results` job will fail since nothing was moved from an S3 bucket to `./cache/external.json` in a previous step, which is needed for the comparison.

New steps:
1. Verify a `$S3_BENCH_PATH/external.json` file exists for the `{{ matrix.os }}-${{ matrix.device }}` benchmark being tested.
2. If JSON file exists, it skips the second step to upload initial data.
3. If no JSON exists, it uploads the file generated in the `Run benchmarks` step as a baseline going forward.

An alternative process to putting this in CI would be running the benchmark locally, creating a new folder in S3 with the proper path, then manually moving the local benchmark file there with the name `external.json`. All things considered, keeping it in CI would keep things simpler and more consistent.

---

The jobs ran earlier today as expected with the `g6.xlarge` instances:

- Run 1:
  - Ran the `(New Benchmarks) Upload Initial Benchmark Data` step and successfully uploaded a baseline JSON file in the correct path 
  - https://github.com/risc0/risc0/actions/runs/9521185195/job/26248070239

- Run 2:
  - Skipped the `(New Benchmarks) Upload Initial Benchmark Data` step since a baseline file was found after being generated from the previous run.
  - https://github.com/risc0/risc0/actions/runs/9521185195/job/26251450354?pr=1974

morgangibbs87:
@flaub - This worked. We are ready to promote the label to `prod`.

When adding labels to test Benchmarks of SPECIFIC device types, using `os`/`feature`/`device`  labels via `include` should avoid runner overlap issues as long as `cpu` is not one of the runner's labels.

This should work for promotion to Prod:
```
labels: [self-hosted, prod, bench, "${{ matrix.os }}", "${{ matrix.device }}"]
if: needs.changes.outputs.trigger-bench == 'true'
needs: changes
strategy:
  fail-fast: false
  matrix:
    os: [Linux, macOS]
    feature: [prove]
    device: [cpu]
    include:
      - os: [Linux]
         feature: [cuda]
         device: [g6.xlarge]
```

If that works for you, I'd love to update the labels and test this over the weekend to ensure there is no overlap so we can start landing these.

.github/workflows/bench_pr.yml
morgangibbs87: This will be removed from this PR since the `g6.xlarge` runners will only be added to `bench_reports.yml` and 'bench_trends.yml`. 

It will be added to documentation for temporary use in tests of new all instances and to generate the initial `external.json` file for devices that will also be added to `bench_pr.yml`.

morgangibbs87:
- Benchmark files generated in `ghpages` as expected after testing:
  - [Application Benchmarks .csv](https://github.com/risc0/ghpages/blob/main/dev/benchmarks/Linux-g6.xlarge.csv)
  - [Datasheet .json](https://github.com/risc0/ghpages/blob/main/dev/datasheet/Linux-g6.xlarge.json)

- The `Application Benchmarks` page pulls the new type as expected for the `next` release: [PREVIEW](https://reports-and-benchmarks-git-morgan-g6xlarge-risczero.vercel.app/main/applications-benchmarks/Linux-g6.xlarge)

- [PR 2004](https://github.com/risc0/risc0/pull/2004) is needed before the `Datasheet` Reports page will pull reports for additional benchmark types.
  - This is required for the page to correctly parse and render benchmarks that were available during previous releases versus the current one.
  - Without this, adding a new benchmark type to the existing array in `datasheet-content.tsx` will throw a `prerender-error` for each previous release that the benchmark type did not exist.
  - The same was previously done for `applications-benchmarks-contents.tsx`, adding a distinguishing array as a [constants.tsx](https://github.com/risc0/risc0/blob/main/web/apps/benchmarks-and-reports/src/app/(benchmarks-and-reports)/%5Bversion%5D/applications-benchmarks/%5Bslug%5D/_utils/constants.tsx) file.

---

Documented process for adding and testing new benchmarks in CI and on Reports page [HERE](https://www.notion.so/risczero/Test-and-Add-New-Self-Hosted-AWS-EC2-Runner-Instance-Types-with-Benchmarks-b1db67775dcc4d4d8dec54ef0d20592b)

---

Leaving this as a Draft so it does not get landed yet, as we will vet all of this when everyone is back.

@flaub @ssh-nkar @nahoc @carterbrett @SchmErik @nvzqz

nahoc (approved):
LGTM 👍

flaub (approved):

PR #1975, opened 2024-06-14
Executor speedup

flaub:


risc0/circuit/rv32im/src/prove/emu/pager.rs
```diff
@@ -76,7 +79,8 @@ impl PagedMemory {
     pub fn new(image: MemoryImage) -> Self {
         Self {
             image,
-            page_cache: HashMap::new(),
+            page_table: vec![INVALID_IDX; 256 * 1024],
```
mothran, line 82: I might suggest moving the magic values to const's for documentation purposes.

mothran (approved):

risc0/circuit/rv32im/src/prove/emu/pager.rs
flaub: Fair enough

PR #1983, opened 2024-06-18
Migrate bonsai-sdk to maybe_async

mothran:
This PR attempts to replace the old async module of bonsai-sdk with the maybe-async (https://crates.io/crates/maybe_async) method of proc-macro rewriting.

Additionally it drops the alpha / alpha_async modules to simplify the code base

## TODO:

- [x] Currently we don't test the async mode in cargo-test, I need to work out how best to test that interface correctly in parallel with the sync interface.


rlukata (approved):

bonsai/sdk/Cargo.toml
```diff
@@ -30,6 +31,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 uuid = { version = "1.8", features = ["v4"] }
 
 [features]
-default = ["std"]
+default = ["std", "sync_client"]
 std = ["risc0-groth16/std"]
-async = ["dep:tokio"]
+sync_client = ["maybe-async/is_sync"]
+async_client = ["dep:tokio", "reqwest/stream"]
```
flaub, line 37: What if a user does:

```
features = ["sync_client", "async_client"]
```

Does this even make sense?

flaub (approved):
Interesting!

Only comment is whether the `async_client` feature flag is needed. Seems odd to have `sync_client` and `async_client` together.

bonsai/sdk/Cargo.toml
mothran: This is explicitly a know problem of maybe_async and not supported: https://github.com/fMeow/maybe-async-rs/issues/6

mothran:
> Interesting!
> 
> Only comment is whether the `async_client` feature flag is needed. Seems odd to have `sync_client` and `async_client` together.

I agree it could be cleaner, the reason I went that path was because maybe_async defaults to async while our default is sync currently. So I wanted to make it explicit to toggle one or the other but its not a clean fit into feature flags.

Do you happen to have a better layout of the flags? I would be happy for a suggestion for how to refactor that.

SchmErik:
@mothran I pushed a few changes including a trick I found to get the docstest to pass. It's not the most pretty but I think it works. If this looks good to you, feel free to merge!

SchmErik (approved):

PR #1988, opened 2024-06-18
Fix Benchmark Summary Annotations for Failed GH Actions Jobs

morgangibbs87:
Context:
https://risc0.slack.com/archives/C0407CEBK5Y/p1718745015459479


.github/workflows/bench_pr.yml
```diff
@@ -13,9 +13,9 @@ concurrency:
   cancel-in-progress: true
 
 # this is needed to gain access via OIDC to the S3 bucket for caching
-permissions:
-  id-token: write
-  contents: read
+permissions: write-all
+#  id-token: write
+#  contents: read
```
flaub, line 18: @kevinnassery is this safe to do?

rtxsecurity: I'm ok with it if it's a temporary debugging tool based on Morgan's comment, however - if it solves the issue, then we should proceed identifying which of the granular job permissions are required.

kevinnassery: For temporary debugging I'm ok with it, in the long term the granular permission table should be used:
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

It might be faster to simply enumerate the specific permissions that could be killing the job (id-token, actions, contents) seem most likely.

morgangibbs87: Temporary debugging was the intention, it would never be given these permissions.

flaub:
Do we still need this? It's not clear why we need it, could you add some context to the ticket (slack links aren't sufficient). You can link to a private ticket or issue

morgangibbs87:
@flaub - Yes, we still need it.

The change in here so far was just the initial test to determine whether the `Error: Resource not accessible by integration` error being seen during `Analyze Benchmark` failures without posting annotations was permissions-related or actions-related.

After that test, we know it to be related to the action.

Testing with full permissions with a forced a failure doesn't generate a new table, just part of one in a comment.

Instead of running through all of the tests every run, the action fails and stops all testing when the first threshold is exceeded. That's part of why it doesn't generate the full table when a threshold is exceeded.  `summary-always: true` just means that the comment output gets put on the Summary page as an Annotation, not that it will "always summarize" a full table.

To get around this, we can set the action NOT to fail when the threshold is exceeded, then add another step that runs after it, checks for tests crossing the threshold, and fails if that is detected. That should give us what we want.

I'll add this functionality and open an Issue for the action so they can look at adding the desired feature.

morgangibbs87:
After testing several different ways to trick the action into always showing a summary table, it looks like the action is not set up to create a summary table at all if a performance alert occurs - whether job technically "fails" or "succeeds".

The Summary does not appear for devices where the alert-threshold is exceeded, even when `fail-on-alert: false` is set and `summary-always: true` is enabled.

This is true whether `comment-on-alert: true` is set or not, and even when a very high `fail-threshold` is set to ensure it is not using the `alert-threshold` value by default.

Marking Blocked for now.

Issue created for Action here:
https://github.com/benchmark-action/github-action-benchmark/issues/254

flaub:
Can we close this or do we still need it?

morgangibbs87:
@flaub - Will close this until issue is resolved: https://github.com/benchmark-action/github-action-benchmark/issues/254

PR #2006, opened 2024-06-22
Add @devel and @demos teams to codeowners for examples

nuke-web3:
https://github.com/orgs/risc0/teams - looking there this seems like the best fit for at least one of us to approve, mostly this is to ensure that those teams are notified of changes as examples are mission critical to @risc0/devrel and likely impact or even suggest breaking changes to @risc0/demos .

Ont thing though - we should not block merges - so maybe we want to add `@risc0/platform` or even `@risc0/dev` to ensure changes can be made without waiting on us.

Closes https://github.com/risc0/devrel/issues/99 .

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,15 +3,16 @@
 # Note: Order is important; the last matching pattern takes the most
 # precedence.
 
+# Platform team is requires approval for all PRs
 *                                      @risc0/platform
 
+# Specific paths requiring teams and/or individual approval
 /bonsai/                               @risc0/bonsai
+/examples/                             @risc0/devrel @risc0/demos
+/groth16_proof/groth16/                @flaub @nategraf @tzerrell
 /rzup/                                 @risc0/platform @hmrtn
+/SECURITY.md                           @kevinnassery
 /web/                                  @risc0/platform @nahoc
 /website/                              @risc0/devrel @pdg744
-/examples/                             @risc0/devrel @risc0/demos
-
-/groth16_proof/groth16/                @flaub @nategraf @tzerrell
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -9,6 +9,8 @@
 /rzup/                                 @risc0/platform @hmrtn
 /web/                                  @risc0/platform @nahoc
 /website/                              @risc0/devrel @pdg744
+/examples/                             @risc0/devrel @risc0/demos
+
```
nuke-web3, line 13:
```suggestion
```

.github/CODEOWNERS
```diff
@@ -9,6 +9,7 @@
 /rzup/                                 @risc0/platform @hmrtn
 /web/                                  @risc0/platform @nahoc
 /website/                              @risc0/devrel @pdg744
+/examples/                             @risc0/devrel @risc0/demos
```
flaub, line 12:
```suggestion
/examples/                             @risc0/devrel @risc0/demos @risc0/platform
```

Also could we sort the dirs?

flaub (approved):

.github/CODEOWNERS
nuke-web3: added comments to sections and did a sort ascending. LMK if that is what you wanted?

nuke-web3: https://github.com/risc0/risc0/blob/86dd3dba48d601316a239f97fae34b56162af1c4/.github/CODEOWNERS

nuke-web3: Note that there is a global requirement on platform team already :
image

flaub (approved):

PR #2007, opened 2024-06-22
chore: update yarn, website dev deps, fix links

nuke-web3:
What it says on the tin


website/package.json
```diff
@@ -1,7 +1,9 @@
 {
-  "name": "docusaurus",
-  "version": "0.0.0",
+  "name": "RISC Zero Developer Website",
+  "description": "RISC Zero Documentation, Education Hub, and Resources",
+  "packageManager": "yarn@4.3.1",
```
nuke-web3, line 4: NOTE: you will need to clear node_modules, and run `yarn` that should self update to the LTS version set in the package.json here.

nahoc (commented):
Some thoughts:

- Any thoughts about prefering bun instead of yarn? 
- Will people on yarn v1/v2 be facing bugs?
- Does this support caching on CI to make the website pipeline faster?

nuke-web3:
IDK about caching - is that explicitly needing yarn 1.x as we used before? 

bun would be fine, and from my experience is faster with it's cache, but requires a larger refactor to get working and IDK if Docusaurus plays nice with it. Are we using bun anywhere else now? 

Yarn self-installs updates when it sees the included files (IIUC the included binary is yarn itself)

nahoc:
so did a bit of digging, turns out vercel doesn't support yarn v2 and up:
image
https://github.com/vercel/vercel/discussions/4223
https://vercel.com/docs/deployments/builds/package-managers

although this says it should work for static builds:
https://vercel.com/guides/does-vercel-support-yarn-2

but then you might need to enable corepack to pickup the package manager defined in `package.json`, and yada yada.

here's my PR with bun benchmarks:
https://github.com/risc0/risc0/pull/2010

which is significantly faster than yarn v1

nuke-web3:
Thanks for the detailed report @nahoc :star_struck: !! 

I am all for moving to bun, just need the rest of the team not to veto requiring the change & replacement of yarn in the workflow. (but it's just about a drop in `alias` for yarn in our use case - should be fine for the crew :crossed_fingers: )

PR #2010, opened 2024-06-23
chore: use bun instead of yarn, website dev deps, fix links

nahoc:
context: https://github.com/risc0/risc0/pull/2007#issuecomment-2184996865
to compare against https://github.com/risc0/risc0/pull/2007/files

dep installs:
`bun:     3.33s`
`yarn@v1: 22.05s`

total pipeline time:
`bun+swc: 1m34s`
`yarn@v1: 2m12s`

---

Other possible pipeline time saves:
- biome instead of prettier


nuke-web3:
very nice - supersedes my PR yes? 

So what about the CI cache issue? I have used a bun action before and the cache is built in... but it was cleared every time for when I was using it as something invalidated it every time... https://github.com/nuke-web3/book/blob/main/.github/workflows/pages.yml#L30-L35

Perhaps https://github.com/oven-sh/setup-bun/releases/tag/v2 that is hot off the presses would not have this behavior :crossed_fingers:

nuke-web3:
> biome instead of prettier

I am a big fan of https://dprint.dev/ and would strongly advocate for that. Also blazingly fast and easy to use, see https://github.com/nuke-web3/book/blob/main/.github/workflows/checks.yml#L16-L26 for how I have used it before (maybe needs a version bump)

flaub (approved):

nuke-web3 (approved):

nuke-web3:
followup with ci update to this you think?

nahoc:
> followup with ci update to this you think? note [#2010 (comment)](https://github.com/risc0/risc0/pull/2010#issuecomment-2185438445)

can be in a separate PR of course, but yeah afaik the oven/bun github action caches and works properly (last I checked).

from what I see we don't really use npm/yarn in any github actions currently though?

PR #2019, opened 2024-06-25
Add editor config for nvim and CoC

nategraf:
Here is the repo-specific editor config I use for Neovim[^1] with the CoC LSP[^2] integration. This is the editor I use, and I've just been maintaining these configs locally, but @nuke-web3 and I were discussing this and I figured I'd open a PR to see if we want to include it for others.

Note the simmilarities between this and [.vscode/settings.json](https://github.com/risc0/risc0/blob/main/.vscode/settings.json). CoC is intended to be very similar to the VS Code plugin system and this uses the `rust-analyzer` plugin.

[^1]: https://neovim.io/
[^2]: https://github.com/neoclide/coc.nvim

### DIFF
--- a/.vim/coc-settings.json
+++ b/.vim/coc-settings.json
@@ -1,0 +1,19 @@
+{
+  "rust-analyzer.linkedProjects": [
+    "Cargo.toml",
+    "benchmarks/Cargo.toml",
+    "examples/Cargo.toml",
+    "risc0/zkvm/methods/guest/Cargo.toml",
+    "risc0/zkvm/methods/std/Cargo.toml",
+    "tools/crates-validator/Cargo.toml"
+  ],
+  "rust-analyzer.check.extraEnv": {
+    "RISC0_SKIP_BUILD": "1",
+    "RISC0_SKIP_BUILD_KERNELS": "1",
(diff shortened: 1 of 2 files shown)


.vim/coc-settings.json
```diff
@@ -0,0 +1,19 @@
+{
+  "rust-analyzer.linkedProjects": [
+    "Cargo.toml",
+    "benchmarks/Cargo.toml",
+    "examples/Cargo.toml",
+    "risc0/zkvm/methods/guest/Cargo.toml",
+    "risc0/zkvm/methods/std/Cargo.toml",
```
SchmErik, line 7: does this run the rust analyzer on guest code as well as the host side?

nategraf: I'm actually not sure. I copied these lines from the `.vscode/settings.json`
https://github.com/risc0/risc0/blob/main/.vscode/settings.json#L71-L77

SchmErik: Oh right, (that's the editor I use, haha). For vscode rust analyzer is running on the guest code as seen by type annotations and code completion suggestions

SchmErik (approved):

Issue #832, opened 2023-08-31
CI run build test for crate-validator tool 


SchmErik:
The crate validator tool broke due to a lack of testing. We should add a job in CI to detect when the crate validator tool breaks. The breakage we saw recently was caused by a missing `cargo risczero install` command. A CI workload can build a small number of crates (10) and does not need to publish results.

PR #2022, opened 2024-06-25
Add `crates-validation` Job to Check 10 Crates to Main Workflow

morgangibbs87:


morgangibbs87:
For [Issue 832](https://github.com/risc0/risc0/issues/832):

- New job added in `main.yml` to validate 10 crates. Number can be adjusted by changing `TOP_N_CRATES`.
- Job is almost identical to steps in `nightly.yml`, but has no user inputs and does not upload anything.
- Job triggers when changes are made to `base` files, `risc0/**`, and/or `tools/crates-validator/**`
- A job already exists that runs `cargo test` against `tools/crates-validator`, which will be kept separate
- Total job run takes about [10-12 mins](https://github.com/risc0/risc0/actions/runs/9668203401/job/26673030441?pr=2022) with sccache, with profile validation taking the most time.

@SchmErik - Let me know if the crates validation should trigger after any other changes. Otherwise, this should be good to land.

@flaub

SchmErik (approved):

PR #2026, opened 2024-06-25
Apply checked_bytes fixes to recursion circuit

shkoo:
Apply checked_bytes recursion circuit fixes from risc0/risczero-wip#598


tzerrell (commented):
This matches what my own local bootstrapping is showing (up to some formatting that isn't working correctly on my machine).

Probably worth coordinating with platform folks about when this is ok to land, as at least some times in the past I have needed to coordinate timing bootstrapping changes with downstream folks -- so I'm not clicking approve to leave them time to comment, but this does LGTM and I can change it to an official approve if needed.

SchmErik (approved):

PR #2028, opened 2024-06-25
risc0-build: change `risc0_data` to support revamped `rzup`

SchmErik:
rzup will store toolchains in the ~/.rzup directory. Check if this new directory does not exist, fallback to the old directory.


SchmErik:
@hmrtn is this the directory you're using? I've placed the ctoolchain in `risc0_data.join("cpp/bin/riscv32-unknown-elf-gcc")` so it will look in `~/.rzup/cpp/bin/riscv32-unknow-elf-gcc` with this change

hmrtn (approved):
LGTM

SchmErik:
@nvzqz can you give this a review?

nvzqz (approved):
Seems good to me. Just two minor thoughts:
- I'm not a fan of polluting the home directory, though it seems we already decided on this approach for `rzup`.
- This function would maybe read better and could probably perform better if we relied on early returns instead.

SchmErik:
> Seems good to me. Just two minor thoughts:
> 
>     * I'm not a fan of polluting the home directory, though it seems we already decided on this approach for `rzup`.
> 

so, the reason why we did this was that rustup has occupied the same directory as rustup in that it occupies ~/.rustup so I guess we can say that we borrowed this from rustup...

>     * This function would maybe read better and could probably perform better if we relied on early returns instead.

addressed!

PR #2030, opened 2024-06-26
Fix replacement string in the guest's zkvm dependency in case of no-std

leonardoalt:
Currently `cargo-risczero --nostd` doesn't replace the "risc0_feature_std" placeholder, leading to the following snippet in `guest/Cargo.toml`, which doesn't compile.

```toml
[dependencies]
risc0-zkvm = { version = "1.0.1", default-features = false{{ risc0_feature_std }} }
```


SchmErik (approved):

flaub:
It seems that our CI didn't catch this, we should probably add more tests to get better coverage (@risc0/platform)

SchmErik:
> It seems that our CI didn't catch this, we should probably add more tests to get better coverage (@risc0/platform)

yeah - I'll add a case that exercises no-std

PR #2031, opened 2024-06-26
Simplify `Program` creation in `testutil`

nvzqz:
This adds a `program_from_instructions` helper that takes an iterator of instructions. It was inspired by the `large_text` implementation.

I did not add this helper directly to `Program` because I wanted this change to only affect private internals, not touch public API.

### DIFF
--- a/risc0/circuit/rv32im/src/prove/emu/testutil.rs
+++ b/risc0/circuit/rv32im/src/prove/emu/testutil.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::BTreeMap;
-
 use anyhow::Result;
 use risc0_binfmt::Program;
 use risc0_zkvm_platform::WORD_SIZE;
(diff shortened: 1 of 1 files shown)


SchmErik (approved):

risc0/circuit/rv32im/src/prove/emu/testutil.rs
```diff
@@ -36,19 +34,36 @@ impl Syscall for NullSyscall {
...
+
     Program {
-        entry: 0x4000,
-        image: BTreeMap::from([
-            (0x4000, 0x1234b137), // lui x2, 0x1234b000
-            (0x4004, 0xf387e1b7), // lui x3, 0xf387e000
-            (0x4008, 0x003100b3), // add x1, x2, x3
-            (0x400c, 0x000045b7), // lui a1, 0x4
-            (0x4010, 0x00000073), // ecall(halt)
-        ]),
```
flaub, line 48: It's actually more convenient to have it in this form since sometimes I need to check that jump offsets are correct and when I'm looking at a specific instruction during debugging it's good to know the PC.

PR #22, opened 2022-03-23
Replace unsafe casts with serde (rust)

flaub:

PR #2035, opened 2024-06-27
serde: serializer: implement collect_str

SchmErik:
The panic in this function was causing chrono's `NaiveDate` to fail serialization. In other words, this prevented users from passing `NaiveDate` as an input for the guest.

### DIFF
--- a/risc0/zkvm/src/serde/serializer.rs
+++ b/risc0/zkvm/src/serde/serializer.rs
@@ -438,7 +438,6 @@ impl<'a, W: WordWrite> serde::ser::SerializeStructVariant for &'a mut Serializer
 mod tests {
     use alloc::string::String;
 
-    use chrono::NaiveDate;
     use serde::Serialize;
 
     use super::*;
@@ -507,11 +506,5 @@ mod tests {
             second: "abc".into(),
         };
         assert_eq!(expected, to_vec(&input).unwrap().as_slice());
-    }
-
-    #[test]
-    fn test_date() {
-        let date: NaiveDate = NaiveDate::parse_from_str("2015-09-05", "%Y-%m-%d").unwrap();
-        let _date_vec = to_vec(&date).unwrap();
     }
 }
--- a/risc0/zkvm/src/serde/serializer.rs
+++ b/risc0/zkvm/src/serde/serializer.rs
@@ -18,6 +18,8 @@ use risc0_zkvm_platform::WORD_SIZE;
 
 use super::err::{Error, Result};
 
+use crate::alloc::string::ToString;
+
 /// A writer for writing streams preferring word-based data.
 pub trait WordWrite {
     /// Write the given words to the stream.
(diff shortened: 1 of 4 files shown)


nuke-web3:
Thanks for the patch! Seems kinda "dirty" to implement _just_ for this type for one crate, but IIUC this should help serialize with `ToString` now used that is much more generally applicable to people?

risc0/zkvm/src/serde/serializer.rs
```diff
@@ -120,11 +122,11 @@ impl<'a, W: WordWrite> serde::ser::Serializer for &'a mut Serializer<W> {
...
     }
 
-    fn collect_str<T>(self, _: &T) -> Result<()>
+    fn collect_str<T>(self, value: &T) -> Result<()>
     where
-        T: core::fmt::Display + ?Sized,
+        T: ?Sized + core::fmt::Display,
     {
-        panic!("collect_str")
+        self.serialize_str(&value.to_string())
```
nuke-web3, line 129: - I see all the way back in https://github.com/risc0/risc0/pull/22 is where this panic was intentionally introduced. I don't know why this was done... perhaps there was a fundamental reason @flaub?

risc0/zkvm/src/serde/serializer.rs
```diff
@@ -18,6 +18,8 @@ use risc0_zkvm_platform::WORD_SIZE;
 
 use super::err::{Error, Result};
 
+use crate::alloc::string::ToString;
```
nuke-web3, line 21: TIL the guest does use `alloc` in the inverse of this PR: https://github.com/risc0/risc0/blob/8753f2c712280c2688e6b4492e8184e4572882fc/risc0/zkvm/src/serde/deserializer.rs#L15 
Should we ensure we can do a round trip of `NaiveDate` or perhaps some more generic type in tests?

SchmErik: I did a round-trip test locally and this didn't require any code changes. Now that i'm looking at this more closely,  round-trip tests are in the mod.rs so I'll remove this test and add a round trip test there. Thanks!

risc0/zkvm/src/serde/serializer.rs
SchmErik: This code was inspired by what the serde crate does. I'm not sure why it was implemented as a panic but in the past, I've fixed floating point numbers panic on serde so I think it was left out in the initial implementation. The tests are passing so we're definitely not regressing AFAIK

SchmErik:
> Thanks for the patch! Seems kinda "dirty" to implement _just_ for this type for one crate, but IIUC this should help serialize with `ToString` now used that is much more generally applicable to people?

I based this solution off of the serde crate. Serialization was failing for NaiveDate because it was hitting this collect_str function so I think this change (which is modeled after the serde crate itself) is a solution that is generally applicable

nvzqz (approved):

Issue #1597, opened 2024-03-25
[Feature] C library: support c-kzg in guest


SchmErik:
## Feature
Support https://github.com/ethereum/c-kzg-4844 in guest.

## Motivation
Customer request

## Implementation
Ensure that the rust binding for this will compile and run in the guest

PR #2045, opened 2024-07-01
Add c-kzg test as an example

SchmErik:
Some unknown issue (likely with cargo, workspaces, and or build scripts) is causing the c-kzg guest to not build in the test directory. Add c-kzg as an example as a temporary measure until we can get to the bottom of why this crate won't build when placed under the `risc0/zkvm/methods` directory.

closes: #1597


flaub:
Looks like CI is passing

flaub (approved):

Issue #1982, opened 2024-06-17
Chore: Handle links from defunct foundry template repo


nuke-web3:
Replace or remove links from now replaced foundry template repo in older version docs

_Originally posted by @nuke-web3 in https://github.com/risc0/risc0/pull/1972#discussion_r1643474156_

-  v0.18: [bonsai-starter.sol]: https://github.com/risc0/bonsai-foundry-template/blob/release-0.18/contracts/BonsaiStarter.sol
-  v0.19: [relay directory of the Bonsai Foundry Template]: https://github.com/risc0/bonsai-foundry-template/blob/release-0.19/relay/examples/offchain_request.rs

PR #2063, opened 2024-07-08
chore: remove links from defunct foundry template repo

bouewnm:
Remove links from now replaced foundry template repo. I didn't find replacement for these links.

Will close: #1982


pdg744 (approved):
Thank you! 

I opened #2072 to suggest a CI improvement so that we can avoid this moving forward.

PR #2066, opened 2024-07-09
Automatically set the branch for `risc0-ethereum`

capossele:
As currently `risc0-ethereum` main is (and should be) consistent with `risc0` main, whenever we create a back port PR to some other branch, we might see the risc0-ethereum CI failing. To solve that, this PR instructs CI to conditionally pick the branch of `risc0-ethereum` based on the name of the branch of `risc0` the PR is being merged into.


.github/workflows/risc0-ethereum.yml
```diff
@@ -82,6 +82,14 @@ jobs:
...
         with:
           path: 'risc0'
+      # Determine the branch for risc0-ethereum
+      - name: Set risc0-ethereum branch
+        run: |
+          if [[ "${{ github.event.pull_request.base.ref }}" == release-* ]]; then
+            echo "RISC0_ETHEREUM_BRANCH=${{ github.event.pull_request.base.ref }}" >> $GITHUB_ENV
+          else
+            echo "RISC0_ETHEREUM_BRANCH=main" >> $GITHUB_ENV
+          fi
```
nategraf, line 92: This feels a bit hacky to me, and raises questions like "what if that environment variable is already in the file". What do you think about using the Github Actions logical operators instead https://docs.github.com/en/actions/learn-github-actions/expressions#example

SchmErik (approved):

Issue #1658, opened 2024-04-11
Add docs about reproducible builds


Cardosaum:
## Overview

Users should understand what's the concept of a reproducible build, why is it important, how do we achieve reproducibility etc. 

## Tasks to be done

- [ ] Write a new page on the [dev docs](https://dev.risczero.com/api) about reproducible builds
  - [ ] Should be under "Writing your zkVM application" section (the in depth explanations about the zkVM)
- [ ] Explain what is the a reproducible build
- [ ] Explain why is the reproducibility of the guest code important
- [ ] Explain how we achieve reproducibility (mention we need docker for it, `cargo` is not deterministic by default etc)
- [ ] Explain how users can build their ELFs in a reproducible way (mention `cargo risczero build`, `cargo risczero deploy` and also `risc0_build` if they want to build elfs from rust code)

pdg744:
transferring this from dev rel board to platform board

pdg744:
I think it's probably best to hold off on adding a dedicated page for this until we do some engineering re: #2075 -- until then, we don't have a good story for how foundry template users can access determinstic builds. 

In the mean time, we have the terminology entry: https://dev.risczero.com/terminology#deterministic-builds

The revisions to the terminology entry in #2074 improve the terminology entry to reflect the what/why/how items in the checklist above.

pdg744:
Closing because this issue is redundant with #1084.

Issue #1084, opened 2023-09-19
Document `cargo risczero build`


tzerrell:
Prior discussion:
> I think there's instructions for how to run the tool, but not good documentation about why it's necessary, what the limitations are, etc.

Current plan is to create a page on dev.risczero.com covering why one would want to use this (i.e. for reproducible build) and what it's functionality and limitations are.

nategraf:
Adding some more context on the requirements here:

Verifying image IDs by building from source is part of the security model for many zkVM applications. In particular, applications integrating with smart contracts will generally need to provide users with a process by which they can reproduce the image ID from source.

We need to have a workflow that we can provide zkVM app developers, and the app developers can provide the consumers of their code, to run this reproducible build. This workflow should be generic across projects, both using the Foundry template and not, and easy to follow.

Cardosaum:
Related to #1658

pdg744:
I think it's probably best to hold off on adding a dedicated page for this until we do some engineering re: #2075 -- until then, we don't have a good story for how foundry template users can access determinstic builds. 

In the mean time, we have the terminology entry: https://dev.risczero.com/terminology#deterministic-builds

PR #2074, opened 2024-07-09
Revising terminology entry for deterministic builds

pdg744:
We have a couple of open tickets (#1084, #1658) related to improving our documentation around reproducible/deterministic builds. 

This PR makes some minimal revisions to the existing terminology entry in order to clarify the meaning/importance of deterministic builds and how to use them.

IMO, we should hold off on adding a dedicated page on this topic until we address the engineering question of how to integrate deterministic builds into the foundry template developer workflow (which currently does not include `cargo risczero`). See #2075 & https://github.com/risc0/risc0-ethereum/issues/141

### DIFF
--- a/website/docs/terminology.md
+++ b/website/docs/terminology.md
@@ -77,7 +77,7 @@ See also: [Code reference for control root]
 
 A compilation process is called "deterministic" (or "reproducible") if it reliably produces the same binary file, on a byte-by-byte level.
 
-In the context of RISC Zero application development, deterministic builds are necessary to ensure a clear linkage between the [guest program] and the resulting [ImageID].
+In the context of RISC Zero application development, deterministic builds are necessary to ensure a clear linkage between the source code for the [guest program] and the resulting [Image ID][ImageID].
 
 To access deterministic builds for your zkVM application, use [`cargo risczero build`]. Deterministic builds are made possible by running the `rustc` compiler inside a Docker container.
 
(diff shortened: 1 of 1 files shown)


SchmErik (approved):
Looks good aside from the small suggestion

website/docs/terminology.md
```diff
@@ -75,10 +75,11 @@ See also: [Code reference for control root]
 
 ### Deterministic Builds
 
-Our [cargo risczero] tool offers a deterministic compilation from user-written Rust code to RISC-V binary.
-Compiled binary files are identified by their [imageID].
+A compilation process is called "deterministic" (or "reproducible") if it reliably produces the same binary file, on a byte-by-byte level.
 
-Deterministic builds are made possible by running the `cargo` compiler inside a Docker container. Without deterministic builds, the zkVM only proves correct execution of a given binary file. With deterministic builds, users can prove correct execution of Rust code.
+In the context of RISC Zero application development, deterministic builds are necessary to ensure a clear linkage between the [guest program] and the resulting [ImageID].
```
nategraf, line 80: I'd recommend clarifying this to be in reference to the source code, because I first read this as referring to the ELF, from which there is already a deterministic translation to an image ID. (Also, in prose, I feel like there should be a space in "Image ID")
```suggestion
In the context of RISC Zero application development, deterministic builds are necessary to ensure a clear linkage between the source code for [guest program] and the resulting [Image ID].
```

nategraf (approved):

website/docs/terminology.md
pdg744:
```suggestion
In the context of RISC Zero application development, deterministic builds are necessary to ensure a clear linkage between the source code for the [guest program] and the resulting [Image ID][ImageID].
```

pdg744:
> Also, in prose, I feel like there should be a space in "Image ID"

hmm ya i guess i agree. this applies to the whole terminology page; will leave fixing it elsewhere for a future PR.

PR #2084, opened 2024-07-13
ZKVM-480: Add experimental support for sys_fork

flaub:


risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -88,8 +88,14 @@ pub trait SyscallContext {
         Ok(region)
     }
 
+    /// Load a page from memory at the specified page index.
```
shkoo, line 91: Can we document what the difference is between peek_page and peek_region and why we'd use one over the other?

risc0/zkvm/src/host/server/exec/syscall/fork.rs
shkoo, line 1: This seems like a fair amount of duplicated code. :(  But I guess copy-paste is an expedient solution.

flaub: Agreed, might be worth a refactor at some point. It's complicated by the `ExecutorEnv` being pretty far removed from the rv32im circuit. It wouldn't be too much trouble to move `ExecutorEnv` down into `risc0-circuit-rv32im` and re-export it, except for the things related to assumptions.

risc0/zkvm/src/host/server/exec/syscall/fork.rs
```diff
@@ -0,0 +1,333 @@
...
+impl<'a, 'b> ChildExecutor<'a, 'b> {
+    pub fn new(ctx: &'b mut dyn SyscallContext<'a>) -> Result<Self> {
+        // Start the child process one instruction past the ecall.
+        let pc = ByteAddr(ctx.get_pc() + WORD_SIZE as u32);
+
+        let mut registers = std::array::from_fn(|idx| ctx.load_register(idx));
+        // The return value of sys_fork is the pid, so set a0 to PID_CHILD.
+        registers[REG_A0] = PID_CHILD;
+
+        let syscall_table = ctx.syscall_table().clone();
```
SchmErik, line 75: This code is pretty new to me so I might be wrong here but it seems like this allows the child to inherit the parent process's syscall and I'm wondering if this means that it will allow the child to do things like write to the journal. Writing to the journal will result in a failure but I'm not sure when it will fail. It would be nice to catch something like this during the execution of the child process..

risc0/zkvm/src/host/server/exec/syscall/fork.rs
```diff
@@ -0,0 +1,333 @@
...
+        bytes.clone_from_slice(&self.0[byte_addr..byte_addr + WORD_SIZE]);
+        let data = u32::from_le_bytes(bytes);
+        tracing::trace!("load({addr:?}) -> 0x{data:08x}");
+        data
+    }
+
+    fn store(&mut self, addr: WordAddr, data: u32) {
+        let word_addr = (addr.0 % PAGE_WORDS as u32) as usize;
+        let byte_addr = word_addr * WORD_SIZE;
+        tracing::trace!("store({addr:?}, 0x{data:08x})");
```
nvzqz, line 330: The formatting specifier `{data:#010x}` would also work like `0x{data:08x}`, but perhaps it's less clear.

risc0/zkvm/src/host/server/exec/syscall/pipe.rs
```diff
@@ -0,0 +1,67 @@
...
+};
+
+use anyhow::{anyhow, Result};
+
+use crate::host::client::posix_io::PosixIo;
+
+use super::{Syscall, SyscallContext};
+
+// This is an arbitrary maximum number of open file descriptors allowed.
+const MAX_FD: u32 = 1000;
```
nvzqz, line 29: Would we want to make this configurable in the future?

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -785,3 +787,45 @@ pub unsafe extern "C" fn sys_verify_integrity(
...
+/// `sys_fork()` creates a new process by duplicating the calling process. The
+/// new process is referred to as the child process. The calling process is
+/// referred to as the parent process.
+///
+/// The child process and the parent process run in separate memory spaces. At
+/// the time of `sys_fork()` both memory spaces have the same content.
+///
+/// # Return Value
+///
+/// On success, the PID of the child process (1) is returned in the parent, and
```
nategraf, line 800: Is this modeled as a PID simply to align with POSIX conventions? In practice I imagine it is essentially a boolean to say whether the current process is parent or child.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -785,3 +787,45 @@ pub unsafe extern "C" fn sys_verify_integrity(
...
+/// 0 is returned in the child. On failure, -1 is returned in the parent, no
+/// child process is created.
```
nategraf, line 802: Why return on failure, rather than simply ending execution? In the case of an honest prover, it's unclear when you'd want to continue. In the case of a dishonest prover, success and failure don't mean very much.

risc0/zkvm/src/host/server/exec/syscall/fork.rs
SchmErik: > > if this means that it will allow the child to do things like write to the journal
> 
> Given that `fork(2)` makes child processes inherit all file descriptors, I believe this would be expected behavior

Great call out - I didn't really think of it this way

> The intention with inheriting the syscall table is to allow disabling of syscalls in the child, such as to prevent fork bombs. Currently syscall data is coupled with its logic. Perhaps we should make a clear separation for data that gets copied vs copy-on-write vs shared?

I think this is a great approach! - maybe it might be good to allow SYS_WRITE to stdout/stderr for easier debugging but do some copy on write or no-op for journal? Anyway, thanks for the response, this was very helpful

nategraf (commented):
Forking interacts with proving in some potentially interesting ways. We are imagining use cases where the the parent is proven and the child is not. But a prover can prove the child code path instead, or both. What's especially interesting is that this might lead to two semantically different outcomes.

As a concrete example, we could imagine a guest that checks some signed data to be valid and halts with an empty journal and exit code `0` if everything checks out. If they use an accelerator for the signature check, and use an unconstrained function with `sys_fork` in that flow, there could be an issue. If the child process halts with an empty journal and exit code `0`, the prover could chose to prove the child process path instead of the parent. It would still have the same image ID, journal, and exit code (i.e. it would have the same receipt claim), even if it actually bypasses all the meaningful logic.

One way to mitigate this would be to use a different exit code for the end of the child process. Another would be to add an `assert!(false)` to the child process, which an executor can skip but a prover cannot.

Another option would be to decide we don't actually want to use `fork` semantics for unconstrained functions, and instead have an ecall that creates a clone of the process with a new PC, jumping into a block that is unreachable in the proven parent execution.

nategraf:
Related to https://github.com/risc0/risc0/pull/2084#pullrequestreview-2197703844, I just realized to other solutions to this question when using `sys_fork` for unconstrained functions.

One is to end the child with a hard loop and some kind of syscall that doesn't halt, but indicates to the host that the process is done. The other is to return to the same execution path as the parent. In either case, the child code does not contain a halt instruction. I personally favor the case of having the guest go into a hard loop.

In general, the halt instruction is kind of special in the zkVM. In many ways, the proofs we generate are "reach-ability proofs", in that the statement says "I know an execution path that reaches a halt instruction with the given journal". So when there are multiple paths in the program that reach a halt instruction, the statement is that one of those paths was taken, without any indication of which one.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -218,8 +220,8 @@ macro_rules! impl_syscall {
                 ::core::arch::asm!(
                     "ecall",
                     in("t0") $crate::syscall::ecall::SOFTWARE,
-                    inout("a0") from_host => a0,
-                    inout("a1") from_host_words => a1,
+                    inlateout("a0") from_host => a0,
```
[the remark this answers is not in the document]
flaub: me too!

risc0/zkvm/platform/src/syscall.rs
flaub: Because this is the POSIX spec (I copied all this from the man page)

risc0/zkvm/platform/src/syscall.rs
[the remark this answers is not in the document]
nvzqz: Going off of [the `asm!` RFC](https://rust-lang.github.io/rfcs/2873-inline-asm.html#late-output-operands), it seems to just be an optimization. It seems `inout` like before would have also been correct.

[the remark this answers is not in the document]
flaub: That was my take as well. In any case, the compiler was emitting incorrect instructions (it was missing the load of the lower bits of the immediate value). An alternative fix could be to use `.s` files so that we have more precise control of the instructions. 

I'll try the alternative out because I don't trust this `inlateout` fix.

risc0/zkvm/platform/src/syscall.rs
flaub: Yep

risc0/zkvm/src/host/server/exec/syscall/pipe.rs
flaub: Perhaps

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
flaub: I added to the `peek_region`:

> A region may span multiple pages.

risc0/zkvm/platform/src/syscall.rs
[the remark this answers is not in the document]
nvzqz: For reference, could you please share the previous `inout` codegen and new `inlateout` version? It might be worth filing an issue

risc0/zkvm/src/host/server/exec/syscall/fork.rs
flaub: I pushed an update that removes the JOURNAL fd from the child, it also removes the sys_fork entry to prevent fork bombs.

risc0/zkvm/methods/guest/src/bin/multi_test.rs
```diff
@@ -343,5 +349,35 @@ fn main() {
...
         }
+        MultiTestSpec::SysFork => {
+            const MSG: &[u8] = b"hello";
+            let mut pipe = [0u32; 2];
+            unsafe { sys_pipe(pipe.as_mut_ptr()) };
+            let pid = sys_fork();
+            if pid == 0 {
+                env::log("child");
+                let mut writer = FdWriter::new(pipe[1], |_| {});
+                writer.write_slice(&MSG);
```
nvzqz, line 360:
```suggestion
                writer.write_slice(MSG);
```

flaub:
I tried to address all the comments. The main thing is changing the exit condition for the child from being `sys_halt` to a new syscall called `sys_exit`. This both preserves POSIX semantics and provides a model whereby the zkvm proves the reach-ability of the entire machine getting into a halted state. `sys_exit` is merely one of the processes in that machine ending.

risc0/zkvm/platform/src/syscall.rs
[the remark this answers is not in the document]
shkoo: Issue filed: https://github.com/rust-lang/rust/issues/128212

nategraf (approved):

Issue #2020, opened 2024-06-25
[BUG] yarn/bun build reports broken anchors


pdg744:
## Bug Report
Running `yarn build` from the `website/` directory shows the following issues:
<img width="885" alt="image" src="https://github.com/risc0/risc0/assets/15272444/89a2ff5a-6fa4-4330-8fe1-30ceeb33abd3">

## Expected behavior
Would be great if `yarn build` returned a clean report.

pdg744:
Similar output when running `bun run build`:

```
Please check the pages of your site in the list below, and make sure you don't reference any anchor that does not exist.
Note: it's possible to ignore broken anchors with the 'onBrokenAnchors' Docusaurus configuration, and let the build pass.

Exhaustive list of all broken anchors found:
- Broken anchor on source page path = /api/0.18/bonsai/eth-examples:
   -> linking to /api/zkvm/quickstart#remote-proving
- Broken anchor on source page path = /api/0.19/bonsai/eth-examples:
   -> linking to /api/zkvm/quickstart#remote-proving
- Broken anchor on source page path = /api/0.20/bonsai/eth-examples:
   -> linking to /api/zkvm/quickstart#remote-proving
- Broken anchor on source page path = /api/0.21/blockchain-integration/contracts/verifier:
   -> linking to /terminology#zkvm-program
- Broken anchor on source page path = /api/next/blockchain-integration/contracts/verifier:
   -> linking to /terminology#zkvm-program
- Broken anchor on source page path = /api/blockchain-integration/contracts/verifier:
   -> linking to /terminology#zkvm-program
- Broken anchor on source page path = /contributors-guide:
   -> linking to /contributors-guide#categories-of-documentation
- Broken anchor on source page path = /faq:
   -> linking to #faq (resolved as: /faq#faq)
- Broken anchor on source page path = /terminology:
   -> linking to #control-columns (resolved as: /terminology#control-columns)
```

pdg744:
Perhaps we should be asking CI to check for existence of these broken anchors...

pdg744:
@nahoc i think this is done now?

nahoc:
> @nahoc i think this is done now?

yep!

PR #2094, opened 2024-07-16
fix(website): broken links

nahoc:
This fixes all broken links that showed up when running `bun run build` aswell as the warning for `remark/rust`.

Solves https://github.com/risc0/risc0/issues/2020

Some links were just flat out removed, as I couldn't even find a trace of them working in the git history.

### Before

image

### After

image


website/api/blockchain-integration/eth-examples.md
```diff
@@ -18,8 +18,8 @@ The [RISC Zero Foundry Template][foundry-template] provides a minimal applicatio
 
 This [example app][governance-example] uses Bonsai as an Ethereum coprocessor. The protocol, based on the OpenZeppelin [Governor smart contract standard], batches signature verifications off-chain for a DAO governance vote. The end result is that in [~160 lines of Rust][signature-aggregation], a gas savings of 66% is achieved with significant room for optimizations.
 
-[zkvm-examples]: ../zkvm/examples.md
-[remote-proving]: ../generating-proofs/remote-proving.md
+[zkvm-examples]: /api/zkvm/examples
```
pdg744, line 21: For docs in `/website/api` that are pointing to other docs in `website/api`, we should use relative links, for two reasons: 

1. The contents in `website/api/` get rendered under `next` on the website, and links on those pages should point to other docs on `next`. 
2. When we push a new release and move the docs into `website/api_versioned_docs/`, the relative paths will still work.

nahoc: does that mean that _all_ absolute links are wrong? Because I see quite a lot of them atm

nahoc: what I mean is should I change all occurences of `[zkvm-examples]: /api/zkvm/examples` to `[zkvm-examples]: ../zkvm/examples`

pdg744: The only place we need to be using absolute links is for links that go back and forth between docs in the versioning system (i.e. in `website/api/` or in `website/api_versioned_docs/`) and docs that aren't in the versioning system (i.e. those in `website/docs/`, including the terminology page & faq). 

I'm not super concerned about getting rid of all the absolute links from `website/api_versioned_docs/`, but it would be nice from a maintenance perspective for those links to point to the associated content in that version's docs.

Does that answer your question?

website/api_versioned_docs/version-0.18/bonsai/eth-examples.md
```diff
@@ -11,7 +11,7 @@ While all of the [zkVM examples][zkvm-examples] can be run on Bonsai by [configu
 This [example app][governance-example] uses Bonsai as an Ethereum coprocessor. The protocol, based on the OpenZeppelin [Governor smart contract standard], batches signature verifications off-chain for a DAO governance vote. The end result is that in [~160 lines of Rust][signature-aggregation], a gas savings of 66% is achieved with significant room for optimizations.
 
 [zkvm-examples]: /api/zkvm/examples
-[remote-proving]: /api/zkvm/quickstart#remote-proving
+[remote-proving]: /api/generating-proofs/remote-proving
```
pdg744, line 14: continuing the conversation about absolute vs. relative paths -- this works for now but will break if we ever do a docs re-organization. 

pointing to the associated reference on the `0.18` page would be better. I think that link is:
```suggestion
[remote-proving]: ../zkvm/quickstart#remote-proving
```

website/api_versioned_docs/version-0.19/bonsai/eth-examples.md
```diff
@@ -11,7 +11,7 @@ While all of the [zkVM examples][zkvm-examples] can be run on Bonsai by [configu
 This [example app][governance-example] uses Bonsai as an Ethereum coprocessor. The protocol, based on the OpenZeppelin [Governor smart contract standard], batches signature verifications off-chain for a DAO governance vote. The end result is that in [~160 lines of Rust][signature-aggregation], a gas savings of 66% is achieved with significant room for optimizations.
 
 [zkvm-examples]: /api/zkvm/examples
-[remote-proving]: /api/zkvm/quickstart#remote-proving
+[remote-proving]: /api/generating-proofs/remote-proving
```
pdg744, line 14:
```suggestion
[remote-proving]: ./bonsai-overview
```

pdg744 (approved):
Thank you! 

I don't think we need to worry about getting rid of the absolute links in old versions of docs at the moment, so you don't necessarily need to incorporate the additional suggestions above for the links in `api_versioned_docs/`

website/api_versioned_docs/version-0.18/bonsai/eth-examples.md
pdg744: adding on to that thought-- if we had been using a relative link in the first place, the content migration from `zkvm/quickstart#remote-proving` to `generating-proofs/remote-proving` wouldn't have broken our link

website/api/blockchain-integration/eth-examples.md
nahoc: yeah it does, thanks for that!

OK, well I have another big PR that includes remark linting [here](https://github.com/risc0/risc0/pull/2096), so maybe I add a rule to enforce relative links somehow.. will do some research

PR #2112, opened 2024-07-20
Examples: add version warning to prevent users from running incompatible code

SchmErik:
This change mitigates issues that many users run into with our examples. When new users try our software, they clone the repository and try the examples but they run into runtime errors because the example code they are trying to run uses pre-released code, their server is incompatible with the code. The failure will happen during runtime. This change adds functionality in the build script to detect this version mismatch during build time instead of runtime. This is done by exposing a `risc0_zkvm::get_server_version`that the build script calls in order to get the major and minor version of the server they're running and compares this semver to `risc0_zkvm::version` which represents the client version. If there's a mismatch, there's a warning printed on the terminal that directs the user to run `git checkout origin/release-$(MAJOR_VERSION).$(MINOR_VERSION)`. I hope this is clear enough for users to follow. I've also added the ability to turn off this version check if users chose to run the examples using the `-F prove` flag.


flaub:
I think we also need to forward the `cuda` feature flag.

flaub (approved):

flaub:
Going to use https://github.com/risc0/risc0/pull/2123 instead.

PR #2123, opened 2024-07-24
client/server: improve version mismatch error

SchmErik:
This change implements version checking on the client side. Upon creation of the ParentProcessConnection, the client will query the r0vm binary for its version and emit a verbose error message upon a version mismatch. We have previously discussed that this could be a handshake bewteen the client and server but I saw a downside to this solution: if we change the server to get this to work using `Client.connect`, it means that this solution will not work for older servers and will only work for new servers. For example, if a user decides to use an older server (i.e. v0.21) on a newer client, we will end up getting the old error message which was difficult to understand:

```
thread 'main' panicked at risc0/r0vm/src/lib.rs:192:18:
called `Result::unwrap()` on an `Err` value: incompatible client version: 1.1.0-alpha.1, server version: 1.0.3
thread 'main' panicked at host/src/main.rs:41:10:
called `Result::unwrap()` on an `Err` value: failed to fill whole buffer
```

Improving client/server compatibility based off of changing both client and server will only work for new servers. Where as changing only the client will allow the version check to work for older servers.

This implementation will result in an error during proving or execution. We discussed implementing this check as a part of `default_prover()` but this function returns a trait object. Since `default_prover` is infallable, including the compatibility check as a part of this function will either force panics or change the type signature of this function. Therefore, I went with adding this check at the time of the subprocess connection which happens at the beginning of execution.


flaub:
I think working for only new servers is acceptable

risc0/zkvm/Cargo.toml
```diff
@@ -70,6 +70,7 @@ nvtx = { version = "1.3", optional = true }
 prost = { version = "0.13", optional = true }
 rand = { version = "0.8", optional = true }
 rayon = { version = "1.5", optional = true }
+regex = "1.10"
```
flaub, line 73: Maybe make this optional so that guest-side doesn't need to pull it in?

flaub (approved):
Other than avoiding `regex` from being pulled into guest-side, this looks ok to me

PR #2128, opened 2024-07-25
feat: add google search console

nahoc:


website/docs/proof-system/stark-by-hand.md
```diff
@@ -429,6 +429,6 @@ Whew! Congratulations and thank you for making it this far!
 
 Got questions, feedback, or corrections? Find us on [Twitter](https://twitter.com/risczero) and [Discord](https://discord.gg/risczero).
 
-[cryptographic security model]: api/security-model
+[cryptographic security model]: /api/security-model
```
pdg744, line 432: I think this is a situation where we want the absolute path because it's going back-and-forth between `docs` and `api`

pdg744 (approved):

website/docs/proof-system/stark-by-hand.md
nahoc: I get a build error from docusaurus saying the URL is broken otherwise 🤔

pdg744: I feel like I got myself turned around. 
IIUC, we want the absolute path here, and your suggested change uses the absolute path correctly.

PR #2136, opened 2024-07-28
correction of pathway for getting started page

Budalebah:
Correction of pathway for getting started page. 
The page referred to is the Getting Started page but the link leads to quickstart.


Budalebah:
Check please

flaub:
This link is correct and not broken

PR #2149, opened 2024-07-30
CI: replace cargo risczero install with rzup install

SchmErik:
This  replaces uses of `cargo risczero install` with the new `rzup install rust $toolchain_version`. There are several workloads that utilize the cargo risczero utility so I left the CI rules for those intact. I will follow up with release.yml changes separately.


SchmErik:
putting this into draft mode until I do more experimentation

pdg744:
Does this change merit any changes to installation instructions, or any other docs?

SchmErik:
> Does this change merit any changes to installation instructions, or any other docs?

no, not yet. we haven't deployed rzup and docs will be updated when it goes live

risc0/build/src/lib.rs
```diff
@@ -39,6 +39,11 @@ use serde::Deserialize;
 
 pub use docker::{docker_build, BuildStatus, TARGET_DIR};
 
+// in order to tell the difference between the rzup bash script and the new rust
+// rzup implementation, the new implementation will place a file with this name
+// under risc0_data. This is used to indicate to the client and risc0-build that
```
flaub, line 44:
```suggestion
// under `$risc0_home`. This is used to indicate to the client and risc0-build that
```

rzup/src/main.rs
```diff
@@ -116,7 +118,24 @@ async fn main() {
     }
 }
 
+fn init() -> Result<()> {
+    // this is a rust revamp of rzup. Before this version, rzup was a bash
+    // script. In order to facilitate compatibility within our tools, we touch a
+    // file in a known location to indicate to the client that the rust
+    // impelemtnation of rzup is being utilized.
+    let new_rzup_indicator = utils::rzup_home()?.join("new-rzup");
```
flaub, line 126: Can we depend on `risc-build` so that we don't have to duplicate this and have the possibility of it getting out of sync?

risc0/build/src/lib.rs
```diff
@@ -39,6 +39,11 @@ use serde::Deserialize;
...
+// the new rust implementation of rzup is being used.
+const NEW_RZUP_INDICATOR_NAME: &str = "new-rzup";
```
flaub, line 46:
```suggestion
const NEW_RZUP_INDICATOR_NAME: &str = ".rzup";
```

flaub (approved):

PR #2150, opened 2024-07-30
Receipt now supports borsh serialization

ercecan:
This pr introduces borsh serialization for zkvm receipts

I can make required changes to the pr if needed


ercecan:
I can also make this a feature and derive serializations with 
```rust
#[cfg_attr(feature="borsh", derive(BorshSerialize, BorshDeserialize))]
```
You can lmk if that's better

flaub:
I think we will need to make a feature flag for this since some of our tests and users want to run in a no-std/WASM environment which apparently doesn't work with borsh. I could be wrong about that, but one of our WASM tests are failing: https://github.com/risc0/risc0/actions/runs/10162990313/job/28144237730?pr=2150

flaub:
Additionally, to pass the CI checks, we require that all the `Cargo.lock` files are up-to-date. You can do:
```
cd examples && cargo check
cd benchmarks && cargo check
```
To have the `Cargo.lock` updated, then commit the results.

austinabell:
> I think we will need to make a feature flag for this since some of our tests and users want to run in a no-std/WASM environment which apparently doesn't work with borsh. I could be wrong about that, but one of our WASM tests are failing:

So borsh does support no_std, but you need to conditionally enable the "std" feature for borsh, and turn off default features.

Seems reasonable to still put borsh under a feature flag in the risc0 crates, to avoid breaking changes or dep bloat imo though.

ercecan:
> Additionally, to pass the CI checks, we require that all the `Cargo.lock` files are up-to-date. You can do:
> 
> ```
> cd examples && cargo check
> cd benchmarks && cargo check
> ```
> 
> To have the `Cargo.lock` updated, then commit the results.

Will do


> So borsh does support no_std, but you need to conditionally enable the "std" feature for borsh, and turn off default features.
> 
> Seems reasonable to still put borsh under a feature flag in the risc0 crates, to avoid breaking changes or dep bloat imo though.

I will enable the "std" feature and put borsh under a feature flag

risc0/binfmt/Cargo.toml
```diff
@@ -36,4 +39,6 @@ std = [
   "serde/std",
   "tracing/log",
   "tracing/std",
+  "borsh/std",
```
flaub, line 42:
```suggestion
  "borsh?/std",
```

ercecan: Got it, fixing now

flaub:
Looks like the lock files need to get updated again.

PR #2156, opened 2024-07-30
CI: add workflow to upload `rzup` binary to s3

SchmErik:
This solution enables the deployment of rzup via a manual workflow dispatch. The idea here is that we can land this, manually trigger the workflow now. I'm thinking that we can do this as a push event to main and add paths filters in the near future but this and gives us a starting place to experiment with the deployment before we lock it down further..


.github/workflows/rzup_release.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          lfs: true
+      - uses: ./.github/actions/rustup
+      - uses: ./.github/actions/sccache
+        with:
+          key: ${{ env.TARGET }}
+      - name: Generate `rzup` Binary
+        run: cargo build --release -p rzup
+      - name: Upload Binary to Artifacts S3 Bucket
+        run: aws s3 cp $BINARY_OUTPUT_PATH $S3_ARTIFACTS_PATH/
```
flaub, line 36: Should we also upload the install script?

PR #2158, opened 2024-07-30
Extend `testutil` loop to support dynamic iters

nvzqz:
This is implemented by packing the iteration count as a `u32` directly inside the guest program's RISC-V instruction stream.


flaub (approved):

nvzqz:
Closing this in favor of the approach I went with in #2196, which is to instead patch a pre-compiled ELF.

PR #2166, opened 2024-07-31
Add `embed_metdata` build option to reduce build times

preston-evans98:
# Introduction
As of right now, a standard Risc0 build consists of three components:
1. Build the guest `elf` with a recursive `cargo build` invocation
2. Compute the ImageID of the guest elf
3. Generate a Rust package exporting the ImageID and elf `const METHOD_ELF: &[u8] = ...`.

This PR adds "light" builds - an opt-in mechanism to skip steps 2 and 3. This drastically reduces the incremental build time, making it much more feasible to compile Risc0 applications as part of the development loop. Unlike the naive strategy of skipping Risc0 compilation, `light` builds still allow detection of compilation issues in the guest binaries - and even proof generation - without waiting on a full build cycle.

## Performance

The timings in this section were generated by compiling the Sovereign SDK demo-rollup. This rollup builds two guest `elf` files, each one of which is about 20MB in size. All timings shown come from hot incremental builds - meaning that each command was invoked twice in succession and only the second output is shown below.

### No Guest Build
Skipping the `risc0_build` invocation entirely, a hot incremental build of the rollup takes 1 second:
```
~/Developer/sovereign/v1/sovereign-sdk$ time SKIP_GUEST_BUILD=1 cargo check                  
   Compiling sov-module-schemas v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/crates/module-system/module-schemas)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.97s
SKIP_GUEST_BUILD=1 cargo check  0.40s user 0.24s system 60% cpu 1.062 total
```

Note: In this configuration the guest binary is not built at all - so any issues in the guest code are undetectable.

### Light Build
Using the light build (skipping steps 2-3) an incremental build takes 2-3 seconds:
```
~/Developer/sovereign/v1/sovereign-sdk$ time cargo check                   
   Compiling risc0 v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/examples/demo-rollup/provers/risc0)
   Compiling sov-module-schemas v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/crates/module-system/module-schemas)
sov-demo-prover-guest-celestia: Starting build for riscv32im-risc0-zkvm-elf, risc0(build)                                                      
sov-demo-prover-guest-celestia:     Finished `release` profile [optimized + debuginfo] target(s) in 0.27s
sov-demo-prover-guest-mock: Starting build for riscv32im-risc0-zkvm-elfisc0(build)                                                             
sov-demo-prover-guest-mock:     Finished `release` profile [optimized + debuginfo] target(s) in 0.24s                                          
Finished risc0 build script in 691 ms
    Checking sov-demo-rollup v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/examples/demo-rollup)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.93s
cargo check  1.25s user 0.75s system 98% cpu 2.026 total
```

### Standard Build
Running an incremental build with the standard setup (`embed_methods_with_options`) takes between 30 and 60 seconds, depending on the run.
```
~/Developer/sovereign/v1/sovereign-sdk$ time cargo check
   Compiling risc0 v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/examples/demo-rollup/provers/risc0)
   Compiling sov-module-schemas v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/crates/module-system/module-schemas)
sov-demo-prover-guest-celestia: Starting build for riscv32im-risc0-zkvm-elfschemas(build)                                                      
sov-demo-prover-guest-celestia:     Finished `release` profile [optimized + debuginfo] target(s) in 0.44s                                      
sov-demo-prover-guest-mock: Starting build for riscv32im-risc0-zkvm-elf
sov-demo-prover-guest-mock:     Finished `release` profile [optimized + debuginfo] target(s) in 0.26s
Finished risc0 build script in 6080 ms 
    Checking sov-demo-rollup v0.3.0 (/Users/prestonevans/Developer/sovereign/v1/sovereign-sdk/examples/demo-rollup)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 32.25s
cargo check  23.03s user 4.84s system 86% cpu 32.390 total
```

## Analysis
As you can see in the timings, the vast majority of the total build time is spent outside of the build script. In other words, invoking cargo to rebuild the guest binaries is not the bottleneck! Based on the simple timing information available here, it looks like...
- Invoking cargo recursively accounts for about 3% of the total work
- Computing image IDs is about 15% of the total
- Compiling/linking the `"methods.rs"` file accounts for the ~80% of the total build time.

 Given this data, I'd propose adding a method to opt-out of ImageID and generation of the `const METHOD_ELF`. The prototype in this PR shows how such functionality can be implemented.


flaub:
Amazing! I wonder if we could toggle this with an environment variable rather than a feature flag since it seems like a dev might want to switch between modes depending on the situation. It'd also be nice to be able to commit the 'production-ready' or 'CI-ready' code to the repository.

preston-evans98:
> Amazing! I wonder if we could toggle this with an environment variable rather than a feature flag since it seems like a dev might want to switch between modes depending on the situation. It'd also be nice to be able to commit the 'production-ready' or 'CI-ready' code to the repository.

Unfortunately, we can't do a runtime switch without sacrificing a bit of safety unless we're willing to break semver. The core issue is that `embed_methods_with_options` returns*  `Vec<GuestListEntry>`, and `GuestListEntry` has public fields for `elf` and `imaged_id`. If we want to return that type, we'll have to put placeholder values in the `elf` and `image_id` fields, which seems like a footgun. In the current draft, I've opted instead to add a separate return type for light builds - but that means that we can't very easily switch between build methods at runtime.

preston-evans98:
> Why not fill in the actual values for the return of the `Vec<GuestListEntry>`? It seems like the slow part is related to the `methods.rs` file that we generate, so it doesn't seem like there's any harm in returning the full ELF contents there. 

To clarify, you're suggesting that we fill the ELF contents in the return value but omit them from the `codegen_consts` for "light" builds? That seems reasonable to me, although that will mean that enabling the `guest-list` feature essentially undoes the benefit of using the lighter build process.

> Also, is it OK to still do compute_image_id or was this deemed too slow? We used to have a caching mechanism, whereby we'd do a sha256 of the ELF to obtain a key that we could use to load an existing image_id from disk so we could avoid recomputing it if the ELF hasn't changed since last build.

Even if there was caching, I think computing the image ID would be a mild pain point. It adds 6-10 seconds to every build, which is pretty significant when you think about how often you want to run `cargo check`. So my preference would be to have some option to skip it - but that's definitely a lower priority than addressing the `methods.rs` compile time.

flaub:
It seems like the `guest-list` feature is orthogonal. We could avoid writing to the `methods.rs` even in that case, if the `light` feature is set.

flaub:
When we did caching of `image_id`, we used a `sha256` of the entire ELF, and this takes about 0.017s on my machine for a 40MB file. So the first run would take ~6 seconds, but every subsequent run should be on the order of 20ms.

flaub:
Oops, now I see that the `guest-list` also writes out the ELF contents, I guess I hadn't realized that.

preston-evans98:
> When we did caching of `image_id`, we used a `sha256` of the entire ELF, and this takes about 0.017s on my machine for a 40MB file. So the first run would take ~6 seconds, but every subsequent run should be on the order of 20ms.

Yeah, the cache hits will be no problem at all - the concern here is just that the caching won't help when there's any change at all to the binary - which is the common case for `cargo check`.

flaub:
I had been working on a proc-macro based way to embed ELF binaries, but found that has the fatal flaw of not being able to know when the guest ELF needs to be rebuilt.

I think this is a reasonable solution for now.

flaub (approved):
LGTM, however it looks like some of the clippy lints are failing

preston-evans98:
Clippy lints should be fixed now - thanks @SchmErik for the patch!

PR #2174, opened 2024-08-02
feat(zkVM): Feature flag to use `embedded-allocator` as a dynamic heap allocator

hashcashier:
This adds a `heap-embedded-alloc` feature to `risc0-zkvm` and `risc0-zkvm-platform` crates. When used, the default bump allocator is replaced with a linked-list allocator managed by the `embedded-allocator` crate, allowing for reuse of deallocated memory.

This PR also modifies `send_recv_slice` to allocate memory using the global allocator rather than by explicitly invoking the bump allocator's `sys_alloc_words` function to ensure the method stays compatible, with a note that `send_recv_slice` will not reclaim memory used to store host responses.


risc0/zkvm/platform/src/rust_rt.rs
```diff
@@ -80,23 +77,50 @@ mod entrypoint {
...
+    #[cfg(feature = "heap-embedded-alloc")]
+    pub mod embedded_allocator {
+        use critical_section::RawRestoreState;
+        use embedded_alloc::Heap;
+
+        struct MyCriticalSection;
+
+        critical_section::set_impl!(MyCriticalSection);
+
+        unsafe impl critical_section::Impl for MyCriticalSection {
```
nvzqz, line 117: Perhaps worth adding a comment that it's fine for these to be no-ops since all code in the guest is single-threaded/serialized.

risc0/zkvm/src/guest/env.rs
```diff
@@ -322,10 +323,15 @@ pub fn verify_assumption(claim: Digest, control_root: Digest) -> Result<(), Infa
...
 ///
 /// On the host side, implement SliceIo to provide a handler for this call.
+///
+/// NOTE: This method never frees up the buffer memory storing the host's response.
 pub fn send_recv_slice<T: Pod, U: Pod>(syscall_name: SyscallName, to_host: &[T]) -> &'static [U] {
     let syscall::Return(nbytes, _) = syscall(syscall_name, bytemuck::cast_slice(to_host), &mut []);
     let nwords = align_up(nbytes as usize, WORD_SIZE) / WORD_SIZE;
-    let from_host_buf = unsafe { core::slice::from_raw_parts_mut(sys_alloc_words(nwords), nwords) };
+    let from_host_buf = unsafe {
+        let layout = Layout::from_size_align(nwords * WORD_SIZE, 4).unwrap();
```
nvzqz, line 332: Perhaps creating a `WORD_ALIGN` constant so that when we support 64-bit this gets automatically updated?

hashcashier: possibly, but might be out of scope for this PR since this alignment of 4 is also explicitly used in the c-kzg example and the sha2 accelerator?

risc0/zkvm/platform/Cargo.toml
```diff
@@ -11,12 +11,22 @@ repository = { workspace = true }
 # used as many places as possible to share the platform definitions.
 [target.'cfg(target_os = "zkvm")'.dependencies]
 bytemuck = { version = "1.13", optional = true }
+critical-section = { version = "1.1.2", optional = true }
```
nvzqz, line 14: As we add more optional crate features, I think it might be worth running `cargo clippy` in CI over every combination of features (or at least every individual feature) to ensure we don't accidentally ship a breaking change

risc0/zkvm/src/guest/env.rs
flaub, line 332:
```suggestion
        let layout = Layout::from_size_align(nwords * WORD_SIZE, WORD_SIZE).unwrap();
```

risc0/zkvm/platform/Cargo.toml
hashcashier: There seems to be a `clippy.py` script run in CI with some default feature flags, but given a feature it expects it to be present across all packages, so I successfully ran `cargo clippy -p risc0-zkvm -F heap-embedded-alloc` (and for `risc0-zkvm-platform`) manually for now.

risc0/zkvm/platform/src/rust_rt.rs
```diff
@@ -80,23 +77,54 @@ mod entrypoint {
...
```
flaub, line 117: What is the purpose of this? It doesn't seem to be in use.

flaub (approved):

risc0/zkvm/platform/src/rust_rt.rs
SchmErik: this is required by the embedded allocator - it needs to know how to lock/unlock to get exclusive access to the heap so that it doesn't run into race conditions

PR #2182, opened 2024-08-04
Bakeoff with various codecs and approaches for host/guest comms

flaub:
Results:

```
| name             | user_cycles | total_cycles | comment
| read             |      137567 |       262144 | use env::read
| read_framed      |       64763 |       262144 | use env::read_framed
| read_buffered    |      336317 |       524288 | use env::read_buffered
| bincode          |      294994 |       524288 | 
| bincode_framed   |       59821 |       262144 | 
| bincode_buffered |      343044 |       524288 |
| borsh            |       11872 |       131072 | framed
| cbor             |     1178661 |      1572864 | framed
| message_pack     |      171431 |       524288 | framed
| postcard         |       53791 |       262144 | framed
```

Also included:
* Use `stability::unstable` to mark new experimental APIs
* Add unstable `env::read_frame()` and `env::read_framed()`
* Drop the `borsh` feature flag
* Drop obsolete `loop` example, it's replaced by `datasheet`
* Update `guest_run` bench to run a program in the guest and print cycles. Use a one-hot approach because the number of cycles should never change
* Update `datasheet` to allow `--max-po2`. Default is `20` but can be `[16, 24]`. This allows testing large segment sizes


risc0/zkvm/src/host/client/env.rs
```diff
@@ -282,6 +282,20 @@ impl<'a> ExecutorEnvBuilder<'a> {
         self
     }
 
+    /// Write a frame to the zkVM guest via stdin.
+    ///
+    /// A frame contains a length header along with the payload. Reading a frame
+    /// can be more efficient than deserializing a message on-demand. On-demand
+    /// deserialization can cause many syscalls, whereas a frame will only have
+    /// two.
+    #[stability::unstable]
```
shkoo, line 291: :+1:  Sweet, I like this unstable tag.

SchmErik (approved):
Looks good to me, I really like the `unstable` feature!

PR #2188, opened 2024-08-06
rzup: remove extension and create a separate r0vm command instead

SchmErik:
This PR removes the extension command and instead adds a r0vm command.

TODO's (not necessary for this PR but I'll work on these later today)
I ripped out the download of cargo risczero so I need to put this back.
Need to fix the update command
Write CI for r0vm and cargo risczero release
add a test mode


.vscode/settings.json
```diff
@@ -74,7 +74,8 @@
     "examples/Cargo.toml",
     "risc0/zkvm/methods/guest/Cargo.toml",
     "risc0/zkvm/methods/std/Cargo.toml",
-    "tools/crates-validator/Cargo.toml"
+    "tools/crates-validator/Cargo.toml",
+    "rzup/Cargo.toml"
```
flaub, line 78: Why is this needed? It's already part of the top-level workspace.

SchmErik: So, adding this activates rust analyzer - without it, i don't seem to get the full functionality.

SchmErik:
will come back to this later. closing for now

PR #2189, opened 2024-08-06
fix: enable `std` for `hex`

Wollac:
This PR fixes building `cargo-risczero` without default features. This is currently causing CI failures in https://github.com/risc0/risc0-ethereum.

Using `?` to convert a `FromHexError` into an `anyhow:Error` requires the `std` feature of the the `hex` crate. This happens in the following line: 
https://github.com/risc0/risc0/blob/b57db72db8c79a6f6bc52ad581189d167ba7eeef/risc0/cargo-risczero/src/commands/verify.rs#L73

This currently leads to compilation failures when building `cargo-risczero` without default features, otherwise the correctly configured `hex` will be pulled in via some other dependencies.

Since `std` is the only feature enabled with `default` in hex, removing `default-features = false` seams preferable to explicitly adding the `std` feature.

### DIFF
--- a/risc0/cargo-risczero/Cargo.toml
+++ b/risc0/cargo-risczero/Cargo.toml
@@ -27,7 +27,7 @@ downloader = { version = "0.2", default-features = false, features = [
 flate2 = "1"
 fs2 = "0.4"
 fs_extra = "1.3.0"
-hex = { version = "0.4", default-features = false }
+hex = { version = "0.4" }
 regex = "1.10.6"
 reqwest = { version = "0.11", default-features = false, features = [
   "json",
--- a/risc0/cargo-risczero/Cargo.toml
+++ b/risc0/cargo-risczero/Cargo.toml
@@ -27,7 +27,7 @@ downloader = { version = "0.2", default-features = false, features = [
 flate2 = "1"
 fs2 = "0.4"
 fs_extra = "1.3.0"
-hex = { version = "0.4" }
+hex = "0.4"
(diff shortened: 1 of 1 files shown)


flaub:
I'm curious about a few things:
1. Why don't we see these CI failures in `risc0`?
2. Why does `risc0-ethereum` need to build `cargo-risczero`?

Wollac:
This bug can be reproduced in `risc0` by calling `cargo build -p cargo-risczero --no-default-features`
or `cargo install --path risc0/cargo-risczero --no-default-features` (which is what the risc0-ethereum CI does). However, everything builds without the `--no-default-features` which is probably why this was not detected here.

The `risc0-ethereum` CI used to pass because `metal` somehow also enabled `default` or `std`, but since it is now a NOOP it is no longer the case, but @capossele would know more about this now...

capossele:
I think this does not happen on `risc0` as the `metal` feature has been dropped from CI in favour of `default`. We could certainly (and eventually will) drop that as well, but for the sake of semver we tried to keep it. Now, `default` brings in `std` as it used to do also `metal` and I think that is why we never catched this issue before. 
`risc0-ethereum` uses `cargo-risczero` to install the `r0vm` from the latest `risc0` main branch.

capossele:
As a side note, I also think that the `metal` feature, semver wise, should still bring in the `prove` feature, so that is as feature complete as before

flaub (approved):

PR #2193, opened 2024-08-06
Fix note about paging in RISC-V cycle table

nategraf:
In the table of cycle counts per instruction, in the optimization guide, there is a note that says "1 if [paged-in](#paging) 1094 to 5130 otherwise". In the case of load instructions, this is true. In the case of store instructions, this is slightly inaccurate. It's instead based on whether the memory is _dirty_ rather than if it is paged in.


flaub (approved):

website/api/zkvm/optimization.md
```diff
@@ -421,9 +421,9 @@ cycle counts added.
 | LW rd,offset(rs1)   | Load Word                          | rd ← s32[rs1 + offset]                         | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
 | LBU rd,offset(rs1)  | Load Byte Unsigned                 | rd ← u8[rs1 + offset]                          | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
 | LHU rd,offset(rs1)  | Load Half Unsigned                 | rd ← u16[rs1 + offset]                         | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
-| SB rs2,offset(rs1)  | Store Byte                         | u8[rs1 + offset] ← rs2                         | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
-| SH rs2,offset(rs1)  | Store Half                         | u16[rs1 + offset] ← rs2                        | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
-| SW rs2,offset(rs1)  | Store Word                         | u32[rs1 + offset] ← rs2                        | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
+| SB rs2,offset(rs1)  | Store Byte                         | u8[rs1 + offset] ← rs2                         | 1 if [dirty](#paging) 1094 to 5130 otherwise    |
```
pdg744, line 424: hmm i think this will be confusing for readers b/c the term "dirty" isn't explained anywhere. 
can we add some explanation of that term to the #paging section before we land this revision?

website/api/zkvm/optimization.md
```diff
@@ -421,9 +421,9 @@ cycle counts added.
 | LW rd,offset(rs1)   | Load Word                          | rd ← s32[rs1 + offset]                         | 1 if [paged-in](#paging) 1094 to 5130 otherwise |
```
pdg744, line 421: is `paged-in` correct in lines 419-423, or should those also be revised to say `dirty`?

website/api/zkvm/optimization.md
flaub: Also it's 1152 cycles per page.

website/api/zkvm/optimization.md
flaub: For loads, paged-in is correct.

website/api/zkvm/optimization.md
flaub: Basically it's the last store of a page that will be marked as dirty. i.e. if multiple stores to the same page occur, it's only the last store that causes lots of cycles to be used for paging. It's not strictly load/store operations, it's really any memory access. So for instance, if the PC moves to a page that hasn't been read, then there will be 1152 cycles to read in the page associated with the PC for that page.

flaub: Also, the paging cycles don't occur in-line, they occur at the beginning and end of a segment, disjoint from where the memory access occurs.

nategraf: > hmm i think this will be confusing for readers b/c the term "dirty" isn't explained anywhere.

Oh, I will fix that. I thought I had explained it.

> Also it's 1152 cycles per page.

Ah, I'll revise this to use the correct number. I must have gotten the math wrong when I wrote this the first time.

> It's not strictly load/store operations, it's really any memory access. So for instance, if the PC moves to a page that hasn't been read, then there will be 1152 cycles to read in the page associated with the PC for that page. Also, the paging cycles don't occur in-line, they occur at the beginning and end of a segment, disjoint from where the memory access occurs.

Ya, it's definitely more complicated than can be accounted for in a simple table and attributing cycles is always a pain 😅. The "if dirty" part of this is trying to capture the idea that if you write to a page twice in one segment, yo only pay once for page-out.

flaub:
This needs to be updated for 1.1 docs

pdg744:
@nategraf LMK if/when you want me to take another look at this; I had forgotten that it was still lingering

nategraf:
I've been meaning to finish this up, but keep not quite getting to it.

website/api/zkvm/optimization.md
nategraf: @flaub I'm looking at the executor source code, but not sure where 1152 comes from versus 1094, which is what I calculate from the formula I see.

https://github.com/risc0/risc0/blob/main/risc0/circuit/rv32im/src/prove/emu/pager.rs#L32-L41

The reason I ask about this is that I want to also calculate the upper bound on what a page-in costs (the 5130 number), which is probably wrong as well.

flaub: OK, I'm also seeing 1094, perhaps 1152 was an earlier number that got stuck in my head.

nategraf: Alright, we'll stick with 1094 to 5130. And I've added a sentence that defines "dirty" ([3da522c](https://github.com/risc0/risc0/pull/2193/commits/3da522ce65b1e59899d3b42411030fa9b2ba7a4d))

PR #2195, opened 2024-08-07
Restore `prove` into `metal` feature

capossele:
With #2092 dropping the `metal` feature and #2179 restoring it as an no-op feature, this PR restores the `prove` feature into it, with the goal of maintaining semver compatibility.


risc0/zkvm/Cargo.toml
```diff
@@ -136,7 +136,7 @@ dual = []
 # The zkVM exposes a getrandom implementation that panics by default. This will
 # expose a getrandom implementation that uses the `sys_random` ecall.
 getrandom = ["risc0-zkvm-platform/getrandom"]
-metal = []
+metal = ["prove"]
```
flaub, line 139:
```suggestion
metal = ["prove", "risc0-zkp/metal"]
```

risc0/zkp/Cargo.toml
```diff
@@ -58,7 +58,7 @@ features = ["prove", "std"]
 circuit_debug = []
 cuda = ["dep:cust", "prove", "risc0-sys/cuda"]
 default = []
-metal = []
+metal = ["prove"]
```
flaub, line 61:
```suggestion
metal = ["prove", "risc0-sys/metal"]
```

capossele: I haven't originally included `risc0-sys/metal` as it is an empty feature.

risc0/zkvm/Cargo.toml
capossele: Also here, I was under the assumption that by including `prove` we get 
```
"risc0-circuit-recursion/prove",
 "risc0-circuit-rv32im/prove",
 "risc0-zkp/prove",
 ```
 Wouldn't that be enough?

flaub (approved):
I haven't been able to test this. How can we test this?

capossele:
I've only checked that by invoking the `metal` feature on any test gated behind `prove` works, e.g.,: 
`cargo test -F metal --package risc0-zkvm --lib -- receipt_claim::tests::merge_receipt_claim --exact --show-output`

capossele:
thanks to @SchmErik for suggesting to add a specific test for it

PR #2196, opened 2024-08-07
Add `cargo risczero datasheet` command

nvzqz:
This code was adapted from [`risc0/zkvm/examples/datasheet.rs`][datasheet], with notable changes:

- Added `--r0vm` flag for providing a path to r0vm, instead of just the `RISC0_SERVER_PATH` environment variable.

- Added min size of 2^15 cycles.

- Uses `ApiClient` to run external r0vm process, instead of `get_prover_server` and `ExecutorImpl`. This requires initially leaving out:
  - RAM usage reporting
  - groth16 and stark2snark benchmarks

- No reliance on the `risc0-zkvm-methods` crate for `BENCH_ELF`. Instead this uses an ELF from a hand-written assembly program `loop.s`, which is pre-compiled, stripped to 640 bytes, and checked-in as `loop.bin`.

[datasheet]: https://github.com/risc0/risc0/blob/v1.0.5/risc0/zkvm/examples/datasheet.rs


risc0/cargo-risczero/src/commands/datasheet/mod.rs
```diff
@@ -0,0 +1,487 @@
...
+            .copied()
+    }
+
+    /// Creates a connection to the zkVM server (`r0vm` executable).
+    ///
+    /// The path is provided by `--r0vm`, `$RISC0_SERVER_PATH`, or searching
+    /// `$PATH`.
+    fn connect_to_server(&self) -> Result<ApiClient> {
+        let server_path = self.r0vm.as_deref().unwrap_or("r0vm".as_ref());
+        ApiClient::new_sub_process(server_path)
```
flaub, line 113:
```suggestion
        ApiClient::from_env()
```

nvzqz: Do we not want this configurable with a CLI flag?

flaub (approved):

risc0/cargo-risczero/src/commands/datasheet/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+
+        let cycles = info.stats.total_cycles;
+        assert_eq!(
+            cycles, expected_cycles,
+            "Incorrect cycle count for po2={po2}"
+        );
+
+        Ok(BenchmarkData {
+            // XXX: This differs from the filter. Should we have these match?
+            name: "rv32im",
```
nvzqz, line 299: @flaub `composite` and `identity_p254` differ between their output names and filter names. Shouldn't these match?

flaub: what'a a filter name and an output name?

nvzqz: `cargo risczero datasheet identity` vs `identity_p254` in the stdout table and JSON file.

Issue #1759, opened 2024-04-30
In BonsaiProver, reduce latency from workflow completion to returning the result


nategraf:
## Feature

Reduce the latency overhead of polling Bonsai for workflow status

## Motivation

Currently, the `BonsaiProver` implements polling with a 5 sec interval to check the status of the workflow. Especially as our proving times improve, and small proofs can complete in less than 5 seconds of proving time, this adds latency to the user experience.

https://github.com/risc0/risc0/blob/9a90de90cae3e7cf9621d3c7da33bdc15720b236/risc0/zkvm/src/host/client/prove/bonsai.rs#L87

## Implementation

We likely do not want to simply increase the polling frequency, as that will add a wasteful increase in load the API servers. Implementing long polling, or web sockets would allow us to improve upon this without additional load.

PR #2198, opened 2024-08-07
Enable env variable to configure Bonsai poll time and reduce default

austinabell:
Tested on low-latency proof workflows, and cuts those e2e snark times using this prove API by ~5s (as expected).

As for what the default is, just going with the upper limit of what @mothran suggested Bonsai can handle. 

Keeping the TODO and #1759 open given this is still a very naive solution and could be improved. Before we are able to consider adding a WS API for this, we could do some of the following:
- [ ] Add a backoff function to the stark polling by default, to minimize server requests for long proofs
	- Possibly also having each configurable 	
- [ ] Separate the Snark and Stark polling intervals, have them both be configurable
- [ ] Add an initial delay for Starks or Snarks for the minimum duration we estimate

But I decided to keep it as simple as possible because it is unclear the direction of this in the long term.


mothran (approved):

flaub (approved):

risc0/zkvm/src/host/client/prove/bonsai.rs
```diff
@@ -81,13 +81,18 @@ impl Prover for BonsaiProver {
         let session = client.create_session(image_id_hex, input_id, receipts_ids, false)?;
         tracing::debug!("Bonsai proving SessionID: {}", session.uuid);
 
+        // TODO(#1759): Improve upon this polling solution.
+        let polling_interval = if let Ok(ms) = std::env::var("BONSAI_POLL_INTERVAL_MS") {
```
nvzqz, line 85: I find seconds to be a more natural unit than milliseconds to put in places like env vars and JSON. And then the default value is simply `1` (previously `5`).

austinabell: The issue is we want to support sub-second polling times for some low-latency use cases. How were you thinking about handling that? This be a floating point value?

nvzqz: Yeah I think 1e-3 represented with 53 bits of floating precision is fine

austinabell: I'm unopinionated, this shouldn't be a commonly used env var anyway, if 👍 given by those who have already reviewed are given, happy to make that change

nvzqz: Just something to consider, not an r-

austinabell:
Is this able to land before next release? Would be very helpful. Unclear if comment is viewed as blocking, happy to change logic if so.

flaub:
This didn't make it into 1.1.0, but we can do 1.1.1

Issue #2049, opened 2024-07-03
async Bonsai prover API


austinabell:
So the issue with using the reqwest blocking client as the default (and only option) for `BonsaiProver`, is that it will cause issues when running in async contexts unless explicitly ran in a task that can block ([context](https://github.com/seanmonstar/reqwest/issues/1017)). Ideally there is an option to have an async version of https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.BonsaiProver.html#method.prove_with_opts to avoid these gymnastics and have these requests actually be async if possible.

Low prio feature ask, it complicates the API with all the prove variants, complicates the features of using the bonsai-sdk sync/async, and there is a workaround that looks like:

```rust
let prove_info = tokio::task::block_in_place(|| prover.prove(env, GUEST_ELF))?;
```

And block in place has to be used over spawn blocking since the prover types are not `Sync`.

You can (and should) use `spawn_blocking` if you can afford creating a new prover/environment instance, though

(mainly opening issue as a reference for anyone else trying to run within async runtime)

austinabell:
> Just hit this issue and it was hard to debug as the error message looks like this:
> 
> ```
> thread 'tokio-runtime-worker' panicked at /Users/leonidlogvinov/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.38.1/src/runtime/blocking/shutdown.rs:51:21:
> Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.
> ```
> 
> Would be nice to have this either fixed or described in the function docs.

What types of docs would you expect to be visible for this, some Rustdocs on the prove method, or some high-level docs around running this in async contexts?

The workaround for your issue would be to run the prove call within a blocking task (`spawn_blocking` ideally). I assume here you are running the prover with Bonsai inside of an async function?

PR #2204, opened 2024-08-08
Bonsai prove async APIs

austinabell:
Very much WIP, have to think about how to remove duplicate code for maintainability.

Main issue right now is that the core types aren't threadsafe (`VerifierContext` specifically), which makes it hard to move blocking compute to a blocking task to avoid starving the executor. It's a bit tricky because the logic in this prove switches between async code to prove and then sync code to verify.

Options:
- Leave as is, we don't really care about starving the async executor (not heavily blocking)
- Make some internal types `Send`
  - think only `VerifierContext` matters here, but others would be nice for other use cases
  - Likely not a feasible option post 1.0
- Change APIs to return an "unvalidated" receipt type, and give the user the option for if they want to verify and if they want in a blocking task or not
  - I personally like the idea of this because it gives the flexibility to ignore even if trusted, but complicates the API for sure
  
Opens:
  
Some initial thoughts would be very helpful, doesn't really seem clean, but even as is does cut e2e times. For a simple snark proof request, https://github.com/risc0/risc0/commit/005123cc3f22ee5ab575d43bcd20b4cbe8cdfd76 cuts the e2e time from ~28s to ~22s. For reference, manually removing the stark download and any verifications with the same polling time is around 19.5-22s.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4172,6 +4172,7 @@ dependencies = [
  "tar",
  "tempfile",
  "test-log",
+ "tokio",
  "tracing",
  "tracing-forest",
  "tracing-subscriber 0.3.18",
--- a/benchmarks/Cargo.lock
+++ b/benchmarks/Cargo.lock
@@ -3394,7 +3394,19 @@ dependencies = [
  "num_cpus",
  "pin-project-lite",
  "socket2",
+ "tokio-macros",
  "windows-sys 0.48.0",
+]
+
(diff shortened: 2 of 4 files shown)


austinabell:
Closing because I do not have the bandwith to explore more, and not a clear win here

austinabell:
linking to issue for discoverability #2049 if the manual implementation is helpful to anyone

Issue #1685, opened 2024-04-18
[BUG] Guest build.rs logs output overwrites Helix editor via Rust analyzer


NukeManDan:
## Bug Report

When using rust-analyzer as the LSP with [Helix as terminal-based editor](https://helix-editor.com/), risc0 guest crate build output will be displayed overriding the display of the code loaded in the editor. See a screen recording here: https://github.com/helix-editor/helix/discussions/10038

I assumed this was an issue with the editor only at first (and it should not allow this to happen IMHO) but this would not be an issue if risc0 didn't write out to the stdout/err buffer in such a way that causes this. Note that no other crates I have used with helix have had such an issue for me, but all risc0 ones with `methods/build.rs` do.

## Steps to Reproduce
1. [Setup helix with rust analyzer](https://github.com/helix-editor/helix/wiki/Language-Server-Configurations#rust) 
2. Open any risc0 project (includes the standard `methods/build.rs` tooling that writes logs to some buffer)
3. Text overriding display of editor

## Expected behavior

No output should be displayed on the editor with the LSP running in a background task.
I would propose that the build tooling is modified to direct to stderr or other buffer that should avoid this :crossed_fingers: 

## Your Environment
- risc0-zkvm version: 0.2 and above 
- Rust version: 1.75 and above
- Platform/OS: Linux

## Additional context

I have not pushed to get the Helix side of things patched to mitigate this, but I would think other editors like nvim might have the same issue too...?

NukeManDan:
I think the call below and maybe few other `StdIo` fn calls might be the culprit, not sure what alternative methods to get the logs printed when executed directly to show but not mess with the editor process :thinking: Would be happy to implement one that the team thinks would be best

https://github.com/risc0/risc0/blob/df99cddbc5c8790b2a7d467f9552a554a264792e/risc0/build/src/lib.rs#L379

NukeManDan:
Ah I think I found the culprit:

https://github.com/risc0/risc0/blob/df99cddbc5c8790b2a7d467f9552a554a264792e/risc0/build/src/lib.rs#L454-L458

https://github.com/risc0/risc0/blob/df99cddbc5c8790b2a7d467f9552a554a264792e/risc0/build/src/lib.rs#L476-L481

So I would expect all tty based editors to have this issue, would you agree?

nuke-web3:
[...]

PR #2208, opened 2024-08-09
Speed up risc0-build

### LINKED ISSUE #1985: [BUG] risc0-build panics when RISC0_SKIP_BUILD=1 is set and a guest is present in the workspace
## Bug Report

When building a workspace or running cargo clippy with RISC0_SKIP_BUILD=1 set it triggers a unwrap() to panic in risc0-build because guest_methods() assumes the existance of the guest elfs on disk. 

```
  thread 'main' panicked at /home/user/src/risc0/risc0/build/src/lib.rs:220:14:
  called `Result::unwrap()` on an `Err` value: No such file or directory (os error 2)
```

## Steps to Reproduce
1. in `risc0` repo
2. `cargo clean`
3. `RISC0_SKIP_BUILD=1 cargo clippy`

## Expected behavior

When RISC0_SKIP_BUILD is set the guest build should not panic

flaub:
Two major issues:
* build scripts are built with the dev profile by default, so compute_image_id is 20x slower than a release build. So call out to `r0vm` to compute the image_id.
* `include_bytes!()` is much faster at scale than embedding the string form of the ELF into methods.rs
* Fixes #1985

This helps especially for null builds

### DIFF
--- a/risc0/build/src/lib.rs
+++ b/risc0/build/src/lib.rs
@@ -30,10 +30,10 @@ use std::{
     process::{Command, Stdio},
 };
 
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use cargo_metadata::{Message, MetadataCommand, Package};
 use risc0_binfmt::compute_image_id;
-use risc0_zkp::core::digest::DIGEST_WORDS;
+use risc0_zkp::core::digest::{Digest, DIGEST_WORDS};
 use risc0_zkvm_platform::memory;
 use serde::Deserialize;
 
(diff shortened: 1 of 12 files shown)


nuke-web3 (approved):
Didn't test locally, but LGTM with a light review of the diff - love to hear this will make (some) build workflows faster!

risc0/build/src/lib.rs
```diff
@@ -490,6 +512,30 @@ fn build_staticlib(guest_pkg: &str, features: &[&str]) -> String {
     }
 }
 
+// HACK: Attempt to bypass the parent cargo output capture and
+// send directly to the tty, if available.  This way we get
+// progress messages from the inner cargo so the user doesn't
+// think it's just hanging.
+fn tty_println(msg: &str) {
```
nuke-web3, line 519: Love this, although related to https://github.com/risc0/risc0/issues/1685

Aside: I think we need better loading /progress information, as spinners don't give an idea if you are stuck or not in many cases, like `rzup` and `cargo risczero install`

flaub: We don't have that option since we're running inside the context of a cargo build process.

PR #2209, opened 2024-08-10
Enable forwarding custom rustc flags to guest builds.

hashcashier:
The current build system defines the `CARGO_ENCODED_RUSTFLAGS` environment variable when compiling guest programs. This prevents cargo from using rust flags specified in other sources, such as `config.toml` files.

This prevents users from fully utilizing conditional rust compilation in guests, even when emitting `cargo::rustc-flags=FLAGS` in a build script for the guest binary, because those flags do not affect how the guest's dependencies are built.

This PR enables specifying custom rustc flags to pass to the guest build process through specifying additional metadata within a guest binary's `Cargo.toml` file:
```cargo
[package.metadata.risc0]
rustc-flags = ['--cfg', 'risc0_guest_allocator="embedded"']
```

NOTE: The metadata is ignored when using `cargo-risczero` build and deploy commands.


hashcashier:
This currently breaks semver. Looking at alternatives.

flaub:
Could we add a test that uses this new feature? Maybe just one of the methods crates could be updated.

hashcashier:
@flaub Just added a guest named `cfg` to `risc0-zkvm-methods` that won't compile unless given the proper config input.

risc0/zkvm/methods/build.rs
```diff
@@ -37,27 +38,22 @@ fn main() {
...
-        ),
-        (
-            "risc0-zkvm-methods-cpp-crates",
-            GuestOptions {
-                features: vec![],
-                use_docker: None,
+                ..Default::default()
             },
         ),
+        ("risc0-zkvm-methods-cpp-crates", GuestOptions::default()),
```
flaub, line 56: I think you can just remove it from the map if it's all default settings.

flaub (approved):

PR #2221, opened 2024-08-14
Add Alerts for Nightly `Benchmark Reports` Job Failures

morgangibbs87:
- Add alert when `Benchmark Reports` workflows fail
- Test alerts when `bench` jobs fail
- Test alerts when `publish` jobs fail

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -593,7 +593,7 @@ checksum = "c6fa2087f2753a7da8cc1c0dbfcf89579dd57458e36769de5ac750b4671737ca"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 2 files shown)


.github/workflows/bench_reports.yml
```diff
@@ -1,6 +1,9 @@
 name: Benchmark Reports
 
 on:
+
+  pull_request:
+    branches: [main]
```
flaub, line 6: This causes a bottleneck on our benchmark machine, let's not do this

flaub: you can run this manually, we don't need to make temporary changes like this i think

flaub (approved):

PR #2238, opened 2024-08-15
cryptographic security model: add verbiage about changing parameters 

pdg744:
This PR adds a brief note to the security model doc to clarify that our assertion of bits of security is based on default system parameters. Additionally, this PR encourages users who are adjusting system parameters to use the security calculator to check the security level.


tzerrell (approved):

flaub:
This needs to be updated for 1.1 docs

flaub (approved):

flaub:
@pdg744 Is this still needed?

pdg744:
Not needed anymore -- we've revised this page to reflect the worst-case soundness based on what the on-chain verifier contract will accept. This approach is preferable to relying on the calculator, since soundness is a property of the verification process rather than the proof generation process (and the calculator happens on the prover side).

PR #2239, opened 2024-08-16
Run clippy on all targets

flaub:
The bigint op_tests have been refactored to use a macro to be more concise.


tzerrell (approved):
Awesome, thank you!

PR #2240, opened 2024-08-16
fix(examples/zkevm-demo): allow dead-code methods

keroro520:
The below error was throw when run `cargo build` within `examples/zkevm-demo/methods/guest/`:

```
error: methods `reset` and `set` are never used
  --> risc0/examples/zkevm-demo/core/src/lib.rs:42:12
   |
37 | / impl<T> ResTrack<T>
38 | | where
39 | |     T: Clone,
   | |_____________- methods in this implementation
...
42 |       pub fn reset(&mut self) {
   |              ^^^^^
...
53 |       pub fn set(&mut self, elm: &T) {
   |              ^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(dead_code)]`

error: could not compile `zkevm-core` (lib) due to 1 previous error
```

### DIFF
--- a/examples/zkevm-demo/core/src/lib.rs
+++ b/examples/zkevm-demo/core/src/lib.rs
@@ -38,6 +38,7 @@ impl<T> ResTrack<T>
 where
     T: Clone,
 {
+    #[allow(dead_code)]
     #[cfg(not(target_os = "zkvm"))]
     pub fn reset(&mut self) {
         self.idx = 0;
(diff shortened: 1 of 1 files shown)


flaub:
It's not intended for the guest to be built with the host compiler. The `risc0-build` crate manages launching `cargo` for the guest side. It can also be manually run with `cargo +risc0 build` in the guest directory.

PR #2241, opened 2024-08-16
chore(examples/voting-machine): enhance vote logic

keroro520:
- Flip the vote and update the count when the voter changes their vote.


keroro520:
Hi, ping for review

flaub:
@nategraf WDYT?

nategraf:
The updated logic isn't more correct in any way, as it simply changes the handling of the same voter submitting two ballots. As implemented on `main`, only the first ballot is accepted. As it is implemented here, the vote is updated. I don't see this as an improvement, and so I'm not inclined to merge this PR.

PR #2244, opened 2024-08-16
Add a bigint nondet inverse op

tzerrell:
Adds testing and support for a nondet (multiplicative modular) inverse op in bigint.


risc0/circuit/bigint/src/test_harness.rs
```diff
@@ -92,31 +90,28 @@ pub(crate) fn test_zkr(ctx: BigIntContext, zkr_name: &str) -> Result<()> {
...
+    tracing::error!("test_zkr: Prover made");
     prover.add_input(&[0u32; 8]); //control id
     prover.add_input(&z.to_u32_words());
     prover.add_input(&all_coeffs);
+    tracing::error!("test_zkr: Inputs added, all_coeffs is {all_coeffs:?}");
     let receipt = prover.run()?;
-    tracing::error!("chkpt C");
 
-    tracing::error!("Test receipt: {receipt:?}");
+    tracing::error!("test_zkr: receipt is {receipt:?}");
```
shkoo, line 114: Do we want to change these to tracing::debug instead of tracing::error?

tzerrell: Yes, or even `trace`, but I want to show them in CI because CUDA is broken and I don't know how yet, and I don't know what log levels CI shows. But yes these should change before landing.

flaub (approved):

risc0/circuit/bigint/src/op_tests.rs
```diff
@@ -250,4 +250,27 @@ bigint_tests! {
         [],
         [108881840, 768445877, 1224562786, 1348962832]
     ),
+
+    nondet_inv(
+        nondet_inv_test_8,
+        ["21", "2f", "0a"],  // TODO: but... don't we need to test the answer is "0a"?
```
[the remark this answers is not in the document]
tzerrell: Thanks! Comment is outdated -- there was an oversight in the test and this used to read just `["21", "2f"]` and I didn't fix the comment when I fixed the test.

risc0/circuit/bigint/src/byte_poly.rs
```diff
@@ -96,6 +96,17 @@ impl BytePoly {
         Self::from_biguint(rem, coeffs)
     }
 
+    pub fn nondet_inv<Rhs: Borrow<BytePoly>>(&self, rhs: Rhs, coeffs: usize) -> BytePoly {
+        // Computes the inverse of LHS mod RHS via the `inv = lhs^{rhs - 2} % rhs` algorithm
```
semenov-vladyslav, line 100: Inverse of LHS mod RHS makes sense for coprime LHS and RHS. Is that assumed? If so, INV = LHS^{phi(RHS) - 1} mod RHS where phi is Euler's totient function. INV = LHS^{RHS-2} mod RHS only for prime RHS. Is that assumed too?

risc0/circuit/bigint/src/byte_poly.rs
```diff
@@ -96,6 +96,17 @@ impl BytePoly {
...
+        // TODO: Is num_bigint's `modpow` algorithm efficient?
```
semenov-vladyslav, line 101: Looks reasonably efficient, it uses 4-bit windowed exponentiation using Montgomery reduction for odd RHS, otherwise for even RHS it uses plain binary modular exponentiation. Modular inverse using extended Euclidean algorithm would be much more efficient on average as numbers tend to get small fast. One concern is (maybe, for some reason) the algorithm must be constant-time, in this case Euclidean algorithm wouldn't work and neither does the num_bigint's modpow which is also non-constant time.

risc0/circuit/bigint/src/byte_poly.rs
tzerrell: Good point! I've added clarifying comments to indicate that this requires prime RHS (which implies coprime LHS & RHS as well). We may generalize the algorithm later, but at the moment, I am using this with prime RHS; and this is not a soundness issue, as this is a nondet and the correctness of the inversion must be checked anyway.

risc0/circuit/bigint/src/byte_poly.rs
tzerrell: Thank you! This all sounds good to me

PR #2246, opened 2024-08-19
Add test for new heap allocator feature

flaub:


risc0/zkvm/src/guest/mod.rs
```diff
@@ -148,19 +148,10 @@ macro_rules! entry {
...
-        }
-        let heap_pos: usize = unsafe { (&_end) as *const u8 as usize };
-        let heap_size: usize = risc0_zkvm_platform::memory::GUEST_MAX_MEM - heap_pos;
-        unsafe {
-            risc0_zkvm_platform::rust_rt::heap::embedded_allocator::HEAP.init(heap_pos, heap_size)
-        }
-    }
+    risc0_zkvm_platform::heap::embedded::init();
+
+    env::init();
```
SchmErik, line 154: Great call on this reordering!

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -690,6 +690,7 @@ pub unsafe extern "C" fn sys_argv(
 }
 
 #[cfg_attr(feature = "export-syscalls", no_mangle)]
+#[deprecated]
```
SchmErik, line 693: small nit: could we provide a message to offer an alternative function for them to use? https://doc.rust-lang.org/nightly/reference/attributes/diagnostics.html#the-deprecated-attribute

SchmErik (approved):
looks great, I had a small nit but nothing to hold this back from landing

risc0/zkvm/platform/src/syscall.rs
flaub: There's no alternative, it shouldn't have been made a public syscall to begin with (IMHO). Also, too late but we should not have `sys_alloc_aligned` either as a syscall.

PR #2262, opened 2024-08-22
Add `seal_size` method to receipt types

nvzqz:


flaub (approved):
Would be good to hunt down all the places the do this computation in an ad-hoc fashion. I'm thinking the datasheet, benchmarks, maybe some examples?

PR #2268, opened 2024-08-24
docs: format components into table and link to audits

carterbrett:
Cleaned up components on the "security model" page into a table and lined to the relevant audit for each component for easy discoverability. Will ping @kevinnassery and @morgangibbs87 for additional review as this change will require some upkeep.

### DIFF
--- a/website/api/security-model.md
+++ b/website/api/security-model.md
@@ -9,17 +9,15 @@ The rest of this document offers an overview of the security model for applicati
 RISC Zero offers the following components, each of which is ready for use on testnet.
 
 ## Overview of Components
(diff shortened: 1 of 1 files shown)


nahoc (approved):

kevinnassery (commented):
@pdg744 's call on whether process info should be in this soundness doc ; links to auditing are currently in the SDLC doc next to this one

kevinnassery (approved):

pdg744 (approved):
LGTM! Will sync with Kevin about whether we can improve the organization/findability of these assets.

PR #2276, opened 2024-08-27
Include powers of two (po2) up to 21 in the default verifier parameters

nategraf:
As part of our continual security analysis work, we've created a calculated for the "bits of security" given all the system parameters relevant to the soundness of the proof of knowledge property guaranteed to the verifier. With this, we find that the lookup argument may not achieve our targeted 97 bits of soundness under worse case assumptions grounded in the Schwartz-Zippel lemma. Under these assumptions, up to 3 bits of security are given up by including segment sizes up to power of two 24, relative to po2 21.

This PR implements the cautious approach of removing po2s above 21 from the default verifier parameters. Developers that wish to use po2 22, 23, or 24 may opt-in by using a non-default `VerifierContext`.

Additionally, this PR fixes a pair of off-by-one errors that had removed the rv32im control IDs for po2 24. Technically, this constitutes a breaking change in the `risc0-circuit-rv32im` crate since the public type `ControlIds` is not a fixed array of length n+1.
https://github.com/risc0/risc0/pull/2276/files#diff-ce2293a7cee10116f619bedf007ecfedd524505b764cfe5383785b7bd17dd7c4

Note: It is expected that the `risc0-ethereum` CI job will fail right now, because the control root is changed. It will pass again once we've merged this and https://github.com/risc0/risc0-ethereum/pull/205 to their respective `main`.

### DIFF
--- a/risc0/circuit/recursion/src/control_id.rs
+++ b/risc0/circuit/recursion/src/control_id.rs
@@ -15,6 +15,9 @@
 use risc0_zkp::core::digest::Digest;
 use risc0_zkp::digest;
 
+/// Smallest cycle limit, as a power of two (po2), supported as a lift program.
+pub const MIN_LIFT_PO2: usize = 14;
+
 /// Control IDs allowed in the default set of recursion programs. Includes control IDs for the base
 /// set of recursion programs, and each power-of-two of the rv32im circuit, using Poseidon2.
 pub const ALLOWED_CONTROL_IDS: &[Digest] = &[
--- a/risc0/circuit/recursion/src/control_id.rs
+++ b/risc0/circuit/recursion/src/control_id.rs
@@ -27,8 +27,6 @@ pub const ALLOWED_CONTROL_IDS: &[Digest] = &[
     digest!("17d75c070f373f333bbbaf33a688bb74fc05670435cd5d6378d37b608448e300"), // rv32im po2=19
     digest!("f3cf1217485c403cea526b1e52e70835c31bdc6e3016ee12916f6c6561fc0977"), // rv32im po2=20
     digest!("4043ce1426811d4f0e6a9c27256d052513c95739e9b80c74f628f313c739b75c"), // rv32im po2=21
-    digest!("b4edfc3f2c49f10285bcdc7493eb81063959820a59569356fb73b62c1586e01f"), // rv32im po2=22
-    digest!("cc1bd6753bb8fd41d41bba14bb56e90ffa9ddf4f20f7727734d140675a9cc52b"), // rv32im po2=23
     digest!("0e51536ec08ac01e42de3d2a0b2a552d8ff2c75e4ef2b5285eed23582a13ef5d"), // recursion identity.zkr
     digest!("d7b56f6ef29d18203ac26960eb21160d60da6f1975124261fbe38d1364fa0e6b"), // recursion join.zkr
     digest!("1131f74f290f9d52f9aa2e7343b0445dff1765597a1e1849ed5f3f0afaba9f4e"), // recursion lift_14.zkr
@@ -39,15 +37,12 @@ pub const ALLOWED_CONTROL_IDS: &[Digest] = &[
     digest!("c607b103592326302802cf1ff1c6ee5dab7f112e85c6106b4085e55c46535f1d"), // recursion lift_19.zkr
     digest!("deca5917ab949d3769a17415d4b5753d0a49f818af600815ad92c85067322217"), // recursion lift_20.zkr
     digest!("cc4245191072ea38989b65721112584cd548385f0eda402d76c70e0e718e0f73"), // recursion lift_21.zkr
-    digest!("dd6cd51e766f484d625df405e74e11067283c1269dbdae21bc6a031b1daa952e"), // recursion lift_22.zkr
-    digest!("4726ab732cc10d26e233aa6c82a32001f2d83a1878e5da7531017b737054dc6d"), // recursion lift_23.zkr
-    digest!("8ea32467c6445926da67365554e3b64be92a5076d071ff0632f45b294787253f"), // recursion lift_24.zkr
     digest!("18386311e4b86b527f3d966de55df3271f7e52209c037b607a41a62c8b4c9f61"), // recursion resolve.zkr
 ];
 
 /// Root of the Merkle tree constructed from [ALLOWED_CONTROL_IDS], using Poseidon2.
 pub const ALLOWED_CONTROL_ROOT: Digest =
-    digest!("9a3767040e4cf554112afa68bc043274a8636a06565e1d5e2b7fa90fda941218");
+    digest!("8b6dcf11d463ac455361b41fb3ed053febb817491bdea00fdb340e45013b852e");
 
 /// Control ID for the identity recursion programs (ZKR), using Poseidon over the BN254 scalar field.
 pub const BN254_IDENTITY_CONTROL_ID: Digest =
--- a/risc0/circuit/rv32im/src/control_id.rs
+++ b/risc0/circuit/rv32im/src/control_id.rs
@@ -69,9 +69,7 @@ pub const BLAKE2B_CONTROL_IDS: ControlIds = [
 
 #[inline]
 pub const fn control_id_sha256(po2: usize) -> Option<Digest> {
-    if po2 < risc0_zkp::MIN_CYCLES_PO2 {
-        None
-    } else if po2 > risc0_zkp::MAX_CYCLES_PO2 {
+    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
         None
     } else {
         Some(SHA256_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
@@ -80,9 +78,7 @@ pub const fn control_id_sha256(po2: usize) -> Option<Digest> {
 
 #[inline]
 pub const fn control_id_poseidon2(po2: usize) -> Option<Digest> {
-    if po2 < risc0_zkp::MIN_CYCLES_PO2 {
-        None
-    } else if po2 > risc0_zkp::MAX_CYCLES_PO2 {
+    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
         None
     } else {
         Some(POSEIDON2_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
@@ -91,9 +87,7 @@ pub const fn control_id_poseidon2(po2: usize) -> Option<Digest> {
 
 #[inline]
 pub const fn control_id_blake2b(po2: usize) -> Option<Digest> {
-    if po2 < risc0_zkp::MIN_CYCLES_PO2 {
-        None
-    } else if po2 > risc0_zkp::MAX_CYCLES_PO2 {
+    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
         None
     } else {
         Some(BLAKE2B_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
--- a/risc0/circuit/rv32im/src/control_id.rs
+++ b/risc0/circuit/rv32im/src/control_id.rs
@@ -66,30 +66,3 @@ pub const BLAKE2B_CONTROL_IDS: ControlIds = [
     digest!("d0a94a254af9bfc8e457d5f84afe5150d680579cff65e37140720aa112e63b9f"), // rv32im po2=23
     digest!("46b765782f325e05cf04188e530fa6f33c6a8c33050f3b929bd6eedb541eb5a0"), // rv32im po2=24
 ];
-
-#[inline]
-pub const fn control_id_sha256(po2: usize) -> Option<Digest> {
-    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
-        None
-    } else {
-        Some(SHA256_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
-    }
-}
-
-#[inline]
-pub const fn control_id_poseidon2(po2: usize) -> Option<Digest> {
-    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
-        None
-    } else {
-        Some(POSEIDON2_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
-    }
-}
-
-#[inline]
-pub const fn control_id_blake2b(po2: usize) -> Option<Digest> {
-    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
-        None
-    } else {
-        Some(BLAKE2B_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
-    }
-}
--- a/risc0/circuit/rv32im/src/control_id.rs
+++ b/risc0/circuit/rv32im/src/control_id.rs
@@ -15,7 +15,7 @@
 use risc0_zkp::core::digest::Digest;
 use risc0_zkp::digest;
 
-const CONTROL_ID_ENTRIES: usize = risc0_zkp::MAX_CYCLES_PO2 - risc0_zkp::MIN_CYCLES_PO2;
+const CONTROL_ID_ENTRIES: usize = risc0_zkp::MAX_CYCLES_PO2 - risc0_zkp::MIN_CYCLES_PO2 + 1;
 
 pub type ControlIds = [Digest; CONTROL_ID_ENTRIES];
 
@@ -32,6 +32,7 @@ pub const SHA256_CONTROL_IDS: ControlIds = [
     digest!("0d85ecb4d16be50857cf50883e5b66e3f7185b2aeb5a7479cfad24ec61306b9b"), // rv32im po2=21
     digest!("a60d1c94faf9ca4be54d70518502139d260882d0d981d6199152b9e1468db984"), // rv32im po2=22
     digest!("5a10d231bc367d19130b29acb569eb7f3827b67e47a271f947e3e209945283f1"), // rv32im po2=23
+    digest!("da30893f75e7b069cb0004874a0bcb0f7b713aa30be88b3135eda6758bfd780b"), // rv32im po2=24
 ];
 
 /// Control IDs for each power-of-two of the rv32im circuit using Poseidon2.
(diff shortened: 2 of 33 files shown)


nategraf:
PR is marked as draft, since some more work is needed to make is reasonable easy to construct a `VerifierContext` and `ProverOptions` with all po2s.

xtask/src/templates/control_id_rv32im.rs.tmpl
```diff
@@ -27,3 +27,30 @@ pub const POSEIDON2_CONTROL_IDS: ControlIds = [{}];
...
+}}
+
+#[inline]
+pub const fn control_id_blake2b(po2: usize) -> Option<Digest> {{
+    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2  {{
+        None
+    }} else {{
+        Some(BLAKE2B_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
+    }}
+}}
```
flaub, line 56: I'm not sure these helpers should live in generated code

flaub:
Don't we need to continue to compute the IDs for larger PO2s? I would imagine that we'd want to have multlple control roots, with the default being the one that only goes up to 21.

xtask/src/bootstrap.rs
```diff
@@ -80,7 +81,7 @@ impl Bootstrap {
             Loader::compute_control_id_table(&CpuHal::new(Blake2bCpuHashSuite::new_suite()));
 
         let contents = format!(
-            include_str!("templates/control_id_rv32im.rs"),
+            include_str!("templates/control_id_rv32im.rs.tmpl"),
```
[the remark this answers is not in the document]
nategraf: I did this because the new functions will get formatted in a way that breaks their ability to be used as a template. Changing the final extension to `.tmpl` keeps most syntax highlighting (for me) while avoiding having the editor or our tooling try to enforce formatting.

xtask/src/templates/control_id_rv32im.rs.tmpl
nategraf: I was having trouble thinking of a better place to put them. Probably next-best in my mind would be `risc0/circuit/rv32im/src/lib.rs`

risc0/circuit/rv32im/src/control_id.rs
```diff
@@ -62,4 +64,32 @@ pub const BLAKE2B_CONTROL_IDS: ControlIds = [
...
+    digest!("46b765782f325e05cf04188e530fa6f33c6a8c33050f3b929bd6eedb541eb5a0"), // rv32im po2=24
 ];
+
+#[inline]
+pub const fn control_id_sha256(po2: usize) -> Option<Digest> {
+    if po2 < risc0_zkp::MIN_CYCLES_PO2 || po2 > risc0_zkp::MAX_CYCLES_PO2 {
+        None
+    } else {
+        Some(SHA256_CONTROL_IDS[po2 - risc0_zkp::MIN_CYCLES_PO2])
+    }
```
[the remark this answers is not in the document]
nategraf: Yes, but the result is a panic instead of returning `None` if this is where the check happens.

risc0/circuit/recursion/src/control_id.rs
```diff
@@ -39,15 +37,12 @@ pub const ALLOWED_CONTROL_IDS: &[Digest] = &[
     digest!("c607b103592326302802cf1ff1c6ee5dab7f112e85c6106b4085e55c46535f1d"), // recursion lift_19.zkr
     digest!("deca5917ab949d3769a17415d4b5753d0a49f818af600815ad92c85067322217"), // recursion lift_20.zkr
     digest!("cc4245191072ea38989b65721112584cd548385f0eda402d76c70e0e718e0f73"), // recursion lift_21.zkr
-    digest!("dd6cd51e766f484d625df405e74e11067283c1269dbdae21bc6a031b1daa952e"), // recursion lift_22.zkr
-    digest!("4726ab732cc10d26e233aa6c82a32001f2d83a1878e5da7531017b737054dc6d"), // recursion lift_23.zkr
-    digest!("8ea32467c6445926da67365554e3b64be92a5076d071ff0632f45b294787253f"), // recursion lift_24.zkr
```
flaub, line 44: Seems like we still need these

nategraf:
> Don't we need to continue to compute the IDs for larger PO2s? I would imagine that we'd want to have multlple control roots, with the default being the one that only goes up to 21.

That's what this PR does. In the recursion control IDs file, they are indeed removed from the `ALLOWED_CONTROL_IDS`, because that is the default set computed during bootstrapping [1]. They are retained in the full list of `POSEIDON2_CONTROL_IDS` just below [2]. What I am working on now is a new function to calculate a list up to a given po2 at runtime, which will be used for the `SuccinctReceiptVerifierParameters` similar to what I've done for `SegmentReceipt`[3]

[1] https://github.com/risc0/risc0/pull/2276/files#diff-61408acc9679dad83d6c4a648d4f29bf8ad2d2dcadf06497efdd08d9e6d02bc9R18-R41
[2] https://github.com/risc0/risc0/pull/2276/files#diff-61408acc9679dad83d6c4a648d4f29bf8ad2d2dcadf06497efdd08d9e6d02bc9R51-R103
[3] https://github.com/risc0/risc0/pull/2276/files#diff-7cad5396cabf80ff7797b87eab9a21fdd3baaf84e04e9e437ec4fbfa6f5d277eR152-R182

risc0/circuit/recursion/src/control_id.rs
nategraf: See https://github.com/risc0/risc0/pull/2276#issuecomment-2313212347 my approach here.

xtask/src/templates/control_id_rv32im.rs.tmpl
flaub: That seems like a good place. And also it feels like we could write a generic function instead of having one for each hash.

nategraf: I'll go ahead and move them, and also revert the rename of this file.

nategraf:
@flaub, this PR technically introduces a breaking change to `risc0-circuit-rv32im` (see PR description). Do we want to worry about that, or does this fall under the threshold of things we can pragmatically ignore?

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -238,6 +242,8 @@ impl From<ProverOpts> for pb::api::ProverOpts {
             prove_guest_errors: opts.prove_guest_errors,
             receipt_kind: opts.receipt_kind as i32,
             control_ids: opts.control_ids.into_iter().map(Into::into).collect(),
+            // conversion from usize to u64 will always work.
+            segment_limit_po2: opts.segment_limit_po2.try_into().unwrap(),
```
flaub, line 246:
```suggestion
            segment_limit_po2: opts.segment_limit_po2 as u64,
```

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -196,19 +202,47 @@ impl Default for ProverOpts {
             prove_guest_errors: false,
             receipt_kind: ReceiptKind::Composite,
             control_ids: ALLOWED_CONTROL_IDS.to_vec(),
+            segment_limit_po2: 21,
```
flaub, line 205: I suggest adding `DEFAULT_SEGMENT_LIMIT_PO2` constant

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -163,6 +167,8 @@ pub struct ProverOpts {
     /// programs that are allowed to run and is a key field in the
     /// [SuccinctReceiptVerifierParameters][crate::SuccinctReceiptVerifierParameters].
     pub control_ids: Vec<Digest>,
+    /// Maximum cycle count, as a power of two (po2) that these prover options support.
+    pub(crate) segment_limit_po2: usize,
```
flaub, line 171: I think of this more as the `max_po2` rather than directly associated with the `segment_limit_po2`. Like `segment_limit_po2` on the `ExecutorEnv` can be smaller than `max_po2`, they seem distinct to me.

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -172,6 +172,7 @@ message ProverOpts {
   bool prove_guest_errors = 2;
   ReceiptKind receipt_kind = 3;
   repeated base.Digest control_ids = 4;
+  uint64 segment_limit_po2 = 5;
```
flaub, line 175:
```suggestion
  uint64 max_po2 = 5;
```

risc0/zkvm/src/receipt/succinct.rs
```diff
@@ -213,6 +229,52 @@ where
...
+    hash_name: impl AsRef<str> + 'static,
+    po2_max: usize,
+) -> anyhow::Result<impl Iterator<Item = Digest>> {
+    // Recursion programs (ZKRs) that are to be included in the allowed set.
+    // NOTE: Although the rv32im circuit has control IDs down to po2 13, lift predicates are only
+    // generated for po2 14 and above, hence the magic 14 below.
+    let allowed_zkr_names: BTreeSet<String> = ["join.zkr", "resolve.zkr", "identity.zkr"]
+        .map(str::to_string)
+        .into_iter()
+        .chain((14..=po2_max).map(|i| format!("lift_{i}.zkr")))
```
flaub, line 243: Suggest we use a `const MIN_RECURSION_PO2: usize = 14`

risc0/circuit/rv32im/src/lib.rs
```diff
@@ -54,3 +57,35 @@ impl TapsProvider for CircuitImpl {
...
+pub fn control_id(hash_name: impl AsRef<str>, po2: usize) -> Option<Digest> {
+    if !(risc0_zkp::MIN_CYCLES_PO2..=risc0_zkp::MAX_CYCLES_PO2).contains(&po2) {
+        return None;
+    }
+    let idx = po2 - risc0_zkp::MIN_CYCLES_PO2;
+    match hash_name.as_ref() {
+        "sha-256" => Some(SHA256_CONTROL_IDS[idx]),
+        "poseidon2" => Some(POSEIDON2_CONTROL_IDS[idx]),
+        "blake2b" => Some(BLAKE2B_CONTROL_IDS[idx]),
+        _ => None,
```
flaub, line 76: nit: It'd be nice to somehow have a single list of strings for all hashes supported, perhaps even defined as an enum. For example we have this also: https://github.com/risc0/risc0/blob/main/risc0/zkp/src/core/hash/mod.rs#L91

risc0/zkvm/src/host/client/prove/mod.rs
flaub: That one is already taken, how about `DEFAULT_MAX_PO2`

nategraf: Addressed in 582001f5

risc0/zkvm/src/host/client/prove/mod.rs
nategraf: Addressed in cc0c9dce

risc0/zkvm/src/host/protos/api.proto
nategraf: Addressed in https://github.com/risc0/risc0/commit/cc0c9dce0fe90ab8ed20bdd03583b26757ea7ac1

risc0/circuit/rv32im/src/lib.rs
nategraf: I went ahead and wrote the `enum` part of that while waiting for some builds. Using `AsRef<str>`, this can be passed into a lot of the places where we use strings right now, and possible used for the match arms as well. Actually using it throughout the codebase will touch quite a lot of places, and it's getting late, so I'll open a new PR with that tomorrow.
```patch
index 08b2ea85..b885bfcc 100644
--- a/risc0/zkp/src/core/hash/mod.rs
+++ b/risc0/zkp/src/core/hash/mod.rs
@@ -21,6 +21,7 @@ pub mod poseidon_254;
 pub mod sha;

 use alloc::{boxed::Box, rc::Rc, string::String};
+use core::str::FromStr;

 use risc0_core::field::{baby_bear::BabyBear, Field};

@@ -86,6 +87,51 @@ impl<F: Field> Clone for HashSuite<F> {
     }
 }

+/// Identifiers for hash function used in the RISC Zero codebase.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum HashId {
+    /// Identifier for the SHA-256 hash function.
+    Sha256,
+    /// Identifier for RISC Zero implementation of Poseidon2 over BabyBear.
+    Poseidon2,
+    /// Identifier for the Blak2b hash function.
+    Blake2b,
+    /// Identifier for RISC Zero implementation of Poseidon over the BN254 scalar field.
+    PoseidonBn254,
+}
+
+impl HashId {
+    /// Convert to the string representation for the identifier.
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::Sha256 => "sha-256",
+            Self::Poseidon2 => "poseidon2",
+            Self::Blake2b => "blake2b",
+            Self::PoseidonBn254 => "poseidon_254",
+        }
+    }
+}
+
+impl AsRef<str> for HashId {
+    fn as_ref(&self) -> &str {
+        self.as_str()
+    }
+}
+
+impl FromStr for HashId {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<HashId, Self::Err> {
+        match s {
+            "sha-256" => Ok(Self::Sha256),
+            "poseidon2" => Ok(Self::Poseidon2),
+            "blake2b" => Ok(Self::Blake2b),
+            "poseidon_254" => Ok(Self::PoseidonBn254),
+            _ => Err(()),
+        }
+    }
+}
+
 /// Construct a supported hash function given its name. Returns None is the name does not
 /// correspond to a supported hash function.
 pub fn hash_suite_from_name(name: impl AsRef<str>) -> Option<HashSuite<BabyBear>> {
 ```

File: risc0/zkvm/src/receipt/succinct.rs (hunk 9 above)
Comment (nategraf) [reply risc0/zkvm/src/receipt/succinct.rs]: Addressed in f44d4ffc7dd647dd2f8bb8c7ed22c419455f58db

File: risc0/circuit/rv32im/src/lib.rs [hunk 11]
```diff
@@ -54,3 +57,35 @@ impl TapsProvider for CircuitImpl {
...
+    }
+}
+
+/// Fetch all precomputed control IDs using the given hash, by name, and up to the cycle limit as a
+/// power of two (po2). If po2 is larger than the max supported, only supported po2s will be
+/// returned.
+pub fn control_ids(
+    hash_name: impl AsRef<str> + 'static,
+    po2_max: usize,
+) -> impl Iterator<Item = Digest> {
```
nvzqz, line 86: This could instead return `Option<&'static [Digest]>` by doing:
```rs
let all_control_ids = ...; // e.g. SHA256_CONTROL_IDS
let po2_idx = po2_max.checked_sub(risc0_zkp::MIN_CYCLES_PO2)?;
all_control_ids.get(..po2_idx)
```

risc0/circuit/rv32im/src/lib.rs
nvzqz: I also recommend an enum for the list of hash algorithms, which would have some benefits:
- `control_ids` no longer requires static or allocated strings
- `SegmentReceiptVerifierParameters::from_max_po2` no longer tracking this same list of algorithm names

Thoughts on your current enum (in no particular order):
- It's not being treated as a numerical ID, so perhaps a better name is `HashAlg`
- From a semantics perspective, it should not implement `AsRef<str>` since it is itself not a string buffer
- `name` might be better than `as_str`
- `FromStr::Err` should be a real error type, not `()`
- I recommend making `control_id` and `control_ids` be methods on it
- Methods on small `Copy` types that don't borrow should generally take `self` instead of `&self`

risc0/circuit/rv32im/src/lib.rs
```diff
@@ -54,3 +57,35 @@ impl TapsProvider for CircuitImpl {
...
+    (risc0_zkp::MIN_CYCLES_PO2..=po2_max)
+        .map(move |po2| control_id(hash_name.as_ref(), po2))
+        .take_while(Option::is_some)
+        .map(Option::unwrap)
```
nvzqz, line 90: The reason for `.take_while(Option::is_some).map(Option::unwrap)` here is nuanced. Please add a note of why this pattern is being used if this is what the final code ends up doing

risc0/circuit/rv32im/src/control_id.rs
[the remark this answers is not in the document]
nvzqz: Slices have a [`get`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get) method that returns `None` instead of panicking

risc0/circuit/rv32im/src/lib.rs
nategraf: This method is not part of any performance critical path, so I prefer the more flexible and general signature than something optimized for the specific implementation like this.

risc0/circuit/rv32im/src/lib.rs
```diff
@@ -54,3 +57,35 @@ impl TapsProvider for CircuitImpl {
...
+        "poseidon2" => Some(POSEIDON2_CONTROL_IDS[idx]),
+        "blake2b" => Some(BLAKE2B_CONTROL_IDS[idx]),
+        _ => None,
+    }
+}
+
+/// Fetch all precomputed control IDs using the given hash, by name, and up to the cycle limit as a
+/// power of two (po2). If po2 is larger than the max supported, only supported po2s will be
+/// returned.
+pub fn control_ids(
```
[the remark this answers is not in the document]
nategraf: I prefer `control_id` and `control_ids` personally. Distinction should be clear enough from the context it's used in, and I try to avoid naming methods with `get`.

risc0/circuit/rv32im/src/lib.rs
nategraf: Thanks, those are some helpful comments. I'll take them into account for the follow-up PR. As for constraints, we are not going to change interfaces covered by 1.0 semver, so the solution can't be perfect.

risc0/circuit/rv32im/src/lib.rs
nategraf: Addressed in 8b7dd5315932823ca299c3bd332f2439cbb73dd2

tzerrell (commented):
I appreciate the new APIs for managing which control IDs you want to allow.

risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -163,6 +167,8 @@ pub struct ProverOpts {
     /// programs that are allowed to run and is a key field in the
     /// [SuccinctReceiptVerifierParameters][crate::SuccinctReceiptVerifierParameters].
     pub control_ids: Vec<Digest>,
+    /// Maximum cycle count, as a power of two (po2) that these prover options support.
+    pub(crate) segment_po2_max: usize,
```
flaub, line 171:
```suggestion
    pub(crate) max_segment_po2: usize,
```

or just `max_po2` so that it aligns with `from_max_po2` and `DEFAULT_MAX_PO2`, etc.

flaub (approved):
nice!

risc0/zkvm/src/host/client/prove/mod.rs
nategraf: Addressed in b19f5e2779b3930b2512f393c87aef0cd4436991

PR #2280, opened 2024-08-29
docs: add readme for prover example

carterbrett:
resolves SPR-98


flaub:
I'll try and come up with a more accurate description, but this example isn't primarily about how to handle multiple segments in parallel.

nuke-web3:
If I understand correctly, this showcases how to implement the Bonsai API for a toy remote proving service that matches https://api.bonsai.xyz/swagger-ui/ . Is that correct?

flaub:
It doesn't really match swagger since there's no REST API or anything like that. It's really to show how you could use the API to get the hooks needed to perform the orchestration needed to distribute the work to a fleet of machines.

PR #2286, opened 2024-08-29
fix stability for sys_execute_zkr

SchmErik:
This marks `sys_execute_zkr` as unstable by gating it behind a feature flag. I've gone with the feature flag approach rather than using the stability crate because this function is implemented in `risc0-zkvm-platform` and we want to keep the dependencies small. Because risc0-zkvm-platform is built on the host-side, adding the `stability` crate results in changes to host-side lock files. I've decided to go with the feature flag approach to keep the delta small.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4342,6 +4342,7 @@ dependencies = [
  "embedded-alloc",
  "getrandom",
  "libm",
+ "stability",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


flaub (changes requested):
I've actually already made the change to add `#[unstable]` to a branch I haven't pushed yet, and the dependency is very small. I think in this case the benefits outweigh the cost. I'll update this branch soon so that it's uniform with our other unstable flags.

PR #2288, opened 2024-08-30
Support for deferred assumptions

flaub:
This change defers the processing of on-demand ZKR proofs so that they occur later in the pipeline rather than inline during the execution phase. The benefit of this approach is that proving services can distribute the ZKR proof workload out to a worker pool.

A `CoprocessorCallback` can be configured on the `ExecutorEnv` which allows users to receive a notification when a ZKR proof needs to be produced. The ZKR proof can happen asynchronously and does not need to block the executor. However, the assumption receipt needs to be produced by the time `resolve` is called to resolve any outstanding assumptions.


risc0/zkvm/src/host/api/mod.rs
```diff
@@ -193,10 +195,14 @@ fn get_server_version<P: AsRef<Path>>(server_path: P) -> Result<Version> {
...
 impl Connector for ParentProcessConnector {
     fn connect(&self) -> Result<ConnectionWrapper> {
+        let passthru_vars: HashMap<_, _> = std::env::vars()
+            .filter(|(key, _)| key == "RUST_LOG" || key == "RUST_BACKTRACE")
+            .collect();
         let addr = self.listener.local_addr()?;
         let child = Command::new(&self.server_path)
             .arg("--port")
             .arg(addr.port().to_string())
+            .envs(passthru_vars)
```
nategraf, line 205: Is the intention here to _only_ pass the `passthru_vars`? Is so, I think `env_clear` needs to be called. https://doc.rust-lang.org/std/process/struct.Command.html#method.env_clear

risc0/zkvm/src/host/client/env.rs
```diff
@@ -75,6 +66,29 @@ impl SegmentPath {
...
+
+/// A trait that supports the ability to be notified of ZKR proof requests
+/// on-demand.
+#[stability::unstable]
+pub trait CoprocessorCallback {
+    /// Request that a proof of a ZKR is produced.
+    fn prove_zkr(&mut self, request: ProveZkrRequest) -> Result<()>;
+}
+
+pub type CoprocessorCallbackRef<'a> = Rc<RefCell<dyn CoprocessorCallback + 'a>>;
```
nategraf, line 90: Another option here would be to make `prove_zkr` take an immutable `&self` and let the implementer chose to use a `Cell` or `RefCell` if they need interior mutability. It's a matter of whether or not the callback semantically carries state. Given the return type, I imagine the answer is "yes", so maybe this signature is best anyways.

risc0/zkvm/src/host/client/env.rs
```diff
@@ -388,4 +399,11 @@ impl<'a> ExecutorEnvBuilder<'a> {
         self.inner.input_digest = Some(digest);
         self
     }
+
+    /// Add a callback for coprocessor requests.
+    #[stability::unstable]
+    pub fn coprocessor_callback(&mut self, handler: CoprocessorCallbackRef<'a>) -> &mut Self {
```
nategraf, line 405: Could this be `dyn CoprocessorCallback + 'a` or `impl CoprocessorCallback + 'a`, similar to how we use `impl Write + 'a` above?

nategraf (commented):
I've started to work through this. I'll get through the rest tomorrow.

examples/prover/src/main.rs
```diff
@@ -71,15 +108,48 @@ fn prover_example() {
...
+        let assumption_receipt = coprocessor.receipts.get(&assumption.claim).unwrap().clone();
+        let opts = ProverOpts::default();
+        succinct_receipt = client
+            .resolve(
+                &opts,
+                succinct_receipt.try_into().unwrap(),
+                assumption_receipt.try_into().unwrap(),
+                AssetRequest::Inline,
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
flaub: This emulates the way Bonsai works. I think @mothran might have a better sense of whether this will go into the work plan but last we talked it sounds like it was going to be part of TaskDB, not necessarily the work plan.

risc0/zkvm/src/host/api/client.rs
```diff
@@ -664,6 +713,21 @@ impl Client {
...
+        &self,
+        env: &ExecutorEnv<'_>,
+        coprocessor_request: pb::api::CoprocessorRequest,
+    ) -> Result<()> {
+        match coprocessor_request.kind.ok_or(malformed_err())? {
+            pb::api::coprocessor_request::Kind::ProveZkr(proof_request) => {
+                let proof_request = proof_request.try_into()?;
+                let coprocessor = env.coprocessor.clone().ok_or(malformed_err())?;
+                let mut coprocessor = coprocessor.borrow_mut();
+                coprocessor.prove_zkr(proof_request)
```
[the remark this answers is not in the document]
flaub: Ideally the request back to the server happens on a different machine. The main idea is that it's up to the callback handler to decide how it wants to handle it.

risc0/zkvm/src/host/api/mod.rs
flaub: I was really just looking for a way to enable logging on the server-side based on the client-side settings. It's not clear that we need to clear the env because otherwise, wouldn't the RUST_LOG have been inherited?

risc0/zkvm/src/host/client/env.rs
flaub: I've found it's much easier to just assume that the handler will mutate state.  But I've gone back and forth on this and I think this is consistent with other callbacks we have.

risc0/zkvm/src/host/client/env.rs
flaub: It could, but I basically need to make a copy of the ref anyways, so immediately I need to throw it into a Rc/RefCell. I'm finding from an API perspective it's just easier for both caller and callee to assume Rc/RefCell.

examples/prover/src/main.rs
```diff
@@ -20,26 +20,62 @@ mod plan;
...
+}
+
 fn prover_example() {
     println!("Submitting proof request...");
 
     let mut task_manager = TaskManager::new();
     let mut planner = Planner::default();
 
-    let iterations = 100;
+    let n = from_hex(b"9c98f9aacfc0b73c916a824db9afe39673dcb56c42dffe9de5b86d5748aca4d5");
```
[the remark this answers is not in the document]
flaub: @shkoo  or @tzerrell might know better than I. I basically copied this code from other examples. I really just needed something I could use as a workload for this integration.

risc0/r0vm/src/lib.rs
```diff
@@ -112,10 +112,6 @@ pub fn main() {
         .init();
 
     let args = Cli::parse();
-    if let Some(port) = args.mode.port {
-        run_server(port);
-        return;
-    }
 
     if args.id {
```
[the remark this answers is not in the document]
flaub: Turns out this isn't the case. If you do `--id` with the existing code, you get:
```
error: the following required arguments were not provided:
  <--port <PORT>|--elf <ELF>|--image <IMAGE>>
```

This is because `Mode` is already marked as required.

I tried adding `requires = "elf"` but the behavior didn't change.

examples/prover/src/main.rs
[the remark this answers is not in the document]
tzerrell: Yes. This verifies (or fails to verify, as the case may be) that `m == s^65537 (mod n)` (see https://github.com/risc0/zirgen/blob/507ed36bf3827ebef56c5796b30695c634c831e6/zirgen/circuit/bigint/rsa.cpp#L23). This is the algorithm for verifying that a signature `s`is valid for message `m` under public key modulus `n` (see https://en.wikipedia.org/wiki/RSA_(cryptosystem)#Signing_messages for more).

And yes, this is 256-bit RSA -- by definition, this is the bitwidth of `n`, which we can see is 32 bytes i.e. 256 bits. It's perhaps worth noting that 256-bit RSA isn't secure (2048 is the key size I most commonly hear recommended for RSA, and I believe is the current NIST recommendation).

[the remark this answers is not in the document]
nategraf: Indeed, NIST recommends either RSA 2048 or 3072
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-78-5.pdf

risc0/zkvm/src/host/api/mod.rs
nategraf: Ya, this one is odd to me. Reading the docs, it seems like an env var like `RUST_LOG` should already be inherited from the current process. If that was accurate, this this would do nothing 😕

risc0/zkvm/src/host/server/prove/prover_impl.rs
```diff
@@ -96,31 +98,48 @@ impl ProverServer for ProverImpl {
...
             .digest();
 
-        // Collect the proven assumption receipts from the Session.
-        // TODO(#982): Support unresolved assumptions here.
-        let assumption_receipts = session_assumption_receipts
+        let mut zkr_receipts = HashMap::new();
+        for proof_request in session.pending_zkrs.iter() {
+            let receipt = prove_zkr(&proof_request.control_id, &proof_request.input)?;
+            let claim_digest = receipt.claim.digest();
+            zkr_receipts.insert(claim_digest, receipt);
```
nategraf, line 105: Using only the claim digest as the key might not distinguish two incompatible receipts. In particular, if two versions of the same accelerator are used, or if there simply happens to be a collision in the claim format between two ZKRs (not something we can prevent in general, as it's easy to create a second ZKR that matches the claim structure of another), then this will break down. Using the `Assumption` (i.e. the `(control_root, claim_digest)` tuple) here would be one way to avoid this. The `control_id` could also be instead of the `control_root` to create the key.

risc0/zkvm/src/host/server/session.rs
```diff
@@ -171,16 +180,22 @@ impl Session {
         // Construct the Output struct for the session, checking internal consistency.
         // NOTE: The Session output is distinct from the final Segment output because in the
         // Session output any proven assumptions are not included.
+        self.claim_with_assumptions(self.assumptions.iter().map(|(_, x)| x))
+    }
+
+    pub(crate) fn claim_with_assumptions<'a>(
```
nategraf, line 186: Instead of adding this method, it should be possible (and semantically consistent) to add a `filter_map` below that removed assumptions that have a `pending_zkr`. The framing here being that `session.claim()` returns the claim that the caller can "expect to prove" with this session object. I.e. `session.claim == prover.prover_session(&ctx, &session)?.receipt.claim()`. The code in `ProverImpl::prove_session` and `CompositeReceipt::claim` handles this, and if we filter out assumptions that have a pending ZKR, then the results from `Session::claim` should match.

risc0/zkvm/src/host/client/env.rs
```diff
@@ -75,6 +66,29 @@ impl SegmentPath {
...
 }
 
+/// A ZKR proof request.
+#[stability::unstable]
+pub struct ProveZkrRequest {
+    /// The digest of the claim that this ZKR program is expected to produce.
+    pub claim_digest: Digest,
+
+    /// The control ID uniquely identifies the ZKR program to be proven.
+    pub control_id: Digest,
```
nategraf, line 76: After reading through the PR and thinking about it for a bit, I think I'm realizing the control root is essentially required here, while the control ID is an optional hint.

If we were to drop the `control_id` here and include only `control_root`, then it would be up to the host to decide which ZKR within the set identified by `control_root` to run. In our immediate use cases, this selection will be trivial; there is only one choice since each coprocessor will only have one ZKR. Probably the first example where the host would have some discretion is with RSA, where we might have a ZKR for RSA 2048, one for 3072, and one for 4096. We could place all of them (including variants with different batch sizes if that's useful) into a control root and let the host decide which one to run based on the input. Maybe we still want to provide the control ID from guest to host as a hint, but the point here is that the control root is actually the constraint to satisfy, and the control ID is unconstrained within that set (as long as the host produces a receipt).

I think providing the control root will make the linking of ZKR requests and assumptions / assumption receipt simpler and less likely to have edge cases. This is related to my comments on `prover_impl.rs`. If we have the `control_root` on the `ProveZkrRequest`, then we can decide exactly which assumption it resolves by constructing `Assumption { claim_digest, control_root }`. With only the `control_id`, it's not decidable because a single control ID can be part of many control roots.

```suggestion
    /// The control root identifies the coprocessor, as a set of ZKRs, to be executed and proven.
    pub control_root: Digest,
```

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        if let Some(coprocessor) = &ctx.syscall_table().coprocessor {
+            coprocessor.borrow_mut().prove_zkr(proof_request)?;
+        } else {
+            get_registered_zkr(&control_id)?;
+            ctx.syscall_table()
+                .pending_zkrs
+                .borrow_mut()
+                .push(proof_request);
+        }
```
nategraf, line 62: With this implementation, there is a bit of a dilemma for developers that want to add a new coprocessor to their proving context in that they need to provide a single handler for _all_ coprocessor requests. In the case of client-server with IPC, and the ZKR registry being a process-level mapping, this may be an issue. E.g. if someone is using `r0vm` through `LocalProver`, it might have the the RSA and Keccak coprocessors installed in the `r0vm` process' registry. If the client wants to add ED25519 verify (possibly a ZKR they wrote) the callback mechanism provides a way to do this. However, adding their coprocessor effectively "uninstalls" RSA and Keccak because all coprocessor requests will now be forwarded to the client. If they want to use those, they need to go install those as well.

Since we are focused on integration with Bonsai and other prover services here, that will want to take over ZKR proving with custom code, maybe this fits the needs of the use cases at hand. Another option here would be to have a mapping of `control_root` to coprocessor handler, such that each new coprocessor could indicate which requests it's taking ownership of. A default handler could be installed, which would take all requests not taken by a more specific handler.

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        let assumption = Assumption {
+            claim: claim_digest,
+            control_root,
+        };
+        ctx.syscall_table()
+            .assumptions
+            .borrow_mut()
+            .push(AssumptionReceipt::Unresolved(assumption));
```
nategraf, line 71: We might want to check if a proven assumption is already available on the syscall table before adding an unresolved assumption receipt here. This might happen if the application has pre-proven an assumption (as is required for rv32im proofs today) and added it via `env.add_assumption`. In this case, we could also skip calling the coprocessor handler. We also would not want the `SysVerify` call to pick up the unresolved assumption instead of the proven one. (In practice, I think this would not happen because the proven assumption would be first in the list, but this feels like a coincidence rather than by design)

nategraf: It definitely feels like the assumption cache should be a map instead of a list.

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
```diff
@@ -0,0 +1,75 @@
...
+        let input: Vec<u8> = ctx.load_region(input_ptr, input_len * WORD_SIZE as u32)?;
+
+        let proof_request = ProveZkrRequest {
+            claim_digest,
+            control_id,
+            input,
+        };
+
+        if let Some(coprocessor) = &ctx.syscall_table().coprocessor {
+            coprocessor.borrow_mut().prove_zkr(proof_request)?;
```
nategraf, line 55: Connecting this to our discussion earlier. In cases where we need execution to be a signal for whether or not something is provable, `prove_zkr` is required to check that the input will result in the provided claim. No changes required here, I just was thinking this through and figured it was worth mentioning.

risc0/zkvm/src/host/recursion/prove/mod.rs
```diff
@@ -40,14 +48,21 @@ use crate::{
...
-    ProverOpts, ReceiptClaim,
+    ProverOpts, ReceiptClaim, Unknown,
 };
 
 use risc0_circuit_recursion::prove::Program;
 
 /// Number of rows to use for the recursion circuit witness as a power of 2.
 pub const RECURSION_PO2: usize = 18;
 
+pub(crate) type ZkrRegistryEntry = Box<dyn Fn() -> Result<Program> + Send + 'static>;
```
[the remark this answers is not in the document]
flaub: Good note! Will think about this for a future PR.

risc0/zkvm/src/host/server/prove/prover_impl.rs
```diff
@@ -96,31 +98,48 @@ impl ProverServer for ProverImpl {
...
+        }
+
+        let inner_assumption_receipts: Vec<_> = session_assumption_receipts
             .into_iter()
-            .map(|a| a.into_receipt())
+            .map(|assumption_receipt| match assumption_receipt {
+                AssumptionReceipt::Proven(receipt) => Ok(receipt),
+                AssumptionReceipt::Unresolved(assumption) => {
+                    let receipt = zkr_receipts
+                        .remove(&assumption.claim)
```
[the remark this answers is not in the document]
flaub: We need to add more test cases for situations like this.

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
flaub: Let's go with this for now and see how well it integrates with Bonsai. Then I think we can come back and see if we even want to add the ability for users to add their own callback handlers. I actually started with this approach (but the key was `control_id` rather than `control_root`) and the part that was annoying was how to register the handlers to accept a wildcard (i.e. all requests like it is now).

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
flaub: Fair enough, I can try to see how a map would work out here.

risc0/zkvm/src/host/api/mod.rs
flaub: My interpretation is that the environment is already clean, and that we don't need to use `env_clear`.

risc0/zkvm/src/host/recursion/prove/mod.rs
```diff
@@ -228,6 +243,63 @@ pub fn identity_p254(a: &SuccinctReceipt<ReceiptClaim>) -> Result<SuccinctReceip
...
+    ))?;
+
+    let control_id_group = MerkleGroup::new([*control_id].into())?;
+    let control_inclusion_proof = control_id_group.get_proof_by_index(0, &*suite.hashfn);
+    let succinct_receipt = SuccinctReceipt {
+        seal: receipt.seal,
+        hashfn: hashfn.to_string(),
+        control_id: *control_id,
+        control_inclusion_proof,
+        claim: MaybePruned::<Unknown>::Pruned(claim_digest),
```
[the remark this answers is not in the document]
flaub: OK, let's fix this in a follow-up PR, I'm not exactly sure how this API works.

risc0/zkvm/src/host/client/env.rs
flaub: Let's see how this does with a few more coprocessors to see what the pattern is, it feels a bit premature to me.

risc0/zkvm/src/host/server/session.rs
flaub: I'm not sure follow. I think we can revisit this as we iterate on this. I wouldn't call this integration done for a few more PRs.

risc0/zkvm/src/host/client/env.rs
flaub: My main issue that seems unresolved in my mind is how we might want to group different ZKRs together. It seems arbitrary and I think that's fine. In which case, I think we'll want to use the `control_id` as the unique thing.

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
```diff
@@ -0,0 +1,82 @@
...
+            control_id,
+            input,
+        };
+
+        for assumption in ctx.syscall_table().assumptions.borrow().iter() {
+            if assumption.claim_digest()? == claim_digest {
+                // This assumption is already known, no need to create another proof.
+                return Ok((0, 0));
+            }
+        }
```
flaub, line 59: @nategraf Is this sufficient to detect duplicates?

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
flaub: I made an attempt to detect this situation, but using a map means we either need for `AssumptionReceipt` to be hashable or partially ordered, which brings up the question, how would we uniquely identify an `AssumptionReceipt`? I think we should punt on this and consider this for later.

risc0/circuit/bigint/examples/rsa.rs
```diff
@@ -0,0 +1,46 @@
...
+use risc0_circuit_bigint::zkr::register_zkrs;
+use risc0_circuit_bigint_test_methods::RSA_ELF;
+use risc0_zkvm::{ExecutorEnv, ExecutorImpl};
+
+fn main() {
+    tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .init();
+
+    register_zkrs();
```
SchmErik, line 26: this doesn't have to be addressed right now at all but I wonder if there's a way to avoid having the host-side code call this and instead have it included in some sort of existing sequence of functions that the host side already calls. it seems awkward for the user to have to call this in something like a starter template... one place would be to call this function when building the executor env but I'm not sure that's a great place for something like this

risc0/zkvm/src/host/client/env.rs
nategraf: I see that. Indeed, we don't have a clear view of the patterns yet. 👍

As a note, the practical problem that spurred this thought is as a way to make unique linking of prove ZKR requests and assumptions easier (related [1]). Adding the `control_root` to this would be one way to do that. I'm sure there are others as well.

[1] https://github.com/risc0/risc0/pull/2288/files/23aab6f936c9a4fdad9d5eaf92794d572d830f8c#r1744281831

risc0/zkvm/src/host/server/exec/syscall/prove_zkr.rs
nategraf: Oh, my thought was to use the `Assumption` as the key. But definitely no reason this must be a map.

risc0/zkvm/src/host/server/session.rs
nategraf: Sounds good, and can definitely address it later.

The core idea of this suggestion is that each `pending_zkr` represents a receipt that will be generated as part of proving the execution (e.g. using `prove_session` or using Bonsai's workflow implementation). When we generate the claim for the session, we can assume that will happen, and so drop any assumptions that will be proven by a `pending_zkr`.

There is a tricky bit about determining whether or not a given `ProveZkrRequest` resolves a particular assumption, because assumptions encode the `control_root` and ZKR requests the `control_id`. If we make the assumption that every `ProveZkrRequest` is proven against the singleton set of control IDs (which is all we handle right now and is assumed elsewhere to be the case), then this linking can happen simply be constructing the `control_root` for the singleton set `[control_id]`.

nategraf (approved):

PR #2302, opened 2024-09-02
docs: fix invalid links

osrm:
Some links within the document are not working. In many cases, Incorrect path of 'terminology' caused the links to be invalid.

Fixed.


pdg744:
Thank you for the attempted fix, but these links are formatted as intended & working as intended. 

The contents of the `website` directory are used to populate the [dev docs site](https://dev.risczero.com), and the links are working properly in that context.

osrm:
Understand, Thank you for your kind reply

PR #2310, opened 2024-09-03
BONSAI: Added cycle limit + SNARK response refactor

mothran:
This PR adds configurable execution cycle limits to bonsai session creation and refactors how snarks are delivered back to customers.

This does slightly break the API of how SNARKs work so it should be considered a API breaking change. But this release also moves a major version for the bonsai-sdk.

### DIFF
--- a/bonsai/sdk/Cargo.toml
+++ b/bonsai/sdk/Cargo.toml
@@ -16,7 +16,6 @@ reqwest = { version = "0.12", default-features = false, features = [
   "rustls-tls",
   "stream",
 ] }
-risc0-groth16 = { workspace = true }
 serde = { version = "1.0", features = ["derive"] }
 thiserror = "1.0"
 tokio = { version = "1", features = ["fs"], optional = true }
@@ -34,8 +33,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 uuid = { version = "1.10", features = ["v4"] }
 
 [features]
-default = ["std"]
-std = ["risc0-groth16/std"]
+default = []
 non_blocking = ["dep:tokio"]
 
 [package.metadata.docs.rs]
--- a/bonsai/sdk/README.md
+++ b/bonsai/sdk/README.md
@@ -86,6 +86,7 @@ use std::time::Duration;
 
 use anyhow::Result;
 use bonsai_sdk::blocking::Client;
+use risc0_zkvm::Receipt;
 
 fn run_stark2snark(session_id: String) -> Result<()> {
     let client = Client::from_env(risc0_zkvm::VERSION)?;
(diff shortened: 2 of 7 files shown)


rlukata (approved):

risc0/zkvm/src/host/client/prove/bonsai.rs
```diff
@@ -192,14 +197,8 @@ impl Prover for BonsaiProver {
...
-        let groth16_receipt = Receipt::new(
-            InnerReceipt::Groth16(Groth16Receipt {
-                seal: snark_receipt.snark.to_vec(),
-                claim: succinct_prove_info.receipt.claim()?,
-                verifier_parameters: ctx.groth16_verifier_parameters.digest(),
-            }),
-            succinct_prove_info.receipt.journal.bytes,
-        );
+        let receipt_buf = client.download(&snark_receipt_url)?;
+        let groth16_receipt: Receipt = bincode::deserialize(&receipt_buf)?;
```
nategraf, line 201: image

nategraf (approved):

SchmErik (approved):
LGTM - @mothran could you document details for the breaking change in the commit message when you squash/merge this? I'd like to mention the breaking change in the CHANGELOG

mothran:
> LGTM - @mothran could you document details for the breaking change in the commit message when you squash/merge this? I'd like to mention the breaking change in the CHANGELOG

Sure thing! can you point me to a good example of what to write aka what you have used in the past for breaking changes reference material?

bonsai/sdk/Cargo.toml
```diff
@@ -16,7 +16,6 @@ reqwest = { version = "0.12", default-features = false, features = [
   "rustls-tls",
   "stream",
 ] }
-risc0-groth16 = { workspace = true }
```
flaub, line 19: Fantastic!

bonsai/sdk/README.md
```diff
@@ -101,8 +102,8 @@ fn run_stark2snark(session_id: String) -> Result<()> {
                 continue;
             }
             "SUCCEEDED" => {
-                let snark_receipt = res.output;
-                eprintln!("Snark proof!: {snark_receipt:?}");
+                let receipt_buf = client.download(&res.output.unwrap())?;
+                let snark_receipt: Receipt = bincode::deserialize(&receipt_buf)?;
```
flaub, line 106: Is this here just to test that we can deserialize properly?

flaub:
I do wonder how this will work when we want to make changes to the `Receipt` in the future.

bonsai/sdk/README.md
mothran: The testing of the serialization happens in the bonsai integration test suite which is wrapping up final touches now.

mothran:
> I do wonder how this will work when we want to make changes to the `Receipt` in the future.

Currently we also deliver STARK's via `Receipt` as well so we currently already have that problem elsewhere in the API. Whats nice about this is dropping any explicit types in favor of buffers means we can make that change later. It would impact all the callsites but thats true of our STARKS. Right now without making a circular dep inside of risc-zkvm <-> bonsai-sdk I don't know a good way to make it explicit types.

bonsai/sdk/README.md
flaub: This seems like dead code, but maybe the diff is throwing me off.

mothran: Yeah its not used because its just a README example.

PR #2311, opened 2024-09-04
CI: add a workflow to test rzup

SchmErik:
This workflow will be used to test changes to the rzup implementation and will download the latest release to run a simple e2e test using the downloaded utilities.


SchmErik:
converting to draft until i see that the workflow is working as intended

flaub:
Seems redundant to all the other places we use `rzup` in our workflows.

SchmErik:
> Seems redundant to all the other places we use `rzup` in our workflows.

Now that I'm thinking about it and we have the paths filter, I also think this is redundant. Closing!

PR #2312, opened 2024-09-04
backport: updating to 97 bits (#2293)

SchmErik:
This docs change is being backported because it touches source code. backporting this commit to the release branch will make these appear on docs rs when the crates are published.


pdg744 (approved):
Thank you!

PR #2327, opened 2024-09-09
Add tests for deferred assumptions

flaub:


risc0/zkvm/src/receipt.rs
```diff
@@ -533,6 +533,27 @@ impl From<Assumption> for AssumptionReceipt {
...
+#[cfg(feature = "prove")]
+impl TryFrom<AssumptionReceipt> for Assumption {
+    type Error = anyhow::Error;
+
+    fn try_from(receipt: AssumptionReceipt) -> Result<Self> {
+        Ok(match receipt {
+            AssumptionReceipt::Proven(ref inner) => {
+                let control_root = match inner {
+                    InnerAssumptionReceipt::Succinct(receipt) => receipt.control_root()?,
+                    _ => Digest::ZERO,
```
flaub, line 545: @nategraf Does this make sense?

I'm not sure that this is the right thing to do, but it does seem to work.

flaub: Avoiding this approach

nategraf (approved):

PR #2336, opened 2024-09-11
Show unstable risc0-zkvm features on docs.rs

nategraf:
I went looking for the new `env::read_frame` function and realized that unstable methods are not shown on `docs.rs`. Assuming we want to show them, here is a PR to do so.


flaub (approved):
We probably need to do this on all our crates

nategraf:
> We probably need to do this on all our crates

Do we have the `unstable` feature anywhere else? When I implemented this I went looking for crates that use `stability`. There was one other, and it already had `unstable` in it's docs.rs config.

flaub:
Yes, `risc0-zkvm-platform`

PR #2342, opened 2024-09-12
Plumb ReceiptClaim through SessionInfo and client-server

### LINKED ISSUE #2267: `SessionInfo` should include a method to construct the associated `ReceiptClaim`

- [X] I have searched the existing issues

- [X] This is not a support question.

### Motivation

_No response_

### Request

Implement a `claim` method or field on `SessionInfo` to get the full `ReceiptClaim`.

### Solution

_No response_

### Are you willing to help with this request?

Yes!

nategraf:
This PR plumbs the `session.claim()` through to the `SessionInfo` such that it is available from executions carried out with the `Executor` trait. Motivation for this PR is that I am writing some tests of composition and want to be able to run the executor and convert the result into a fake receipt. This is possible with `ExecutorImpl`, but not with `Executor` right now.

More context on https://github.com/risc0/risc0/issues/2267#issuecomment-2347372529

Closes #2267

### DIFF
--- a/risc0/zkvm/src/host/api/mod.rs
+++ b/risc0/zkvm/src/host/api/mod.rs
@@ -41,7 +41,7 @@ use lazy_regex::regex_captures;
 use prost::Message;
 use semver::Version;
 
-use crate::{get_version, ExitCode, Journal};
+use crate::{get_version, ExitCode, Journal, ReceiptClaim};
 
 mod pb {
     pub(crate) mod api {
@@ -356,6 +356,10 @@ pub struct SessionInfo {
 
     /// The [ExitCode] of the session.
     pub exit_code: ExitCode,
+
+    /// The [ReceiptClaim] associated with the executed session. This receipt claim is what will be
+    /// proven if this session is passed to the Prover.
+    pub receipt_claim: ReceiptClaim,
 }
 
 impl SessionInfo {
--- a/risc0/zkvm/src/host/client/prove/local.rs
+++ b/risc0/zkvm/src/host/client/prove/local.rs
@@ -66,10 +66,12 @@ impl Executor for LocalProver {
             });
             Ok(Box::new(NullSegmentRef))
         })?;
+        let receipt_claim = session.claim()?;
         Ok(SessionInfo {
             segments,
             journal: session.journal.unwrap_or_default(),
             exit_code: session.exit_code,
+            receipt_claim,
         })
     }
 }
--- a/risc0/zkvm/src/host/protos/api.proto
+++ b/risc0/zkvm/src/host/protos/api.proto
@@ -206,6 +206,7 @@ message SessionInfo {
   uint32 segments = 1;
   bytes journal = 2;
   base.ExitCode exit_code = 3;
+  Asset receipt_claim = 4;
 }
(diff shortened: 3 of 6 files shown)


risc0/zkvm/src/host/api/mod.rs
```diff
@@ -356,6 +356,10 @@ pub struct SessionInfo {
 
     /// The [ExitCode] of the session.
     pub exit_code: ExitCode,
+
+    /// The [ReceiptClaim] associated with the executed session. This receipt claim is what will be
+    /// proven if this session is passed to the Prover.
+    pub receipt_claim: ReceiptClaim,
```
nategraf, line 362: Note that, because all the fields on this struct as `pub`, this is technically a breaking change. In particular, it would break users code if the use the `SessionInfo { _ }` syntax to make a `SessionInfo` object themselves. I would be somewhat surprised if anyone is actually doing this, although I would also understand if we do not want to accept that.

risc0/zkvm/src/host/client/prove/local.rs
```diff
@@ -66,10 +66,12 @@ impl Executor for LocalProver {
             });
             Ok(Box::new(NullSegmentRef))
         })?;
+        let receipt_claim = session.claim()?;
         Ok(SessionInfo {
             segments,
             journal: session.journal.unwrap_or_default(),
             exit_code: session.exit_code,
+            receipt_claim,
```
flaub, line 74:
```suggestion
        Ok(SessionInfo {
            segments,
            journal: session.journal.unwrap_or_default(),
            exit_code: session.exit_code,
            receipt_claim: session.claim()?,
```

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -206,6 +206,7 @@ message SessionInfo {
   uint32 segments = 1;
   bytes journal = 2;
   base.ExitCode exit_code = 3;
+  Asset receipt_claim = 4;
```
flaub, line 209: Why is this an Asset?

nategraf: It could be bytes instead. I just used Asset because that's how we package the other opaque messages.

risc0/zkvm/src/host/client/prove/local.rs
nategraf: That unfortunately makes the Rust borrow checker unhappy, because of the `session.journal` line above it.

risc0/zkvm/src/host/protos/api.proto
flaub: Why is it opaque?

risc0/zkvm/src/host/api/mod.rs
flaub: I don't like the idea of a receipts being associated with execution.

risc0/zkvm/src/host/protos/api.proto
nategraf: Ah, because the `ReceiptClaim` type is in `core.proto`

flaub (approved):

PR #2344, opened 2024-09-16
Ethereum Examples: updated governance links

sashaaldrick:
Updated the relevant links on "Blockchain Integration/Ethereum Examples" to point to the refactored Governance example.


sashaaldrick:
I forsaw these errors when editing 1.0/1.1 to point to `risc0-ethereum/main`. 

I find these checks too constraining when wanting to make changes to the docs. In the current situation, I'd have to open a new PR to update the risc0-ethereum `release-1.0` branch to include the governance example code and link to that, in order to match the versioning.

I don't personally see much value from this strict tracking of versions between the monorepo and risc0-ethereum but I am just in the process of finding out the procedure on the `risc0-ethereum` side of things, perhaps it makes sense to open that PR to `release-1.0`.

website/api_versioned_docs/version-1.0/blockchain-integration/eth-examples.md
```diff
@@ -24,12 +24,12 @@ This [example app][governance-example] uses Bonsai as an Ethereum coprocessor. T
 
 [blockchain-examples]: https://github.com/risc0/risc0-ethereum/tree/release-1.0/examples
 [foundry-template]: https://github.com/risc0/risc0-foundry-template
-[governance-example]: https://github.com/risc0/risc0/tree/release-0.20/bonsai/examples/governance
+[governance-example]: https://github.com/risc0/risc0-ethereum/tree/main/examples/governance
```
pdg744, line 27: hmm... do we want to point to `main` or `release-1.0`?

pdg744:
> I forsaw these errors when editing 1.0/1.1 to point to `risc0-ethereum/main`.
> 
> I find these checks too constraining when wanting to make changes to the docs. In the current situation, I'd have to open a new PR to update the risc0-ethereum `release-1.0` branch to include the governance example code and link to that, in order to match the versioning.
> 
> I don't personally see much value from this strict tracking of versions between the monorepo and risc0-ethereum nevertheless I am in the process of finding out the procedure on the `risc0-ethereum` side of things, perhaps it makes sense to open that PR to `release-1.0` and then this would solve the current issue.

hmm i think we do probably want users to land on a stable example rather than `main`. 

my thinking is that if we want to point to `main` on risc0-ethereum, we would probably also need to insist that the web3 teams makes sure the examples on `main` don't break as they iterate. e.g., if the upcoming version of risc0-ethereum includes a re-design on the publisher architecture, users trying out the governance demo shouldn't be impacted. 

i do agree though that the friction here is pretty annoying. i've come around to thinking that this may be necessary friction in order to be able to offer high quality docs in the context of an actively changing system architecture, but i'd love to figure out how to make this feel easier...

pdg744:
cc @flaub

sashaaldrick:
thanks for your comment @pdg744, I now understand the reasoning entirely behind this friction. The solution here is to follow the friction and point to the right release.

sashaaldrick:
Waiting on this PR to be reviewed and merged: https://github.com/risc0/risc0-ethereum/pull/247. Once this is done, I will revisit this PR to point to the correct example links in `release-1.0`.

flaub:
If we don't have an example that works against a particular stable release, then we shouldn't have a link in the docs. However, once the example is backported to that stable release and is included in CI, then I'd say it's safe to add the link.

flaub:
I think in the future, we should plan to add examples as part of a release and only backport them if it's not going to take a lot of extra work. There's an ongoing maintenance cost for older versions that we might not be prepared to support.

PR #2345, opened 2024-09-16
[ON HOLD] fix(docs): make 1.1 latest in docs

nahoc:
resolves: https://linear.app/risczero/issue/DR-330

now https://docs-website-git-nahoc-docs-versioning-risczero.vercel.app/api/generating-proofs/dev-mode/ works


website/docusaurus.config.js
```diff
@@ -70,7 +71,7 @@ export default async function createConfigAsync() {
             // Otherwise we risk losing the update on the next release.
             return `https://github.com/risc0/risc0/edit/main/website/api/${docPath}`;
           },
-          lastVersion: "1.0",
+          lastVersion: apiVersions[0],
```
nahoc, line 74: automatically maps to latest version, so that way we don't forget to update it

website/package.json
```diff
@@ -63,5 +63,5 @@
     "remark-toc": "9.0.0",
     "remark-wiki-link": "2.0.1"
   },
-  "packageManager": "bun@1.1.21"
+  "packageManager": "bun@1.1.27"
 }
```
nahoc, line 67: I also tried to update to docusaurus to 3.5, but getting ESM errors. Looking into it.

website/docusaurus.config.js
flaub: I specifically pinned this to `1.0` since we already snapped `1.1` but we are pausing the final release. Once `1.1` is fully released, we can just remove `lastVersion` since the default already does this.

flaub (changes requested):
We don't want to make 1.1 the latest just yet.

nuke-web3:
fixes DR-330

PR #2356, opened 2024-09-18
Support wider range of client/server use cases

flaub:


mothran (approved):

mothran:
I added two more commits to the sibling PR: so we should hold off on merging until those are included:
93cf264557014524295e7c995cdf112d3c31a601
82d44f0e08e61c02330e667d7cb7de3f3687c2be

PR #2358, opened 2024-09-18
Support wider range of client/server use cases

flaub:


.github/workflows/main.yml
```diff
@@ -223,7 +223,7 @@ jobs:
       - uses: ./.github/actions/rustup
       - uses: ./.github/actions/sccache
         with:
-          key: ${{ matrix.os }}-${{ matrix.feature }}
+          key: ${{ matrix.os }}-${{ matrix.feature }}-1.0
```
flaub, line 226: @mothran @morgangibbs87 It looks like we have some sccache invalidation issues. We need this to apparently force the cache between different versions to be distinct. Otherwise we get weird linking issues which makes it seem like the CUDA caching isn't working right.

mothran: Interesting, I will double check out ssccache based builds in our docker agent builds since I think that builds the kernels.

mothran:
Small nit: This PR title does not quite match the contents since this is just back porting the verify and identity fixes.

mothran (approved):

PR #2366, opened 2024-09-23
feat(reports): add copy to benchmarks

nahoc:
Solves https://linear.app/risczero/issue/FRONT-3/benchmarks-copy-edits

image

also updated fonts in regards to https://linear.app/risczero/issue/WEBR0-25/text-formatting-bug-on-zkvm-design-blogpost


flaub (changes requested):
The `benchmarks` directory is only involved with the Application Benchmarks page. I'm also not sure we should link to the source code because it can change over time.

pdg744:
> I'm also not sure we should link to the source code because it can change over time.
Hmm I hear you, but it feels pretty essential to me that users (on our team and externally) can make meaning of the reports, and it seems to me that adding a link to the source code is the most straight-forward path to addressing that issue. I think the maintenance cost is worth the benefit, but definitely open to other solutions if you have suggestions for a better way to address the problem of "what do these reports mean?"

pdg744:
> The benchmarks directory is only involved with the Application Benchmarks page.

good catch... some notes below on the 4 status of the 4 reports, in my eyes: 

- crates validation page is good as is; i don't think users will struggle to make meaning of that info. 
- datasheet page is good as is; there's already a pointer to the recursive proving architecture page and an "example" at the top of the page
- application benchmarks & benchmarks are currently very opaque to interpret. 
  - `Application Benchmarks` would be clarified by a pointer to `risc0/benchmarks`
  - For `Benchmarks`, I guess I don't actually what to point users to in order to clarify meaning here. I guess we could just include a link to the [Recursion page on dev docs](https://dev.risczero.com/api/recursion) and call that sufficient.

pdg744 (approved):

pdg744 (changes requested):
Whoops -- hit "approve" but it looks like there are some suggestions above that haven't been incorporated yet.

PR #2371, opened 2024-09-24
Add Elliptic Curve acceleration operations

tzerrell:
Adds Elliptic Curve (EC) functions/ZKRs based on the BigInt dialect, with operations for verifying add/sub/neg/mul/double plus point equality and verifying point on curve. Includes ZKRs over the secp256k1 curve as well as over a tiny custom curve for testing purposes.

This work is done in conjunction with https://github.com/risc0/zirgen/pull/41, which is the source of the generated bigint code.

The code in this PR is largely code testing these EC ZKRs (and @shkoo's less dynamic codegen, see below). To this end, I've migrated more of the ZKR-testing code to shared infrastructure (in `test_harness.rs`). I've expanded that to include the option for `bigint_short_tests` which in contrast to `bigint_tests` _only_ test for a successful proof and do not test against a fixed golden set of witness & z values (which can get quite long for large ZKRs). There's also now an option for `bigint_should_fail_tests` for testing that inputs which _should_ cause a failed constraint _do_ cause a failed constraint. I've tried to keep these tests trimmed to achieve a good balance of testing perf & coverage, but I'm definitely open to suggestions for improvement here if this ends up being slow in CI or on other people's local systems.

The new EC ZKRs this PR introduces are designed for unit tests and as components for larger systems. No larger system uses them yet, which is probably for the best as this system needs more review and these are not fully production-ready.

Includes @shkoo's work for improved compile times via reducing dynamic codegen (e.g., with bigint op macros) (see the included commits https://github.com/risc0/risc0/commit/673c5584f55209b2fe115485a0f8b68dcea020f4 and https://github.com/risc0/risc0/commit/5b6e88e9461ea1a5ac2fdd808f52fc96f3b3dbc9).


flaub:
I'm noticing that the tests are taking 43m to run in CI. We'd like to target more like 20m total time, so I wonder if we need to adjust the tests to be a bit faster or smaller. I'm also not entirely sure yet if it's this PR that is going slow or if our caching got broken somewhere.

SchmErik:
> I'm noticing that the tests are taking 43m to run in CI. We'd like to target more like 20m total time, so I wonder if we need to adjust the tests to be a bit faster or smaller. I'm also not entirely sure yet if it's this PR that is going slow or if our caching got broken somewhere.

I think the caching broke for linux

tzerrell:
> I'm noticing that the tests are taking 43m to run in CI. We'd like to target more like 20m total time, so I wonder if we need to adjust the tests to be a bit faster or smaller. I'm also not entirely sure yet if it's this PR that is going slow or if our caching got broken somewhere.

👍 I will trim down until the tests are in an acceptable time frame -- it certainly could be due to this PR, the tests were running quite slow locally until I did a first pass of trimming so there might be more trimming needed. I'm OOO until Friday so I'll wait until then to see if any caching fixes have happened.

If it really is taking a full 23 extra minutes the cuts will be substantial (possibly all the way down to "test mul on secp256k1 and nothing else" or maybe even on the 8 bit test curve instead). Mul does exercise most of the other ops so that's probably ok for CI, but I think we should leave the other tests available but off in CI to make root causes easier to find if/when there is a bug.

Cargo.toml
```diff
@@ -61,7 +61,22 @@ lto = true
 
 # Always optimize; otherwise tests take excessively long.
 [profile.dev]
+debug = 0
```
shkoo, line 64: I think we only want `debug = 0` for the bigint package, right?

Cargo.toml
```diff
@@ -61,7 +61,22 @@ lto = true
...
 opt-level = 3
 
+[profile.test]
```
shkoo, line 67: I think we only want `debug = 0` for the bigint package, right?

shkoo (approved):

SchmErik:
I think this caused by caching breaking - i'm seeing long build times for r0vm cc @morgangibbs87

tzerrell:
CI Perf:

Over the weekend I made a pair of PRs comparing a minimal suite of bigint tests to the full suite bigint of tests that we've written (in [0da35f8](https://github.com/risc0/risc0/pull/2371/commits/0da35f827cbc405bd37df6ace5d616710dffb6a5) and [6773686](https://github.com/risc0/risc0/pull/2371/commits/6773686e0326320d0236a356693fc01c501d02be)). Unfortunately, I wasn't able to learn much from this: for the linux-cpu/cuda/metal tests I saw 37/32/11 min for the version with few tests and 34/19/13 min for the version with a full test suite. If these timings were giving us useful data, we should not see any of these times be _longer_ in the fewer tests case.

To be cautious with our CI performance, I commented out some semi-redundant tests. This reduces coverage a little (we now mostly have just one test per op, and are mostly testing on just a small 8-bit test curve), but it still covers the main functionality, and I did leave a secp256k1 test for EC mul, which is a pretty thorough case (for a single test). By leaving the other tests inactive but available as comments, I'm hoping if/when problems do arise whoever is debugging can uncomment those tests and drill deeper on the cause of the problems.

One thing I didn't do, but perhaps could, is modify https://github.com/risc0/zirgen/pull/41 to generate fewer ZKRs. I would expect this to only matter when compiling the bigint ZKRs, which I don't think ought to be all that often -- although maybe I'm mistaken, and also maybe the problems with caching will make these recompile when they otherwise wouldn't.

tzerrell:
In the CI runs after https://github.com/risc0/risc0/pull/2371/commits/3cade8ca36084ff28d94b93a3bd8a91dbd3778b5 I'm seeing test failures that seem unrelated to this PR, e.g.
```
error: creating a shared reference to mutable static is discouraged
   --> risc0/zkvm/src/guest/env/mod.rs:122:9
    |
122 |         HASHER.set(Sha256::new()).unwrap();
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^ shared reference to mutable static
```
(others as well)

Since these seem to be related to files this PR hasn't touched, I'm inclined not to look to closely myself and just continue to pull in from `main` as new commits land.

tzerrell:
I've added the contents of the branch https://github.com/risc0/risc0/tree/tzerrell/ec-mul-guest to this PR -- this adds an e2e prove_and_verify test for EC Mul and creates an example EC Mul guest.

Beyond that, my understanding of what we've discussed on the build/test perf issues of this PR are:
* These perf issues are real, and we want to address them
* They are nontrivial, and not part of the same "core idea" as this PR
* While this PR has temporary slowness, it will not create large, permanent problems for general users of this repo
* Therefore, it probably makes sense to land this PR and iterate

@risc0/platform if you agree with that summary, can you review/approve this PR and I'll land it?

flaub (approved):

PR #2372, opened 2024-09-25
fix: rzup install --version <VERSION>

hmrtn:
This fixes `rzup install --version <VERSION>` such that if a version is specified, `rzup` will  install the default toolchains and the specified version of `cargo-risczero`. 

Currently, the documentation around this does not indicate the user must specify the name of the thing they are trying to install, i.e. `rzup install install cargo-risczero/rust/cpp --version <VERSION>`. 

Note: this still supports `rzup install cargo-risczero --version <VERSION>`.


0xinit:
the docs [here](https://dev.risczero.com/api/zkvm/install#:~:text=For%20a%20specific%20version%2C%20use%20rzup%20install%20%2D%2Dversion%20%3Cversion%3E%2C%20where%20the%20%3Cversion%3E%20is%20a%20release%20tag.) say that users can use rzup install --version 1.0.5, but it is currently not working. This PR hopefully resolves that?

Any timeline to merge this ? @hmrtn

flaub (approved):

PR #2375, opened 2024-09-26
Makes Bonsai Dependencies Optional under client feature flag

austbot:
The zkvm `client` feature flag includes the `bonsai-sdk`, which in turn includes the `bonsai-sdk` dependancy tree.
Because the `bonsai-sdk` includes `reqwest` and its `rustls-tls` users may find that they have version conflicts with various low level packages such as `zeroize` and other cryptography related crates. This is especially henious when working with other blockchain sdk crates such as solana. 

I propose the following change which is to make the bonsai-sdk only active when a developer specifically asks for it. This change will reduce compiled binary size, allow devs to unshackle themselves from forked versions of risc0 such as https://github.com/anagrambuild/risc0/tree/v1.0.1-bonsai-fix, and allow devs implementing their own hosts to not have to bleed so much.

Its likely that some docs need to be updated and perhaps examples aswell in order to fully integrate this change, but this PR is meant to see if the Risc0 team will accept this type of change.

Note the `bonsai` flag will default to on so that current sdk consumoooors will not be affected.


flaub:
This seems reasonable for me, and we were planning to do something similar. @mothran any issues with this?

mothran:
> This seems reasonable to me, and we were planning to do something similar. @mothran any issues with this?

Nope, makes sense to me.

flaub (approved):

SchmErik:
@austbot could you update your branch? we'd like to merge it but it's out of date

flaub:
The base branch needs to be updated again. There's an option in github that would allow us to update your branch.

PR #2376, opened 2024-09-26
change visibility of groth16 proof

wz14:
This PR changes the visibility of some structures and functions of groth16.
It's helpful to extract the *real* groth16 proof from r0 proof receipt for those teams writing on-chain verifier themselves.

### DIFF
--- a/risc0/groth16/src/verifier.rs
+++ b/risc0/groth16/src/verifier.rs
@@ -158,7 +158,7 @@ impl Verifier {
 
 /// Verifying key for Groth16 proofs.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct Fr(#[serde(with = "serde_ark")] pub(crate) ark_bn254::Fr);
+pub struct Fr(#[serde(with = "serde_ark")] pub ark_bn254::Fr);
 
 impl Digestible for Fr {
     /// Compute a tagged hash of the [Fr] value.
@@ -178,7 +178,7 @@ impl Digestible for Fr {
 
 /// Verifying key for Groth16 proofs.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct VerifyingKey(#[serde(with = "serde_ark")] pub(crate) ark_groth16::VerifyingKey<Bn254>);
+pub struct VerifyingKey(#[serde(with = "serde_ark")] pub ark_groth16::VerifyingKey<Bn254>);
 
 /// Hash a point on G1 or G2 by hashing the concatenated big-endian representation of (x, y).
 fn hash_point<S: Sha256>(p: impl AffineRepr) -> Digest {
--- a/risc0/groth16/src/verifier.rs
+++ b/risc0/groth16/src/verifier.rs
@@ -160,6 +160,13 @@ impl Verifier {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Fr(#[serde(with = "serde_ark")] pub ark_bn254::Fr);
 
+impl Fr {
+    #[stability::unstable]
+    pub fn ark_fr(&self) -> ark_bn254::Fr {
+        self.0
+    }
+}
+
 impl Digestible for Fr {
     /// Compute a tagged hash of the [Fr] value.
     fn digest<S: Sha256>(&self) -> Digest {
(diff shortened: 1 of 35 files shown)


nategraf:
Hi @wz14! These fields and functions are not `pub` because they are not part of the stable API surface. We actually have plans to change the underlying implementation of these types from Arkworks to another implementation. If they were `pub`, this would be a breaking change.

In order to extract the Groth16 proof from its encoded form as the seal. This is what we do in `risc0-ethereum` [1].

[1] https://github.com/risc0/risc0-ethereum/blob/ac4a73fbed87ebad34306e7808fd951dbfae3f75/contracts/src/groth16/RiscZeroGroth16Verifier.sol#L163

flaub:
@nategraf could we make them `pub` but also mark as `stability::unstable`? That way we can unblock today, with no guarantee that these functions will exist in a future release.

nategraf:
@flaub, sure! That seems like a good way to handle this. @wz14, does that work for you? Here is an example of what we are referring to. https://github.com/risc0/risc0/blob/a917620feaa03b030db8e5f6d38d7c8b764cdf99/risc0/zkvm/src/guest/env/mod.rs#L444

If you mark these are unstable, I'll re-review this PR and would be happy to approve it.

wz14:
@nategraf Hi, I've marked `unstable` to these functions. Please help review this PR.

risc0/groth16/src/verifier.rs
```diff
@@ -158,7 +158,14 @@ impl Verifier {
 
 /// Verifying key for Groth16 proofs.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct Fr(#[serde(with = "serde_ark")] pub(crate) ark_bn254::Fr);
+pub struct Fr(#[serde(with = "serde_ark")] pub ark_bn254::Fr);
```
flaub, line 161:
```suggestion
pub struct Fr(#[serde(with = "serde_ark")] ark_bn254::Fr);
```

This needs to remain private since `ark_fr` provides access in an unstable way.

risc0/groth16/src/verifier.rs
```diff
@@ -178,7 +185,7 @@ impl Digestible for Fr {
 
 /// Verifying key for Groth16 proofs.
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct VerifyingKey(#[serde(with = "serde_ark")] pub(crate) ark_groth16::VerifyingKey<Bn254>);
+pub struct VerifyingKey(#[serde(with = "serde_ark")] pub ark_groth16::VerifyingKey<Bn254>);
```
flaub, line 188: ditto here

wz14:
Thanks. @flaub Already Corrected.

nategraf (approved):
Looks good! Thanks for making those changes

wz14:
Hi, here seems to need a Vercel approval from you @nategraf

nategraf:
> Hi, here seems to need a Vercel approval from you @nategraf 

Vercel is for previewing the website. You shouldn't need it here. Merging is currently blocked because the branch is out of date with `main`. If you merge `main`, it should be unblocked. Let me know you need another review at that point

wz14:
> > Hi, here seems to need a Vercel approval from you @nategraf
> 
> Vercel is for previewing the website. You shouldn't need it here. Merging is currently blocked because the branch is out of date with `main`. If you merge `main`, it should be unblocked. Let me know you need another review at that point

Already catch up to `main` branch. Seems still to be blocked.

nategraf:
Ah, I see now. `cargo sort --check` us failing. Let me know when you've resolved that. Alternatively I could push to your branch to resolve minor things. Whichever works for you.
https://github.com/risc0/risc0/actions/runs/11210344610/job/31164655768#step:11:1

wz14:
> Ah, I see now. `cargo sort --check` us failing. Let me know when you've resolved that. Alternatively I could push to your branch to resolve minor things. Whichever works for you. https://github.com/risc0/risc0/actions/runs/11210344610/job/31164655768#step:11:1

Really thx. I'm guessing the problem is the dependencies for groth16 are not sorted. It should be passed now. @nategraf  Need another approval to run the CI.

wz14:
Hi, update Cargo.lock for benchmark and examples. Need another approval to run CI. @nategraf

nategraf (approved):

PR #2380, opened 2024-09-27
risc0-zkvm: add syscalls to accelerate keccak

SchmErik:
This PR adds plumbing to accelerate keccak. Placing this in draft mode for now. in order to land this, I need to do the following
- [ ] add claim when calling `sys_keccak_read`
- [x] create input transcript for accelerator
- [ ] figure out ways to support multiple hasher instances (might defer for a later PR)


SchmErik:
closed in favor of #2503

PR #2381, opened 2024-09-30
fix: fix macOS x86 reqwest build issue

norton-zh:
To solve the problem that occurs when executing “cargo install --path risc0/cargo-risczero” on the macOS x86 system.
issue:
image


SchmErik:
Thanks for your contribution! There were a few things that needed to be adjusted and the issue was preventing our CI from working so this has been superseded by #2388 - I took some of your work so I added a Co-authored-by tag to credit you for your contribution. So I'll close this

PR #2096, opened 2024-07-16
[ready to merge] feat(website): add remark linting configuration

nahoc:
follows https://github.com/risc0/risc0/pull/2094

## What's this PR for?

This PR leverages `remark` to lint/format all .md files instead of prettier.

Pros:
- we already use remark under the hood for docusaurus
- more consistent markdown files across the risc0's codebase
- vscode extension to lint files + format on save if you want
- catch old removed links

Cons:
- might break the CI a bit more, although for VSCode users you format on save

## What this PR does

- This PR adds 2 new scripts `check-markdown` and `fmt-markdown` and runs the former on CI.
- This auto-checks all the .md files and gives them a consistent markdown _without changing anything visually on the website_.
- I took out some rules to make it as close to what we already had in the codebase.
- I added a check in the CI
- I added the `remark` extension to the suggested vscode extensions

## Follow-up items

- Remove all occurences of `[here]` since it makes links harder to find?

PR #2395, opened 2024-10-01
`bun run fmt-markdown` run on `/website`

nuke-web3:
I suppose we don't have this in CI for a reason? :thinking: @nahoc  https://github.com/risc0/risc0/pull/2096/files#r1679667431 seems to indicate we _should_ have the lint but perhaps missed with the cli flags used in the `bun run lint` command? 

I run `"fmt-markdown-full": "remark . --output" with this PR and get no more changes on top of this PR's fixes


examples/profiling/README.md
```diff
@@ -64,21 +64,20 @@ This can be helpful in understanding the efficiency of various algorithms and th
 Use the pprof web interface to compare the performance of the 3 Fibonacci implementations.
 Refer to the [pprof docs] for more information about the web interface.
 
-[^1]:
-    Here "sampling" is in quotes because the profiler actually captures the call
+[^1]: Here "sampling" is in quotes because the profiler actually captures the call
```
nuke-web3, line 67: All other pages work fine with remark formatting... except this one. no mater what way I try to format it, it seems to decide footnotes are "not a thing" and formats to this (without the > quote prefix):

```markdown

> \[^1]: Here "sampling" is in quotes because the profiler actually captures the call
> stack at every cycle of program execution. Capturing a call stack on every
> cycle of execution is not done in most programs on physical CPUs for a few
> reasons:
> 
> ```
> - It would be cost prohibitive to do so for all but quite short program
>   executions.
> - Introducing such heavy profiling would actually alter the performance
>   characteristics in significant ways.
> 
> In zkVM execution, executions are generally short and all execution is
> synchronous and is not subject to any deviations in behavior due to
> measurement overhead.
> ```

```

unclear how to resolve that all... IMHO all MD files in the repo should use the same formatting & linting if possible

flaub (approved):

examples/profiling/README.md
nahoc: yes, I have faced the same issue, I haven't quite found a way to fix this.

The only thing I was able to do is manually disable auto-save, revert he change and save again. Yes I agree it sucks.

Might revisit this today, maybe I'll have to write a custom remark plugin for this.

website/api_versioned_docs/version-0.21/zkvm/optimization.md
```diff
@@ -468,17 +468,20 @@ below.
...
+
     <!-- HACK: This comment prevents the list below from being interpreted to be a code block -->
+
     - It would be cost prohibitive to do so for all but quite short program
       executions.
     - Introducing such heavy profiling would actually alter the performance
       characteristics in significant ways.
+
     <!-- -->
+
```
nahoc, line 484: all of these changes will break tests -- basically the `<!-- HACK: -->` comment gets disabled, and this code gets treated as a test block.. Same issue. I spent 5 hours trying to debug this and no matter what, remark keeps formatting footnotes wrong.

Will need to dig a bit deeper.

nahoc (changes requested):
Need to figure out a way to not break tests with footnotes

examples/profiling/README.md
nuke-web3: Yeah, bummer on that behavior... have we filed a bug report yet?
I would advocate to get a md formatter for the  whole repo, and it is implicitly with the `.vscode` extension settings (if you use that IDE) 

But if that is much work to get configured to be sure it will not get in the way of writing docs (IMHO friction to get docs done is a _no no_ as it's usually a chore eng. don't want to do anyway... best devex for docs possible is a goal worth pursuing IMHO, even if it means a bit less polish)

nahoc (approved):
ok, think i fixed it

PR #2407, opened 2024-10-07
Add redis option to client/server model

rlukata:
This PR adds a new feature to the r0vm that enables adding `segments` to a redis instance instead of returning the segment via the callback function. Instead of returning the full segment, the callback function returns only a redis key.

Tasks:
- [x] AssetRequest `Redis` value
- [x] Protobuf `Redis` value
- [x] Handle conversions
- [x] Tests (unit, integration, or example)
- [x] Concurrency or threads


risc0/zkvm/Cargo.toml
```diff
@@ -34,6 +34,7 @@ borsh = { version = "1.5", default-features = false, features = ["derive"] }
 bytemuck = { version = "1.13", features = ["extern_crate_alloc"] }
 getrandom = { version = "0.2", features = ["custom"] }
 hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
+redis = { version = "0.27.2", default-features = false, features = ["tls-native-tls"]}
```
flaub, line 37: Would be good to make this enabled by a feature flag

risc0/zkvm/src/host/api/server.rs
```diff
@@ -323,39 +324,89 @@ impl Server {
...
-
-                Ok(Box::new(NullSegmentRef))
-            })?;
+            // use redis if available
+            let session = if let AssetRequest::Redis(url, key, ttl) =
+                AssetRequest::try_from(segments_out.clone())?
+            {
+                let client = redis::Client::open(url)?;
+                let mut connection = client.get_connection()?;
+                exec.run_with_callback(|segment| {
```
flaub, line 333: Could we use a chain of responsibility pattern for this?

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -226,6 +226,12 @@ message ProveSegmentResult {
   Asset receipt = 1;
 }
 
+message RedisRequest {
```
flaub, line 229:
```suggestion
message RedisParams {
```

rlukata: [75d1fb8](https://github.com/risc0/risc0/pull/2407/commits/75d1fb84a670402dda2cd67f0b84c79af759a172)

risc0/zkvm/Cargo.toml
rlukata: Great advice, cleaned up nicely all the lock files: [75d1fb8](https://github.com/risc0/risc0/pull/2407/commits/75d1fb84a670402dda2cd67f0b84c79af759a172)

risc0/zkvm/src/host/api/server.rs
rlukata: Could I bug you for an example/pseudocode? I'm aware of the pattern, just not sure how to approach it for this specific use case.

risc0/zkvm/src/host/protos/api.proto
```diff
@@ -226,17 +226,25 @@ message ProveSegmentResult {
...
+  string url = 1;
+  string key = 2;
+  uint64 ttl = 3;
+}
+
 message Asset {
   oneof kind {
     bytes inline = 1;
     string path = 2;
+    bytes redis = 3;
```
flaub, line 239:
```suggestion
    RedisAsset redis = 3;
```

```
message RedisAsset {
  string key = 1;
  uint64 index = 2;
}
```

risc0/zkvm/methods/guest/Cargo.toml
```diff
@@ -34,3 +34,6 @@ debug = 1
 
 [package.metadata.release]
 release = false
+
+[features]
+redis = []
```
flaub, line 39: why do we need this?

rlukata: we don't, but my rust-analyzer kept complaining that it doesn't have a redis feature and would stop working

rlukata: I [removed it](https://github.com/risc0/risc0/pull/2407/commits/5e5938602154f54233d77b4cfd0557569c23e43b) and now rust-analyzer isn't complaining anymore 🤷🏽

risc0/zkvm/src/host/api/server.rs
```diff
@@ -835,12 +833,27 @@ fn execute_redis(
     exec: &mut ExecutorImpl,
     params: super::RedisParams,
 ) -> Result<crate::Session> {
-    let client = redis::Client::open(params.url)?;
-    let mut connection = client.get_connection()?;
-    exec.run_with_callback(|segment| {
+    let (sender, receiver) = kanal::unbounded();
```
flaub, line 836: Might want to bound this, unclear what the bound should be however.

risc0/zkvm/src/host/api/server.rs
```diff
@@ -835,12 +833,27 @@ fn execute_redis(
...
+                .map_err(anyhow::Error::new)
+                .unwrap();
+        }
+    });
+
+    let session = exec.run_with_callback(|segment| {
         let segment_bytes = bincode::serialize(&segment)?;
         let segment_key = format!("{}:{}", params.key, segment.index);
-        connection.set_ex(segment_key.clone(), segment_bytes, params.ttl)?;
+        sender.send((segment_key.clone(), segment_bytes))?;
```
flaub, line 856: Can we send the Segment directly and do the serialization over on the other thread?

risc0/zkvm/src/host/api/server.rs
rlukata: I'll test a few values, but the [crate's benchmark](https://github.com/fereidani/rust-channel-benchmarks) didn't show a big improvement vs unbounded.

risc0/zkvm/src/host/api/server.rs
rlukata: [d90aaba](https://github.com/risc0/risc0/pull/2407/commits/d90aaba376ae3960dbc49a422102e0abca8dca56)

risc0/zkvm/src/host/api/server.rs
rlukata: [d90aaba](https://github.com/risc0/risc0/pull/2407/commits/d90aaba376ae3960dbc49a422102e0abca8dca56)

I set the size to 100, to mimic what we have in Bonsai. I didn't notice a difference in performance between a bound of 10, 50, and 100.

mothran: I would suggest have the default be override-able via env vars so we can tune it externally to our machine memory IMHO.

rlukata: [Done](https://github.com/risc0/risc0/pull/2407/commits/6ba76b94c7ea0cffebb7ed824290c7fcf4ffd962#diff-6a4cb9d6d6a46162414dc817ee4e765428b256aa3307d039484053b40cbb800cR836-R840)

risc0/zkvm/methods/std/Cargo.toml
```diff
@@ -34,6 +34,7 @@ release = false
 test_feature1 = []
 test_feature2 = []
 test_feature3 = []
+redis = []
```
flaub, line 37: Not sure why we need this?

risc0/zkvm/src/host/api/client.rs
```diff
@@ -666,7 +665,7 @@ impl Client {
                                     .try_into()?,
                                 }),
                                 None => Err(malformed_err()),
-                            }
+                            };
```
flaub, line 668: would be nice to revert changes to this file, to make backporting easier down the road

tools/crates-validator/Cargo.toml
```diff
@@ -29,3 +29,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
 [dev-dependencies]
 rstest = "0.18"
+
+[features]
+redis = []
```
flaub, line 34: should drop this

benchmarks/Cargo.toml
```diff
@@ -33,6 +33,7 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 cuda = ["risc0-zkvm/cuda"]
 default = []
 prove = ["risc0-zkvm/prove"]
+redis = []
```
flaub, line 36: I think this can be dropped

risc0/zkvm/Cargo.toml
```diff
@@ -62,10 +62,12 @@ bincode = { version = "1.3", optional = true }
 bonsai-sdk = { workspace = true, optional = true }
 bytes = { version = "1.7", features = ["serde"], optional = true }
 elf = { version = "0.7", default-features = false, optional = true }
+kanal = { version = "0.1.0-pre8", default-features = false, optional = true }
```
flaub, line 65: Can we use a stable release?

flaub (approved):

PR #2415, opened 2024-10-11
Disable environment variables and arguments by default in guests

nategraf:
This PR disables syscalls `sys_getenv`, `sys_argc` and `sys_argv` by default, with the option to enable them via a feature flag on the `risc0-zkvm-platform` crate. In host programs, environment variables and arguments are generally considered trusted. In contrast, the guest does not trust the host, which leads to risk that code designed for running on the host may result in insecure behavior when run in the guest. Disabling environment variables and args by default mitigates this risk.


nategraf:
PR is currently marked as draft because I'm seeing some issues with `cargo risczero test`. I'm not sure if the issues are related, or is simply now being seen because I added a test.

nategraf:
I've decided to disable the new `cargo risczero test` test in https://github.com/risc0/risc0/pull/2415/commits/b23de205caa44aa5af467d361a85f39b82ecfd88. I confirmed that the changes do not introduce new issues. However, there is an issue with the standard test library provided by the toolchain in that it is trying to use thread despite a `cfg` to turn it off for the zkvm. This is something that was previously not an issue, so I am not sure what's going on. However this tool is mostly just used by me right now, so I'm going to revisit later and see if it either resolves itself or I'll figure it out with fresh eyes when I need it next.

https://github.com/risc0/rust/blob/b2060c0fa7b0f6f42a180e9c85e28bddc174c777/library/test/src/lib.rs#L594

flaub (approved):

PR #2423, opened 2024-10-16
docker builds: fix cpp toolchain support

SchmErik:
This change fixes errors that occur from building guests that use C code (such as c-kzg) using the docker reproducible build image. This change set adds the `CC` and `CFLAGS` to the appropriate location and uses a new docker image that includes the `build-essentials` package.

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -154,7 +154,7 @@ fn create_dockerfile(
     .join(" ");
 
     let build = DockerFile::new()
-        .from_alias("build", "risczero/risc0-guest-builder:r0.1.79.0-2")
+        .from_alias("build", "risczero/risc0-guest-builder:r0.1.79.0-3")
         .workdir("/src")
         .copy(".", ".")
         .env(manifest_env)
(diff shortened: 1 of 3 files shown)


SchmErik:
cc @hashcashier for vis

risc0/build/src/docker.rs
```diff
@@ -154,14 +154,19 @@ fn create_dockerfile(
...
         .copy(".", ".")
         .env(manifest_env)
         .env(rustflags_env)
         .env(&[("CARGO_TARGET_DIR", "target")])
         // Fetching separately allows docker to cache the downloads, assuming the Cargo.lock
         // doesn't change.
+        .env(&[(
+            "CC",
+            "/root/.local/share/cargo-risczero/cpp/bin/riscv32-unknown-elf-gcc",
+        )])
```
flaub, line 168: I would imagine we'd only want to configure the guest-side, this will also include the host-side.

SchmErik: Good point - we don't have any spaces in the gcc path so I pushed a change just now to use the `CC_riscv32im_risc0_zkvm_elf` flag to keep it confined to the guest

risc0/build/src/docker.rs
```diff
@@ -154,14 +154,19 @@ fn create_dockerfile(
...
         .env(manifest_env)
         .env(rustflags_env)
         .env(&[("CARGO_TARGET_DIR", "target")])
         // Fetching separately allows docker to cache the downloads, assuming the Cargo.lock
         // doesn't change.
+        .env(&[(
+            "CC_riscv32im_risc0_zkvm_elf",
+            "/root/.local/share/cargo-risczero/cpp/bin/riscv32-unknown-elf-gcc",
+        )])
+        .env(&[("CFLAGS_riscv32im_risc0_zkvm_elf", "-march=rv32im -nostdlib")])
```
flaub, line 169: Could we move these lines up before the comment? I think the comment should stay near the `.run(&fetch_cmd)`

SchmErik: thanks for catching this!

flaub (approved):

PR #2425, opened 2024-10-16
rzup install: run source on the profile

SchmErik:
Users are running into issues when installing rzup due to the profile not being sourced. This forces them to run the source command manually or start a new shell. Fix this by running `source $PROFILE` to add `rzup` to the path.


hmrtn (approved):
lgtm! Thanks!

flaub (approved):

PR #2427, opened 2024-10-17
Update remote-proving.md

nategraf:
First two bullets on how to use Bonsai are confusing to me. There was also some out of date language about "Bonsai as a zk coprocessor"


sashaaldrick (approved):
Just had a dev ask about bonsai proving and how to send jobs with default prover, so this is great timing.

PR #2429, opened 2024-10-18
replaced use of risc0_zkvm::Bytes with Vec<u8> to fix running tests in examples/waldo/core

sashaaldrick:
[related discord issue](https://discord.com/channels/953703904086994974/1294228671527387136)

Running `cargo test` in `examples/waldo/core` gives errors:

```rust
error[E0412]: cannot find type `Bytes` in crate `risc0_zkvm`
   --> waldo/core/src/image.rs:132:71
    |
132 |     ) -> impl Fn(risc0_zkvm::Bytes) -> risc0_zkvm::Result<risc0_zkvm::Bytes> + '_ {
    |                                                                       ^^^^^ not found in `risc0_zkvm
```

This is because [Bytes](https://docs.rs/risc0-zkvm/latest/risc0_zkvm/struct.Bytes.html) is only available for feature flags `client` or `prove`.

Instead of adding a feature flag, this PR replaces the use of `risc0_zkvm::Bytes` with `Vec<u8>`.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -5822,6 +5822,7 @@ dependencies = [
 name = "waldo-core"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "bincode",
  "divrem",
  "elsa",
--- a/examples/waldo/core/Cargo.toml
+++ b/examples/waldo/core/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = "1.0"
 bincode = "1.3"
 divrem = "1.0"
 elsa = "1.7"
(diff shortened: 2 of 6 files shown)


flaub:
This seems to break other things unfortnately.

sashaaldrick:
@flaub the other issues are now fixed

SchmErik:
I wonder if we should add the cargo test command as a part of ci...

flaub:
I actually have a pretty simple solution to this issue, when I get a chance I'll push here to have some review.

As far as regression testing, the issue is that sometimes running tests from a specific crate rather than from the workspace can cause a different set of dependency features to be enabled. I don't know how to get coverage on this without having CI run every single crate as a separate `cargo test` invocation, not very feasible IMHO.

flaub:
I prefer #2437 over this one

sashaaldrick:
closing this PR in favour of #2437

PR #2437, opened 2024-10-22
Fix missing host-side dependency for waldo-core

flaub:


flaub:
@sashaaldrick this is an alternative solution to #2429

sashaaldrick (approved):
this is much simpler, thanks @flaub

PR #2444, opened 2024-10-23
zkp changes to support extended field element use by zirgen codegen

shkoo:
* add Shift as a poly ext step opcode to return {0, 1, 0...}
* Implement a more complete set of operators for field elements
* Add debugging capabilities to PolyExtStep


risc0/zkp/src/adapter.rs
```diff
@@ -210,45 +212,74 @@ impl PolyExtStep {
...
             PolyExtStep::Add(x1, x2) => {
-                fp_vars.push(fp_vars[*x1] + fp_vars[*x2]);
+                let val = fp_vars[*x1] + fp_vars[*x2];
+                tracing::trace!("[{}] {self:?} -> {val:?}", fp_vars.len());
+                fp_vars.push(val);
             }
             PolyExtStep::Sub(x1, x2) => {
-                fp_vars.push(fp_vars[*x1] - fp_vars[*x2]);
+                let val = fp_vars[*x1] - fp_vars[*x2];
+                tracing::trace!("[{}] {self:?} -> {val:?}", fp_vars.len());
```
flaub, line 235: If these aren't needed anymore, let's comment them out.

flaub:
Do we need corresponding changes to cuda/metal?

risc0/zkp/src/adapter.rs
shkoo: I put them behind a const so they're easy to re-enable.  Does that work?

shkoo:
> Do we need corresponding changes to cuda/metal?

I don't think so; best I can tell, poly ext only gets run on CPU

flaub (approved):

PR #2393, opened 2024-10-01
Deterministic builds page

nuke-web3:
Fixes https://linear.app/risczero/issue/DR-349/document-reproducible-builds-and-alternatives-to-cargo-risczero-build
Closes #1084 

Might be blocked by https://github.com/risc0/risc0-ethereum/issues/141

PR #2459, opened 2024-10-29
Fix det builds docs

nuke-web3:
targets #2393


SchmErik:
@nuke-web3 it looks like you're based on the wrong branch. I think you need to do something like `git rebase risc0:n/det-builds-docs` or something like `git rebase --onto HEAD~2 risc0:n/det-builds-docs` or something like that. basically, take your last two commits and apply it to the one in the risc0 branch...

nuke-web3:
I merged main at [8bb2593](https://github.com/risc0/risc0/pull/2459/commits/8bb2593faf826e94be205bcfda805f27a1712564) and tried to push to the original branch (but cannot) I think it's safe to merge this into the original PR and it should figure out the correct diff. Otherwise I would hand it over to the maintainers to use the patch from the last two commits & decide how to integrate it :pray:

PR #2460, opened 2024-10-30
Split syscalls into separate modules

flaub:


SchmErik (approved):
Thanks! Looks good to me!

PR #2464, opened 2024-10-30
rzup: replace toolchain `reqwest` download with `downloader`

### LINKED ISSUE #2446: "rzup install" needlessly consumes memory
I encountered this problem when trying to install Risc0 on a VPS with only a few GB of memory. `rzup install` quickly consumes the machine's entire available memory.

rzuip

This code is the culprit:

```rust
        let mut file = fs::File::create(&temp_file_path)?;
        let content = response.bytes().await?;

        verbose_msg!(format!(
            "Writing contents to file {}",
            temp_file_path.display()
        ));

        file.write_all(&content)?;
```

https://github.com/risc0/risc0/blob/2ba504fddd84376235d335ec4db6b2353d967fc9/rzup/src/toolchain.rs#L191

When downloading large tarballs, the download should be streamed to a file on disk, not buffered into memory. 

If you wanted to be even more efficient, you could stream the download through a tarball decompression stream handler. But i think the bare minimum is at least buffering to disk instead of memory.

hmrtn:
fixes #2446

This reduces memory usage for memory constrained devices, from >800MB to <70MB by using the `downloader` crate. Attached are two `top` logs over a 60s interval with a 1s sample rate. 

New: [rzup_downloader_memory_log.txt](https://github.com/user-attachments/files/17579890/rzup_downloader_memory_log.txt)
Current: [rzup_memory_log.txt](https://github.com/user-attachments/files/17579891/rzup_memory_log.txt)

We should also make this change for all `rzup` download handlers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2508,6 +2508,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper-tls"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70206fc6890eaca9fde8a0bf71caa2ddfc9fe045ac9e5c70df101a7dbde866e0"
+dependencies = [
+ "bytes",
+ "http-body-util",
+ "hyper 1.3.1",
+ "hyper-util",
+ "native-tls",
+ "tokio",
+ "tokio-native-tls",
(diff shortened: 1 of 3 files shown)


flaub (approved):
Nice!

PR #2481, opened 2024-11-05
Add RSA syscall

tzerrell:
Extends our RSA acceleration support to enable accelerated computation of the RSA exponentiation (instead of just accelerated verification of the result of an RSA exponentiation). This is accomplished by adding an RSA syscall which allows the guest to request that the host provide the result of the RSA operation, which it can then verify via the recursion circuit bigint RSA accelerator.

This PR also adds a convenient API for accessing this RSA acceleration: `modpow_65537`. This takes in two `BigUint`s, a `base` and a `modulus`, and computes `base^65537 % modulus` (where `^` is exponentiation) via this syscall + bigint proof mechanism. For convenience in patching into other crates, two bigint libraries are available for this API: By default, it will use `num_bigint`, but if the `bigint-dig-shim` feature of `risc0-circuit-bigint` is enabled it will instead use `num_bigint_dig`.


risc0/circuit/bigint/methods/guest-dig/src/bin/rsa_dig.rs
```diff
@@ -0,0 +1,28 @@
...
+use risc0_circuit_bigint::rsa;
+use risc0_zkvm::guest::env;
+
+fn main() {
+    // Computes and proves the result of modpow with exponent of 65537
+    let input: Vec<[BigUint; 2]> = env::read();
+    let result: Vec<BigUint> = input
+        .into_iter()
+        // .map(|[base, modulus]| rsa::modpow_65537(&base, &modulus).unwrap())
+        .map(|[base, modulus]| rsa::modpow_65537(&base, &modulus).expect("here?"))
```
flaub, line 25: Are these both still needed?

risc0/circuit/bigint/methods/guest-dig/src/bin/rsa_dig.rs
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use num_bigint::BigUint;
+use risc0_circuit_bigint::rsa;
+use risc0_zkvm::guest::env;
+
+fn main() {
+    // Computes and proves the result of modpow with exponent of 65537
+    let input: Vec<[BigUint; 2]> = env::read();
```
flaub, line 21: Could this be `Vec<(BigUint, BigUint)>`?

risc0/circuit/bigint/src/rsa.rs
```diff
@@ -16,14 +16,138 @@
...
+    let mut base = base.to_bytes_le();
+    if base.len() % 4 != 0 {
+        base.resize(base.len() + (4 - (base.len() % 4)), 0);
+    }
+    let mut modulus = modulus.to_bytes_le();
+    if modulus.len() % 4 != 0 {
+        modulus.resize(modulus.len() + (4 - (modulus.len() % 4)), 0);
+    }
+    // Convert inputs to Vecs of u32s
+    let mut base_vec = Vec::<u32>::new();
```
flaub, line 84:
```suggestion
    let mut base_vec = Vec::new();
```

or 
```
let mut base_vec: Vec<u32> = vec![];
```

risc0/zkvm/src/host/server/exec/syscall/rsa.rs
```diff
@@ -0,0 +1,53 @@
...
+        tracing::debug!("SYS_RSA");
+        // Load inputs into BigUint
+        let base_ptr = ByteAddr(ctx.load_register(REG_A3));
+        let modulus_ptr = ByteAddr(ctx.load_register(REG_A4));
+        let base = ctx.load_region(base_ptr, rsa::WIDTH_BYTES.try_into()?)?;
+        let modulus = ctx.load_region(modulus_ptr, rsa::WIDTH_BYTES.try_into()?)?;
+        let base = BigUint::from_bytes_le(&base);
+        let modulus = BigUint::from_bytes_le(&modulus);
+        // Compute result
+        let result = base.modpow(&BigUint::from(rsa::RSA_EXPONENT), &modulus);
```
flaub, line 43:
```suggestion
        let result = base.modpow(&rsa::RSA_EXPONENT.into(), &modulus);
```

risc0/zkvm/src/host/server/exec/syscall/rsa.rs
```diff
@@ -0,0 +1,53 @@
...
+        &mut self,
+        _syscall: &str,
+        ctx: &mut dyn SyscallContext,
+        to_guest: &mut [u32],
+    ) -> Result<(u32, u32)> {
+        tracing::debug!("SYS_RSA");
+        // Load inputs into BigUint
+        let base_ptr = ByteAddr(ctx.load_register(REG_A3));
+        let modulus_ptr = ByteAddr(ctx.load_register(REG_A4));
+        let base = ctx.load_region(base_ptr, rsa::WIDTH_BYTES.try_into()?)?;
```
flaub, line 38:
```suggestion
        let base = ctx.load_region(base_ptr, rsa::WIDTH_BYTES as u32)?;
```

risc0/circuit/bigint/methods/guest-dig/src/bin/rsa_dig.rs
tzerrell: whoops, no

PR #2482, opened 2024-11-06
Adjust JWT example to use RSA acceleration

tzerrell:
Makes use of the RSA acceleration from https://github.com/risc0/risc0/pull/2481 and https://github.com/risc0/RustCrypto-RSA/pull/1 to accelerate the JWT example.

Note that I had to adjust the risc0 crate dependencies of this example to point at github rather than including them via relative path; this is because the [RSA patch](https://github.com/risc0/RustCrypto-RSA/pull/1) this uses depends on a pinned commit of risc0, and if I don't use a github dependency here, I end up with errors like
```
error: the `#[global_allocator]` in risc0_zkvm_platform conflicts with global allocator in: risc0_zkvm_platform
```


.github/workflows/main.yml
```diff
@@ -326,6 +326,9 @@ jobs:
       - name: test in dev mode
         run: RISC0_DEV_MODE=1 cargo test --locked -F $FEATURE --workspace --exclude prover-example
         working-directory: examples
+      - name: test JWT validator separately
+        run: RISC0_DEV_MODE=1 cargo test --locked -F $FEATURE
+        working-directory: examples/jwt-validator
       - name: test prover example
```
flaub, line 332: we'll need to explicitly exclude this from the line above if we're going to do it separately, but why do we need to do this?

examples/Cargo.toml
```diff
@@ -16,8 +16,6 @@ members = [
   "hello-world",
   "json",
   "json/core",
-  "jwt-validator",
-  "jwt-validator/core",
```
flaub, line 20: Why do we need this?

examples/jwt-validator/Cargo.toml
```diff
@@ -3,12 +3,17 @@ name = "jwt-validator"
 version = "0.1.0"
 edition = "2021"
 
+[workspace]
+
 [dependencies]
 jwt-core = { path = "core" }
 jwt-methods = { path = "methods" }
-risc0-zkvm = { path = "../../risc0/zkvm" }
+risc0-zkvm = { git = "https://github.com/risc0/risc0" }
```
flaub, line 11: why do we do this?

tzerrell:
Will not work, replaced by #2488

PR #2486, opened 2024-11-08
Update the document with best practice for vscode

nishuzumi:
By default, vscode cannot add type recognition to a guest when creating a risczero application. This documentation supplements this and allows guest to be recognized as well.


nishuzumi:
@pdg744 @sashaaldrick

PR #2487, opened 2024-11-09
Redo RSA Acceleration

tzerrell:
We discovered that the approach used in #2481 was not playing well with linking. This PR redoes the RSA acceleration wrappers to expect outside crates to interact with an `extern "C"` API utilizing pointers to `[u32; WIDTH_WORDS]` rather than pulling in Rust wrapping functions, moving the work of translating bigints into u32 words to the patched crates.

Because this moves wrapping code to patched crates, the `bigint-dig-shim` feature was dropped as unused and unnecessary.

See also related PRs #2488 and https://github.com/risc0/RustCrypto-RSA/pull/2 which will follow this one


risc0/circuit/bigint/methods/guest/src/bin/rsa.rs
```diff
@@ -12,16 +12,24 @@
...
-        .collect();
-    env::commit(&result);
+    let input: Vec<u32> = env::read();
+    let base: [u32; rsa::WIDTH_WORDS] = input.try_into().expect("Inputs must be 96 words");
+    let input: Vec<u32> = env::read();
+    let modulus: [u32; rsa::WIDTH_WORDS] = input.try_into().expect("Inputs must be 96 words");
+    const fn zero() -> u32 {
+        0
+    }
+    let mut result = [zero(); rsa::WIDTH_WORDS];
```
flaub, line 27:
```suggestion
    let mut result = [0u32; rsa::WIDTH_WORDS];
```

PR #2488, opened 2024-11-09
Update JWT Example

tzerrell:
This updates the JWT example to use #2487. Note that there are some sharp edges which may need further thought, and that the while #2487 and https://github.com/risc0/RustCrypto-RSA/pull/2 are in review this isn't fully ready (i.e. can't patch correctly).

Replaces #2482


examples/jwt-validator/methods/guest/src/main.rs
```diff
@@ -45,3 +46,13 @@ fn main() {
...
+
+// Dead code for the linker
+// The linker will "helpfully" optimize away our weakly linked function if we don't directly use it somewhere. This is a fake function to prevent such an optimization. Do not call this function, it is both useless and unsafe.
+fn _linker_helper() {
+  assert!(false, "Do not call this function");
+  // Null pointers break safety but this is dead code so it's ok
+  unsafe {
+    modpow_65537(std::ptr::null_mut(), std::ptr::null(), std::ptr::null());
+  }
+}
```
flaub, line 58: Is this still needed?

tzerrell: It's working for me locally without it, but I think that's because locally I have the RSA crate patched as
```
[patch.crates-io]
rsa = { path = "/Users/tzerrell/code/RustCrypto/RSA" }
```

I have a couple of other branches that try to use `whole-archive` to avoid this, but I'm not happy with either of them:
* The code in https://github.com/risc0/risc0/tree/tzerrell/jwt-whole-archive works but gives a warning that it's being deprecated:
```
validator: warning: attribute should be applied to an `extern` block with non-Rust ABI
validator:   --> src/main.rs:17:1
validator:    |
validator: 17 | #[link(name = "risc0_circuit_bigint", kind = "static", modifiers = "+whole-archive")]
validator:    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
validator: 18 | use risc0_circuit_bigint::rsa::modpow_65537;
validator:    | -------------------------------------------- not an `extern` block
validator:    |
validator:    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
validator:    = note: `#[warn(unused_attributes)]` on by default
```
* The code in https://github.com/risc0/risc0/tree/tzerrell/whole-archive-extern tries to address the comments from that warning, but doesn't link:
```
validator:   = note: rust-lld: error: unable to find library -lrisc0_circuit_bigint
```

tzerrell:
Marking as draft b/c https://github.com/risc0/risc0/pull/2487 and https://github.com/risc0/RustCrypto-RSA/pull/2 are needed first

examples/jwt-validator/methods/guest/src/main.rs
tzerrell: I think you're right that this can just be dropped, I will do that, and we can come back if we hit problems.

tzerrell:
Status: investigating why the ZKR doesn't seem to be registered if I run the test from the `examples` directory, but it does seem to be registered if I run the test from the `examples/jwt-validator` directory.

flaub:
Superseded by #2505

PR #2489, opened 2024-11-09
Add the bigint2 accelerator circuit

flaub:


risc0/circuit/rv32im-sys/cxx/extern.cpp
```diff
@@ -334,4 +347,39 @@ extern_plonkReadAccum_bytes(void* ctx, size_t cycle, const char* extra, std::arr
...
+
+std::array<Fp, 16>
+extern_syscallBigInt2Witness(void* ctx, size_t cycle, const char* extra, std::array<Fp, 5> args) {
+  PreflightTrace* trace = static_cast<MachineContext*>(ctx)->trace;
+  size_t extraIdx = trace->cycles[cycle].extraIdx;
+  uint32_t a0 = trace->extras[extraIdx + 0];
+  uint32_t a1 = trace->extras[extraIdx + 1];
+  uint32_t a2 = trace->extras[extraIdx + 2];
+  uint32_t a3 = trace->extras[extraIdx + 3];
+  return {
```
jbruestle, line 365: Nit: you do this as a look for the cuda version, why not here?

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -541,15 +540,23 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
     fn ecall_bigint2(&mut self) -> Result<bool> {
         let blob_ptr = self.load_guest_addr_from_register(REG_A0)?.waddr();
         let nondet_program_ptr = self.load_guest_addr_from_register(REG_T1)?;
-        let nondet_program_size = self.load_u32_from_guest((blob_ptr + 0u32).baddr())?;
+        let verify_program_ptr = self.load_guest_addr_from_register(REG_T2)?;
```
jbruestle, line 543: Maybe we should schedule a discussion around registers/calling convention for bigint2 sometime before we get to a release, just to make sure we are happy with things as it sort of becomes part of the ABI we might need to support in the future (although maybe it's not relevant since we can make a 'compatibility' kernel for v2, and change whatever we want).

risc0/circuit/rv32im/src/prove/emu/preflight/bigint2.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub term: BytePolynomial,
+    pub total: BytePolynomial,
+    pub total_carry: BytePolynomial,
+}
+
+impl Instruction {
+    // instruction encoding:
+    // 3  2   2  2    1               0
+    // 1  8   4  1    6               0
+    // mmmmppppcccaaaaaoooooooooooooooo
```
jbruestle, line 68: TODO for me: write a .md talking about this format.

jbruestle (approved):
Amazing!

risc0/circuit/rv32im-sys/cxx/extern.cpp
flaub: Ah, I guess I could do a loop here too.

PR #2503, opened 2024-11-13
Implement plumbing for keccak acceleration

SchmErik:
This PR introduces a batching mechanism for keccak circuit proofs and supersedes #2380


risc0/zkvm/src/guest/env/mod.rs
```diff
@@ -78,26 +79,27 @@ use alloc::{
     vec,
 };
 
+use crate::{
+    sha::{
+        rust_crypto::{Digest as _, Sha256},
+        Digest, Digestible,
+    },
+    Assumptions, MaybePruned, Output,
+};
```
flaub, line 88: can you move this down to near the `self` location? Basically we try to group uses by:

* std
* external
* internal

flaub (approved):

PR #2504, opened 2024-11-13
Add risc0-bigint2 guest-side wrapper crate

flaub:
Related to ZKVM-630


risc0/bigint2/src/ec/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+#[cfg(test)]
+mod tests;
+
+use crate::ffi::sys_bigint2_2;
+
+const DOUBLE_BLOB: &[u8] = include_bytes!("double.blob");
+
+pub fn double(point: &[u32], result: &mut [u32]) {
```
tzerrell, line 22: Not for this PR, but as followup: This is a specific EC double, and we're going to want more than one as we patch multiple crates. So we'll need to genericize the naming and add the alternates (or genericize this one with more parameters).

The EC Double function at https://github.com/risc0/zirgen/blob/040fcd133094974a9a218c13f717c98be0fbada9/zirgen/circuit/bigint/bigint2c.cpp#L420 that this is based on hard-codes the curve parameters to be those of `secp256k1` (given in affine coordinates and Weierstrass form). Note that the underlying function call https://github.com/risc0/zirgen/blob/040fcd133094974a9a218c13f717c98be0fbada9/zirgen/circuit/bigint/elliptic_curve.cpp#L382 makes the curve parameters generic (always affine + Weierstrass though), allowing us to run this on other curves as long as we generate them.

tzerrell (approved):
🎉

PR #2507, opened 2024-11-14
Drop risc0-circuit-bigint and add executor metrics

flaub:


andrewcoder666:
I have a question about [the change of workers number in the example](https://github.com/risc0/risc0/pull/2507/files#diff-829fd505f1def9176a9013748f4935fcba653d7790e5646aef42830e064de7b0R59).

Is it necessary to turn it down to only 1 thread?

flaub:
We only did that to get around an issue on mac where it's easy to overwhelm the GPU. It should be fine on a larger machine with a robust GPU and lots of VRAM.

PR #2508, opened 2024-11-14
Add union recursion program

nategraf:
Updates to the ZKR archive and control root to include the new `union` prgram implemented in https://github.com/risc0/zirgen/pull/60

PR includes recursion program updates from the Zirgen repo:
* https://github.com/risc0/zirgen/pull/58 introduced changes to the codegen effecting all recursion programs.
* https://github.com/risc0/zirgen/pull/60 adds the `union` program

Note: We expect the `risc0-ethereum` CI job to fail here due to mismatching control root and identity bn254 control ID. I will be following up with a PR to `risc0-ethereum` to resolve this.

To-do:
- [x] Run bootrap from zirgen once https://github.com/risc0/zirgen/pull/60 lands.
- [x] Update pinned digests for verifier parameters in all of our tests.


nategraf:
PR is currently marked as draft because it seems all the recursion control IDs changed. This may be an issue with how I ran `bootstrap` in the Zirgen repo.

flaub (approved):

PR #2510, opened 2024-11-14
Constrain buffer length for RSA bigint values

austinabell:
Digging more into how this is called, but there was an assumption that the bigint buffer would always have the max word length. This was my initial thought for a quick fix, but will now go back to double check and see if a more sound way to fix.

Example error:
```
2024-11-14T21:01:55.553897Z  INFO risc0_circuit_rv32im::prove::emu::exec: split: 962517 + 82464 + 3594 = 1048575, padding: 1, pending: PendingState { pc: 0x000817aa(0x00205ea8), insn: 4160954083, cycles: 1, syscall: None, output_digest: None, exit_code: None, events: {} }
2024-11-14T21:01:55.596132Z  INFO risc0_circuit_rv32im::prove::emu::exec: split: 986585 + 58396 + 3594 = 1048575, padding: 1, pending: PendingState { pc: 0x000803d0(0x00200f40), insn: 21048323, cycles: 1, syscall: None, output_digest: None, exit_code: None, events: {} }
2024-11-14T21:01:55.638487Z  INFO risc0_circuit_rv32im::prove::emu::exec: split: 984397 + 60584 + 3594 = 1048575, padding: 1, pending: PendingState { pc: 0x000817a4(0x00205e90), insn: 48592483, cycles: 1, syscall: None, output_digest: None, exit_code: None, events: {} }
2024-11-14T21:01:55.675177Z  INFO risc0_circuit_rv32im::prove::emu::exec: split: 909212 + 134976 + 3594 = 1047782, padding: 794, pending: PendingState { pc: 0x00089839(0x002260e4), insn: 4244701459, cycles: 1, syscall: None, output_digest: None, exit_code: None, events: {} }
2024-11-14T21:01:55.691562Z  INFO risc0_circuit_rv32im::prove::emu::exec: split: 764155 + 280478 + 3594 = 1048227, padding: 349, pending: PendingState { pc: 0x00081178(0x002045e0), insn: 17146915, cycles: 1, syscall: None, output_digest: None, exit_code: None, events: {} }
thread 'main' panicked at /Users/austinabell/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/8fc8437/risc0/bigint2/src/rsa/mod.rs:65:5:
assertion `left == right` failed
  left: 32
 right: 96
Error: Failed to generate proof: Guest panicked: assertion `left == right` failed
  left: 32
 right: 96
```

tested by patching that this change fixes for the example I tested.

- [x] Add test for diff buffer sizes of `BigUint` (didn't check the larger than buffer case and what happens)
- [x] Experiment with refactoring this to be an array and check cycle counts (preferred to be more strict)


flaub:
Do you have a test case where this is causing the assertions to fire?

tzerrell (commented):
Thank you! I thought I already had logic for that ... it must have been in just an earlier draft. This is the right solution, I have one caveat inline that I think we should address.

risc0/bigint2/src/rsa/mod.rs
```diff
@@ -45,12 +45,16 @@ fn to_u32_digits(input: &BigUint) -> Vec<u32> {
         digits.push(word);
     }
 
+    digits.resize(RSA_3072_WIDTH_WORDS, 0);
```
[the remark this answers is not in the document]
austinabell: you're completely right, and it should be a panic for larger values. Will have a test case for the larger bits and make sure the final impl handles it correctly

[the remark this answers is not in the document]
austinabell: yes, that is correct that the `sys_bigint2_3` allows smaller buffers, but I wonder if that is intentional to not be strict?

That is also the most efficient way in terms of cycle count, so it would be preferred if there is no security concern (I can't think of one, but haven't dug into internals of the syscall yet) [`4c0e739` (#2510)](https://github.com/risc0/risc0/pull/2510/commits/4c0e739cb6b851e160d4ba13ef2571dc25277f0a)

risc0/bigint2/src/rsa/mod.rs
```diff
@@ -62,8 +62,8 @@ pub fn modpow_65537(base: &BigUint, modulus: &BigUint) -> BigUint {
 }
 
 pub fn raw_modpow_65537(base: &[u32], modulus: &[u32], result: &mut [u32]) {
-    assert_eq!(base.len(), RSA_3072_WIDTH_WORDS);
-    assert_eq!(modulus.len(), RSA_3072_WIDTH_WORDS);
+    assert!(base.len() <= RSA_3072_WIDTH_WORDS);
+    assert!(modulus.len() <= RSA_3072_WIDTH_WORDS);
```
tzerrell, line 66: I don't think it's safe to call this function with lengths < RSA_3072_WIDTH_WORDS, and I'm surprised `modpow_65537_small_buffers` doesn't break. What this is telling the accelerator to do is read 3072 bits starting from the address `base.as_ptr()` and read 3072 bits starting from the address `modulus.as_ptr()`, do modpow_65537 on them, and write the resulting 3072 bits to `result.as_mut_ptr()`. I really don't see how that could be safe if any of `base`, `modulus`, or `result` aren't the full 3072 bits. Maybe the test is getting "lucky" and has a bunch of zeros after each of `base` and `modulus`?

Oh, maybe we're getting the right answer because the prime `7` is so small, so it's relatively easy to get an answer by shear luck. How about trying out a roughly 32 bit prime instead of 7? You could use BabyBear, for example. Or maybe we need to have more code in the guest, so that when we read too far we run into something other than zeros.

tzerrell (changes requested):
The exact equality in the `raw_modpow_65537` asserts is important, we should keep it and do resizing (but first fix the `modpow_65537_small_buffers` test so it notices the problem!)

risc0/bigint2/src/rsa/mod.rs
tzerrell: (I don't think you need to know this to make these changes, but if you want context on how much will be read: Look at https://github.com/risc0/zirgen/blob/main/zirgen/circuit/bigint/bigint2c.cpp#L405 and the surrounding lines. `11`, `12`, and `13`, used in the `LoadOp`s and `StoreOp` indicate reading/writing to the 1st, 2nd, and 3rd buffers provided in `sys_bigint2_3`, and the `bits = 3072` used with the loads controls how much will be read.)

flaub: Agreed that these asserts should not be changed, they need to check that the full 3072 bits are available to be read.

risc0/bigint2/src/rsa/mod.rs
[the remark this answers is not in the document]
tzerrell: Ah, sorry, I didn't mean adjust the existing assert to be weaker, I meant adding a new assert before resizing so that you would only ever resize up

flaub:
I think we can avoid a resize (which could lead to misalignments, and is expensive), by replacing:
```
let mut digits = Vec::with_capacity(RSA_3072_WIDTH_WORDS);
```

with:
```
let mut digist = vec![0; RSA_3072_WIDTH_WORDS];
```

risc0/bigint2/src/rsa/mod.rs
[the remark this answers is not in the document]
flaub: The interpreter on the host side will read values that it's not intended to, which is a bug but not a soundness issue. However the verifier circuitry will read incorrect values which is will be a soundness problem.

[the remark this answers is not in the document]
flaub: We also need to avoid a resize entirely.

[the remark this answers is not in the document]
austinabell: gotcha, sorry I misread last night, and I assumed the host function was implemented passing a buffer length so that there was never a read past the buffer. Will fix now

risc0/bigint2/src/rsa/mod.rs
austinabell: Trying to debug the fact that the tests pass, I got a result that was unexpected. Digging in more to understand, as I might just be misunderstanding something.

Printing the memory that's used for the syscall without fixing the buffer sizes:

```
    let base = to_u32_digits(base);
    let modulus = to_u32_digits(modulus);
    let _tmp = std::hint::black_box(Box::new([0u32; RSA_3072_WIDTH_WORDS]));
    unsafe {
        let modulus_ptr = modulus.as_ptr();
        let base_ptr = base.as_ptr();

        println!("Memory after modulus:");
        for i in 0..RSA_3072_WIDTH_WORDS {
            let addr = modulus_ptr.add(i);
            println!("0x{:x}: {:08x}", addr as usize, *addr);
        }

        println!("\nMemory after base:");
        for i in 0..RSA_3072_WIDTH_WORDS {
            let addr = base_ptr.add(i);
            println!("0x{:x}: {:08x}", addr as usize, *addr);
        }
    }

```

Yields:

```
Memory after modulus:
0x230948: 7fffffff
0x23094c: 00000000
0x230950: 00000000
0x230954: 00000000
0x230958: 00000000
0x23095c: 00000000
0x230960: 00000000
0x230964: 00000000
...

Memory after base:
0x230938: 00000001
0x23093c: 00000000
0x230940: 00000000
0x230944: 00000000
0x230948: 7fffffff
0x23094c: 00000000
0x230950: 00000000
...
```

And the test still passes here for some reason. (modulus being `2^31-1`)

In either case, these tests fail if there is data allocated after these are. This test wasn't causing issues previously because there was nothing allocated after these, and memory is zeroed by default. I don't have an explanation for why the above case passes, though

austinabell:
> I think we can avoid a resize (which could lead to misalignments, and is expensive), by replacing:
> 
> ```
> let mut digits = Vec::with_capacity(RSA_3072_WIDTH_WORDS);
> ```
> 
> with:
> 
> ```
> let mut digist = vec![0; RSA_3072_WIDTH_WORDS];
> ```

Do we care about the micro-optimization to avoid zeroing the bytes at the cost of some minimal unsafe code? Not sure this calls `alloc_zeroed` which for the bump allocator will avoid zeroing given memory is already zeroed.

austinabell:
So proposal to make these low level APIs array based instead of slice and asserting at runtime the length [`53432e7` (#2510)](https://github.com/risc0/risc0/pull/2510/commits/53432e71cf812097412c0dc5e58e51a18e3a3b26)

It's my assumption that this crate is not currently released or used, so the API breaking change is acceptable, is that correct? Would you prefer it to be as it was, or do you agree this is a good direction to go? Seems better to prevent misuse

tzerrell:
> So proposal to make these low level APIs array based instead of slice and asserting at runtime the length [`53432e7` (#2510)](https://github.com/risc0/risc0/pull/2510/commits/53432e71cf812097412c0dc5e58e51a18e3a3b26)
> 
> It's my assumption that this crate is not currently released or used, so the API breaking change is acceptable, is that correct? Would you prefer it to be as it was, or do you agree this is a good direction to go? Seems better to prevent misuse

Yeah, please refactor the APIs to whatever would be ideal (and the array approach makes a lot of sense to me given that these really are fixed size things) -- this code was only initially written a few days ago, and while we've pointed a few external people at it with "check out what we're building" we have not reached a stage where we're committing to stable APIs.

> Do we care about the micro-optimization to avoid zeroing the bytes at the cost of some minimal unsafe code? Afaik this doesn't call alloc_zeroed which for the bump allocator will avoid zeroing given memory is already zeroed.

Hmm. The perf benefits seem somewhat marginal, but the unsafe code doesn't look that risky either. That said, I'm not super familiar with unsafe Rust memory manipulation functions and don't immediately know off the top of my head if there are any problems with e.g. alignment for what you have written -- so my slight preference would be to not add this micro-optimization just to not have to worry about it. But happy to be overruled on that, esp. by someone with more unsafe Rust experience than me.

tzerrell:
Although speaking of APIs ... should `raw_modpow_65537` even be a `pub` function? Seems like we could leave it internal and have our only API surface be `modpow_65537` instead. The disadvantage of this suggestion is it forces users to use one of the supported big integer libraries (i.e. `num-bigint` or `num-bigint-dig`) rather than being able to hand over plain u32 arrays; not sure how much that matters.

Also, this all is making me wonder if we should mark `raw_modpow_65537` as `unsafe` ... even with the new array-based API (which is safer), it still has a contract with the caller that the `result` memory must be zero'd when passed in -- that seems like a memory assumption that might benefit from being flagged with `unsafe`.

tzerrell (approved):
We still have some design decisions we're working through (see my previous comments), but also I think this is now in a state where we could reasonably either make those changes in this PR _or_ land and iterate in future PRs, so I'm approving.

risc0/bigint2/src/rsa/mod.rs
tzerrell: It's exponentiation to the 65537 power, not multiplication, and I asked python (which has out-of-the-box modpow):
```
>>> pow(0x7fffffff00000000000000000000000000000001, 65537, 2147483647)
1
```
Finite fields have a lot of structure encoded in their exponentiation (which is how RSA works at all), so you will hit 1 as a result more often than you would just by rolling dice. If instead of the Mersenne prime 2147483647 you used the baby bear prime 2013265921 then python suggests it should make the test fail:
```
>>> pow(0x7fffffff00000000000000000000000000000001, 65537, 2013265921)
606819399
```

austinabell:
> Although speaking of APIs ... should `raw_modpow_65537` even be a `pub` function? Seems like we could leave it internal and have our only API surface be `modpow_65537` instead. The disadvantage of this suggestion is it forces users to use one of the supported big integer libraries (i.e. `num-bigint` or `num-bigint-dig`) rather than being able to hand over plain u32 arrays; not sure how much that matters.

One other reason you might want it pub is for an optimization where you do multiple muls with the same value, you don't have to do the conversion multiple times. Although for that, we would probably want the BigUint -> array functions available.

I wouldn't personally argue against making the `raw` method private, since it's such a small wrapper around the FFI function, which is already exposed.

> Also, this all is making me wonder if we should mark `raw_modpow_65537` as `unsafe` ... even with the new array-based API (which is safer), it still has a contract with the caller that the `result` memory must be zero'd when passed in -- that seems like a memory assumption that might benefit from being flagged with `unsafe`.

This is unsafe because the syscall only writes to the words it needs, or because the result buffer values are not just written to?

tzerrell:
> This is unsafe because the syscall only writes to the words it needs, or because the result buffer values are not just written to?

I think I was wrong here, actually -- I meant that "the syscall only writes to the words it needs", but fundamentally, it's filling these words from a StoreOp, which will know about the maximum size the result could hold, not the actual size, so I think that means it actually does write all the zeros. I withdraw my "mark `unsafe`" suggestion -- I think by making all the inputs arrays you have made this function safe.

risc0/bigint2/src/rsa/mod.rs
tzerrell: Discussed in DMs and we realized that my approach doesn't work because the modulus is the thing stomping the memory, so the stomped base value changes. Fundamentally, we're exponentiating (1 + 2^n * modulus)^65537 % modulus which will always be 1. So, what we'll need to do is create a test where the modulus is large enough that the stomped memory overflows the extra memory left in base, thus breaking that formula and causing the test to fail.

austinabell: Resoled in DMs, `pow(0x7fffffff00000000000000000000000000000001, 65537, 2013265921)` turned out as not 1 result because the prime is different in the mod and mangled base, `pow(0x7800000100000000000000000000000000000001, 65537, 2013265921) == 1`

@tzerrell had a great suggestion about using a full length mod value such that only part of the data when read past the buffer is used. Otherwise, can't make `BigUint(1)` large enough to split the value given it's always truncated to 1 word in this implementation. 

Modified test case [`86d0780` (#2510)](https://github.com/risc0/risc0/pull/2510/commits/86d0780fcedf9f0f6a5e98a6adc4ee8cbec62cc4) such that it fails before [`53432e7` (#2510)](https://github.com/risc0/risc0/pull/2510/commits/53432e71cf812097412c0dc5e58e51a18e3a3b26) and also added a test for the inverse pattern, just in case it's ever changed that the mod is allocated before the base

SchmErik (approved):

PR #2521, opened 2024-11-19
feat(reports): graphs in mhz/khz

nahoc:
image

solves https://linear.app/risczero/issue/ZKVM-542/adjust-y-axis-on-reports-page-to-show-results-in-khzmhz


flaub (changes requested):
I'm seeing some issues on `prove/poseidon2` here: https://reports-and-benchmarks-git-nahoc-hz-risczero.vercel.app/benchmarks/Linux-nvidia_rtx_3090_ti

flaub:
Could we display the actual number (raw or prefixed) in the tooltip text? Generally I'd like to be able to see the exact value for a particular point.

flaub (approved):

PR #2525, opened 2024-11-20
rzup: add support for cross-device toolchain copying

hmrtn:
this resolves the `Invalid cross-device link (os error 18)` issue when extracting and moving toolchains by adding support for cross-device copying (rather than move) in `move_toolchain`. See [slack](https://risc0.slack.com/archives/C03J8034W10/p1732098080189069) for more context.

### DIFF
--- a/rzup/src/toolchain.rs
+++ b/rzup/src/toolchain.rs
@@ -260,7 +260,7 @@ impl Toolchain {
                 ));
 
                 // Move subdir contents to toolchain dir
-                self.move_toolchain(&subdir, &toolchain_dir)?;
+                Toolchain::move_toolchain(&subdir, &toolchain_dir)?;
             }
         }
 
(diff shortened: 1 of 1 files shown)


mothran (approved):
This appears to me to solve the problem of cross-device issues, but I will defer to @flaub for final approval.

SchmErik (approved):

flaub:
It'd be better if this was an atomic operation. Can we not move the root directory across devices in a single operation?

mothran:
> It'd be better if this was an atomic operation. Can we not move the root directory across devices in a single operation?

From `man 2 rename`
```
       EXDEV  oldpath  and  newpath  are not on the same mounted filesystem.  (Linux permits a
              filesystem to be mounted at multiple points, but rename() does not  work  across
              different mount points, even if the same filesystem is mounted on both.)
```

So a good number of linux distro's have /tmp mounted `tmpfs` which means the rename's will fail and can't be reliably used.

saurik:
When will this be pushed to production / an installable release? I just installed rzup; I get 0.2.2, and still seem to have this issue.

PR #2526, opened 2024-11-20
Reduce verbosity of debug impls for protobuf messages

flaub:
This should get forward-ported to main as well


mothran (approved):

rlukata (approved):

flaub:
The `tracing::info!` issue is only on main

PR #2531, opened 2024-11-21
Fix concurrent access to TcpSocket client/server

flaub:


rlukata (approved):

risc0/zkvm/src/host/api/server.rs
```diff
@@ -833,39 +817,67 @@ fn execute_redis(
...
+        if let Err(send_err) = sender.send((segment_key, segment)) {
+            let mut redis_err_opt = redis_err.lock().unwrap();
+            let redis_err_inner = redis_err_opt.take();
+            return Err(match redis_err_inner {
+                Some(redis_thread_err) => anyhow!(redis_thread_err),
+                None => send_err.into(),
+            });
+        }
         Ok(Box::new(NullSegmentRef))
     });
```
mothran, line 883: Note in this file, there is a left over `.expect()` on the join handle that I think we should consider replacing with a non-panic'ing: 
```rust
join_handle
        .join()
        .map(|_| session)
        .map_err(|err| anyhow::anyhow!("redis task join failed: {err:?}"))?
```

PR #2532, opened 2024-11-21
Fix concurrent access to TcpSocket client/server

flaub:


rlukata (approved):

risc0/zkvm/src/host/api/server.rs
```diff
@@ -729,39 +715,67 @@ fn execute_redis(
...
+        if let Err(send_err) = sender.send((segment_key, segment)) {
+            let mut redis_err_opt = redis_err.lock().unwrap();
+            let redis_err_inner = redis_err_opt.take();
+            return Err(match redis_err_inner {
+                Some(redis_thread_err) => anyhow!(redis_thread_err),
+                None => send_err.into(),
+            });
+        }
         Ok(Box::new(NullSegmentRef))
     });
```
mothran, line 780: This file includes the .expect() which can cause a hang if the r0vm process exits un-safely.

PR #2534, opened 2024-11-21
Genericize EC Support Over Curves

tzerrell:
Change our EC acceleration from being secp256k1 specific to now support EC operations on any 256-bit prime field curve.


risc0/bigint2/src/ec/mod.rs
```diff
@@ -16,124 +16,240 @@
...
 const DOUBLE_BLOB: &[u8] = include_bytes_aligned!(4, "double.blob");
 
+/// The secp256k1 curve's prime as u32 digits, least significant digit first
+const SECP256K1_PRIME: [u32; 8] = [
+    0xFFFFFC2F, 0xFFFFFFFE, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF,
+];
+const SECP256K1_CURVE: &WeierstrassCurve<8> =
+    &WeierstrassCurve::<8>::new(SECP256K1_PRIME, [0u32; 8], [7, 0, 0, 0, 0, 0, 0, 0]);
+
 pub const EC_256_WIDTH_WORDS: usize = 256 / 32;
```
austinabell, line 32: Not currently used anywhere, I made some use in #2542 but we could also not expose this if we want

risc0/bigint2/src/lib.rs
```diff
@@ -12,9 +12,75 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(feature = "num-bigint-dig")]
-extern crate num_bigint_dig as num_bigint;
-
 pub mod ec;
```
SchmErik, line 15:
```suggestion
#[cfg(features = "unstable)]
pub mod ec;
```
I think gating `ec`, `ffi` and `rsa` modules on the `unstable` feature will give you more flexibility to change things as more crates are patched.

tzerrell: thank you!

SchmErik (approved):
the unstable stuff looks good to me

PR #2537, opened 2024-11-21
Switch curve to be static ref over Rc

austinabell:
Very optional. Assuming no reason to set curve at runtime


tzerrell (approved):

tzerrell:
Seems good to be able to support `no_std`, and I don't see why you'd need to ever delay initializing a curve until runtime, so I say include this

PR #2538, opened 2024-11-21
mark affine constructor unchecked, remove curve from type

austinabell:
- Affine constructor now `new_unchecked` to indicate that points not on a curve can be constructed
- Removed curve field from `AffinePoint`
  - Removed curve check on ec add
- Change exported secp curve type (now a const fn on the curve type)
- Changed API to have a mutable ref to the return, consistent with RSA APIs (cycle optimization for re-using buffers)

### DIFF
--- a/risc0/bigint2/src/ec/mod.rs
+++ b/risc0/bigint2/src/ec/mod.rs
@@ -72,7 +72,8 @@ pub struct AffinePoint<const WIDTH: usize> {
 }
 
 impl<const WIDTH: usize> AffinePoint<WIDTH> {
-    /// Constructs an affine point, without checking that it is on a specific curve.
+    /// Constructs an affine point from x and y coordinates, without checking that it is on
+    /// a specific curve.
     pub fn new_unchecked(x: [u32; WIDTH], y: [u32; WIDTH]) -> AffinePoint<WIDTH> {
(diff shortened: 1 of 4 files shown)


risc0/bigint2/src/ec/mod.rs
```diff
@@ -89,26 +88,22 @@ impl<const WIDTH: usize> AffinePoint<WIDTH> {
...
-        }
+    pub fn from_u32s(data: [[u32; WIDTH]; 2]) -> AffinePoint<WIDTH> {
+        AffinePoint { buffer: data }
     }
 }
 
 pub fn mul<const WIDTH: usize>(
     scalar: &[u32; WIDTH],
     point: &AffinePoint<WIDTH>,
-) -> AffinePoint<WIDTH> {
```
tzerrell, line 106: Did you switch back to putting the result as a mutable parameter instead of as a return value for performance reasons?

tzerrell (approved):
Looks good

risc0/bigint2/src/ec/mod.rs
austinabell: yes, and for consistency with RSA. Can always add higher-level APIs later for ergonomics?

PR #2539, opened 2024-11-22
reduce bigint2 log level

austinabell:
This log is very verbose, assume meant to be debug (could be trace?)


flaub:
I was intending to keep this in until I finish a way to print metrics on a per-op and per-accelerator grouping is done.

austinabell:
> I was intending to keep this in until I finish a way to print metrics on a per-op and per-accelerator grouping is done.

Cool, I'll close this for now then, feel free to open back if you want

PR #2542, opened 2024-11-22
refactor AffinePoint to be generic over the curve

austinabell:
Context is that in the future we may way to check that a point is on the curve on construction, and this generic would ensure that you don't perform arithmetic on points on different curves, to make it harder to misuse.

This trades off having an extra generic for the `AffinePoint` type for removing the need to pass the curve as a parameter to the fn.

Changes are very much optional, if we decide we want this, I will go through to update the upstream patches to ensure compatibility.


tzerrell (approved):
I can see arguments for both the old API style and this PR's API style, but I think this is good.

PR #2543, opened 2024-11-22
split modpow into different bit lengths

austinabell:
Alternative being having a trait impl on the array type, but that involves introducing a new trait and changing the API slightly.

TODO:
- [ ] Update blob


austinabell:
closing in favour of #2545

PR #2550, opened 2024-11-25
Add `r0vm-ver-compat` feature to relax r0vm server version checks

mothran:
Allows cross client/server minor version's when using rc/pre releases for testing upcoming ZKVM releases.


flaub (approved):

mothran:
@flaub / @SchmErik 

Ok I pushed an update that:
1. Adds more logging on r0vm redis thread failures to help debug live issues.
2. comments out the recache flag for future reference for folks hitting this.
3. Adds a ignore'd test for r0vm redis that uses external redis, this is super useful for our testing so we can just leverage that unit test for benching / error testing in the future.

Give it a second review pass when you can.

Additionally this would be a target for backporting to `release-1.0` and `release-1.1` when its ready.

SchmErik (approved):

flaub (approved):

PR #2564, opened 2024-11-26
fix: ran bun run docusaurus docs:version:api 1.2.0

nahoc:


SchmErik:
Thanks! I realized that i gave you the wrong command, it was supposed to be `1.2` rather than `1.2.0`. Anyway, I got it to work and so I'll close in favor of #2569

PR #2567, opened 2024-11-26
Set RISC0_FEATURE_bigint2 in risc0-build to signal zkVM features and fix usage of `stability::unstable` with `no_mangle`

nategraf:
This is a proposal for how we might improve build debugging when someone tries to use bigint2, but it is not available. In this PR, an env var is added to invocations of `cargo` in `risc0-build` that is a positive signal to downstream crates that that a new "hardware" feature is available.

In the case of upcoming cryptography patches that only work with `bigint2`, we'd add the following snippet to the `lib.rs` file. The advantage being that we give an error and link to our docs rather than a linker error at the end of the build process. From there, we can provide better guidance on either upgrading the `risc0-zkvm` version or downgrading the cryptography patch version.

```rust
#[cfg(all(target_os = "zkvm", target_arch = "riscv32"))]
const _: () = {
    assert!(
        core::option_env!("RISC0_FEATURE_bigint2").is_some(),
        r#"
RISC Zero zkVM feature bigint2 is not available, and is required by this crate.

If you'd like to use bigint2, please upgrade to risc0-zkvm and risc0-build and ensure the required
feature flags are enabled. See the RISC Zero dev docs for more information.
https://dev.risczero.com/api/zkvm/acceleration
"#
    );
};
```

This is written with bigint2 in mind. It could also be used when transitioning to the new ABI with zkVM revision 2.

When I went to test this I found that my ECDSA program linked and built without setting `unstable` on `risc0-zkvm`, which was unexpected. Digging into the `stability` crate I found that the way the implement their macro [1], it doesn't play nice with `no_mangle` and the symbol will still be included.

https://docs.rs/stability/latest/src/stability/unstable.rs.html#85


flaub:
OK, this seems reasonable. Minor nit about assert error message: should be 1.2 and all future versions.

nategraf:
Need to gate the setting of the env var behind an `unstable` flag

nategraf:
Ok @flaub, I added an `unstable` feature flag as a gate to `risc0-build`. I also added the compile-time assert (updated message also in PR description) to `risc0-bigint2` so that it doesn't need to be added to each individual crate. I dropped mention of a specific zkVM version from the message and we can instead keep that information up to date in the docs.

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -895,9 +895,8 @@ pub extern "C" fn sys_exit(status: i32) -> ! {
 /// `control_id` must be aligned and dereferenceable.
 ///
 /// `input` must be aligned and have `input_len` u32s dereferenceable
-#[cfg(feature = "export-syscalls")]
+#[cfg(all(feature = "export-syscalls", feature = "unstable"))]
 #[no_mangle]
-#[stability::unstable]
```
flaub, line 900: Why does this need to be changed?

flaub: This breaks the ability for the `unstable` attribute to adjust the docs with an unstable warning

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -1018,7 +1015,7 @@ macro_rules! impl_sys_bigint2 {
         ///
         /// `blob_ptr` and all arguments must be aligned and dereferenceable.
         #[cfg_attr(feature = "export-syscalls", no_mangle)]
-        #[stability::unstable]
+        #[cfg(feature = "unstable")]
```
flaub, line 1018: These are not equivalent.

nategraf:
@flaub , sorry I should have called it out specifically with a comment instead of just adjusting the description.

When I went to test this I found that my ECDSA program linked and built without setting unstable on risc0-zkvm, which was unexpected. Digging into the stability crate I found that the way the implement their macro [1], it doesn't play nice with no_mangle and the symbol will still be included.

So I made the changes to use `cfg` instead of `stability` to ensure that the symbol does not get included when the unstable flag is not set. I see now that this isn't quite right. Let me try another variation.

[1] https://docs.rs/stability/latest/src/stability/unstable.rs.html#85

flaub:
It looks like CI isn't happy about it either: https://github.com/risc0/risc0/actions/runs/12060414268/job/33630808653?pr=2567

nategraf:
> It looks like CI isn't happy about it either: https://github.com/risc0/risc0/actions/runs/12060414268/job/33630808653?pr=2567

So this I think is due to the `unstable` flag not being set on the `jwt` example guest. So similar to ECDSA, I was expecting this to fail as it is on `main` but it was working. 100439a8ab57c8aefa5850ff47c9b1c281f0b235

nategraf:
Ok, I tried a different variant with https://github.com/risc0/risc0/pull/2567/commits/91bd5208e316ad27ba74a05b774ac25de40f8fb2. It results in the expected behavior of the `jwt` example working when the guest `Cargo.toml` has `unstable` set and not without it. It is still not the exact same as whats on `main`, now the rust exports are provided even when the `export-syscall` feature is not set. I could add an extra `cfg` to avoid that if you think that would be better.

examples/jwt-validator/methods/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [build-dependencies]
-risc0-build = { path = "../../../risc0/build" }
+risc0-build = { path = "../../../risc0/build", features = ["unstable"] }
```
flaub, line 7: Won't this cause the entire workspace to allow unstable to be visible?

risc0/zkvm/platform/src/syscall.rs
```diff
@@ -895,8 +895,7 @@ pub extern "C" fn sys_exit(status: i32) -> ! {
 /// `control_id` must be aligned and dereferenceable.
 ///
 /// `input` must be aligned and have `input_len` u32s dereferenceable
-#[cfg(feature = "export-syscalls")]
-#[no_mangle]
+#[cfg_attr(all(feature = "export-syscalls", feature = "unstable"), no_mangle)]
```
flaub, line 898: Could we just drop the `export-syscall` feature? I'm not sure what it's for.

examples/jwt-validator/methods/Cargo.toml
nategraf: Hmmm, quite possibly yes. At least in the build script, so the `RISC0_FEATURE_bigint2` env var will be set. This won't make the `sys_bigint2` symbols available though. This is an interesting result of a common workspace.

risc0/zkvm/platform/src/syscall.rs
nategraf: We certainly could. The only reason I see for it is to give a way to prevent those symbols from ending up in the build, in case that were to cause some issues e.g. if another package implements them.

nategraf:
Looks like CI is failing with the `cargo-platform` related build failures. One solution to this in in https://github.com/risc0/risc0/pull/2584

PR #2570, opened 2024-11-27
bump version to 1.2.0

SchmErik:
Putting this in draft now until we get a green light from stakeholders


.github/workflows/release.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - uses: risc0/cargo-install@9f6037ed331dcf7da101461a20656273fa72abf0
         with:
           crate: cargo-binstall
-          version: "1.4"
+          version: "1.10.15"
```
SchmErik, line 70: to do in the future: drop binstall and use rzup to download cargo-risczero

SchmErik: 1.4 doesn't compile using rust 1.81 so I'm bumping it for now. this version build successfully when I tested locally.

PR #2571, opened 2024-11-27
Add redis retry logic

rlukata:
Single retry, then return error if the second one also fails.

During local testing I added a sleep before the reconnect and then stopped and started the docker instance. The executor was able to recover without issues.


SchmErik (approved):

mothran (approved):

flaub:
This job is hanging, @morgangibbs87?

PR #2572, opened 2024-11-27
ec bigint2 handles zero/infinity point

austinabell:
Tried refactoring to make code cleaner by having the AffinePoint type be an enum, and was ~1-2% slower, so kept as is.

Also tried changing how the buffer initializations happen within `mul`, this was fastest. Perf overall with these extra checks seems to be about 5% slower tested with one signature verify in k256

Going to add some docs before opening, as draft for visibility


risc0/bigint2/src/ec/mod.rs
```diff
@@ -160,31 +177,55 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
     pub fn double(&self, result: &mut Self) {
         let curve = C::CURVE;
-        double_raw(self.as_u32s(), curve.as_u32s(), &mut result.buffer);
+        if self.is_zero {
+            *result = *self;
+        } else {
+            unsafe {
+                double_raw(self.as_u32s(), curve.as_u32s(), &mut result.buffer);
+            }
+            result.is_zero = false;
```
tzerrell, line 192: This will work for the current curves, but if we ever do anything on a curve with points of order 2, we'll need to check if the y coordinate is zero. Perhaps something like the following:
```suggestion
            if self.buffer[1] != [0u32; WIDTH] {
                unsafe {
                    double_raw(self.as_u32s(), curve.as_u32s(), &mut result.buffer);
                }
                result.is_zero = false;
            } else {
                result.is_zero = true;
            }
```

risc0/bigint2/src/ec/mod.rs
```diff
@@ -160,31 +177,55 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
+        }
     }
 
     #[stability::unstable]
     pub fn add(&self, rhs: &AffinePoint<WIDTH, C>, result: &mut AffinePoint<WIDTH, C>) {
         let curve = C::CURVE;
         // TODO: Do we want to check for P + P, P - P? It isn't necessary for soundness -- it will fail
         // an EQZ if you try -- but maybe a pretty error here would be good DevEx?
```
tzerrell, line 200: This PR resolves this TODO
```suggestion
```

tzerrell (commented):
:tada:

risc0/bigint2/src/ec/mod.rs
austinabell: Sorry I read this after pushing a commit, I think https://github.com/risc0/risc0/pull/2572/commits/8b31b229f8300ccd38e2af48ff2a8617a7bde230 is that diff (maybe double check I didn't apply in the wrong place)

Should this also be done in the add fn? (it calls raw method since the is_zero already checked)

risc0/bigint2/src/ec/mod.rs
```diff
@@ -160,31 +179,71 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
+            *result = *self;
+        } else {
+            if self.buffer[1] != [0u32; WIDTH] {
+                unsafe {
+                    double_raw(self.as_u32s(), curve.as_u32s(), &mut result.buffer);
+                }
+                // DO NOT REMOVE: the result is unchecked, and only the buffer is updated above
+                result.is_zero = false;
+            } else {
+                // DO NOT REMOVE: the result is unchecked, and only the buffer is updated above
```
tzerrell, line 200: Not sure this comment is quite accurate. Maybe this?
```suggestion
                // DO NOT REMOVE: in this case a zero has been computed and the buffer is ignored
```

risc0/bigint2/src/ec/mod.rs
```diff
@@ -160,31 +179,71 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
+    /// Elliptic curve doubling of the affine point.
     #[stability::unstable]
     pub fn double(&self, result: &mut Self) {
         let curve = C::CURVE;
-        double_raw(self.as_u32s(), curve.as_u32s(), &mut result.buffer);
+        // If the point is zero, can short-circuit and return the identity point.
+        if self.is_zero {
+            *result = *self;
+        } else {
+            if self.buffer[1] != [0u32; WIDTH] {
```
nategraf, line 193: Is this correct? I think with double there shouldn't be a branch here, if `is_zero == false` we process with the `sys_bigint` call.

risc0/bigint2/src/ec/mod.rs
tzerrell: LGTM

The add fn is good as-is. This is the analog of the `self.buffer[0] == rhs.buffer[0]` and `self.buffer[1] != rhs.buffer[1]` case over in `add` -- that case checks for `A + -A = 𝒪` (where `𝒪` is the zero point), and this case checks for when `2A = 𝒪` because that happens exactly when the y coordinate is zero (because `-(x, y)` is `(x, -y)` on elliptic curves, so `A = -A` when `y = 0`)

risc0/bigint2/src/ec/mod.rs
austinabell: From https://github.com/risc0/risc0/pull/2572#discussion_r1861306172

tzerrell: This isn't checking for `A` is zero, this is checking for `A` is order 2, i.e. `2A = 𝒪`. This is because `-(x, y) = (x, -y)`, so we can check `A = -A` by checking `y = 0`

tzerrell: See also this thread: https://github.com/risc0/risc0/pull/2572#discussion_r1861306172

tzerrell (commented):
LGTM but will wait to hit approve until @nategraf is happy too

risc0/bigint2/src/ec/mod.rs
```diff
@@ -160,31 +179,71 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
+                );
+            }
+            // DO NOT REMOVE: the result is unchecked, and only the buffer is updated above
+            result.is_zero = false;
+        }
     }
 }
 
-fn double_raw<const WIDTH: usize>(
+unsafe fn double_raw<const WIDTH: usize>(
```
flaub, line 246: Why is this unsafe?

risc0/bigint2/src/ec/mod.rs
```diff
@@ -35,10 +35,14 @@ const SECP256K1_CURVE: &WeierstrassCurve<EC_256_WIDTH_WORDS> =
...
+/// Generic static curve configuration.
 pub trait Curve<const WIDTH: usize> {
     const CURVE: &'static WeierstrassCurve<WIDTH>;
 }
 
+/// An implementation of [Curve] for secp256k1.
+///
+/// This type should be used as a generic for [AffinePoint].
 #[derive(Debug, Eq, PartialEq, Ord, PartialOrd)]
 pub enum Secp256k1Curve {}
```
flaub, line 47: I feel like this should go into a different module

risc0/bigint2/src/ec/mod.rs
```diff
@@ -160,31 +179,71 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
+            } else {
+                // DO NOT REMOVE: in this case a zero has been computed and the buffer is ignored
+                result.is_zero = true;
+            }
+        }
     }
 
+    /// Elliptic curve addition of the affine point.
     #[stability::unstable]
     pub fn add(&self, rhs: &AffinePoint<WIDTH, C>, result: &mut AffinePoint<WIDTH, C>) {
```
flaub, line 208: Why not have this return an AffinePoint?

risc0/bigint2/src/ec/mod.rs
austinabell: Just because the zero point handling isn't done for the raw functions and can panic in the host. Depends what the semantics of unsafe we want to follow, I just assumed that a panic in the host instead of guest would qualify.

Happy to switch back —- but was more of a reason to ensure that in future if refactors, they are done with care, because it can very easily break edge cases without noticing.

I'll add some tests for the edge cases regardless.

risc0/bigint2/src/ec/mod.rs
austinabell: Optimization of not having to move around the affine point on the stack. Having this API and re-using the result buffers has given non-trivial cycle reductions in downstream crate patches.

austinabell:
I've made a few changes since review, so let me summarize the notable ones:
- Moved curve types into separate module [`259c0b5` (#2572)](https://github.com/risc0/risc0/pull/2572/commits/259c0b508f5618211f7936fe879d911f06247a66)
- Added test cases for edge cases added in this PR [`89600ad` (#2572)](https://github.com/risc0/risc0/pull/2572/commits/89600ad99aebe2b6048fcf51a60e4b9b4bea0ad4)
- Changed name of `is_zero` variable to `infinity` [`52a49be` (#2572)](https://github.com/risc0/risc0/pull/2572/commits/52a49be22edc83ec088b52bab635671fc90b24e9)
  - This was already causing confusion internally with the zero coordinates, seems like a less ambiguous name
- Switch `as_u32s` to return an optional ref [`3f42a9b` (#2572)](https://github.com/risc0/risc0/pull/2572/commits/3f42a9b1ddf36ab68e8eaf0d87cc93e5b011ad41)
  - Seems more correct, because the previous API could easily be misused by using the buffer when the point is the identity point
  - Noting this is a breaking change from the previous rc release, but previous impls should not be used anyway (so breakage is almost helpful)

risc0/bigint2/src/ec/mod.rs
flaub: I think `unsafe` has a specific definition around undefined behavior in rust. Extending this definition I think can cause confusion, especially over the long term.

risc0/bigint2/src/ec/curve.rs
```diff
@@ -0,0 +1,70 @@
...
+        SECP256K1_PRIME,
+        [0u32; EC_256_WIDTH_WORDS],
+        [7, 0, 0, 0, 0, 0, 0, 0],
+    );
+
+/// Generic static curve configuration.
+pub trait Curve<const WIDTH: usize> {
+    const CURVE: &'static WeierstrassCurve<WIDTH>;
+}
+
```
flaub, line 32: I guess I was thinking something like a generic Curve trait could live in `ec` but then we'd have a separate module for each kind of specific curve.

austinabell: Sure! Was this what you had in mind? [`abccf42` (#2572)](https://github.com/risc0/risc0/pull/2572/commits/abccf42fe4e617774e91706ad5bdf1d406ff7fe8)

Or did you still want the curve types kept in their same module for organization? Exports are the same either way

tzerrell (commented):
Thank you, I really appreciate the updates here, I think they make the code much nicer and easier to use going forward.

risc0/bigint2/src/ec/mod.rs
```diff
@@ -70,58 +56,77 @@ impl<const WIDTH: usize> WeierstrassCurve<WIDTH> {
...
-        SECP256K1_CURVE
-    }
-}
 
-#[derive(Clone, Debug, Eq, PartialEq, Copy)]
+/// An affine point on an elliptic curve.
+#[derive(Debug, Eq, PartialEq)]
 pub struct AffinePoint<const WIDTH: usize, C> {
     buffer: [[u32; WIDTH]; 2],
+    infinity: bool,
```
tzerrell, line 68: I think this is a good clarification, thank you. Personally, I think a re-name to identity would be even better [insert pedantic math lecture about an ellipitic curve's point at infinity not being the same as infinity], but I'm content to go along with infinity if people prefer it.

risc0/bigint2/src/ec/mod.rs
```diff
@@ -32,7 +32,7 @@ pub use curve::{Curve, Secp256k1Curve, WeierstrassCurve};
 #[derive(Debug, Eq, PartialEq)]
 pub struct AffinePoint<const WIDTH: usize, C> {
     buffer: [[u32; WIDTH]; 2],
-    is_zero: bool,
+    infinity: bool,
```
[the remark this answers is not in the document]
tzerrell: Oops, should have put this on the current revision, resolving in favor of https://github.com/risc0/risc0/pull/2572/files#r1866478006

risc0/bigint2/src/ec/mod.rs
austinabell: Yeah, sounds good, I'll do identity, given it would be the same name as the identity constant we have. I was just defaulting to infinity because of k256 naming

risc0/bigint2/src/ec/mod.rs
austinabell: I'll switch back, missed this comment

risc0/bigint2/src/ec/mod.rs
```diff
@@ -158,36 +195,36 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
+        };
+        let Some(rhs) = rhs.as_u32s() else {
             *result = *self;
-        } else if self.buffer[0] == rhs.buffer[0] {
+            return;
+        };
+
+        if lhs[0] == rhs[0] {
             // X coordinates are the same, so either we double the value if it's the same point,
             // or return the identity if it's different (not on the curve).
```
tzerrell, line 207: An assumption of this function is that both points are on the curve; so if the points are different but have the same x value they must be negatives (because if (x, y) is on the curve, the only other point on the curve with the same x coordinate is (x, -y) which is -(x, y) under the elliptic curve group law).
```suggestion
            // or return the identity if it's different (and so lhs = -rhs).
```

tzerrell (approved):
LGTM, thank you! I only have a minor documentation comment

austinabell:
Tested latest commit on k256 and all tests pass in zkvm https://github.com/risc0/RustCrypto-elliptic-curves/pull/6/commits/2d4edfafc4d50d940e7ad994c28d94d95978a735

PR #2576, opened 2024-11-30
Override gcc version

### LINKED ISSUE #2575: CUDA high gcc version incompatibility

- [X] I have searched the existing issues

- [X] This is not a support question.

### Description of bug

On some systems like `Arch Linux` that have high GCC versions such as 13 and upwards are incompatible with CUDA. Therefore we need a way to override the compiler that nvcc uses for compiling the CUDA kernels.

### Steps to reproduce

Try to compile the CUDA feature using a gcc version of 14.2.1. Will result in:
```
  --- stderr


  error occurred: Command LC_ALL="C" "nvcc" "-ccbin=c++" "-Xcompiler" "-O0" "-Xcompiler" "-ffunction-sections" "-Xcompiler" "-fdata-sections" "-Xcompiler" "-fPIC" "-m64" "-I" "/home/starkbamse/.cargo/git/checkouts/risc0-7d6ab9951628e5b1/71ecab3/risc0/sys/cxx" "-I" "/home/starkbamse/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sppark-0.1.10/sppark" "-Xcompiler" "-Wall" "-Xcompiler" "-Wextra" "-DFEATURE_BABY_BEAR" "-arch=native" "-diag-suppress=177" "-diag-suppress=2922" "-Xcudafe" "--display_error_number" "-Xcompiler" "-Wno-missing-braces,-Wno-unused-function" "-o" "/home/starkbamse/GitHub/sparkle/target/debug/build/risc0-sys-b135acaf3de15a89/out/01ed5306a404dc2c-ntt.o" "-c" "--device-c" "kernels/zkp/cuda/supra/ntt.cu" with args nvcc did not execute successfully (status code exit status: 2).


```

saefstroem:
Closes #2575 
Closes ZKVM-736

This allows the user to specify `NVCC_CCBIN` and override the default gcc compiler that is used by the nvcc compiler as the cc build tool does not adhere to if this variable is simply set from the linux environment.


saefstroem:
@flaub

flaub:
Related: #2695

saefstroem:
> Possibly duplicate: #2695 

That is the duplicate. This PR is 2mo old :)

flaub:
Are they both trying to do the same thing?

saefstroem:
> Are they both trying to do the same thing?

Yes but this PR allows the specification of the native Nvidia env variables instead of creating custom logic. This is a one-line fix.

flaub (approved):

risc0/build_kernel/src/lib.rs
```diff
@@ -197,6 +198,7 @@ impl KernelBuild {
             .cuda(true)
             .cudart(&cudart)
             .debug(false)
+            .ccbin(false)
```
flaub, line 201: This seems to break in our CI environment. Is there a way we can make this conditional? Perhaps only if `NVCC_CCBIN` is defined?

flaub (changes requested):

risc0/build_kernel/src/lib.rs
saefstroem: This is a good idea. I'll update it to only disable if that env var is defined.

saefstroem:
@flaub

risc0/build_kernel/src/lib.rs
```diff
@@ -193,10 +194,13 @@ impl KernelBuild {
...
+        let build = build.cuda(true).cudart(&cudart).debug(false);
+
+        if env::var("NVCC_CCBIN").is_ok() {
+            build.ccbin(false);
+        }
+
         build
-            .cuda(true)
-            .cudart(&cudart)
-            .debug(false)
```
flaub, line 199: How about this?

```
build
    .cuda(true)
    .cudart(&cudart)
    .debug(false)
    .ccbin(!env::var("NVCC_CCBIN").is_ok())
```

saefstroem:
@flaub Good suggestion, simplified even further :)

flaub:
Almost there, can you bump the first line so it reads:
```
// Copyright 2025 RISC Zero, Inc.
```

saefstroem:
> Almost there, can you bump the first line so it reads:
> 
> ```
> // Copyright 2025 RISC Zero, Inc.
> ```

done :)

flaub (approved):

PR #2584, opened 2024-12-02
Update Rust toolchain version in rust-toolchain.toml

nategraf:
I've started to see some issues with crates requiring Rust tool chains newer than 1.77. This PR bumps the Rust toolchain version to 1.81, which coincides with the latest guest toolchain (although there is no strong requirement that these coincide).

I opened this in reaction to build failures in the `risc0-ethereum` job. There is also the following, more targeted solution to that exact problem. https://github.com/risc0/risc0/pull/2586


flaub (approved):

nategraf:
Not sure what's up with the `semver` check
https://github.com/risc0/risc0/actions/runs/12126871150/job/33809916624?pr=2584

nategraf:
I removed `zkevm-demo` example in 0558286236a6bd21cda0f2248b551fe20d86b40f, but can revert this if we don't want to do that. I won't merge this unless @flaub is ok with this.

PR #2593, opened 2024-12-02
backport 1.0: Add redis retry logic (#2571)

SchmErik:
Single retry, then return error if the second one also fails.

During local testing I added a sleep before the reconnect and then stopped and started the docker instance. The executor was able to recover without issues.


flaub (approved):

risc0/zkp/src/hal/metal.rs
```diff
@@ -850,49 +850,49 @@ impl<MH: MetalHash> Hal for MetalHal<MH> {
         self.hash.as_ref().unwrap().get_hash_suite()
     }
 
-    #[cfg(feature = "metal_prefix_products")]
-    fn prefix_products(&self, io: &Self::Buffer<Self::ExtElem>) {
-        let block_size = 256;
-        let grain_size = 16;
```
flaub, line 856: Wondering why these are changing for a backport?

SchmErik: I backported the toolchain changes to get the CI to work and newer toolchains detect features that aren't in the Cargo.toml like this `metal_prefix_products` feature

PR #2600, opened 2024-12-03
Add compile time env vars to docker build

eyusufatik:
Implemented passing of compile time environment variables to Docker builds.

Compile time env vars are very useful for guest code that's desired to be run for different environments or different modes. One example would be light clients for blockchains, which you'd like to support mainnet / testnet. Instead of adding an output stating which network the proof belongs to, (or a genesis hash) you can compile different elf files for different environments. And it'd be up to the verifier to know which method id to use.

We are looking to use this feature in [Citrea](https://github.com/chainwayxyz/citrea) and [risc0-to-bitvm2](https://github.com/chainwayxyz/risc0-to-bitvm2).

 There's a few points I need feedback on so I'm opening the PR as draft.
 
 - `DockerOptions::compile_time_env` is Option<Vec> but it can be implemented as Vec. (empty vec instead of None)
 - A lot of `build_guest_package_docker` calls were made with `GuestBuildOptions::default()` which has `use_docker` set to None. I expected `use_docker` to be some in my patch, however, I can change that behaviour if that's not what you desire.
 - `cargo risczero build` command needs to be updated so that a list of env. var. names can be passed. How should that be implemented? Expect a pair of key values or just keys?

### DIFF
--- a/risc0/build/src/config.rs
+++ b/risc0/build/src/config.rs
@@ -24,6 +24,9 @@ pub struct DockerOptions {
     ///
     /// The current working directory is used if `None` is specified.
     pub root_dir: Option<PathBuf>,
+
+    /// Specify the environemnt variables for the docker build.
+    pub compile_time_env: Option<Vec<(String, String)>>,
 }
 
 /// Options defining how to embed a guest package in
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -255,14 +255,19 @@ mod test {
     use std::path::Path;
 
     use super::{build_guest_package_docker, TARGET_DIR};
-    use crate::config::GuestBuildOptions;
+    use crate::{config::GuestBuildOptions, DockerOptions};
 
     const SRC_DIR: &str = "../..";
 
-    fn build(manifest_path: &str) {
+    fn build(manifest_path: &str, compile_time_env: Option<Vec<(String, String)>>) {
         let src_dir = Path::new(SRC_DIR);
         let manifest_path = Path::new(manifest_path);
-        build_guest_package_docker(manifest_path, src_dir, &GuestBuildOptions::default()).unwrap();
+        let mut build_opts = GuestBuildOptions::default();
+        build_opts.use_docker = Some(DockerOptions {
+            root_dir: Some(src_dir.to_path_buf()),
+            compile_time_env,
+        });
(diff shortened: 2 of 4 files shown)


nategraf:
One important question is: what use case does this solve better than alternatives, such as feature flags? Do you have an thoughts about using this instead of a feature flag?

eyusufatik:
> One important question is: what use case does this solve better than alternatives, such as feature flags? Do you have an thoughts about using this instead of a feature flag?

yes, in our case we want to include a genesis state root (and a bunch of other things) of the rollup in the light client circuit and the batch proof circuit. we either make it a constant inside the elf code or we pass it as an input. using compile time inputs makes it much easier to work with an elf that will be built for different environments.

if we do it an input, then we have to output it so that a verifier can assert it. if we don't output it, that means we can build on a different chain.

this kind of approach is not really ideal since something else down the line has to check the output -> you have to make whatever you need to check a constant in a different place.

going with a feature flag in a case like this would need you to know the constant before hand, or modify the code when you get to know it. that works of course but it doesn't really work well with testing.

risc0/build/src/config.rs
```diff
@@ -24,6 +24,9 @@ pub struct DockerOptions {
     ///
     /// The current working directory is used if `None` is specified.
     pub root_dir: Option<PathBuf>,
+
+    /// Specify the environemnt variables for the docker build.
+    pub compile_time_env: Option<Vec<(String, String)>>,
```
nategraf, line 29: Hmmm, this is actually a breaking change. And we do use the `DockerOptions { _ }` syntax to construct this type in downstream applications. I do wish we'd marked this as `#[non_exhaustive]` and gave a builder instead like we do for `ProverOpts` [1]. Maybe we accept this as a breaking change, but it'd be good is there is another way.

[1] https://github.com/risc0/risc0/blob/533aff23f633c6102c961d47b9370ec96770de6d/risc0/zkvm/src/host/client/prove/mod.rs#L159

nategraf:
@flaub and @SchmErik are the ones that maintain the build tooling here, so it'd be best to get their comments and suggestions on the approach.

flaub:
Seems similar to https://github.com/risc0/risc0/pull/2709

flaub:
Superceded by #2762

PR #2606, opened 2024-12-04
Fix missing `check-cfg` to validate `risc0_custom_cfg_test` flag

hashcashier:
This fixes the below lint when building with rust >1.79
```
risc0-zkvm-methods-cfg: warning: unexpected `cfg` condition name: `risc0_custom_cfg_test`
```


flaub:
Thanks for this, I ended up resolving this using a lint allowance.

PR #2609, opened 2024-12-04
Enforce < modulus on bigint results

tzerrell:
Enforces that we return the canonical form of modular arithmetic operations from bigint calls (e.g., if we compute the answer to be 2 in the finite field of order 3, don't allow returning 5 even though 2 == 5 mod 3).


jacobdweightman (approved):

risc0/bigint2/src/lib.rs
```diff
@@ -32,6 +32,20 @@ pub trait ToBigInt2Buffer<const WIDTH: usize> {
     fn from_u32_array(array: [u32; WIDTH]) -> Self;
 }
 
+#[cfg(feature = "unstable")]
+// Checks if two u32 arrays representing big integers with little-endian digit order satisfy lhs < rhs
+fn is_less<const N: usize>(lhs: &[u32; N], rhs: &[u32; N]) -> bool {
```
nategraf, line 37: You might try adding `#[inline]` just to see if this changes anything.

nategraf (approved):

PR #2610, opened 2024-12-04
remove redundant prime checks, inline is_less

austinabell:
Only checks the result is within prime once per public function call

inlines `is_less` as it was measured to be an optimization (context: https://github.com/risc0/risc0/pull/2609#discussion_r1870272857)


tzerrell (approved):
Thank you, looks good to me!

PR #2612, opened 2024-12-04
optimize buffer pointer swapping in bigint2 EC

austinabell:
Swapping the memory addresses of the pointers is possible and more optimized than this. This shaves off ~6000 cycles from a signature verify (~2%), but imo also cleaner code.


tzerrell (approved):
Great!

risc0/bigint2/src/ec/mod.rs
```diff
@@ -120,57 +120,49 @@ impl<const WIDTH: usize, C: Curve<WIDTH>> AffinePoint<WIDTH, C> {
...
         }
 
         if first_write {
             // Multiplied by zero, which is the identity point.
             *result = AffinePoint::IDENTITY;
         } else {
-            // Return the result, based on which buffer was written to last.
-            *result = if result_flip { result2 } else { result1 };
+            // Copy the value from the buffer most recently written to.
+            *result = *result_ptr;
```
austinabell, line 159: There might be a way to only copy if the most recent buffer was the scratch buffer, and point the `result_ptr` to the parameter of the fn, but I couldn't think of a clean way to do this, might revisit (only a micro-optimization)

PR #2617, opened 2024-12-05
integrate keccak2

SchmErik:
still a WIP: marking as draft now and will mark as ready when tests/build errors are fixed


risc0/zkvm/src/host/protos/api.proto
```diff
@@ -60,11 +61,10 @@ message ProveSegmentReply {
...
 
 message ProveZkrRequest {
-  base.Digest claim_digest = 1;
-  base.Digest control_id = 2;
-  bytes input = 3;
+  base.Digest control_id = 1;
+  bytes input = 2;
   // This is optional in the context of a CoprocessorRequest
-  AssetRequest receipt_out = 4;
+  AssetRequest receipt_out = 3;
```
flaub, line 67: The ordinals are not allowed to change to maintain compatibility.

risc0/zkvm/src/host/server/session.rs
```diff
@@ -93,6 +93,10 @@ pub struct Session {
     /// A list of pending ZKR proof requests.
     // TODO: make this scalable so we don't OOM
     pub(crate) pending_zkrs: Vec<ProveZkrRequest>,
+
+    /// A list of pending ZKR proof requests.
```
flaub, line 97:
```suggestion
    /// A list of pending Keccak proof requests.
```

risc0/circuit/keccak/Cargo.toml
```diff
@@ -34,6 +34,7 @@ risc0-circuit-keccak-methods = { path = "methods" }
 risc0-zkvm = { workspace = true, features = [
   "std",
   "unstable",
+  "prove",
```
flaub, line 37: If we always assume `prove`, then do we still need a feature flag for this crate?

risc0/zkvm/src/host/client/env.rs
```diff
@@ -79,6 +79,9 @@ pub struct ProveZkrRequest {
 /// A Keccak proof request.
 #[stability::unstable]
 pub struct ProveKeccakRequest {
+    /// The digest of the claim that this ZKR program is expected to produce.
+    pub claim_digest: Digest,
```
flaub, line 83: Why is this needed for a proof request?

risc0/circuit/keccak/Cargo.toml
SchmErik: yes - the prove flag is set for dev-dependencies - I changed the test for this crate to not use coprocessor so it exercises the pending_keccaks code path (which is similar to the pending_zkrs thing we do) in the server.

risc0/zkvm/src/host/client/env.rs
SchmErik: originally, I didn't think this was needed either. so I left it out. As i worked on the prover example, I saw that for prove_zkr in the prover example, this was utilized. I was thinking that this could be computed from the receipt produced from the proof but i saw that the prove_zkr did it thsi way so I added this. This is set by the host syscall handler and is passed from the guest. If there's a mismatch then it means that the guest and the prover disagreed on the digest, which is a good check to have. but I guess this spurs another question: why does this exist on the prove_zkr side?

PR #2618, opened 2024-12-05
semver checks: ignore circuit keccak methods

SchmErik:


flaub:
Are you sure we want to do this? I think the error was only happening on the first time?

flaub (approved):

PR #2619, opened 2024-12-06
ZKVM-1020: Fix verifier parameters metadata check to actually compare verifier context and metadata

nategraf:
Discovered while debugging a different issue, the check that is supposed to report if verifier parameters are mismatched currently does not. Instead it just checks internal consistentency of `receipt.metadata.verifier_parameters` against `receipt.inner.verifier_parameters`, both of which are digests to help the verifier determine which version of the verifier parameters it should use. Note that the digest alone is not enough to verify the receipt, and instead the actual parameters come from the `VerifierContext`.

This PR fixes the check to actually compare against the verifier context so that the correct error message can be shown, instead of a generic verification failure.


flaub (approved):

nategraf:
I realized that a simple digest check doesn't work as well for `SegmentReceipt` as for `SuccinctReceipt` and `Groth16Receipt`, so I've removed the check there and added a note in 7625813ca6c38011af17c0816f53d55b3347bfc0

nategraf:
Not immediately sure why the keccak test is failing. I'll need to come back to this after I take care of some things for Boundless.

nategraf:
> @nategraf Is this still useful?

It is, ya. This is to fix a bug in how verification failures our reported when the parameters don't match.

flaub (approved):

flaub:
@nategraf Updating to main I don't think is the issue, it's really that some of the CI checks need to get fixed so we can land this.

PR #2622, opened 2024-12-07
guest: Implement keccak batching 

SchmErik:


risc0/zkvm/src/guest/env/batcher.rs
```diff
@@ -15,10 +15,107 @@
...
+        let mut claim_digest = self.claim_state;
+        for word in claim_digest.as_mut_words().iter_mut() {
+            *word = word.to_be();
+        }
+        claim_digest
+    }
+
+    fn reset(&mut self) {
+        self.claim_state = SHA256_INIT;
+        self.inputs = vec![];
```
flaub, line 90:
```suggestion
        self.inputs.clear();
```

This will potentially avoid a re-alllocation.

flaub (approved):

flaub (approved):

saurik:
@SchmErik @flaub The way this batcher works (as coded), it has to cache -- in the guest :( -- all of the keccak states that need to be proved. This involves a ton of memory copies that are otherwise useless, as the only reason for this caching is to be able to pass them to a later syscall... I think about a thousand cycles worth of copying? Instead, mightn't it be a lot better to break sys_prove_keccak into two steps: sys_queue_prove_keccak and sys_commit_prove_keccaks, where the first can be called on every keccakf, to cache the data in the host instead of the guest until the latter consumes all of the inputs? (Or do I have a misunderstanding about the performance tradeoffs involved with syscalls?)

flaub:
Thinking about this a little more, perhaps because we have an incremental sha2 hash of the transcript, the guest wouldn't need all the intermediate states.

@SchmErik and @nategraf WDYT?

saurik:
@flaub Yeah: all the guest is doing is passing that giant buffer of cached states to sys_prove_keccak and then discarding it. It is also keeping the sha2 hash of the claim_digest for the keccak circuit already, and so isn't using that buffer to verify any trust. Essentially, what I'm suggesting is very similar to how writing to the journal doesn't require buffering all of the data in the guest, as the guest also keeps the running sha2 hash of the data it is writing... you are already keeping that running sha2 hash of the pre/post states of the keccak operation, so storing this buffer in the guest really seems pointless (and when I wrote that comment I hadn't yet realized just how many such states are being stored... (1<<KECCAK_DEFAULT_PO2)/KECCAK_PERMUTE_CYCLES is 655, omg ;P. Even just the paging overhead of this sounds brutal, as that's 131kB of buffer; but, again: I might have a bad intuition for the cost model here. Certainly, though, it sticks out as a store thumb as being extremely inelegant vs. all of the other running hashes you have, as it is the only thing that actually requires a giant allocation from your basic set of behaviors.)

saurik:
FWIW, I started playing with implementing this, to get a feel for the prover server code, and realized that maybe there would be some major architecture issue with this, as there could be segment boundaries between the call to queue the data and when it is later committed...

(I also figured out that the reason that copy was taking me 1000+ cycles was merely because of a ridiculously suboptimal memcpy I threw together for my setup; I've added a better copy operator for my keccak state and got it down to 105 cycles; it still feels awkward to cache that much data in the guest--particularly if the po2 is to be separately configured inside the guest by the prover using the environment variable, as it will be the only thing I need anything like an allocator for--but, again: I readily admit I don't have a good intuition yet for the cost model of this VM.)

On such note, another advantage of the queue syscall idea is that, with a small modification to how it works, you can also avoid having the po2 abstraction leak into the guest (which I'm also finding really awkward, as it is making the guest environment feel like an extension of the configuration state of the prover in multiple ways... it just all feels so inelegant) by having the _host_ decide when _it_ has too many keccak operations queued and returning something from the queue syscall, informing the guest that it needs to commit its claim assumption.

saurik:
Another option for this po2 configuration -- if the queue pushback is architecturally miserable -- might be to add a syscall that returns the maximum number of keccaks?

I just really want to stress how much this keccak po2 push-down infects the guest: this is the only feature that requires using environment variables, this is the only feature which requires parsing decimal integers (!), and this is the only feature which hardcodes constants referring to the performance properties of the prover circuit (the math using KECCAK_PERMUTE_CYCLES).

(I am noticing this really hard as I am notably trying very hard to avoid compiling 100+kB of Rust into my Risc-V binary, as the attack surface of all of this code is crazy; I thereby am seeing very transparently all of the logic in the guest, as it is contained in a mere few hundred lines of C++... my hello world example binaries are just ~5kB. I worry that as you make things too easy to blur the line between the guest and the host that you'll lose control, and end up not immediately noticing the guest impact.

It will also be the first/only performance concern of the prover that requires passing configuration to the guest via .set_env() and hoping it takes advantage of it well...

I think I'd feel better about it if this sys_prove_keccak syscall were generalized for other circuits? If you were to later add a new circuit for, say, accelerated blakeb, wouldn't the arguments be the same? You'd want the control root (to look up the circuit), the input to commit to the circuit (backing up the verified claim receipt), and (maybe... ;P) a way to configure the po2 of said circuit (though I'd still wish for that to be handled more entirely on the host side). Like, this API is in some way just the dual of the guest-side API to commit these two values to the guest's assumptions list. Then the prover can look up the circuit by control root, give it the input that was passed (or queued) to calculate its claim receipt, and send it off to do whatever it normally would do... it doesn't really need to know it is specifically for keccak.

SchmErik:
@saurik thanks so much for your feedback.. We're on a break for the year at this point but I've started a draft PR addressing your comments here: https://github.com/risc0/risc0/pull/2682 The plan is to drop the accumulation and the po2 on the guest side and make it so that sys_keccak will return a status indicating that the host-side is "full" so the guest can then call sys_prove_keccak to generate the proof request on the host-side.

I'll get to this after the new year but if you like, feel free to take the commits from my PR and modify it. it's a work in progress so it likely won't compile. Thanks again!

PR #2635, opened 2024-12-09
Add RISC0_FEATURE_bigint2 in the Docker build environment

nategraf:


risc0/build/src/docker.rs
```diff
@@ -153,13 +153,20 @@ fn create_dockerfile(
...
         .copy(".", ".")
         .env(manifest_env)
         .env(rustflags_env)
-        .env(&[("CARGO_TARGET_DIR", "target")])
+        .env(&[("CARGO_TARGET_DIR", "target")]);
+
+    #[cfg(feature = "unstable")]
+    {
+        build = build.env(&[("RISC0_FEATURE_bigint2", "")]);
+    }
```
austinabell, line 167: This behavior would be asymmetric from non-docker builds, because in that case, you never need to use the `unstable` feature flag on `risc0-build`, only `risc0-zkvm`.

This might have been an unintended usage of the previous API, and that the recommended path is to use `unstable` on `risc0-build` and not use `RISC0_FEATURE_bigint2` directly, just noting.

nategraf: So, the intended pattern until `bigint2` is stabilized is that the developer set `unstable` on `risc0-zkvm` and `risc0-build`, and not use the `RISC0_FEATURE_bigint2` directly.

flaub (approved):

PR #2639, opened 2024-12-10
Add precompile & bigint docs

tzerrell:
Update our documentation for precompiles and bigint. Note that this is marked as draft, as some of the links are not yet correct until the updates in our patched crates are landed on appropriate branches.


website/api/zkvm/precompiles.md
```diff
@@ -60,12 +81,14 @@ modular multiplication instruction. E.g. [`FieldElement8x32R0::mul`][field-mul].
...
 [cargo-patch]: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section
 [curve25519-dalek]: https://github.com/risc0/curve25519-dalek/tree/risczero
-[ecdsa]: https://github.com/risc0/risc0/tree/main/examples/ecdsa
-[ecdsa-patched]: https://github.com/risc0/risc0/blob/main/examples/ecdsa/methods/guest/Cargo.toml#L13-L18
+[ecdsa]: https://github.com/risc0/risc0/tree/release-1.2/examples/ecdsa
+[ecdsa-patched]: https://github.com/risc0/risc0/blob/release-1.2/examples/ecdsa/methods/guest/Cargo.toml#L13-L18
 [field-mul]: https://github.com/risc0/RustCrypto-elliptic-curves/compare/k256/v0.13.1..k256/v0.13.1-risczero.1#diff-ab10e01be1d99a874f90c9a6143bb1c64f37e04dcb220b5ab50b9273d99e0a0cR176-R179
 [git-dep]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories
 [k256-diff]: https://github.com/risc0/RustCrypto-elliptic-curves/compare/k256/v0.13.1..k256/v0.13.1-risczero.1
+[lincomb]: TODO
```
pdg744, line 89: flagging this todo so we don't forget it

tzerrell: I'll also flag that most of the links to our own stuff will need review after all the target patches land

website/api/zkvm/precompiles.md
```diff
@@ -1,22 +1,23 @@
-# Cryptography Acceleration
+# Precompiles
 
 RISC Zero's rv32im implementation includes a number of specialized extension
-circuits, including two "accelerators" for cryptographic functions: SHA-256 and
-[256-bit modular multiplication][bigint], referred to as "bigint"
-multiplication. By implementing these operations directly in the "hardware" of
-the zkVM, programs that use these accelerators execute faster and can be proven
+circuits, including "precompiles" for cryptographic and algebraic functions: SHA-256,
+Keccak, RSA, elliptic curve, and modular multiplication operations.
```
pdg744, line 5: let's hold off on mentioning keccak for now

website/api/zkvm/precompiles.md
```diff
@@ -1,22 +1,23 @@
...
+By implementing these operations directly in the "hardware" of
+the zkVM, programs that use these precompiles execute faster and can be proven
 with significantly less resources [^1].
 
 ## Accelerated Crates
 
-The SHA-256 and bigint accelerators are currently integrated in "accelerated"
+Our precompiles are currently integrated in "accelerated"
```
pdg744, line 12: mb we want to reference the term "patched crates" here? not strongly opinionated.

website/api/zkvm/precompiles.md
```diff
@@ -33,24 +34,44 @@ git = "https://github.com/risc0/RustCrypto-hashes"
...
+version updates, you may need to regenerate your `lock` file. For example:
+```toml
+cargo update -p sha2 --precise 0.10.6
+```
+
 When using cryptography indirectly, e.g. via the `cookie`, `oauth2`, or `revm`,
 crates it may be possible to enable acceleration support without code changes by
 applying a [Cargo patch][cargo-patch].
 
+Our elliptic curve, keccak, and RSA precompiles still undergoing revision and
```
pdg744, line 54: Let's move forward with this assuming we'll land it before we land keccak. 

Also, I wonder if we can write this in a way that we won't have to manage granular updates about which precompiles are marked unstable... might be easier to just accept the maintenance work as we remove those flags (and add other precompiles marked similarly) but figured i'd mention the idea...

website/api_versioned_sidebars/version-1.2-sidebars.json
```diff
@@ -148,8 +148,8 @@
         },
         {
           "type": "doc",
-          "label": "Cryptography Acceleration",
-          "id": "zkvm/acceleration"
+          "label": "Precompiles",
+          "id": "zkvm/precompiles"
```
pdg744, line 152: We will probably want to create a redirect to avoid 404s for users who end up at the `zkvm/acceleration` page -- i don't recall the details of how those are set up but we can ask cohan/morgan for help if necessary

website/api_versioned_docs/version-1.2/zkvm/precompiles.md
```diff
@@ -33,6 +34,19 @@ git = "https://github.com/risc0/RustCrypto-hashes"
 tag = "sha2-v0.10.6-risczero.0"
 ```
 
+Make sure that your dependency gives the same version of the crate as listed in
+the patch tag. In this example, the dependency would be
+```toml
+[dependencies]
+sha2 = "0.10.6"
```
austinabell, line 41:
```suggestion
sha2 = "=0.10.6"
```

website/api_versioned_docs/version-1.2/zkvm/precompiles.md
```diff
@@ -33,6 +34,19 @@ git = "https://github.com/risc0/RustCrypto-hashes"
...
+```
+
+In some situations, particularly for crates you've used before with patch
+version updates, you may need to regenerate your `lock` file. For example:
```
austinabell, line 45:
```suggestion
In some situations, for example when a patch is used indirectly, you may
need to update the version specifically to update your lockfile. For example:
```

Personally unclear about this wording, plus this is the main reason why you might need to manage the patch version like this I think.

examples/ecdsa/methods/guest/Cargo.toml
```diff
@@ -6,15 +6,15 @@ edition = "2021"
...
 hex-literal = "0.4"
-k256 = { version = "=0.13.1", features = ["arithmetic", "serde", "expose-field", "std", "ecdsa"], default_features = false }
+k256 = { version = "=0.13.3", features = ["arithmetic", "serde", "expose-field", "std", "ecdsa"], default_features = false }
 
 [patch.crates-io]
 # Placing these patch statement in the workspace Cargo.toml will add RISC Zero SHA-256 and bigint
 # multiplication accelerator support for all downstream usages of the following crates.
 sha2 = { git = "https://github.com/risc0/RustCrypto-hashes", tag = "sha2-v0.10.6-risczero.0" }
-k256 = { git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.1-risczero.1"  }
+k256 = { git = "https://github.com/risc0/RustCrypto-elliptic-curves", branch = "austin/bigint2"  }
```
austinabell, line 17:
```suggestion
k256 = { git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.3-risczero.1" }
```

website/api/zkvm/precompiles.md
```diff
@@ -33,39 +34,59 @@ git = "https://github.com/risc0/RustCrypto-hashes"
...
 [cargo-patch]: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section
 [curve25519-dalek]: https://github.com/risc0/curve25519-dalek/tree/risczero
-[ecdsa]: https://github.com/risc0/risc0/tree/main/examples/ecdsa
-[ecdsa-patched]: https://github.com/risc0/risc0/blob/main/examples/ecdsa/methods/guest/Cargo.toml#L13-L18
-[field-mul]: https://github.com/risc0/RustCrypto-elliptic-curves/compare/k256/v0.13.1..k256/v0.13.1-risczero.1#diff-ab10e01be1d99a874f90c9a6143bb1c64f37e04dcb220b5ab50b9273d99e0a0cR176-R179
+[ecdsa]: https://github.com/risc0/risc0/tree/release-1.2/examples/ecdsa
+[ecdsa-patched]: https://github.com/risc0/risc0/blob/release-1.2/examples/ecdsa/methods/guest/Cargo.toml#L13-L18
 [git-dep]: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories
 [k256-diff]: https://github.com/risc0/RustCrypto-elliptic-curves/compare/k256/v0.13.1..k256/v0.13.1-risczero.1
+[lincomb]: TODO
```
austinabell, line 87: Is this a link you want? Or did you have something else in mind?

https://github.com/risc0/RustCrypto-elliptic-curves/blob/k256/v0.13.3-risczero.1/k256/src/arithmetic/mul.rs#L349-L377

austinabell (commented):
Going to make some docs in addition to this, to clear up some nuances and maybe build a table of supported crates/versions -- and happy to make these changes I comment there, if agreed and preferred

website/api/zkvm/precompiles.md
austinabell: I remove the keccak wording in #2642 because it was not yet released, sounds like I should add it back proactively just so we don't forget when we add keccak to the table?

pdg744 (approved):
Thank you!

pdg744 (changes requested):
whoops hit approve prematurely :)

austinabell:
https://github.com/risc0/risc0/actions/runs/12284121059/job/34279179170?pr=2639 failure is unrelated. If I remember correctly, this was pinned for MSRV reasons, is this documented somewhere, or good to open a separate PR to check newer versions? (Installing this version of cargo-binstall locally also fails)

cc @SchmErik

SchmErik (approved):
looks good so I'll approve - i talked to @austinabell about the test case failure. I think the only remaining open is the redirect but I'll defer to @morgangibbs87 or @nahoc for that...

pdg744 (approved):

PR #2641, opened 2024-12-10
Minor issues found using zeth and better metrics

flaub:
Using `RUST_LOG=info RISC0_INFO=1`, the following will be reported:
```
number of segments: 3
2621440 total cycles
2415056 user cycles (92.13%)
171906 paging cycles (6.56%)
34478 reserved cycles (1.32%)
ecalls
	5252 Sha2 calls, 388648 cycles, (14.83%)
	5 Software calls, 51 cycles, (0.00%)
	0 Input calls, 0 cycles, (0.00%)
	0 BigInt2 calls, 0 cycles, (0.00%)
	0 BigInt calls, 0 cycles, (0.00%)
syscalls
	3 Keccak calls
	1 Read calls
	0 Write calls
	0 VerifyIntegrity calls
	0 ProveKeccak calls
```


risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -220,7 +252,7 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
             }
 
             if let Some(max_cycles) = max_cycles {
-                if self.cycles.user >= max_cycles as usize {
+                if self.cycles.user >= max_cycles {
```
SchmErik, line 255: i noticed that we use usize and u64 for cycle count. might be good to converge on one but definitely something that can be done later

PR #2643, opened 2024-12-10
zkvm: enable configurable po2 for keccak circuit and fix batcher input

SchmErik:


risc0/zkvm/src/guest/env/batcher.rs
```diff
@@ -14,27 +14,48 @@
...
 
 impl Keccak2Batcher {
-    pub const fn init() -> Self {
+    fn max_keccak_inputs(&self) -> usize {
+        let max_keccak_cycles: usize = 1 << self.po2;
+        max_keccak_cycles / KECCAK_PERMUTE_CYCLES
+    }
+
+    pub fn new() -> Self {
+        let po2 = match std::env::var("RISC0_KECCAK_PO2") {
```
flaub, line 42: How will this work in `#![no_std]`?

risc0/zkvm/src/host/client/env.rs
```diff
@@ -175,6 +177,17 @@ impl<'a> ExecutorEnvBuilder<'a> {
             }
         }
 
+        if let Ok(po2) = std::env::var("RISC0_PPROF_OUT") {
```
flaub, line 180: This env var name seems wrong?

risc0/zkvm/src/guest/env/batcher.rs
flaub: We could also call `sys_get_env` directly

risc0/zkvm/src/host/client/env.rs
```diff
@@ -194,6 +207,21 @@ impl<'a> ExecutorEnvBuilder<'a> {
...
+    ///
+    /// Lowering this value will reduce the memory consumption of the prover. Memory consumption is
+    /// roughly linear with the segment size, so lowering this value by 1 will cut memory
+    /// consumpton by about half.
+    ///
+    /// The default value is chosen to be performant on commonly used hardware. Tuning this value,
+    /// either up or down, may result in better proving performance.
+    ///
+    /// Given value must be within [risc0_circuit_keccak::KECCAK_PO2_RANGE].
+    pub fn keccak_po2(&mut self, limit: u32) -> &mut Self {
```
flaub, line 220: This is probably the "max" po2. Seems like we'll need to plumb this thru client/server, but it'd be nice if there was a way to avoid that.

SchmErik: executor env has an env_vars field. we could porbably stuff it in there

risc0/zkvm/methods/guest/src/bin/multi_test.rs
```diff
@@ -521,5 +521,19 @@ fn main() {
...
+                let mut pows = 987654321_u64;
+                for part in state.as_mut_slice() {
+                    *part = pows;
+                    pows = pows.wrapping_mul(123456789);
+                }
+                state
+            }
+            let mut state = test_input();
+
+            env::keccak_update(&mut state);
```
flaub, line 536: I think we should add an assert here that we get the expected output state.

flaub: Wouldn't we want to assert the state after the `keccak_update` call?

PR #2644, opened 2024-12-11
zkvm: ProverServer: expose prove_keccak

SchmErik:


mothran (approved):
Works in my testing

flaub (approved):

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -43,6 +43,11 @@ use crate::{
 /// A ProverServer can execute a given ELF binary and produce a [ProveInfo] which contains a
 /// [Receipt][crate::Receipt] that can be used to verify correct computation.
 pub trait ProverServer {
+    /// Prove the specified keccak request
+    #[cfg(feature = "unstable")]
```
flaub, line 47: Can we use `#[stability::unstable]`?

SchmErik: I was trying to use this a few days ago and ran into some issues. looking at the description https://docs.rs/stability/latest/stability/attr.unstable.html, this attribute changes the visibility from pub crate to pub and since functions in traits are always pub, this doesn't apply. I'm open to other ideas but for now, I'll hit merge for now..

PR #2649, opened 2024-12-13
update RSA patch to tag, use in compat test

austinabell:
Switched rsa tag from specific commit

~~Noticed rsa being used in a compat test, figured I'd change that also (maybe preferred to run software impl, but it is a constrained test vector so maybe fine?)~~ decided to revert this, required too many test/CI changes. Assumed we might do this once it isn't under `unstable`

When doing above, noticed that the keccak impl failed to compile. Assumed this may have been a case where `std` not enabled, in which case the `Vec` usage wouldn't be defined:

```
~/dev/ris/ris/ris/zkv/methods (main|✚4) [101]🌲 cargo c
    Blocking waiting for file lock on package cache
   Compiling risc0-build v1.3.0-alpha.1 (/Users/austinabell/development/risc0/risc0/risc0/build)
    Checking risc0-zkvm v1.3.0-alpha.1 (/Users/austinabell/development/risc0/risc0/risc0/zkvm)
error[E0412]: cannot find type `Vec` in this scope
  --> risc0/zkvm/src/guest/env/batcher.rs:30:13
   |
30 |     inputs: Vec<KeccakState>,
   |             ^^^ not found in this scope
   |
help: consider importing one of these structs
   |
15 + use crate::guest::env::vec::Vec;
   |
15 + use alloc::vec::Vec;
   |

   Compiling risc0-zkvm-methods v1.3.0-alpha.1 (/Users/austinabell/development/risc0/risc0/risc0/zkvm/methods)
```

Happy to revert the other changes and just change the patch tag, but figured I'd note these


austinabell:
Happy to remove and split out the keccak no_std compat change into separate PR, likely that should come with some test/CI change to ensure `no_std` compat with unstable

tzerrell (approved):

flaub (approved):

PR #2651, opened 2024-12-13
Fix typos

VolodymyrBg:
Fixed errors and inaccuracies in the project's textual materials. Improved readability of texts and documentation


pdg744:
Hmm... some of these changes are correcting a typo but others are going to cause confusion. Doesn't seem high-value enough to try to go through and suggest changes on the ones that are in error, so I'm going to close this.

PR #2653, opened 2024-12-13
Add modular arithmetic precompiles

tzerrell:
This adds precompiles for basic modular arithmetic operations (+, -, *, and inverse). Requires https://github.com/risc0/zirgen/pull/125


risc0/bigint2/methods/guest/src/bin/modsub.rs
```diff
@@ -0,0 +1,42 @@
...
+    let lhs = lhs.to_u32_array();
+    let rhs = rhs.to_u32_array();
+    let modulus = modulus.to_u32_array();
+
+    let mut result = [0u32; risc0_bigint2::field::FIELD_256_WIDTH_WORDS];
+    risc0_bigint2::field::modsub_256(&lhs, &rhs, &modulus, &mut result);
+
+    let result = BigUint::from_slice(&result);
+
+    env::commit(&result);
```
austinabell, line 36: Optional: All of these tests could be simplified to not go through `BigUint` and just read arrays of words

risc0/bigint2/src/field/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+    inp: &[u32; FIELD_256_WIDTH_WORDS],
+    modulus: &[u32; FIELD_256_WIDTH_WORDS],
+    result: &mut [u32; FIELD_256_WIDTH_WORDS],
+) {
+    modinv_256_unchecked(&inp, &modulus, result);
+
+    // An honest host will always return a result less than the modulus. A dishonest prover can
+    // sometimes return a result greater than the modulus, so enforce that we're in the honest case.
+    assert!(crate::is_less(&result, &modulus));
+}
```
austinabell, line 125: Thing to note about this is that we don't check that `inp` is not zero, so we are in a similar position to the `ec` things where it would panic in host at runtime, which isn't ideal.

Given that we are passing the result buffer as a param, seems the only way to reasonably have this API if handling this is to return a `Result<(), ()>` (or have the error type be a `&'static str` or something minimal) such that those patterns are nudged towards checking

This is optional, just noting that perhaps we should at least document that behaviour

risc0/bigint2/src/lib.rs
```diff
@@ -17,6 +17,8 @@ pub mod ec;
 #[cfg(feature = "unstable")]
 pub mod ffi;
 #[cfg(feature = "unstable")]
+pub mod field;
```
austinabell, line 20: Thinking out loud that it might be that this crate is getting larger than needed, where it might be ideal before removing `unstable` feature flag, to have these be separated into their own feature flags or separate crates. Nothing to do on this though, and opinionated

austinabell (approved):
Tested `modinv`, `modadd` and `modmul` in patch more extensively

risc0/bigint2/src/field/mod.rs
tzerrell: A bit of trickiness is that "`inp` is nonzero" is not the only thing we need to check here -- this will also fail if `inp` and `modulus` are not relatively prime (in the same way and for the same reason). For instance, someone might set `inp` as `modulus`, and that will make this fail. (And if users set a non-prime `modulus` -- which is not typical but is legal -- then there are even more ways to encounter such a case.)

Therefore, I'm going to just document the behavior instead of adding checks for this.

risc0/bigint2/methods/guest/src/bin/modsub.rs
tzerrell: I think I'll leave that for followup work

flaub (approved):

PR #2656, opened 2024-12-16
Introduce variants of embed_methods that restrict their targets to options passed

wraitii:
The `required_features` support makes it possible to technically do this with pure features, but that seems un-necessarily annoying to me.

If you'd prefer, I can switch the PR around to default to this behaviour (which IMO makes more sense) and just support the cargo.toml metadata for backwards compatibility.


flaub:
There seems to be a trend of needing to use a builder API to make `risc0-build` configuration extensible.

Related:
* #2600
* #2709

flaub:
Now that we've implemented the builder pattern, this feature could be implemented that way.

PR #2660, opened 2024-12-17
keccak: fix range check to include po2 = 18 add `no_mangle` to `risc0_keccak_update`

SchmErik:
The `no_mangle` declaration will allow our patched `tiny-keccak` crate to drop its `risc0` dependency. `keccak_update` sounds too generic and could result in name collisions so I've renamed this to `risc0_keccak_update`.

### DIFF
--- a/risc0/circuit/keccak/src/lib.rs
+++ b/risc0/circuit/keccak/src/lib.rs
@@ -24,7 +24,9 @@ pub use self::control_id::{KECCAK_CONTROL_IDS, KECCAK_CONTROL_ROOT};
 
 pub const KECCAK_DEFAULT_PO2: usize = 17;
 
-pub const KECCAK_PO2_RANGE: core::ops::Range<usize> = 14..18;
+pub const KECCAK_MAX_PO2: usize = 18;
+
+pub const KECCAK_PO2_RANGE: core::ops::Range<usize> = 14..KECCAK_MAX_PO2 + 1;
 
 pub const RECURSION_PO2: usize = 18;
 
--- a/risc0/zkvm/src/guest/env/batcher.rs
+++ b/risc0/zkvm/src/guest/env/batcher.rs
@@ -59,8 +59,9 @@ impl Keccak2Batcher {
             let po2: u32 = po2.parse::<u32>().unwrap();
             if !KECCAK_PO2_RANGE.contains(&(po2 as usize)) {
                 panic!(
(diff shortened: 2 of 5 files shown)


SchmErik:
I'm going to drop the risc0_keccak_update changes for now and focus this one on enabling po2 = 18..

risc0/circuit/keccak/src/lib.rs
```diff
@@ -24,7 +24,9 @@ pub use self::control_id::{KECCAK_CONTROL_IDS, KECCAK_CONTROL_ROOT};
 
 pub const KECCAK_DEFAULT_PO2: usize = 17;
 
-pub const KECCAK_PO2_RANGE: core::ops::Range<usize> = 14..18;
+pub const KECCAK_MAX_PO2: usize = 18;
+
+pub const KECCAK_PO2_RANGE: core::ops::Range<usize> = 14..KECCAK_MAX_PO2 + 1;
```
flaub, line 29:
```suggestion
pub const KECCAK_PO2_RANGE: core::ops::Range<usize> = 14..19;
```

risc0/zkvm/src/guest/env/batcher.rs
```diff
@@ -59,8 +59,9 @@ impl Keccak2Batcher {
             let po2: u32 = po2.parse::<u32>().unwrap();
             if !KECCAK_PO2_RANGE.contains(&(po2 as usize)) {
                 panic!(
-                    "invalid keccak po2 {po2}. Expected range: {:?}",
-                    KECCAK_PO2_RANGE
+                    "invalid keccak po2 {po2}. Expected range: {} - {}",
+                    KECCAK_PO2_RANGE.start,
+                    KECCAK_PO2_RANGE.end - 1
```
flaub, line 64:
```suggestion
                    "invalid keccak po2 {po2}. Expected range: {:?}",
                    KECCAK_PO2_RANGE
```

If we use `RangeInclusive` does that make this error message sane?

SchmErik: I got this:
```
called `Result::unwrap()` on an `Err` value: Guest panicked: invalid keccak po2 19. Expected range: 14..=18
```
I think it looks cleaner if I have the range (but drop the -1 thing). I changed it to look like this now:
```
called `Result::unwrap()` on an `Err` value: Guest panicked: invalid keccak po2 19. Expected range: 14 - 18 (inclusive)
```

flaub (approved):

PR #2669, opened 2024-12-18
add keccak example and change devmode for receipts with assumptions

SchmErik:
In order to add a keccak example, this PR does the following:
- add `no_mangle` to risc0_keccak_update. This allows the patched `tiny-keccak` to drop its `risc0-zkvm` dependency.
- add an example that runs keccak. This uses the patched tiny-keccak crate and is similar to the sha example.
- for dev-mode, add a function for the host to compute claim digests for a given guest input.


SchmErik:
@nategraf I added you as a reviewer because I was wondering if this approach to dev-mode seems reasonable. As of now, in devmode, the claims are matched during verification but in dev mode we're technically not producing valid receipts so would it be sufficient to use verify_integrity and skip the claim match?

nategraf:
I definitely suggest we check the claim. This allows higher degrees of confidence that when something works in dev mode, it'll work when actually proving. I'd be happy to help with any challenges around making this work.

flaub:
Agreed that we shouldn't skip verifying claims since those can be done in devmode without real receipts.

SchmErik:
Thanks for the comments! I'll revert the devmode change and look at what's going on with the claims

SchmErik:
@flaub and @nategraf this is ready for review. I wrote a function that computes the claim digest based off of the input and called that during dev-mode proving. Once the claim is computed, it's included as the receipt's session_claim

flaub (approved):
LGTM, but there's a conflict in `docker.rs`

examples/keccak/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "keccak"
```
nategraf, line 2: Cargo gets unhappy when there are multiple crates with the same name. `keccak` is crate on `crates.io`, so having this something unique would be a good.
```suggestion
name = "risc0-keccak-example"
```

risc0/circuit/keccak/src/lib.rs
```diff
@@ -28,9 +28,61 @@ pub const KECCAK_PO2_RANGE: core::ops::RangeInclusive<usize> = 14..=18;
...
     &KECCAK_CONTROL_IDS[po2 - KECCAK_PO2_RANGE.min().unwrap()]
 }
+
+pub fn max_keccak_inputs(po2: u32) -> usize {
+    let max_keccak_cycles: usize = 1 << po2;
+    max_keccak_cycles / KECCAK_PERMUTE_CYCLES
+}
+
+#[cfg(feature = "prove")]
+pub fn compute_keccak_digest(input: &[u8]) -> Digest {
```
nategraf, line 46: This is kind of a strangely factored function. Not saying we need to change it, but at least a comment would be helpful because I think the name and signature is misleading.

```suggestion
/// Given a slice of `KeccakState`, encoded as `[u8]`, produce the SHA-256 digest matching what is produced by the keccak circuit.
#[cfg(feature = "prove")]
pub fn compute_keccak_digest(input: &[u8]) -> Digest {
```

risc0/zkvm/src/guest/env/mod.rs
```diff
@@ -500,6 +500,7 @@ pub fn read_buffered<T: DeserializeOwned>() -> Result<T, crate::serde::Error> {
 
 /// get an updated keccak state
 #[cfg(feature = "unstable")]
-pub fn keccak_update(state: &mut risc0_circuit_keccak::KeccakState) {
+#[no_mangle]
+pub fn risc0_keccak_update(state: &mut risc0_circuit_keccak::KeccakState) {
```
nategraf, line 504: Since misuse of this function could cause undefined behavior, it should be marked as unsafe. A function that uses `unsafe` internally should only be marked as safe (i.e. not marked `unsafe`) when it internally ensures that memory safety is maintained. (e.g. a deserializer may chose to use `unsafe` internally for performance, but makes sure to check buffer lengths and pointer to ensure safety for anyone using its public API).

If we are making an argument that this is safe because it can only be used inside the zkVM, and this is safe because the zkVM is single-threaded and non-preemptive, then we might want to note this.

```suggestion
pub unsafe fn risc0_keccak_update(state: &mut risc0_circuit_keccak::KeccakState) {
```

nategraf (approved):

risc0/zkvm/src/guest/env/mod.rs
flaub: I'm not sure we should be using `unsafe` in this manner. Is there any way a user could call this function and wind up with undefined behavior?

nategraf: If the zkVM makes provides as a contract that it will not allow preemptive or concurrent execution, and this function is only allowed in the zkVM context, then I think we can argue it is safe and mark it as such. I think that may be the case here, but we don't provide any explanation for why this function, at an API level is safe while using `unsafe` code internally.

flaub: I think for every `unsafe` we should have a `* SAFETY` section.

PR #2670, opened 2024-12-18
ZKVM-1058: Discuss selecting patched backend

tzerrell:


website/api_versioned_docs/version-0.21/zkvm/precompiles.md
```diff
@@ -42,6 +42,16 @@ can be in the [ECDSA example][ecdsa]. Note the [use of the patched
 versions][ecdsa-patched] of `sha2`, `crypto-bigint` and `k256` crates used in
 the guest's `Cargo.toml`.
 
+Some crates will allow you a choice of different backend crates, and you may be
```
pdg744, line 45: hmm i think you meant to make this change in `api_versioned_docs/version-1.2`

pdg744 (approved):

website/api_versioned_docs/version-0.21/zkvm/precompiles.md
tzerrell: Whoops! Good call

SchmErik:
@tzerrell 1.3 is emminent so could you drop the change for 0.21. we can merge this and then it'll be staged for when we do 1.3

tzerrell:
> @tzerrell 1.3 is emminent so could you drop the change for 0.21. we can merge this and then it'll be staged for when we do 1.3

Thanks for the ping, I'd lost track of this PR; updating with your suggestion

PR #2674, opened 2024-12-19
ZKVM-936: rzup: improve rzup installer script

hmrtn:
the installer script has been restructured for better maintainability, testing, and ux. Missing version tracking, error handling, and debug-ability prompted this refactor - a good opportunity to tidy it up.

Changes include:
- clearer script structure 
- added version flag (-v) to display installer version
- added debug mode (-d flag) for better troubleshooting (replace verbose flag)
- colored output for better visual feedback -- no more emojis, just ascii.
- proper cleanup with trap handlers

This install script has been tested in a variety of docker containers to ensure non-interactive compatibility.


hmrtn:
Can be tested with: 
```sh
curl -L https://raw.githubusercontent.com/risc0/risc0/refs/heads/hmrtn/rzup-install-script-1.0.0/rzup/install | bash
```

SchmErik:
> do we have CI/CD setup for the rzup install script?
yes, will follow up with you

rzup/install
```diff
@@ -15,218 +16,227 @@
 
 set -euo pipefail
```
hmrtn, line 17: script will fail if:
- any command fails (-e)
- any variable is undefined (-u) ()
- any part of a pipeline fails (-o pipefail)

0xkr8os:
this would be great to have. currently can't install rzup on buildjet without a workaround

flaub (approved):

PR #2677, opened 2024-12-19
Add used and free methods to bump allocator module

nategraf:
This PR adds `used` and `free` methods to the bump allocator module to allow for collecting memory metrics. It also moves the logic of the bump allocator into the `heap::bump` module, ensuring that it cannot be used if it is not enabled at compile time (which would cause undefined behavior).

Also, this PR adds the `unsafe` specifier to `heap::embedded::init`. This should be specified because calling this function can result in breaking Rust's memory guarentees. This is a breaking change, however any user code that calls this function (if any exists) is likely unsafe/broken right now, since init is called as part of the guest entrypoint and should not be called again.

### DIFF
--- a/risc0/zkvm/platform/src/heap/bump.rs
+++ b/risc0/zkvm/platform/src/heap/bump.rs
@@ -96,7 +96,7 @@ pub(crate) unsafe fn alloc_aligned(bytes: usize, align: usize) -> *mut u8 {
 
 /// Initialize the bump allocator with the memory allocations defined in the [memory][crate::memory] module.
 ///
-/// # Saftey
+/// # Safety
 ///
 /// This function must be called exactly once.
 pub unsafe fn init() {
--- a/risc0/zkvm/platform/src/heap/bump.rs
+++ b/risc0/zkvm/platform/src/heap/bump.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::syscall::sys_alloc_aligned;
+use crate::{memory::GUEST_MAX_MEM, syscall::sys_panic, WORD_SIZE};
 use core::alloc::{GlobalAlloc, Layout};
 
 #[global_allocator]
@@ -28,19 +28,19 @@ extern "C" {
     static _end: u8;
 }
 
-static HEAP_START: usize = unsafe { (&_end) as *const u8 as usize };
+static mut HEAP_START: usize = 0;
 
-/// Pointer to next heap address to use, initialized as the address of the _end symbol.
+/// Pointer to next heap address to use, initialized by the [init] function to the _end symbol.
 ///
 /// The heap grows into higher addresses from it's starting position. Since this is a bump
 /// allocator and memory is never freed, this address strictly increases over time.
-pub(crate) static mut HEAP_POS: usize = HEAP_START;
+static mut HEAP_POS: usize = 0;
 
 pub struct BumpPointerAlloc;
 
 unsafe impl GlobalAlloc for BumpPointerAlloc {
     unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
-        sys_alloc_aligned(layout.size(), layout.align())
+        alloc_aligned(layout.size(), layout.align())
     }
 
     unsafe fn dealloc(&self, _: *mut u8, _: Layout) {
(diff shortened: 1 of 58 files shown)


risc0/zkvm/platform/src/heap/bump.rs
```diff
@@ -35,3 +53,58 @@ unsafe impl GlobalAlloc for BumpPointerAlloc {
...
+    unsafe { HEAP_POS = heap_pos };
+    ptr
+}
+
+/// Initialize the bump allocator with the memory allocations defined in the [memory][crate::memory] module.
+///
+/// # Saftey
+///
+/// This function must be called exactly once.
+pub unsafe fn init() {
```
austinabell, line 102: Perhaps this should be under an underscored name, like `__init()` and documented that it's called automatically by risc0 boilerplate, given that this `init` call was accidentally called in practice for the embedded alloc?

risc0/zkvm/platform/src/heap/embedded.rs
```diff
@@ -31,7 +31,12 @@ unsafe impl critical_section::Impl for CriticalSection {
     }
 }
 
-pub fn init() {
+/// Initialize the embedded-alloc allocator with the memory allocations defined in the [memory][crate::memory] module.
+///
+/// # Saftey
```
austinabell, line 36:
```suggestion
/// # Safety
```

risc0/zkvm/platform/src/heap/bump.rs
```diff
@@ -35,3 +53,58 @@ unsafe impl GlobalAlloc for BumpPointerAlloc {
...
+    }
+
+    // SAFETY: Single threaded, and non-premptive so modification is safe.
+    unsafe { HEAP_POS = heap_pos };
+    ptr
+}
+
+/// Initialize the bump allocator with the memory allocations defined in the [memory][crate::memory] module.
+///
+/// # Saftey
```
austinabell, line 99:
```suggestion
/// # Safety
```

risc0/zkvm/platform/src/heap/mod.rs
```diff
@@ -13,7 +13,29 @@
 // limitations under the License.
 
 #[cfg(not(feature = "heap-embedded-alloc"))]
-mod bump;
+pub mod bump;
 
 #[cfg(feature = "heap-embedded-alloc")]
 pub mod embedded;
```
austinabell, line 19: Wondering why these are exposed at all, seems like a footgun and more API to maintain? Assuming `bump` needs to be exposed for the `init`, but just noting the expansion of the API (and breaking changes for a guest if you enable/disable the embedded alloc)

risc0/zkvm/platform/src/heap/bump.rs
austinabell: Also curious why this doesn't use something like `OnceCell` as a lazy static for the heap meta-data of each. Assume it might just be for code simplicity and avoiding adding more types to the core API?

nategraf: The __ methods are generally symbols that end up (unmangled) in the ELF. I don't think that helps so much with the risk of someone unwittingly calling this function. I do think the `unsafe` modifier is important here, and communicates the risks to a large extent.

risc0/zkvm/platform/src/heap/mod.rs
nategraf: Ah! Because I factored it one way, then changed it. Keeping it private is almost certainly better.
886d85bc32a66eb891f6a11e6261b1c341d8b6c4

risc0/zkvm/platform/src/heap/bump.rs
nategraf: > Also curious why this doesn't use something like OnceCell as a lazy static for the heap meta-data of each. Assume it might just be for code simplicity and avoiding adding more types to the core API?

We could do that. I don't really have a strong intuition either way.

austinabell: > We could do that. I don't really have a strong intuition either way.

Just suggesting in an attempt to make the APIs less prone to misuse -- I don't feel strongly about it.

flaub: OnceCell seems nice

flaub (approved):

risc0/zkvm/platform/src/heap/bump.rs
austinabell: OnceCell might also be an unnecessary complication, as it would still require unsafe code to mutate the value. Avoiding init from being called twice could just be as simple as checking if `HEAP_START != 0` in init before modifying to achieve basically the same thing

PR #2680, opened 2024-12-20
ZKVM-1008: document timing attack precompile disclaimer

austinabell:


website/api/zkvm/precompiles.md
```diff
@@ -101,6 +101,12 @@ versions `>=1.2.0` of `risc0` crates. For users who need a stable, production-re
 working on stablizing these precompiles as soon as possible, and the `"unstable"` feature flag will
 no longer be required.
 
+## Timing Attacks
+
+These precompiles do not currently provide strict guarantees about constant-time execution and
+proving time. Be very careful if using any of these precompiles with private data, such as signing
+a message within the zkvm where an observer can measure the proving time.
```
flaub, line 108: I'm thinking that wallclock time won't be the way that timing attacks manifest, but instead it would be based on cycle counts. The side channel would be how many cycles one operation does relative to others.

nategraf: This is also a consideration, although after compression the cycle counts are no longer visible except through the proving time.

tzerrell: Cycle counts could also come up indirectly in situations where you had a cycle limit -- you could be right at the boundary of the cycle limit and see whether the proof went through or not. (I think just mentioning cycles is probably adequate to address this case).

tzerrell (approved):

website/api/zkvm/precompiles.md
austinabell: Added the statement just to be careful to note, as it may be possible there is an edge case where this could be observed. I'd definitely like to keep the wall clock timing wording in, given having const time guarantees for the executor and prover for the bigint2 precompiles is something I assume we don't have (correct me if I'm wrong)

pdg744 (approved):

PR #2682, opened 2024-12-21
ZKVM-1002: drop accumulation of input keccak states in the guest and remove po2

SchmErik:


risc0/zkvm/src/host/server/exec/syscall/keccak.rs
```diff
@@ -12,31 +12,128 @@
...
+        }
+    }
+}
+
+impl SysKeccak {
+    pub fn new(max_po2: Option<u32>) -> Self {
+        let max_po2 = match max_po2 {
+            Some(po2) => po2 as usize,
+            None => KECCAK_DEFAULT_PO2,
+        };
```
flaub, line 57:
```suggestion
        let max_po2 = max_po2.unwrap_or(KECCAK_DEFAULT_PO2);
```

flaub (approved):

PR #2696, opened 2024-12-26
Including pager cycles in user_cycles

ghostant-1017:
I have observed that in my project
user cycles: 6177246
total_cycles: 9699328
cycle efficiency: 63%
insn_cycles: 6177246
pager_cycles: 3405386

User cycles seems to count only the part of insn_cycles, not the part that contains pager_cycles.
I'm not sure if this is a Bug, or is it designed that way?


ghostant-1017:
Can someone handle this pr?

flaub:
`user_cycles` is intended to represent the cycles that the user requested via their program. The `paging_cycles` are considered to be a system cost of a particular rv32im circuit implementation. It's not likely that the user will be able to influence the paging cycles very much. It's also the case that the circuit implementation may change so that paging costs are less.

PR #2704, opened 2025-01-04
ZKVM-894: rv32im_v2: executor tests are passing

flaub:
Using `rstest` to create side-by-side testing of `rv32im-v1` and `rv32im-v2` executors.

Notes:
* Image IDs are a bit different in v2. This is because there is now a user-mode and kernel-mode portion of the initial memory image. At build time, the `user_id` or `kernel_id` is computed (based on a `package.metadata.risc0.kernel` config setting). At runtime, when the user goes to run an image, they can compute the full `image_id` by combining these two IDs.
* `risc0-build` has been bumped to `2.0.0` to deal with these new changes.
* A `v1compat` kernel is added that provides emulation for the v1 host/guest syscall interface. This kernel needs to be built and used by `risc0-zkvm`, so a new `cargo risczero bake` command is added to support this. The `bake` command will build a guest binary ELF and compute the user or kernel ID, then output the result into a source directory.
* The v2 circuit now includes the `pc` as part of the memory image. However, `ReceiptClaim` uses `SystemState`, and so currently the `pc` is always set to `0` for compatibility. In the future we can see about changing this.


risc0/zkos/v1compat/src/kernel.s
```diff
@@ -0,0 +1,152 @@
...
+    # check that ecall request is within range
+    bge a0, t5, 1f
+    # adjust index so that it points to word sized entries
+    slli a0, a0, 2
+    # compute the table entry
+    add a1, t4, a0
+    # jump into dispatch table
+    jr a1
+1:
+    fence # panic
```
SchmErik, line 89: trying to check my understanding here: this an atomic fence and is intended cause an illegal instruction error if we get to this point?

risc0/circuit/rv32im-v2/src/execute/sha2.rs
```diff
@@ -0,0 +1,243 @@
...
+        let e = ctx.load_u32(sha2.state_in_addr + 7u32 - i)?;
+        sha2.a = a.to_be();
+        sha2.e = e.to_be();
+        old_a.push(sha2.a);
+        old_e.push(sha2.e);
+        ctx.store_u32(sha2.state_out_addr + 3u32 - i, a)?;
+        ctx.store_u32(sha2.state_out_addr + 7u32 - i, e)?;
+    }
+
+    // HERE!
```
SchmErik, line 101:
```suggestion

```
I think we can remove this comment

nategraf:
Probably the biggest question in my mind is how to handle the binding of the kernel ID. Logically the verifier must chose or validate the kernel, as an arbitrary kernel chosen by the prover would not provide any soundness. There are a number of direct and indirect approaches to doing so.

A verifier might have one kernel it accepts, and this could be the default kernel provided as part of a given zkVM version. In this case, the verifier would manage the image ID essentially the same way it does now: it would bind the image ID in it's verifier program (e.g. a Rust program or smart contract) and would provide the full image to the prover (e.g. local, Boundless, or Bonsai) which the prover would execute.

And incremental change over this would be to have the verifier use the full image ID, but the client that requests the proof provide the user image (i.e. the lower half of memory) and the kernel ID, assuming the prover already has the kernel it needs on-hand.

Another incremental change is that if the verifier has a list of kernels that are acceptable, then it will include two pieces of information in the verifier program: a user ID, and a list of kernel IDs (or a commitment to a list of kernel IDs, although this seems like more trouble than its worth). During verification, the verifier would receive the proof and kernel ID from the prover, it would verify the kernel ID is in the list of allowed kernels, then hash it together with the user ID to form the image ID and run verify.

Sometimes there is more than one verifier for a given proof, and they may have distinct requirements on the which kernel(s) are acceptable. If parties A and B, both verifiers, have a protocol where A chooses the user ID and B constrains the kernel ID, there are a couple of different approaches here. One key approach is to let A still chose the image ID, but require that they at some point, either during registration or during each interaction, prove to B that the correct kernel is included in that image ID. This is easy to do, as A simply needs to give B the user ID along with the image ID and B can hash together the user ID and kernel ID to form the image ID. There is the complication, for EVM verifiers, that the hash function used for memory images is Poseidon2. We should try implementing our Poseidon2 in Solidity (or EVM assembly more directly) and see what the gas costs look like. If it is unacceptable, then there may be opportunities to off-load this into the zkVM, at a higher cost in complexity.

> The v2 circuit now includes the pc as part of the memory image. However, ReceiptClaim uses SystemState, and so currently the pc is always set to 0 for compatibility. In the future we can see about changing this.

What would happen if the machine ran with `pc = 0`? IIUC, the lowest page in memory is set to all zeroes. Is an instruction of all zeroes valid in the RISC-V ISA? One of my thoughts here is it would be a somewhat nice assurance if the deprecated `pc` is set to a value that will result in a fault on both v1 and v2 circuit if it's ever mistakenly interpreted as a PC.

PR #2708, opened 2025-01-07
Bug Fix: snarkjs public signals length

andrewcoder666:
There are only 5 output signals ( public witness ) passed to the Groth16 verifier, therefore only first 5 elements should be checked before ecPairing check. The line deleted is meant to check the non-existing 6th element of the array parameter, which obviously is to access to the uninitialized memory since the array itself only contains 5 elements.

Same issue at [risc0-ethereum](https://github.com/risc0/risc0-ethereum/pull/333)


capossele (commented):
It makes sense to me. I believe this behaviour was due to a bug in `snarkjs` (later fixed by https://github.com/iden3/snarkjs/pull/518) where `IC.length` was used instead of the amount of public signals. Should be fixed on `risc0-ethereum` though as I don't recall we use this file somewhere. Also might be worth checking if we need to bump the snarkjs version we use within the docker image so that this issue does not show up again (we use snarkjs to auto-generate this file)

capossele:
Can confirm that `snarkjs` v0.7.5 (latest version) generates the verifier contract with the same fix. I'll recommend updating its version also here https://github.com/risc0/risc0/blob/7e643f767ad9c55778a81c32d599637cb4db3a5c/groth16_proof/docker/ceremony.Dockerfile#L37

nategraf:
Ok, ya sorry for being slow on this one. I've wanted to take the time to double check this, even though it does seem pretty clear reading the code. Looking at it again now, it does indeed look correct and like a good bug fix to me. I think with @capossele having checked this matches the 0.7.5 behavior, the upstream PR, and reviewing it myself we can go ahead and merge this PR and it's related PR on risc0-ethereum. Thank you @andrewcoder666 for submitting this improvement 🙏

nategraf (approved):

nategraf (approved):

PR #2709, opened 2025-01-07
risc0-build: Allow to pass custom env variables to build container

Wiezzel:
Closes ZKVM-970


flaub:
These are great features to add, however they are done in a way that breaks compatibility with 1.0 semver of the `risc-build` crate. I have an upcoming PR that should land soon which 1) bumps the major version to 2.0.0 and 2) makes some breaking changes. I'd like to get this PR in first but then I think we can land this ontop.

https://github.com/risc0/risc0/pull/2704

Wiezzel:
@flaub What do you think about introducing a builder for `DockerOptions` and `GuestOptions` and making the default constructor private so that new options can be added in the future without breaking compatibility?

flaub:
Could you try and add a builder interface? Since we've bumped the major version, now is the time to make this change before a publish the crate.

Wiezzel:
@flaub OK, will do.

Wiezzel:
@flaub Done in 978b41968ff1459baa641ec1fae24012484b653e

flaub:
Thanks for this! I've continued this on a new PR that fixed a few issues I found along the way.

#2762

PR #2712, opened 2025-01-08
ZKVM-997: [rzup] update to version 0.3

hmrtn:
This update aims to create a more robust version management system for RISC Zero components. The new implementation provides improvements over the old version:
- Support for multiple versions
- Better terminal UI
- Less confusing CLI interface

This changes `rzup` from a standalone binary into a library with a CLI frontend. This change allows other tools to find RISC Zero components by using `rzup` as a library.

#### Change Overview
1. Component Management System
   - added modular component system for managing different RISC Zero components
   - support for components: rust, cpp, cargo-risczero, and r0vm
   - consistent versioning across components using semver

2. Installation & Version Management
   - improved version detection and management
   - support for multiple versions of each component
   - atomic installations with cleanup on failure

3. Enhanced CLI Experience
   - suggestions and tips
   - less ambiguous usage pattern
   - more descriptive errors 
   - progress indicators and colorized output
   - Quiet and verbose mode

4. Settings Management
   - persistent settings (`settings.toml`)
   - per-component version management (with `rzup default ...`)

#### Migration (from `rzup` < 0.3)
- Existing component installations and `.risc0/` structure is preserved and backward compatible 
- core cli usage remains largely unchanged (remove confusing extension/toolchain usage)

#### Example usage

As a user:
```
rzup v0.3.0-alpha.1

    ██████  █▀▀▀██
    ██████  █ ▄▀ █
    ██████  ██▄▄▄█
    ██████
    ██████   RISC
    ██████   ZERO

Usage: rzup [OPTIONS] <COMMAND>

Commands:
  install    Install and update components
  check      Check for new component versions
  default    Set a component version as default
  show       Show installed components
  uninstall  Uninstall a component
  build      Build a component
  help       Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose  Enable verbose output
  -q, --quiet    Suppress output
  -h, --help     Print help
  -V, --version  Print version

Discussion:
    Installs and manages software components distributed by Risc Zero.

    Installed components are stored in the path specified by the `RISC0_HOME`
    environment variable, or in `~/.risc0` if not set.

    Rust software components symlink their binaries into ~/.cargo/bin/ when
    made default.
```

```bash
# Install latest versions
rzup install

# Update components to latest versions (note: this is an alias to install)
rzup update

# Install specific version of cargo-risczero
rzup install cargo-risczero 1.0.0

# Install specific version of rust toolchain
rzup install rust 1.81.0

# Check for updates
rzup check

# Show installed versions
rzup show

# Set a specific version of a component as default
rzup default cargo-risczero 1.0.0

# Uninstall a component
rzup uninstall rust 1.79.0
```

As a library:
```rust 
use rzup::{Rzup, Result};
use semver::Version;

fn main() -> Result<()> {
    let mut rzup = Rzup::new()?;
    
    // Get active version
    if let Some((version, path)) = rzup.get_default_version("cargo-risczero")? {
        println!("Default version: {} at {}", version, path.display());
    }
    
    Ok(())
}
```

PR for integration with risc0 crates https://github.com/risc0/risc0/pull/2752


rzup/src/lib.rs
```diff
@@ -11,64 +11,378 @@
...
+        // Test virtual component (r0vm is inside cargo-risczero's bin directory)
+        let virtual_component = "r0vm";
+        let virtual_bin_path = rzup.get_bin_path(virtual_component, &version).unwrap();
+        assert!(virtual_bin_path.is_some());
+        let virtual_bin_path = virtual_bin_path.unwrap();
+        assert!(virtual_bin_path.ends_with(format!("bin/{}", virtual_component)));
+    }
+
+    #[test]
+    #[ignore = "requires GitHub API access"]
```
bobbobbio, line 352: My initial reaction to tests like this are to wonder if there is a way we can run them without doing the remote network access piece.

In the past I've gotten around this in different ways,
one way is to allow passing in the base URLs that are being accessed and in your tests you run a little HTTP server and pass in that address, another way is just dependency injecting the part that does the HTTP requests. What are your thoughts on these sorts of approaces?

rzup/src/lib.rs
```diff
@@ -11,64 +11,378 @@
...
-            .join(".cargo/bin");
+    /// Installs a specific component version.
+    ///
+    /// # Arguments
+    /// * `component` - Component identifier (e.g. "rust", "cargo-risczero")
+    /// * `version` - Specific version to install, or None for latest
+    /// * `force` - If true, reinstalls even if already installed
+    pub fn install_component(
+        &mut self,
+        component: &str,
```
bobbobbio, line 120: Have you considered making this be an `enum` instead of a string? Both here and in other places in the library (the HashMaps could be EnumMap or otherwise). At least from an API standpoint it might make a little nicer to use (easier to discover the valid components, and let you detect when you have chosen an invalid component at build time vs run time).

rzup/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl From<std::io::Error> for RzupError {
+    fn from(e: std::io::Error) -> Self {
+        RzupError::Io(e.to_string())
+    }
+}
+
+impl From<anyhow::Error> for RzupError {
+    fn from(e: anyhow::Error) -> Self {
+        RzupError::Other(e.to_string())
```
bobbobbio, line 48: Maybe it would be worth trying to downcast this error to see if it is an `std::io::Error` and using `RzUp::Io` variant in that case?

rzup/src/lib.rs
```diff
@@ -11,64 +11,378 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-
+#[cfg(feature = "cli")]
 pub mod cli;
```
bobbobbio, line 15: Why have this `cli` module in the library at all? Could it just go in the binary part?

rzup/src/cli/commands.rs
```diff
@@ -0,0 +1,300 @@
...
+        if rzup.version_exists(&self.name, &version)? {
+            rzup.set_active_version(&self.name, version.clone())?;
+            println!(
+                "Successfully set {} version {} as active",
+                self.name, version
+            );
+        } else {
+            println!(
+                "! Version {} of {} is not installed.\n  Please use 'rzup install {} {}' to install",
+                version, self.name, self.name, version
```
bobbobbio, line 225: Here and other places you can also write this sort of thing like this:
```
println!(
    "! Version {version} of {name} is not installed.\n  Please use 'rzup install {name} {version}' to install",
    name=self.name
);
```

bobbobbio: If the format string is a little long it can be a littler easier to understand whats being printed if you do it this way

rzup/src/cli/commands.rs
```diff
@@ -0,0 +1,300 @@
...
+    #[arg(value_parser=component_parser())]
+    name: String,
+    /// Version of the component to uninstall
+    version: String,
+}
+
+impl UninstallCommand {
+    pub(crate) fn execute(&self, rzup: &mut Rzup) -> Result<()> {
+        let version = Version::parse(&self.version)
+            .map_err(|_| RzupError::InvalidVersion(self.version.clone()))?;
```
bobbobbio, line 297: could this maybe instead an implementation of `From<semver::Version> for RzupError`?

rzup/src/cli/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    format!(
+        r#"
+rzup v{}
+
+   {}"#,
+        version,
+        r#"RISC
+   ZERO"#
+            .bold()
+    )
```
bobbobbio, line 124: there is this really great crate called `indoc` that can really help with this sort of thing, it will dedent string literals so you don't have to make it look a little weird like this.

rzup/src/settings.rs
```diff
@@ -0,0 +1,121 @@
...
+                RzupError::Environment(format!("Failed to create settings directory: {}", e))
+            })?;
+        }
+
+        env.emit(RzupEvent::Debug {
+            message: format!("Saving settings to {}", env.settings_path().display()),
+        });
+
+        let contents = toml::to_string_pretty(self)
+            .map_err(|e| RzupError::Environment(format!("Failed to serialize settings: {}", e)))?;
```
bobbobbio, line 65: I believe that the only way this can fail is due to a programming mistake and asserts seem appropriate to me in that situation.

rzup/src/distribution/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+            .join(format!("{}.lock", archive_name.display()));
+
+        // create and lock the file
+        let lock_file = std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .open(&lock_path)?;
+
+        match lock_file.try_lock_exclusive() {
```
bobbobbio, line 110: if you did `map_err(..)?` here instead it would let you dedent the stuff in the `Ok` branch

rzup/src/lib.rs
```diff
@@ -11,64 +11,378 @@
...
+    /// * `version` - Version to uninstall
+    pub fn uninstall_component(&mut self, component: &str, version: Version) -> Result<()> {
+        self.registry
+            .uninstall_component(&self.environment, component, version)?;
+        self.registry.scan_environment(&self.environment)?;
+        Ok(())
+    }
+
+    /// Lists all installed versions of a component.
+    ///
```
bobbobbio, line 143: Maybe it would be good to mention in the doc-comment that the return listed is sorted from newest to oldest

rzup/src/paths.rs
```diff
@@ -0,0 +1,280 @@
...
+                        parts[0].parse().unwrap_or(0),
+                        parts[1].parse().unwrap_or(0),
+                        parts[2].parse().unwrap_or(0),
+                    ))
+                } else {
+                    None
+                }
+            })
+        } else if dir_name.starts_with('v') {
+            if let Some(version_part) = dir_name.strip_prefix('v') {
```
bobbobbio, line 167: rather than `if let Some(_)` here you can use `?`

rzup/src/paths.rs
```diff
@@ -0,0 +1,280 @@
...
+                let parts: Vec<_> = v.split('.').collect();
+                if parts.len() == 3 {
+                    Some(Version::new(
+                        parts[0].parse().unwrap_or(0),
+                        parts[1].parse().unwrap_or(0),
+                        parts[2].parse().unwrap_or(0),
+                    ))
+                } else {
+                    None
+                }
```
bobbobbio, line 164: another way is 
```
(parts.len() == 3).then(|| Version::new(...))
```

rzup/src/paths.rs
```diff
@@ -0,0 +1,280 @@
...
+
+        let result = if dir_name.starts_with("r0.") && component_id == "rust" {
+            // Handle legacy rust format
+            dir_name
+                .strip_prefix("r0.")
+                .and_then(|v| v.split('-').next())
+                .and_then(|v| Version::parse(v).ok())
+        } else if component_id == "cpp" && !dir_name.starts_with('v') {
+            // Handle legacy cpp format (YYYY.MM.DD-risc0-cpp-...)
+            dir_name.split('-').next().and_then(|v| {
```
bobbobbio, line 154: you can use `?` here instead of `and_then`

bobbobbio (commented):
I left some comments and questions. I sincerely hope the comments come across as helpful and not annoying. :)

rzup/src/settings.rs
hmrtn: Good catch - there are a ton of infallible operations that can be refactored.

rzup/src/lib.rs
```diff
@@ -11,64 +11,378 @@
...
+    }
+
+    /// Ensures version directories exist for a component.
+    ///
+    /// Creates the necessary directory structure if it doesn't exist.
+    ///
+    /// # Arguments
+    /// * `component_id` - Component identifier
+    /// * `version` - Version to create directories for
+    pub fn ensure_version_dirs(&self, component_id: &str, version: &Version) -> Result<()> {
```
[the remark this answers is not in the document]
hmrtn: Yeah, you're right - this should be removed. The proper directory creation is already handled by `Paths::create_version_dirs()`.

rzup/src/components.rs
```diff
@@ -0,0 +1,284 @@
...
+        Component::CargoRiscZero => symlink(
+            &version_dir.join("cargo-risczero"),
+            &env.cargo_bin_dir().join("cargo-risczero"),
+        )?,
+        Component::R0Vm => symlink(&version_dir.join("r0vm"), &env.cargo_bin_dir().join("r0vm"))?,
+        Component::RustToolchain => {
+            symlink(&version_dir, &env.rustup_toolchain_dir().join("risc0"))?
+        }
+        Component::CppToolchain => symlink(
+            &version_dir.join("riscv32im-linux-x86_64"),
```
hmrtn, line 194: The hardcoded path `riscv32im-linux-x86_64` doesn't match the platform-specific paths we create, causing this symlink to fail on non-x86 systems

rzup/src/distribution/http.rs
```diff
@@ -0,0 +1,79 @@
...
+    response.json().map_err(|e| RzupError::Other(e.to_string()))
+}
+
+pub fn download_text(url: impl IntoUrl, bearer_token: &Option<String>) -> Result<String> {
+    let response = http_client_get(url, bearer_token)?;
+    let response = error_on_status(response)?;
+    response.text().map_err(|e| RzupError::Other(e.to_string()))
+}
+
+pub fn download_bytes(
```
hmrtn, line 72: nice! this is the correct approach to streaming without `downloader` crate - `top` logs indicate peak memory usage at about 16M

rzup/src/components.rs
bobbobbio: oh yeah, thats a good catch. I need to fill out test coverage for lots of things with a different platform I think

rzup/Cargo.toml
```diff
@@ -10,25 +10,37 @@ repository = { workspace = true }
...
 
+[[bin]]
+name = "rzup"
+path = "src/bin/rzup.rs"
+
+[features]
+default = ["cli", "install"]
+cli = ["dep:clap", "dep:indicatif", "dep:colored", "dep:is-terminal"]
+install = ["dep:reqwest", "dep:tar", "dep:xz", "dep:flate2", "dep:fs2"]
+
```
flaub, line 21: What is the intention with these feature flags?

flaub:
I suggest that we stop making changes to this PR until #2752 merges into this one.

rzup/Cargo.toml
hmrtn: They are intended to support both library usage where installation and cli features aren't necessary - such as in #2752, and the rzup binary. 

We could drop the cli feature since it provides little benefit - library users will end up disabling both features, or just enabling install.

PR #2713, opened 2025-01-09
ZKVM-945: emu: validate address range before allocating `to_guest` vector

SchmErik:


risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -437,6 +437,15 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
         if into_guest_len > 0 && !is_guest_memory(into_guest_ptr.0) {
             bail!("{into_guest_ptr:?} is an invalid guest address");
         }
+
+        if into_guest_len > 0 && !into_guest_ptr.is_null() {
+            let end_addr = into_guest_ptr
+                .0
+                .checked_add(into_guest_len as u32)
+                .ok_or_else(|| anyhow!("invalid guest address range"))?;
```
flaub, line 445: Maybe we could add a `checked_add` to `ByteAddr` and `WordAddr`

SchmErik: i'll address this in the afternoon

flaub (approved):

PR #2715, opened 2025-01-09
ZKVM-776: adding linear workflow

pdg744:


pdg744:
[chatting with parker to address the test failure]

flaub (approved):

PR #2721, opened 2025-01-10
ZKVM-919: Add bounds checks on raw baby bear field elements

shkoo:
Fixes risc0/sec-issues#1

### DIFF
--- a/risc0/core/src/field/mod.rs
+++ b/risc0/core/src/field/mod.rs
@@ -108,12 +108,6 @@ pub trait Elem: 'static
     /// methods, this may be called on an INVALID element.
     fn is_valid(&self) -> bool;
 
-    /// Returns true if this element is represented in reduced/normalized form.
-    /// Every element has exactly one reduced form. For a field of prime order
-    /// P, this typically means the underlying data is < P, and for an extension
-    /// field, this typically means every component is in reduced form.
-    fn is_reduced(&self) -> bool;
-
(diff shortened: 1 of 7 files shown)


nategraf (approved):
Looks good to me.

I did have the thought: if someone hands the prover or verifier a Poseidon2 `Digest` and the prover or verifier passes it into `hash_pair` without checking it, the program will `panic` (this is true on `main` now). This becomes more relevant when user ID is a `poseidon2` hash. Definitely the kind of thing that is easy to forget about.

pdg744:
@shkoo can you test whether there's a performance impact from this change?

risc0/core/src/field/mod.rs
```diff
@@ -108,12 +108,6 @@ pub trait Elem: 'static
     /// methods, this may be called on an INVALID element.
     fn is_valid(&self) -> bool;
 
-    /// Returns true if this element is represented in reduced/normalized form.
-    /// Every element has exactly one reduced form. For a field of prime order
-    /// P, this typically means the underlying data is < P, and for an extension
-    /// field, this typically means every component is in reduced form.
-    fn is_reduced(&self) -> bool;
```
flaub, line 115: This breaks semver; we could leave it stubbed out with an impl that always returns the appropriate value.

flaub (changes requested):
There are some semver breaking changes. It's also not clear if this introduces any perf regressions.

PR #2724, opened 2025-01-13
risc0-build: use CARGO_TARGET_DIR if set

Wiezzel:
When building for non-default target, `cargo-build` gets the target directory wrong. It is assumed that the directory structure is
```
$ROOT/target/$profile/build/$crate/out
```
while it can also be
```
$ROOT/target/$target/$profile/build/$crate/out
```
This fix uses a modified version of logic borrowed from [`cxx-build`](https://docs.rs/cxx-build/latest/src/cxx_build/target.rs.html#10-49) to determine the target directory better:
* If `CARGO_TARGET_DIR` is set and is an absolute path, use that.
* Otherwise, go up from `OUT_DIR` until a directory is found such that:
    * It contains `.rustc_info.json`, or
    * It contains `CACHEDIR.TAG`, or
    * It is named `target` and its parent contains `Cargo.toml`.

Closes ZKVM-971


flaub:
In what situation does the out dir get miscomputed?

Is there some way we can add a test for this change?

Wiezzel:
@flaub You can reproduce it by e.g. building a project with `RISC0_USE_DOCKER=1 cargo build --release --target linux-unknown-musl` or with `RISC0_USE_DOCKER=1 CARGO_TARGET_DIR=/your/custom/target/dir cargo build`

flaub (approved):

PR #2727, opened 2025-01-14
Require addresses used by `ecall_sha`, `ecall_bigint`, and `ecall_bigint2` to be aligned in emulator

nategraf:
Currently there is an issue in that a program can issue an `ecall`, e.g. for bigint (1), with an unaligned pointer and the executor will succeed, but proving will fail. This is an issue both for the smaller impact that it's better to know sooner when an execution cannot be proven, and for the larger impact that it could be leveraged to cause issues on Boundless, where the executor is used to determine whether or not a given job is provable.

This PR also adds a check to the software ecall that the _end_ of the guest buffer is in guest memory. This is related check prevents the host from potentially allocating up to 4 GB of memory when the given buffer len is actually invalid.


risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -590,10 +593,10 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
...
 
     fn ecall_bigint2(&mut self) -> Result<bool> {
-        let blob_ptr = self.load_guest_addr_from_register(REG_A0)?.waddr();
-        let nondet_program_ptr = self.load_guest_addr_from_register(REG_T1)?;
-        let verify_program_ptr = self.load_guest_addr_from_register(REG_T2)?;
-        let consts_ptr = self.load_guest_addr_from_register(REG_T3)?;
+        let blob_ptr = self.load_aligned_guest_addr_from_register(REG_A0)?.waddr();
+        let nondet_program_ptr = self.load_aligned_guest_addr_from_register(REG_T1)?;
+        let verify_program_ptr = self.load_aligned_guest_addr_from_register(REG_T2)?;
+        let consts_ptr = self.load_aligned_guest_addr_from_register(REG_T3)?;
```
nategraf, line 599: I'm not 100% certain that bigint2 addresses need to be aligned, but my understanding of how it works would imply they probably do. If someone could check my understanding here, that would be helpful.

risc0/zkvm/src/host/server/exec/syscall/fork.rs
```diff
@@ -111,10 +111,13 @@ impl<'a, 'b> ChildExecutor<'a, 'b> {
     fn ecall_software(&mut self) -> Result<bool> {
         tracing::debug!("ecall_software");
         let into_guest_ptr = ByteAddr(EmuContext::load_register(self, REG_A0)?);
-        if !is_guest_memory(into_guest_ptr.0) && !into_guest_ptr.is_null() {
+        let into_guest_len = EmuContext::load_register(self, REG_A1)?;
+        if into_guest_len > 0
+            && !(is_guest_memory(into_guest_ptr.0)
+                && is_guest_memory(into_guest_ptr.0 + into_guest_len))
+        {
```
nategraf, line 118: Note that this _is_ a slightly different check now. I noticed this is different than the main emulator, but it's unclear why it would be. @flaub , is there a reason the check here should be different than the main emulator's `ecall_software`?

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
pdg744: cc @jbruestle @tzerrell

flaub: Technically, only the `verify_program_ptr` needs to be aligned since that is the pointer that is dereferenced by the circuit. The rest of these pointers are only loaded by the executor/preflight, so those should be safe to be unaligned. It probably doesn't hurt to check all of these.

risc0/zkvm/src/host/server/exec/syscall/fork.rs
flaub: Probably just forgot to update fork. The main thing is that the rust compiler is allowed to supply an unaligned pointer when the length is 0, so the main emulator was updated to account for this.

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -436,8 +436,11 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
     fn ecall_software(&mut self) -> Result<bool> {
         tracing::trace!("[{}] ecall_software", self.insn_cycles);
         let into_guest_ptr = ByteAddr(self.load_register(REG_A0)?);
-        let into_guest_len = self.load_register(REG_A1)? as usize;
-        if into_guest_len > 0 && !is_guest_memory(into_guest_ptr.0) {
+        let into_guest_len = self.load_register(REG_A1)?;
+        if into_guest_len > 0
+            && !(is_guest_memory(into_guest_ptr.0)
+                && is_guest_memory(into_guest_ptr.0 + into_guest_len))
```
flaub, line 442: Note that this needs to be a checked_add

risc0/zkvm/src/host/server/exec/syscall/fork.rs
flaub: Ditto here about checked_add

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
flaub: I suggested in this PR that we add a `checked_add` to `ByteAddr` and `WordAddr`: https://github.com/risc0/risc0/pull/2713

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -436,8 +436,11 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
     fn ecall_software(&mut self) -> Result<bool> {
         tracing::trace!("[{}] ecall_software", self.insn_cycles);
         let into_guest_ptr = ByteAddr(self.load_register(REG_A0)?);
-        let into_guest_len = self.load_register(REG_A1)? as usize;
-        if into_guest_len > 0 && !is_guest_memory(into_guest_ptr.0) {
+        let into_guest_len = self.load_register(REG_A1)?;
```
flaub, line 439: Why not cast this to `usize` here? This avoids casting noise later.

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
flaub: Also, would prefer to use `check_guest_addr` instead of `is_guest_memory` directly, to avoid dropping down the `u32` and staying in `ByteAddr`

flaub: Note that there is the following check that happens later in this function, which we should remove if we're moving the check up to here:
```
        // The guest uses a null pointer to indicate that a transfer from host
        // to guest is not needed.
        if into_guest_len > 0 && !into_guest_ptr.is_null() {
            Self::check_guest_addr(into_guest_ptr + into_guest_len)?;
            self.store_region(into_guest_ptr, bytemuck::cast_slice(&syscall.to_guest))?
        }
```

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
nategraf: That makes sense. It seems like it doesn't make a huge difference either way then. I'd lean towards checking alignment here.

risc0/zkvm/src/host/server/exec/syscall/fork.rs
nategraf: I've removed these changes.

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
nategraf: I've removed these changes.

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
nategraf: I've removed this change.

PR #2743, opened 2025-01-19
ZKVM-975: Remove default timeout (30s) and replace with None timeout

### LINKED ISSUE #2742: Bonsai SDK times out when uploading large inputs (20-30MB)

### Description of bug

When using the Bonsai client (via default_prover) and requesting proofs with large inputs the request times out with:

```shell
Error: HTTP error from reqwest

Caused by:
    0: error sending request for url (https://bonsai-bucket-prod.s3.us-west-2.amazonaws.com/)
    1: operation timed out
```

### Steps to reproduce

- Relocate to a remote farm in rural Australia
- Connect to internet via Starlink
- Attempt to build a proof with input size in the 20-30 MB range using the `default_prover` with Bonsai SDK env vars set

willemolding:
Closes #2742 

This little fix does the trick. By default the `HttpClient` has a 30s timeout and this is getting triggered for my use-case. Explicitly setting this to None removes the timeout and solves the issue.

It may be the case that some timeout is desired and if this is the case we could introduce another env var to set this.


bonsai/sdk/src/lib.rs
```diff
@@ -490,6 +490,7 @@ pub mod module_type {
         Ok(HttpClient::builder()
             .default_headers(headers)
             .pool_max_idle_per_host(0)
+            .timeout(None)
```
mothran, line 493: I might suggest a configurable timeout with a reasonable default that can be overridden via the API (and/or env var), just so the default behavior is to not hang forever if there is a network issue.

mothran (approved):

PR #2752, opened 2025-01-21
rzup 0.3 as a library integration

bobbobbio:
This contains changes to `risc0-build`, `cargo-risczero`, and `risc0-zkvm` to use the new 0.3 version of rzup

These crates now find their relevant components (Rust toolchain, C++ toolchain, and r0vm) via a library call into `rzup`, this makes things a little easier at the call-site.

This usage as a library to read the contents of the component registry (which is backed by directories and TOML files) does create a bit of a risk from a versioning standpoint. Once it is available as a library crate we have to commit to not breaking older versions by having newer versions of `rzup` always write to the component registry in a backwards-compatible way. This way of doing it though does provide a very natural dependency experience for users (its just a normal crate).

The new version of `rzup` among other features includes the ability to install and manage multiple different version of the previously mentioned components, see the `rzup` 0.3 PR for more details https://github.com/risc0/risc0/pull/2712

`risc0-build` use the "default" version of whichever relevant toolchain. The default can be changed with the `rzup default` command.  Somehow detecting the right version to use here is left as future work.

`risc0-zkvm` know the correct version to use and always tries to use that particular version if it is installed. If the correct version isn't installed it falls back to trying to use whatever `r0vm` is available in `PATH`

`cargo-risczero` has been modified to have the `build-toolchain` and `install` commands instead direct to use to use `rzup` for these actions.  The idea here is to have `rzup` own these actions going forward. There exists documentation that mentions these subcommands that should probably be updated before the next release.


bobbobbio:
Here is an automation run I kicked off manually
https://github.com/risc0/risc0/actions/runs/12919190678/job/36029207745

the `check-semver` jobs are failing because I kicked this job off manually and it wasn't automatically run on the PR, (I'm not sure what makes that happen automatically)

bobbobbio:
I've merged this into the main rzup branch per conversation with @flaub

PR #2756, opened 2025-01-22
Tutorial STARK by Hand: add Sage script

teddav:
I really like your STARK By Hand tutorial but I found the python script a bit messy (sorry @pdg744 😅) so I rewrote the entire tutorial with Sage. I think it could be really helpful for beginners to understand.  
I also corrected a small mistake in part 10, in the order of the DEEP polynomials.


pdg744 (approved):
Thanks for adding this!

SchmErik:
hi @teddav we chatted internally about this and came to the conclusion that we'd like to keep this to python only as adding a sage version increases our maintenance. If you'd like to contribute, we definitely welcome improvements to the existing python version so I'm going to close this

teddav:
Hey @SchmErik @pdg744 that's too bad.  
I don't think there's any "maintenance" needed.  
From what I see, this tutorial hasn't been updated in the past year and a half: https://github.com/risc0/risc0/commits/main/website/docs/proof-system/stark-by-hand.md

The tutorial is really nice and explains well how STARKs work and I don't think it will need updates. This Sage script is just an addition for beginners to understand better  
It was an easy (and nice imo) way to improve an already good tutorial

PR #2760, opened 2025-01-22
ZKVM-995: replace `num-bigint` with better-performing `ibig` for bibc evaluator

mars-risc0:
Running Parker's k256-test benchmark, the cycle rate prior to this change was 6.2618 MHz, and the cycle rate with the new library is 6.7958 MHz.

### DIFF
--- a/risc0/circuit/rv32im/Cargo.toml
+++ b/risc0/circuit/rv32im/Cargo.toml
@@ -34,6 +34,7 @@ crypto-bigint = { version = "0.5", default-features = false, optional = true }
 cust = { version = "0.3", optional = true }
 derive_more = { version = "1.0", features = ["debug"], optional = true }
 enum-map = { version = "2.7.3", optional = true }
+ibig = { version = "0.3.6" }
 lazy-regex = { version = "3.3", optional = true }
 num-bigint = { version = "0.4.6", optional = true, default-features = false }
 num-derive = { version = "0.4", optional = true }
--- a/risc0/circuit/rv32im/Cargo.toml
+++ b/risc0/circuit/rv32im/Cargo.toml
@@ -36,7 +36,6 @@ derive_more = { version = "1.0", features = ["debug"], optional = true }
 enum-map = { version = "2.7.3", optional = true }
 ibig = { version = "0.3.6" }
 lazy-regex = { version = "3.3", optional = true }
-num-bigint = { version = "0.4.6", optional = true, default-features = false }
 num-derive = { version = "0.4", optional = true }
 num-traits = { version = "0.2", optional = true }
 rand = { version = "0.8", optional = true }
@@ -69,7 +68,6 @@ prove = [
   "dep:derive_more",
   "dep:enum-map",
   "dep:lazy-regex",
-  "dep:num-bigint",
   "dep:num-derive",
   "dep:num-traits",
   "dep:rand",
--- a/risc0/circuit/rv32im/src/prove/emu/bibc.rs
+++ b/risc0/circuit/rv32im/src/prove/emu/bibc.rs
@@ -173,7 +173,7 @@ impl Program {
                     let typ = &self.types[op.result_type];
                     let count = typ.coeffs.next_multiple_of(16) as u32;
                     let value = io.load(op.arena(), op.offset(), count)?;
-                    regs[op_index] = IBig::from(value);
+                    regs[op_index] = value.into();
                 }
                 OpCode::Store => {
                     let typ = &self.types[op.result_type];
--- a/risc0/circuit/rv32im/src/prove/emu/bibc.rs
+++ b/risc0/circuit/rv32im/src/prove/emu/bibc.rs
@@ -16,8 +16,8 @@ use std::io::Read;
 
 use anyhow::{anyhow, Result};
 use byteorder::{LittleEndian, ReadBytesExt};
-use ibig::{ibig, IBig, ubig, UBig};
 use ibig::modular::ModuloRing;
+use ibig::{ibig, ubig, IBig, UBig};
 use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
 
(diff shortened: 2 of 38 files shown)


risc0/circuit/rv32im/src/prove/emu/preflight/mod.rs
```diff
@@ -872,31 +872,33 @@ impl<'a> BigInt2Witness<'a> {
...
-        }
-        ensure!(words.len() == word_count);
-        for (i, word) in words.iter().enumerate() {
-            self.witness.insert(addr + i, *word);
+        for i in 0..word_count {
+            let offset = i * 4;
+            let mut word = 0_u32;
+            if offset < bytes.len() {
+                let wordbytes = &bytes[offset..offset + 4];
+                word = u32::from_le_bytes(wordbytes.try_into().unwrap());
```
tzerrell, line 899: Hmm. I would think this could fail if storing something very small (e.g. `1`) as then `bytes` would have a length that wasn't an even multiple of 4 and so this line would try to build `word` from fewer than 4 bytes and fail. But we have tests that should catch this in https://github.com/risc0/risc0/blob/main/risc0/bigint2/src/field/tests.rs ... and I'm not sure those tests are being run. Not this PR's fault if they aren't, but I'm not going to approve until I investigate that more

tzerrell: Update: Yep, this causes those tests to fail, for the reason I outlined. I'll make a separate update with the tests working.

tzerrell: https://github.com/risc0/risc0/pull/2763

mars-risc0: Good catch! I had assumed that a bigint with 32-bit limbs would always return a multiple-of-4 byte array.

flaub: Also, we need to avoid calling `unwrap` when possible and return errors instead.

risc0/circuit/rv32im/Cargo.toml
```diff
@@ -34,6 +34,7 @@ crypto-bigint = { version = "0.5", default-features = false, optional = true }
 cust = { version = "0.3", optional = true }
 derive_more = { version = "1.0", features = ["debug"], optional = true }
 enum-map = { version = "2.7.3", optional = true }
+ibig = { version = "0.3.6" }
 lazy-regex = { version = "3.3", optional = true }
 num-bigint = { version = "0.4.6", optional = true, default-features = false }
```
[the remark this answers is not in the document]
mars-risc0: There are a lot of other places which use it, but now that I look more closely, none of them are inside `rv32im`: so yes, we can actually drop it here!

risc0/circuit/rv32im/src/prove/emu/bibc.rs
```diff
@@ -151,35 +152,34 @@ impl Program {
...
                     }
-                    regs[op_index] = BigInt::from_biguint(Sign::Plus, value);
+                    regs[op_index] = IBig::from(value);
                 }
                 OpCode::Load => {
                     let typ = &self.types[op.result_type];
                     let count = typ.coeffs.next_multiple_of(16) as u32;
                     let value = io.load(op.arena(), op.offset(), count)?;
-                    regs[op_index] = BigInt::from_biguint(Sign::Plus, value);
+                    regs[op_index] = IBig::from(value);
```
flaub, line 176:
```suggestion
                    regs[op_index] = value.into();
```

risc0/circuit/rv32im/src/prove/emu/preflight/mod.rs
```diff
@@ -872,31 +872,43 @@ impl<'a> BigInt2Witness<'a> {
...
+            let byte_offset = i * 4;
+            let mut word = 0_u32;
+            if byte_offset < value_bytes.len() {
+                assert!(byte_offset + 3 < value_bytes.len());
+                let word_bytes: [u8; 4] = [
+                    value_bytes[byte_offset],
+                    value_bytes[byte_offset + 1],
+                    value_bytes[byte_offset + 2],
+                    value_bytes[byte_offset + 3],
+                ];
```
flaub, line 908:
```suggestion
        let word_count = count as usize / WORD_SIZE;
        for i in 0..word_count {
            let byte_offset = i * WORD_SIZE;
            let mut word = 0_u32;
            if byte_offset < value_bytes.len() {
                assert!(byte_offset + 3 < value_bytes.len());
                let word_bytes: [u8; WORD_SIZE] = [
                    value_bytes[byte_offset],
                    value_bytes[byte_offset + 1],
                    value_bytes[byte_offset + 2],
                    value_bytes[byte_offset + 3],
                ];
```

mars-risc0: great idea - I'll update the assert, too

tzerrell (approved):

risc0/circuit/rv32im/src/prove/emu/preflight/mod.rs
```diff
@@ -872,31 +872,43 @@ impl<'a> BigInt2Witness<'a> {
...
+            if byte_offset < value_bytes.len() {
+                assert!(byte_offset + WORD_SIZE <= value_bytes.len());
+                let word_bytes: [u8; WORD_SIZE] = [
+                    value_bytes[byte_offset],
+                    value_bytes[byte_offset + 1],
+                    value_bytes[byte_offset + 2],
+                    value_bytes[byte_offset + 3],
+                ];
+                word = u32::from_le_bytes(word_bytes);
+            }
```
flaub, line 910:
```suggestion
            let word = if byte_offset < value_bytes.len() {
                assert!(byte_offset + WORD_SIZE <= value_bytes.len());
                let word_bytes = [
                    value_bytes[byte_offset],
                    value_bytes[byte_offset + 1],
                    value_bytes[byte_offset + 2],
                    value_bytes[byte_offset + 3],
                ];
                u32::from_le_bytes(word_bytes)
            } else {
                0
            };
```

flaub (approved):

PR #2761, opened 2025-01-23
ZKVM-998: fix panic when using embedded allocator and add more details to OOM message

SchmErik:
This fixes a panic that happens when calling items in std when the embedded allocator is enabled. Our std implementation calls into `sys_alloc_aligned` so we should not panic when this is called. Fix this by calling the embedded allocator's alloc function.


pdg744 (approved):

SchmErik:
putting in draft - I'd like to add a small test for this

flaub (approved):

PR #2762, opened 2025-01-24
ZKVM-999: risc0-build: Use a builder pattern for GuestOptions and DockerOptions

flaub:
Also includes the ability to specify custom environment variables for docker builds.

Supersedes #2709


Wiezzel:
@flaub Why did you decide to use `derive_builder` over `typed-builder`?

flaub:
`typed_builder` wouldn't let me write code like:

```
    let mut guest_opts = GuestOptionsBuilder::default();
    if env::var("RISC0_USE_DOCKER").is_ok() {
        let docker_opts = DockerOptionsBuilder::default()
            .root_dir("../../..")
            .build()
            .unwrap();
        guest_opts.use_docker(docker_opts);
    }
    let guest_opts = guest_opts.build().unwrap();
```

risc0/build/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository = { workspace = true }
 [dependencies]
 anyhow = "1.0"
 cargo_metadata = "0.19.1"
+derive_builder = "0.20.2"
```
SchmErik, line 13: TIL! This is cool!

SchmErik (approved):

PR #2763, opened 2025-01-24
ZKVM-1000: Fix bigint field tests

tzerrell:
The bigint field tests were broken (trying to import ELFs that didn't exist). This fixes the tests and also puts them into CI.


mars-risc0 (approved):
looks good to me

flaub (approved):

PR #2764, opened 2025-01-24
feat(rzup): Add regex pattern matching for toolchain paths

crStiv:
Implements regex pattern matching for toolchain paths instead of requiring exact name matches. This change:
- Adds regex-based path matching
- Improves error handling
- Allows flexible toolchain selection using patterns
- Resolves TODO comment in default.rs


hmrtn:
thank you - closing this for #2712

PR #2766, opened 2025-01-25
feat(rzup): Add path validation for toolchain names

DeVikingMark:
Add regex-based validation for toolchain names in the default command.
This change:
- Validates toolchain names using regex pattern
- Only allows alphanumeric characters, underscores, hyphens and dots
- Checks if toolchain exists before linking
- Provides more informative error messages
- Resolves TODO comment about path validation


hmrtn:
thank you - closing this for #2712

PR #2777, opened 2025-01-28
ZKVM-1010: Replace numb-bigint / crypto-bigint with `malachite` and lots of other small optimizations for bigint2

mothran:
Sample bigint2 workload:
```
exec 3s -> 1s
preflight 500ms -> 291ms 
```


tzerrell (approved):

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -740,25 +741,60 @@ impl<S: Syscall> Executor<'_, '_, S> {
...
-        let mut bytes = value.to_le_bytes();
-        if bytes.len() < count as usize {
-            bytes.resize(count as usize, 0);
-        }
-        ensure!(bytes.len() == count as usize);
-        self.store_region_into_guest(addr, &bytes)
+        let mut bytes = bigint_to_bytes_le(value);
+
+        match bytes.len().cmp(&(count as usize)) {
+            Ordering::Greater => bytes.truncate(count as usize),
```
flaub, line 784: Is this safe to do? It feels like this should be an error if the bytes are larger than expected.

risc0/circuit/rv32im/src/prove/emu/pager.rs
```diff
@@ -285,7 +302,7 @@ impl PagedMemory {
 
 impl Page {
     fn load(&self, addr: WordAddr) -> u32 {
-        let word_addr = (addr.0 % PAGE_WORDS as u32) as usize;
+        let word_addr = (addr.0 & ((PAGE_WORDS - 1) as u32)) as usize;
```
flaub, line 305: Is this actually any faster?

risc0/circuit/rv32im/src/prove/emu/preflight/bigint2.rs
```diff
@@ -154,24 +155,31 @@ impl ProgramState {
...
     fn shift(&self) -> Self {
         let mut ret = self.coeffs.clone();
-        for _ in 0..BIGINT2_WIDTH_BYTES {
-            ret.insert(0, 0);
-        }
-        Self::new(ret)
+        ret.insert_many(0, [0; BIGINT2_WIDTH_BYTES]);
+        // for _ in 0..BIGINT2_WIDTH_BYTES {
+        //     ret.insert(0, 0);
+        // }
```
flaub, line 181: Still needed?

risc0/circuit/rv32im/src/prove/emu/pager.rs
```diff
@@ -81,6 +81,23 @@ impl PagedMemory {
...
+        if idx == INVALID_IDX {
+            self.load_page(page_idx);
+            idx = self.page_table[page_idx as usize];
+        }
+
+        let page = &self.page_cache[idx as usize];
+        let offset = addr.0 % PAGE_SIZE as u32;
+        let bytes_to_read = size.min(PAGE_SIZE as u32 - offset);
+
+        bytes.extend_from_slice(&page.0[offset as usize..(offset + bytes_to_read) as usize]);
```
flaub, line 98:
```suggestion
        let bytes = page.0[offset as usize..(offset + bytes_to_read) as usize].to_vec();
```

Even better if maybe we could have a mutable slice be sent to this function as an argument, so that we could eliminate a vec allocation.

flaub (approved):

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
```diff
@@ -740,25 +741,60 @@ impl<S: Syscall> Executor<'_, '_, S> {
...
         let addr = base + offset * BIGINT2_WIDTH_BYTES as u32;
-        let mut bytes = value.to_le_bytes();
-        if bytes.len() < count as usize {
-            bytes.resize(count as usize, 0);
-        }
-        ensure!(bytes.len() == count as usize);
-        self.store_region_into_guest(addr, &bytes)
+        let mut bytes = bigint_to_bytes_le(value);
+
+        match bytes.len().cmp(&(count as usize)) {
```
bobbobbio, line 783: Is there a reason to do `cmp` instead of using `>`?

risc0/circuit/rv32im/src/prove/emu/pager.rs
mothran: Nope, ended up being not worth it, removed in: 59cc88a2c6d3423e8427d94b41fb3989beab2c8a

risc0/circuit/rv32im/src/prove/emu/preflight/bigint2.rs
mothran: 59cc88a2c6d3423e8427d94b41fb3989beab2c8a

risc0/circuit/rv32im/src/prove/emu/pager.rs
mothran: Good call 59cc88a2c6d3423e8427d94b41fb3989beab2c8a

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
mothran: This a clippy warning I never really agreed with but its this one: https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain we could allow() this one if we like a if better here, I am happy with either.

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
mothran: That was a oddity to me, its roughly taken from how bytes were previously truncated to the size of `count` (https://github.com/risc0/risc0/blob/main/risc0/circuit/rv32im/src/prove/emu/exec/mod.rs#L758) but I did not know that function interface well enough to know when we are "over storing" via the count param and if it should be a fault. Let me know what you think since I don't have a good handle on that interface's call patterns.

risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
bobbobbio: I didn't think it triggered for something with only two branches, but if it does I have to say I also am not sure I agree with it

PR #2779, opened 2025-01-29
ZKVM-1011: Add unstable feature flag to risc0-zkp Cargo.toml to address warning and allow usage of unstable feature

nategraf:
https://github.com/risc0/risc0/pull/2658 introduced some API surface under `#[stability::unstable]` but forgot to add the `unstable` flag. I saw warnings about this in the CI builds and so figured I should open this PR


nategraf:
@shkoo

flaub (approved):

PR #2782, opened 2025-01-30
Use borsh_derive dependency instead of feature

BertrandD:
Hey :wave: 

When using [indexmap](https://docs.rs/indexmap/latest/indexmap/) and borsh, we need to disable the borsh feature "derive" and use the borsh_derive crate directly. But as risc0 enables "derive", we do have a cyclic dependency. Cf Indexmap doc:

> borsh: Adds implementations for [BorshSerialize](https://docs.rs/borsh/1.5.3/x86_64-unknown-linux-gnu/borsh/ser/trait.BorshSerialize.html) and [BorshDeserialize](https://docs.rs/borsh/1.5.3/x86_64-unknown-linux-gnu/borsh/de/trait.BorshDeserialize.html) to [IndexMap](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html) and [IndexSet](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html). Note: When this feature is enabled, you cannot enable the derive feature of [borsh](https://docs.rs/borsh/1.5.3/x86_64-unknown-linux-gnu/borsh/index.html) due to a cyclic dependency. Instead, add the borsh-derive crate as an explicit dependency in your Cargo.toml and import as e.g. use borsh_derive::{BorshSerialize, BorshDeserialize};.

### DIFF
--- a/risc0/binfmt/Cargo.toml
+++ b/risc0/binfmt/Cargo.toml
@@ -14,7 +14,8 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 
 [dependencies]
 anyhow = { version = "1.0", default-features = false }
-borsh = { version = "1.5", default-features = false, features = ["derive"] }
+borsh = { version = "1.5", default-features = false }
+borsh-derive = { version = "1.5", default-features = false }
 derive_more = { version = "1.0", default-features = false, features = [
   "add",
   "add_assign",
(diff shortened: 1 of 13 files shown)


BertrandD:
It didn't exist at that time, so maybe this new feature solves it. We do not use indexmap anymore, so I can't easily check this right now... But according to https://github.com/near/borsh-rs/pull/346 it should be fixed indeed.

PR #2799, opened 2025-02-06
fix: ensure stable sorting for recipients with equal shares

crStiv:
This PR fixes the unstable sorting issue when recipients have equal shares in the allocate() function.

Changes include:
- Added secondary sorting by recipient name for equal shares
- Added property-based test using proptest to verify sorting stability
- Test ensures consistent allocation results regardless of input order
- Maintains correct total amount distribution

The sorting is now deterministic even when multiple recipients have the same share, which is important for reproducible results in the allocation process.

Note: There are two remaining TODO comments in the file regarding rounding tests and CSV whitespace handling. These will be addressed in separate PRs for better change isolation.

Fixes TODO: "In the case of ties in ownership share this is not sufficient to give consistent behavior regardless of input order"

### DIFF
--- a/examples/prorata/core/src/lib.rs
+++ b/examples/prorata/core/src/lib.rs
@@ -92,13 +92,10 @@ impl fmt::Display for AllocationQueryResult {
 /// amount - total amount to distribute
 /// recipients - list of recipients with their share of the total amount
 pub fn allocate(amount: Decimal, recipients: Vec<Recipient>) -> Vec<Allocation> {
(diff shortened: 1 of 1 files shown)


SchmErik:
@crStiv looks like this is failing CI. Please run cargo fmt and address the examples build failure

SchmErik:
we're trying to decrease the amount of open PR's. Since this PR doesn't pass tests due to missing dependencies, I'm going to close

PR #2801, opened 2025-02-07
ZKVM-1035: Improvements to guest profiling

bobbobbio:
This adds two main improvements to guest profiling
- Support for inline functions
- Show page events as function calls

### inline functions
To maintain the way the cycle-accurate profiler works we build a table of all the inline function from the DWARF data and then when we enter an inline function we push a frame to the profiler's call stack for that inline function. Once we leave the range of the inline function we pop it off.

### page events
When we incur something that causes a page to be loaded we now magic up a functions called `[PageIn]` and `[PageOut]`. 
Since this can happen somewhat randomly within a segment, functions that normally would be seen as low-cost can incur a high cycle count due to them triggering page events (you can see this in the profile below with `FdReader::new`, that function is only a single instruction which writes a single word to memory)
The emitting of these events was added to the v1 circuit code. The v2 circuit code needs to be updated to work with the profiler first, and that isn't tackled in this PR.

### comparison
These screenshots are from before and after the changes running `tests::keccak_update2::case_1`

before:
before
after:
after

### DIFF
--- a/risc0/circuit/rv32im/src/trace.rs
+++ b/risc0/circuit/rv32im/src/trace.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 RISC Zero, Inc.
+// Copyright 2025 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


risc0/circuit/rv32im/src/trace.rs
```diff
@@ -47,6 +47,12 @@ pub enum TraceEvent {
         /// Data that's been written
         region: Vec<u8>,
     },
+
+    /// A page is read for the first time in a segment
+    PageIn { cycles: u64 },
+
+    /// A page has been written to for the first time in a segment
```
flaub, line 54: So in this case, it's the last time a page is written to that will actually be performed in the circuit.

flaub (approved):
Nice!

PR #2802, opened 2025-02-07
ZKVM-1036: Replace more occurrences of `.ok_or(..)` with `.ok_or_else(..)`

bobbobbio:
In #2765 I replaced many places where we were creating an `anyhow!` error not in an error-path to pass into the `ok_or` function. This PR fixes some places that I missed originally.

I messed up finding these originally because I mainly searched for `ok_or(anyhow!(..))`, this time I took a look at all places where we are calling `ok_or` in the `risc0` repository. The only uses remaining should be situations where creating the error is free.


SchmErik (approved):

flaub (approved):
Thank you

PR #2813, opened 2025-02-11
ZKVM-1051: Support for using client to create v2 segments

flaub:


risc0/zkvm/compat/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "risc0-zkvm-compat-tests"
+version = "0.1.0"
+edition = "2021"
```
SchmErik, line 4: I think we want to set release = "false" here to prevent this crate from being published

flaub: Good call

SchmErik (approved):

PR #2818, opened 2025-02-12
Drop rv32im-v2

SchmErik:
v2 will be disabled for the upcoming 1.3 release. This PR disables v2 as well as tests that run on v2


flaub:
I realized that there some more bits that need to be removed for this to be effective. I'll try and take a pass at it later on.

flaub (approved):

PR #2821, opened 2025-02-12
ZKVM-1060: rv32im-v2: BigInt soundness

flaub:


risc0/circuit/rv32im-v2-sys/kernels/cuda/defs.cu.inc
```diff
@@ -1,7 +1,7 @@
 SET_FIELD(BabyBear);
-constexpr size_t kRegCountAccum = 80;
+constexpr size_t kRegCountAccum = 103;
```
pdg744, line 2: is this additional 23 columns expected?

jbruestle: Yes, this is the BigInt verification logic.

jbruestle (approved):

PR #2823, opened 2025-02-13
ZKVM-1062: Example using substrate-bn crate

tzerrell:
Demonstrates using the `substrate-bn` crate in the ZKVM. Working but depends on https://github.com/risc0/paritytech-bn/pull/3, so I'm leaving as draft until that lands.

The test takes about 6 minutes to run on my machine, so we may want to keep an eye on CI and see if it takes too long.

### DIFF
--- a/examples/bn254/methods/guest/Cargo.lock
+++ b/examples/bn254/methods/guest/Cargo.lock
@@ -607,6 +607,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "no_std_strings"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a5b0c77c1b780822bc749a33e39aeb2c07584ab93332303babeabb645298a76e"
+
+[[package]]
 name = "num-bigint"
 version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -833,6 +839,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "risc0-zkos-v1compat"
+version = "0.1.0"
+dependencies = [
+ "include_bytes_aligned",
+ "no_std_strings",
+]
+
+[[package]]
 name = "risc0-zkp"
 version = "1.3.0-alpha.1"
 dependencies = [
@@ -872,6 +886,7 @@ dependencies = [
  "risc0-circuit-rv32im-v2",
  "risc0-core",
  "risc0-groth16",
+ "risc0-zkos-v1compat",
  "risc0-zkp",
  "risc0-zkvm-platform",
  "rrs-lib",


pdg744:
Looks like https://github.com/risc0/paritytech-bn/pull/3 has landed -- is this ready for review now?

pdg744:
Hmm... any ideas for how we can reduce the CI impact on this?

examples/bn254/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+
+fn main() {
+    let receipt = prove_pairing();
+    receipt.verify(BN254_VERIFY_ID).unwrap();
+
+    let is_one: bool = receipt
+        .journal
+        .decode()
+        .expect("Journal should contain a single `bool`");
+    println!("Pairing batch should give one; did it? {}", is_one);
```
bobbobbio, line 67: this can be `{is_one}`

examples/bn254/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+name = "bn254-inner"
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
+
+[dependencies]
+bn254-core = { path = "../../core" }
+risc0-zkvm = { path = "../../../../risc0/zkvm", default-features = false, features = ["std", "unstable"] }
+substrate-bn = "0.6.0"
```
bobbobbio, line 11: if its a pre-release version, maybe better to lock it down to exactly 0.6.0 ` = "=0.6.0"`

examples/bn254/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[patch.crates-io]
+# TODO: Get a tag stabilized and use it
```
bobbobbio, line 14: was this a TODO you meant to leave?

tzerrell:
> Hmm... any ideas for how we can reduce the CI impact on this?

Here are the options that occur to me:

1. Decide the CI perf hit doesn't matter. This is my preferred option, because this PR took 12, 13, and 9 minutes on CPU, CUDA, and Metal, respectively; compare that to [this recent PR](https://github.com/risc0/risc0/actions/runs/13321089909/job/37205746046), which took 12, 11, and 9 minutes on those same systems. I don't fully trust those timings -- they aren't meant to be benchmarks and it's weird that only CUDA took 2 minutes longer -- but they suggest the extra CI perf hit isn't really a problem. That said, we should probably also be adding tests that run in CI to bls12-381, k256, and p256, and the aggregate perf impact would presumably be larger
2. Test something smaller. The main functionality of this crate is pairing. I could perhaps run a single pairing rather than a batch of 2 pairings to save a bit of performance, although batch pairing is the more common thing to call, and the test will need to be a little more awkward and opaque in what it does. Beyond that though, we have few options except having the example test some very minor functionality (scalar multiplication on one of the curves?) that doesn't test the main algorithm.
3. Make the CI system more parallel. We could split off this example (and other slow examples) to run on their own machine in parallel to the other example tests. This would reduce the wall clock time of running CI but also make CI a bit more expensive (since each machine would need to duplicate the setup steps). It would also add complexity to the CI code (both in building this change but also in maintaining it).

examples/bn254/methods/guest/Cargo.toml
tzerrell: Yes. Until we have made a tagged release for https://github.com/risc0/paritytech-bn/tree/release/v0.6.0, this TODO needs to remain, and we can't make that tagged release until risc0-bigint2 v1.3 has been released

bobbobbio:
your example made sense to me as someone who doesn't understand `substrate-bn` very well.

> Hmm... any ideas for how we can reduce the CI impact on this?
I'm kinda confused, looking at the most CI run on your PR I see this in the logs
```
     Running unittests src/main.rs (target/debug/deps/bn254_example-d2aeeb7ee48ec6c9)

running 1 test
WARNING: proving in dev mode. This will not generate valid, secure proofs.
WARNING: Proving in dev mode does not generate a valid receipt. Receipts generated from this process are invalid and should never be used in production.
test test_pairing ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.88s
```
this ran in under 2s, is there a different spot in the CI where it is taking longer?

tzerrell:
> this ran in under 2s, is there a different spot in the CI where it is taking longer?

Oh ... CI is running examples in dev mode. That makes a lot of sense and makes the test go _much_ faster. So yeah, no concerns about CI, we should not drop this test.

bobbobbio (approved):

PR #2828, opened 2025-02-13
ZKVM-1068: bls12_381 example

iddo-bentov:


tzerrell (approved):
Looks good!

It might be nice to add a simple test, e.g. one that just ensures `exec_pairing` runs successfully; that way this will get tested in CI. (The only reason I could think not to do that would be if it ran too slow and bogged down CI -- which admittedly does seem possible with pairing.)

A possible upgrade (that we don't necessarily need to do and could be a separate PR if we do decide to do it) would be to have the host pass in expected inputs and the guest write the result to the receipt, which the host can check matches what's expected.

examples/bls12_381/README.md
```diff
@@ -0,0 +1,11 @@
+# BLS12_381 Example
+
+## Quick Start
+
+First, follow the [examples guide] to install dependencies and check out the correct version of the example.
```
tzerrell, line 5: Looks like my error made it into your PR too ... `[examples guide]` needs a link in the file, like
```
[examples guide]: https://dev.risczero.com/api/zkvm/examples/#running-the-examples
```

iddo-bentov: Ok, I added the examples guide link, but I haven't yet added a test because I don't want to make CI slower at least for now (I notice that some examples such as `ecdsa/k256` don't have a test, while others such as `chess` have a test).

SchmErik (approved):

PR #2832, opened 2025-02-14
ZKVM-1090: added bls12_381 to precompiles.md

iddo-bentov:


website/api/zkvm/precompiles.md
```diff
@@ -43,6 +43,18 @@ each fork's repository on GitHub.
 | --------------------------------------------------------- | ------------------ | -------------------------------------------------------------------------------------- | ----------------------- |
 | [`rsa`](https://github.com/risc0/RustCrypto-RSA/releases) | 0.9.6              | `rsa = { git = "https://github.com/risc0/RustCrypto-RSA", tag = "v0.9.6-risczero.0" }` | [Yes]                   |
 
+### SUBSTRATE-BN
```
tzerrell, line 46: What about merging bn254 and bls12-381 under a single "Other Elliptic Curve" header? I.e. because we are listing earlier EC patches under the algorithm names not the curve names
```suggestion
### Other Elliptic Curve
```

website/api/zkvm/precompiles.md
```diff
@@ -43,6 +43,18 @@ each fork's repository on GitHub.
...
+
+| Crate                                                                    | Versions supported  | Patch Statement Example                                                                                   | Requires Unstable Flag? |
+| ------------------------------------------------------------------------ | ------------------- | --------------------------------------------------------------------------------------------------------- | ----------------------- |
+| [`substrate-bn`](https://github.com/risc0/paritytech-bn/releases)        | 0.6.0               | `substrate-bn = { git = "https://github.com/risc0/paritytech-bn", tag = "v0.6.0-risczero.0" }`            | [Yes]                   |
+
+### BLS12_381
+
+| Crate                                                                    | Versions supported  | Patch Statement Example                                                                                   | Requires Unstable Flag? |
+| ------------------------------------------------------------------------ | ------------------- | --------------------------------------------------------------------------------------------------------- | ----------------------- |
```
tzerrell, line 55: If you like my previous comment, you'll need to adjust this too
```suggestion
```

tzerrell (approved):
Thank you!

website/api/zkvm/precompiles.md
pdg744: I think we have way too many tables on this page -- seems like it would be clearer if we consolidate all of this into a single table. Mayyyyybe 2-3 tables. Any more than that seems counterproductive to me.

pdg744: For reference, here's how SP1 handles this: https://docs.succinct.xyz/docs/sp1/writing-programs/patched-crates#supported-libraries

pdg744 (approved):

PR #2833, opened 2025-02-14
ZKVM-1091: Cleanup parallel union support

flaub:


examples/prover/src/plan.rs
```diff
@@ -179,7 +184,13 @@ impl Planner {
         Ok(task_number)
     }
 
-    pub fn finish(&mut self) -> Result<usize, PlannerErr> {
+    pub fn finish(&mut self) -> Result<(), PlannerErr> {
+        self.finish_unions()?;
+        self.finish_joins()?;
```
bobbobbio, line 189: disclaimer: I don't really understand the implications of calling this now in this function instead of when it was called previously, I assume its roughly the same point in time, but now you don't need an RPC

bobbobbio (approved):

SchmErik (approved):

PR #2834, opened 2025-02-14
ZKVM-1: fix ProveKeccakRequest alignment issues

austinabell:
The original error reported:

```
thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck-1.20.0/src/internal.rs:33:3:
cast_slice>TargetAlignmentGreaterAndInputNotAligned
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at ./rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/panicking.rs:74:14
   2: bytemuck::internal::something_went_wrong
             at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck-1.20.0/src/internal.rs:33:3
   3: bytemuck::internal::cast_slice
             at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck-1.20.0/src/internal.rs:257:15
   4: bytemuck::cast_slice
             at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck-1.20.0/src/lib.rs:377:12
   5: risc0_zkvm::host::server::prove::keccak::prove_keccak
             at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/risc0-zkvm-1.2.1/src/host/server/prove/keccak.rs:35:37
   6: <risc0_zkvm::host::server::prove::prover_impl::ProverImpl as risc0_zkvm::host::server::prove::ProverServer>::prove_keccak
             at ./usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/risc0-zkvm-1.2.1/src/host/server/prove/prover_impl.rs:255:9
...
```

Where the issue here is that the input is just `Vec<u8>` and a reference to is being cast to `&[[u64; 25]]` which should require 8 byte alignment, but is not guaranteed by the `Vec<u8>`.

~~I don't love this fix, as it introduces a new type to the API, but was just trying to avoid at all costs making breaking changes to the protobuf format, which has it as `Vec<u8>`. Perhaps could switch the newtype to that struct and cast on conversions, but would require copying bytes as you can't cast the `Vec`.~~ Also hard to repro, as I didn't hit the issue myself, so trying to find a way to test this now. Opening for visibility.


risc0/zkvm/src/host/client/env.rs
```diff
@@ -77,6 +77,12 @@ pub struct ProveZkrRequest {
     pub input: Vec<u8>,
 }
 
+/// Wrapper newtype to enforce alignment to allow casting to KeccakState, which is u64 aligned
+/// (8 bytes).
+#[repr(align(8))]
+#[derive(Clone)]
+pub struct KeccakInput(pub Vec<u8>);
```
bobbobbio, line 84: I'm pretty sure that this `repr(align(8))` here doesn't do what you want. This will affect the alignment of the memory that the `Vec` is in (which is just a size and pointer) but not the buffer that the `Vec` allocates, which IIUC is what is needed.

It seems like we need to somehow either:
1. change the type of the Vec before it allocates memory
2. copy the memory to somewhere that is aligned properly before casting

bobbobbio: might just be easiest to just change the `Vec<u8>` to a `Vec<u64>`

austinabell: resolved in below commits (sorry for not having one commit to fix, was trying to avoid installing latest toolchain)

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -1206,7 +1209,23 @@ impl TryFrom<pb::api::ProveKeccakRequest> for ProveKeccakRequest {
...
                 .ok_or_else(|| malformed_err("ProveKeccakRequest.control_root"))?
                 .try_into()?,
-            input: value.input,
+            input,
         })
     }
 }
+
+#[stability::unstable]
+pub(crate) fn keccak_input_to_bytes(input: &[KeccakState]) -> Vec<u8> {
```
flaub, line 1218: If something is `pub(crate)` then `unstable` doesn't make sense.

austinabell: well, just wanted to make sure the method wasn't dead code and produce a warning. Perhaps I'm misunderstanding the unstable semantics here.

austinabell:
Unfortunately not able to reproduce the exact issue through the interaction with Bento, even when running the exact same proof workloads on multiple machines.

This PR comes with a test that if you revert the implementation will fail (checks deserializing from different byte offsets to check possible memory alignments). Hopefully this is sufficient.

Also noting that backporting this to 1.2 didn't apply directly because of some conflicts. https://github.com/risc0/risc0/compare/release-1.2...austin/1.2-keccak-fix in case it helps resolve when/if backporting.

SchmErik (approved):

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -1206,7 +1209,75 @@ impl TryFrom<pb::api::ProveKeccakRequest> for ProveKeccakRequest {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_keccak_bytes_to_input_alignment() {
+        // Create a buffer with extra padding at the start to test different alignments
+        let padding = 8; // Test all possible alignments (0-7)
```
bobbobbio, line 1241: Generally this can go up to 16 actually

risc0/zkvm/src/host/api/convert.rs
```diff
@@ -1206,7 +1209,75 @@ impl TryFrom<pb::api::ProveKeccakRequest> for ProveKeccakRequest {
...
+        for (i, byte) in test_buffer.iter_mut().enumerate() {
+            *byte = (i % 256) as u8;
+        }
+
+        // Test each alignment
+        for offset in 0..8 {
+            let aligned_slice = &test_buffer[offset..][..(keccak_states * state_size)];
+            let result = try_keccak_bytes_to_input(aligned_slice);
+
+            assert!(result.is_ok(), "Failed to parse at alignment {}", offset);
```
bobbobbio, line 1256: this can be `{offset}` here and other places below

bobbobbio: this message doesn't quite make sense. you don't actually know the alignment of the buffer pointer, and then you add some offset to it, which still makes it kinda unknown alignment, maybe just say "offset" here and below instead

risc0/zkvm/src/host/server/prove/dev_mode.rs
```diff
@@ -74,7 +74,9 @@ impl ProverServer for DevModeProver {
         let mut keccak_assumptions = HashMap::new();
 
         for proof_request in session.pending_keccaks.iter() {
-            let claim = compute_keccak_digest(&proof_request.input);
+            // Note: safe to cast as slice here, given alignment of KeccakState (8 bytes) is greater
+            // than alignment of u8 slice.
```
bobbobbio, line 78: because a `u8` slice has no alignment requirement, its always going to be okay to cast some memory to it.

bobbobbio (approved):
I think the test has misleading assertion messages, it doesn't actually know what alignment it is testing, but other than that LGTM

flaub (approved):

risc0/zkvm/src/host/api/convert.rs
austinabell: Fair enough, but in this case the only that matter are the u64 (8) based alignments?

bobbobbio: Well this is the alignment of the memory you are copying from, which can have an arbitrary alignment, technically it can have any kind of alignment but the ones that matter for real processors are (2, 4, 8, 16) IIUC. 

Thinking it through later, this test was probably just overkill, just testing one situation of a mismatched alignment would probably have been fine. Testing all the ways it can be unaligned is probably unneeded.

PR #2836, opened 2025-02-14
ZKVM-1092: Preparations for switching over to v2 as default

bobbobbio:
This change does some preparation for switching over to make v2 be the default.

- It reduces the amount of places where we need to update the default version.
- It adds more Executor switching to some places that need the new executor.


bobbobbio:
Closing in favor of #2846

PR #2846, opened 2025-02-17
ZKVM-1095: Replace risc0-circuit-rv32im with rv32im-v2

flaub:
The following tests are failing and so are ignored:

* host::server::exec::tests::docker::trace
* host::server::exec::tests::sys_fork
* host::server::exec::tests::sys_fork_fork_panic
* host::server::exec::tests::sys_fork_journal_panic
* host::server::prove::tests::run_unconstrained


risc0/binfmt/src/lib.rs
```diff
@@ -29,9 +29,6 @@ mod image;
 mod image2;
```
SchmErik, line 29: should we drop the image mod and rename image2 -> image?

SchmErik: also, I'm fine doing this on a later pass

flaub: I have a feeling this would be a breaking change.

SchmErik: yeah, maybe we could consider bumping binfmt to 2.0?

SchmErik (approved):

PR #2848, opened 2025-02-18
ZKVM-1101: Change rv32imv2 `eqz` to panic rather than throw an exception

shkoo:
This lets us actually see eqz failures instead of getting difficult-to-track-down memory corruption, since poolstl's parallel for doesn't wait for all threads to finish before propagating exceptions.


tzerrell (approved):

flaub:
Is there a way to wait for threads to finish before throwing? I think we should avoid `abort`, especially in the context of Bonsai/Bento.

flaub:
Looks like CI is hanging, might be related to abort?

flaub:
This might work in a pinch for debugging, but I think we'll need to sort out how to configure poolstl to catch exceptions properly. I'm thinking about downstream uses like Bonsai/Bento may not play well with `abort`.

PR #2849, opened 2025-02-19
ZKVM-1103: updating security model page for po2 22

pdg744:
Currently, our maximium allowable segment size is 2^21. In recent testing, we have found the performance gains from using a segment size of 2^22 are significant. Based on that observation, we are updating the zkVM & verifier contracts to allow this segment size in our upcoming 1.3.0 release. 

Nominally, this change reduces our bits of security by 1 bit using the current methodology in our soundness calculators. This PR updates the docs accordingly. 

Notably, we don't believe this change has any actual impact in terms of the difficulty of mounting an attack against the system: although the expected number of attacks required to construct a false proof would decrease by a factor of 2, the difficult of each attempt would increase by a factor of 2, which results in no impact overall.


pdg744:
This PR also fixes an omission from earlier: the calculation of number of years required to attack the system was written for segment size of 2^20 and never updated to reflect 2^21.

tzerrell (approved):

jacobdweightman (approved):

SchmErik (approved):

Issue #2853, opened 2025-02-19
Possible unsoundness with uninit `to_guest: &mut [u32]` for syscalls


thaliaarchi:
I think it may be unsound for `Syscall` to send data from the host to guest via `to_guest: &mut [u32]`, when the pointer given by the guest is to an uninitialized allocation. I think use of `sys_read` or `sys_read_words` with uninitialized buffers could trigger UB. I think it should instead be `&mut [MaybeUninit<u32>]`.

My motivation for this comes from wanting to implement `Read::read_buf` for the RISC Zero version of `std::io::Stdin` (tracked in https://github.com/rust-lang/rust/issues/136756).

If this sounds good, I could submit a PR here similar to what I did for the Hermit unikernel in https://github.com/hermit-os/kernel/pull/1606 (Use uninitialized buffers for `read` and `recvfrom`). However, since each file descriptor is associated with a `dyn Read` and `Read::read_buf` is currently unstable, this might not be possible.

On the other hand, Rust tentatively does not have [full recursive validity for references](https://github.com/rust-lang/unsafe-code-guidelines/issues/412), so as long as uninitialized data is not read, you *might* be fine. @RalfJung?

### Steps to reproduce

This [patch to `std`](https://github.com/rust-lang/rust/pull/136769/commits/f0764b13bcdd6fbae48b6a2b4155a341bc6be1fc#diff-6ee4d23cdba75d7cefd174768cce8fea78e72e863e7c29ee586f3894c4a11bbf), which implements `Read::read_buf` for RISC Zero stdin, would encounter this issue.

Since I know little about RISC Zero, here's my understanding of the relevant bits, so you can correct me if I'm wrong:

<details><summary>Trace from the C FFI to the underlying <code>Read::read</code> calls</summary>

- Guest code invokes a function in the C FFI, passing a buffer via a `*mut u8` and `usize` length:

  https://github.com/risc0/risc0/blob/711b163551c96ccfbfc8271b7634d945997ff457/risc0/zkvm/platform/src/syscall.rs#L482

- It invokes a syscall via `ecall`, passing the pointer and length by `a0` and `a1` registers and :

  https://github.com/risc0/risc0/blob/711b163551c96ccfbfc8271b7634d945997ff457/risc0/zkvm/platform/src/syscall.rs#L584-L590

  https://github.com/risc0/risc0/blob/711b163551c96ccfbfc8271b7634d945997ff457/risc0/zkvm/platform/src/syscall.rs#L236-L251

[...]

PR #2856, opened 2025-02-20
Fix store peeking

thaliaarchi:
When storing a region of memory in the guest, the host peeks at each word so it can write it back with the byte replaced. When that guest memory is uninitialized, that causes it to read uninitialized memory, when only a store was necessary. Instead, perform such stores directly over word-aligned regions instead of byte-aligned. Since the pointers used in practice are already word-aligned, this makes it simpler.

Also, fix a bounds check which used a byte length when a word length was given.

See https://github.com/risc0/risc0/issues/2853.

### DIFF
--- a/risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
+++ b/risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
@@ -455,7 +455,7 @@ impl<S: Syscall> Executor<'_, '_, S> {
             bail!("{into_guest_ptr:?} is an invalid guest address");
         }
 
-        if into_guest_len > 0 && !into_guest_ptr.is_null() {
+        if into_guest_len > 0 {
             let end_addr = into_guest_ptr
                 .checked_add(into_guest_len as u32)
                 .ok_or_else(|| anyhow!("invalid guest address range"))?;
@@ -490,7 +490,7 @@ impl<S: Syscall> Executor<'_, '_, S> {
 
         // The guest uses a null pointer to indicate that a transfer from host
         // to guest is not needed.
-        if into_guest_len > 0 && !into_guest_ptr.is_null() {
+        if into_guest_len > 0 {
             self.store_region(into_guest_ptr, bytemuck::cast_slice(&syscall.to_guest))?;
         }
 
--- a/risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
+++ b/risc0/circuit/rv32im/src/prove/emu/exec/mod.rs
@@ -485,7 +485,7 @@ impl<S: Syscall> Executor<'_, '_, S> {
         // The guest uses a null pointer to indicate that a transfer from host
         // to guest is not needed.
         if into_guest_len > 0 {
-            self.store_region(into_guest_ptr, bytemuck::cast_slice(&syscall.to_guest))?;
+            self.store_region(into_guest_ptr, &syscall.to_guest)?;
         }
 
         let (a0, a1) = syscall.regs;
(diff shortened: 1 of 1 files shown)


thaliaarchi:
I have more confidence in the correctness of the first two patches, which are local changes, than the third. It seems that the codebase assumes generally deals with memory in terms of `&[u8]`/`Vec<u8>` and tracks writes as such in `SyscallRecord`s among other places. This clashes with memory only being word-addressable. The changes in the third commit, although local, suggest that these records of memory ops should be in terms of `&[u32]`/`Vec<u32>` everywhere, which is a much larger change.

bobbobbio (changes requested):
As far as I can tell, I wasn't able to find an issue with the guest's buffer being uninitialized here.

Also, I think you missed some code in `sys_read` where it was already converting from bytes into words, so at this point in the code you shouldn't need to convert it.

The host is allowed to read any memory used as the backing store for the guest memory at any point because from the host's perspective it is alway initialized.

thaliaarchi:
> Also, I think you missed some code in `sys_read` where it was already converting from bytes into words, so at this point in the code you shouldn't need to convert it.

I was aware that `sys_read` splits the guest slice to be word-aligned. What I was referring to was this snippet:

https://github.com/risc0/risc0/blob/06f9eaea79ddec82d4f53e4a98fcf2afa2c73b73/risc0/circuit/rv32im/src/prove/emu/exec/mod.rs#L459-L460

Here, `into_guest_ptr` is a `ByteAddr`, but `into_guest_len` is the number of words in the slice. This adds that length as if it was bytes. Thus, when the guest pointer is close to the end of the guest memory and the length overruns, this validation will be too short and miss that.

The other alignment stuff I reviewed looked fine.

When I get back to this, I'll rework this PR, since my uninit assumption was wrong. The general approach of avoiding word peeks should be faster and still desirable. I'll make a routine which splits a range into an aligned range with an unaligned prefix and suffix, then only does the peeking for the prefix and suffix.

bobbobbio:
> Here, `into_guest_ptr` is a word-aligned `ByteAddr` and `into_guest_len` is the number of words in the slice. However, this adds that length as if it was in bytes. Thus, when the guest pointer is near the end of the guest memory and the length overruns, this validation will be too short and miss that.

I see, yes my mistake, that does now look wrong to me, thanks for explaining.

> When I get back to this, I'll rework this PR, since my uninit assumption was wrong. The general approach of avoiding word peeks should be faster and still desirable. I'll make a routine which splits a range into an aligned range with an unaligned prefix and suffix, then only does the peeking for the prefix and suffix.

Cool, thank you for offering to do that. I agree that it sounds like a desirable change.

bobbobbio:
Hey, so sorry this is my mistake in not realizing and point this out sooner, but this code in particular is going to be deleted real soon. There is a new version of this code in `rv32im-v2` that I am going to look at to try to verity doesn't have any bounds checking bugs like you found.

thaliaarchi:
No worries! I don't mind spending the time learning how it works and the patches didn't take long. I'll go ahead and close this then.

bobbobbio:
I tracked down this code-path in v2. I didn't find any issues. Here is the path for posterity

The `ecall_software` function is implemented in assembly now.
https://github.com/risc0/risc0/blob/06f9eaea79ddec82d4f53e4a98fcf2afa2c73b73/risc0/zkos/v1compat/src/kernel.s#L134-L140
It immediately bit-shifts the length left by two, which has the effect of multiplying it by 4 (which is word-size).

Then later it forwards this call into the (new I think) `HOST_ECALL_READ` syscall, which is handled here.
https://github.com/risc0/risc0/blob/06f9eaea79ddec82d4f53e4a98fcf2afa2c73b73/risc0/circuit/rv32im-v2/src/execute/r0vm.rs#L216-L227

This does the similar bounds checking as before, but the length is actually in bytes so it should be right now.

PR #2860, opened 2025-02-21
Remove Unused Import: Sha256

NeoByteXx:
This PR removes the unused import Sha256 from the use statement in the file. The Sha256 struct was never used in the code, making its import redundant. Removing it enhances code clarity and prevents unnecessary dependencies.


SchmErik:
this is used. It's not a struct but a trait and including this allows us to use the functions for this trait. closing

PR #2862, opened 2025-02-22
ZKVM-1121: Update arkworks

tzerrell:
~Have our `risc0-groth16` crate use our arkworks fork that's accelerated for the zkVM. The PR (https://github.com/risc0/arkworks-rs-algebra/pull/1) with the acceleration should be landed first (and the patch updated accordingly).~

~Note that this also updates arkworks to 0.5 (from 0.4)~

This PR now just updates arkworks to 0.5, and is more performant than the patched version was.


tzerrell:
I turns out if I patch in the right spot, just updating to arkworks 0.5 goes (slightly) faster than even my patch did. So this is now just an "update arkworks" PR.

bobbobbio (approved):

PR #2863, opened 2025-02-22
ZKVM-1122: implement dev-mode for union/keccak proofs

SchmErik:


flaub:
Could add a dev-mode test so we can catch this sort of thing in the future.

SchmErik:
> Could add a dev-mode test so we can catch this sort of thing in the future.

added a test under examples that exercises this path

flaub (approved):

PR #2866, opened 2025-02-22
zkvm: enable 3gb of guest memory

SchmErik:
This is a simple fix to adjust the GUEST_MAX_MEM. There maybe more that we can do to clean up risc0-zkvm-platform but this could result in semver breaking changes such as removing public symbols. This change is meant to do the bare minimum to enable a larger heap for the guest.

### DIFF
--- a/risc0/zkvm/platform/src/memory.rs
+++ b/risc0/zkvm/platform/src/memory.rs
@@ -17,7 +17,7 @@ use super::WORD_SIZE;
 pub const MEM_BITS: usize = 28;
 pub const MEM_SIZE: usize = 1 << MEM_BITS;
 pub const GUEST_MIN_MEM: usize = 0x0000_0400;
-pub const GUEST_MAX_MEM: usize = SYSTEM.start;
+pub const GUEST_MAX_MEM: usize = 0xC000_0000;
 
 /// Top of stack; stack grows down from this location.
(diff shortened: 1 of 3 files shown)


flaub (approved):
This is fine for now, but we probably need a feature flag before cutting a release.

austinabell:
for my own understanding, what is the reason that the SYSTEM const address is not changed also? `SYSTEM_START` is based on this and is used for register states. Won't this have conflicting memory because registers will be kept within the range of the heap memory, allowing one to overwrite registers?

flaub:
> for my own understanding, what is the reason that the SYSTEM const address is not changed also? `SYSTEM_START` is based on this and is used for register states. Won't this have conflicting memory because registers will be kept within the range of the heap memory, allowing one to overwrite registers?

The values in this file aren't used by `risc0-circuit-rv32im-v2`. There's another round of cleanup that needs to be done at some point.

risc0/zkvm/platform/src/memory.rs
```diff
@@ -17,7 +17,7 @@ use super::WORD_SIZE;
 pub const MEM_BITS: usize = 28;
 pub const MEM_SIZE: usize = 1 << MEM_BITS;
 pub const GUEST_MIN_MEM: usize = 0x0000_0400;
```
flaub, line 19: This actually needs to be `0x0000_4000` in v2.

PR #2875, opened 2025-02-24
ZKVM-1135: Update bigint2 datasheet benchmark to work with v2

bobbobbio:
The number of cycles for the same workload has changed making it no longer in-line with the other execute benchmarks. I just adjusted the workload to make it closer, and updated the assert since I couldn't get under 10k difference.

### DIFF
--- a/risc0/bigint2/methods/guest/build.rs
+++ b/risc0/bigint2/methods/guest/build.rs
@@ -16,7 +16,7 @@ use std::{fmt::Write as _, path::Path};
 
 use k256::ecdsa::{signature::Signer as _, Signature, SigningKey, VerifyingKey};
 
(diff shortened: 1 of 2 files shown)


SchmErik (approved):
looks good to me but I'm wondering if we can add the datasheet (or the asserts) as a part of CI? it would be nice if we can some how add the asserts to the PR CI, we will not encounter this issue on the nightly workflow any more

bobbobbio:
> looks good to me but I'm wondering if we can add the datasheet (or the asserts) as a part of CI? it would be nice if we can some how add the asserts to the PR CI, we will not encounter this issue on the nightly workflow any more

I think some of the asserts are going to be hard to make without running the same work the datasheet runs. I don't see why we can just run the datasheet as part of CI though.

Issue #2439, opened 2024-10-22
Incompatible risc0 version used for 1.0.x


drgora:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [X] This is not a support question.

### Description of bug

I believe there is a problem with the Risc0 dependencies from crates.io, which causes errors when trying to compile a risc0 project based on `risc0-zkvm v1.0.x`.

Ultimately the problem seems to be with semantic versioning: for example, `risc0-zkvm v1.0.5` ends up depending on `risc0-build v1.1.2`, incurring compilation failures. This is expected given that, according to semver, `v1.1.2` should be compatible with `v1.0.5`.

Possible solutions might be using the wildcard syntax to match deps with version `1.1.*`, or using an exact match like `=1.0.5`?

### Steps to reproduce

Follow the hello-world steps, using version `1.0.x`:
```
rzup --version 1.0.5
cargo risczero new foo --guest-name foo_guest
cargo run --release
```
fails with:
```
error: unexpected argument '--id' found

Usage: r0vm <--port <PORT>|--elf <ELF>|--image <IMAGE>>

For more information, try '--help'.

```

flaub:
It's actually just a warning, even though the message says that it's an error. What's going on is that the image_id computation can run faster by running `r0vm` which is built in release mode. But if we try to call `r0vm` and it's an older version you'll see this message. However, it does fallback to computing the image_id in the same process as part of the build and so doesn't actually block anything.

We can work on improving this error message to make it more clear what's going on and how the user should proceed.

drgora:
Thank you for explaining, that makes sense.
Still, with the same steps mentioned above, I get a panick when executing the demo program:
```
    Finished `release` profile [optimized + debuginfo] target(s) in 37.25s
     Running `target/release/host`
thread 'main' panicked at host/src/main.rs:41:10:
called `Result::unwrap()` on an `Err` value: Your installation of the r0vm server is not compatible with your host's risc0-zkvm crate.
Do one of the following to fix this issue:

1. Install r0vm server version 1.0
2. Change the risc0-zkvm and risc0-build dependencies in your project to 1.0

risc0-zkvm version: 1.1.2
r0vm server version: 1.0.5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

hzysvilla:
Hi drgora, I also ran into this issue and have found a temporary method to address it.
[...]

PR #2881, opened 2025-02-25
Constrain versions of internal deps to minor version for 1.2

austinabell:
This prevents 1.3 or any other minor version to be used for indirect references with any imported crate as they are not semver compatible.

related https://linear.app/risczero/issue/ZKVM-741/minor-versions-not-constrained-for-risc0-internal-sub-dependencies and https://github.com/risc0/risc0/issues/2439

although this doesn't resolve for future versions yet, will follow up this PR


flaub:
We're planning to release 2.0 soon

PR #2885, opened 2025-02-25
ZKVM-1147: compat test: fix compat test by adding `--locked` to install command

SchmErik:


bobbobbio:
This seems okay if this is just a temporary fix, I'm not sure we should be using `--locked` in this scenario after we fix things with 1.3

bobbobbio (approved):
Okay, I understand now why this is needed

PR #2890, opened 2025-02-26
ZKVM-1156: yanking 1.3 docs

pdg744:


bobbobbio (changes requested):
This misses some stuff, I think we should just revert https://github.com/risc0/risc0/commit/cfd063a66fbfa2f89ef5dfe93de0ad8681da6f4e

PR #2892, opened 2025-02-26
Fix cargo risczero new to use 2.0.0 for risc0_build

bobbobbio:
This is hopefully temporary until we can replace it with something more robust.

This will need to be backported to the 1.3 branch.


bobbobbio:
btw - it seems the reason why the release CI didn't catch this is that we don't run `cargo risczero new` at all there. In the main CI job we do run it, but with the `--path` flag which doesn't exercise this path at all

flaub (approved):

PR #2896, opened 2025-02-26
Restrict some 1.1 branch dependencies

bobbobbio:
We made a breaking changes to the following crates in 1.2.X:

- risc0-zkp
- risc0-circuit-recursion
- risc0-circuit-recursion-sys
- risc0-circuit-rv32im
- risc0-circuit-rv32im-sys

The 1.1.X crates won't compile with those version, so we have to restrict them.

These crates really shouldn't be depended on directly by non-risc0 crates, so I think we can just restrict these dependencies to get out of these problems for 1.1.1


bobbobbio:
I think there are other crates that have problems too. I need to figure out all the crates first

PR #2897, opened 2025-02-26
Grammatical suggestions for: README.md

0xbasar:
Hi,

I have two suggestions for this file:

First; There is a small issue with "complex workspace or directory structure." It would be better to use "a" to make the sentence more grammatically correct: "However, this crate fails to build when placed in a complex workspace or directory structure",

Second: The use of "panic" could be more consistent. You could use: "This example does not produce a receipt and is intended as a test to ensure that the guest code runs without panicking".

Thanks.


bobbobbio:
Thank you for this, but we aren't accepting spelling or grammar related PRs at this time.

PR #2899, opened 2025-02-27
Disable some failing risc0-circuit-bigint tests in 1.1 release

bobbobbio:
I'm not sure why these are failing, but it seems that we don't care that much about this crate, since it isn't used.

I was originally going to just push to the branch ignoring the failures, but then I realized there is some CUDA testing that won't run unless we fix it somehow.


flaub:
There's likely a caching error somehow in our build for cuda code. In the past, we've blown away the cache to get around it.

bobbobbio:
shoot I just figured this out independently! thanks for the tip. I think that cache issue is that the GPU architecture isn't being included in the inputs, last time this succeeded on a different card.

PR #2900, opened 2025-02-27
Backport 1.3: ZKVM-1121: Update arkworks (#2862)

bobbobbio:


tzerrell (approved):
LGTM (up to the reproducible build test that CI is telling you about, which will need a new expected hash since this does change the dependencies of the guest that we check is stable).

Issue #2871, opened 2025-02-23
`serde::Visitors` should have `visit_map` rather that `visit_seq` called when deserializing structs


jeswr:
### Is there an existing issue?

- [x] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [x] This is not a support question.

### Description of bug

The following error is thrown when trying to pass the type `I2` from the host to the guest; despite the fact that the type will roundtrip when serialized/deserialized using serde; including using `serde_json::from_reader`.

```
called `Result::unwrap()` on an `Err` value: Custom("invalid type: sequence, expected struct I2")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at host/src/main.rs:84:64:
called `Result::unwrap()` on an `Err` value: Guest panicked: called `Result::unwrap()` on an `Err` value: Custom("invalid type: sequence, expected struct I2")
```

### Steps to reproduce

I have created a repro at https://github.com/jeswr/risc0-sparql-poc/tree/chore/minimal-serialization-repro:
 - The deserializer implementation is [here](https://github.com/jeswr/risc0-sparql-poc/blob/94c26f8caa460c2fb3f56f60c23dcf921950e4b3/core/src/lib.rs#L25-L48)
 - Roundtrip tests that pass outside of the ZKVM can be found [here](https://github.com/jeswr/risc0-sparql-poc/blob/94c26f8caa460c2fb3f56f60c23dcf921950e4b3/core/src/lib.rs#L69-L85)
 - CI can be found [here](https://github.com/jeswr/risc0-sparql-poc/actions/runs/13483080451/job/37670777147?pr=5), showing successful roundtripping in the `cargo test` and failure to pass between host and guest with `cargo run --release`

Related issues:
 - #1742
 - #1059

jeswr:
Renaming to reflect core issue discovered in implementing fix #2905

PR #2905, opened 2025-02-28
fix: `serde::Visitors` should have `visit_map` rather that `visit_seq` called when deserializing structs

jeswr:
Working towards a fix of #2871. The underlying issue of which is that in `serde::Visitors` should have `visit_map` rather that `visit_seq` called when deserializing structs.

Marking as ready for review as [feedback is needed on this comment](https://github.com/risc0/risc0/pull/2905#discussion_r1975683152). 

Note, the implementation of `WordRead` for `&str` is co-pilot generated and requires checking before merging.


risc0/zkvm/src/serde/deserializer.rs
```diff
@@ -457,7 +457,10 @@ impl<'de, R: WordRead + 'de> serde::Deserializer<'de> for &'_ mut Deserializer<'
     where
         V: Visitor<'de>,
     {
-        self.deserialize_tuple(fields.len(), visitor)
+        visitor.visit_map(MapAccess {
+            deserializer: self,
+            len: fields.len(),
+        })
```
jeswr, line 463: This is a slight hack; to implement this properly `MapAccess` should be checking that only the fields defined in `fields` are being passed to the map visitor.

jeswr: @flaub Before I proceed to:
 - add this check to `MapAccess`, and
 - add a test cases where there are entries in the struct
 
 Could I please ask you to review this draft PR - to confirm:
  - The previous use of `deserialize_seq` here was a bug rather than the intended behavior
  - That you are comfortable with moving forward by adding an `Optional` `fields` property to `MapAccess`; which would add validation behavior to the keys iterated through by `MapAccess`.

jeswr: Note that the change introduces the following test failures:

```
failures:

---- serde::deserializer::tests::test_struct stdout ----
thread 'serde::deserializer::tests::test_struct' panicked at risc0/zkvm/src/serde/deserializer.rs:576:49:
called `Result::unwrap()` on an `Err` value: NotSupported

---- serde::deserializer::tests::test_str stdout ----
thread 'serde::deserializer::tests::test_str' panicked at risc0/zkvm/src/serde/deserializer.rs:594:49:
called `Result::unwrap()` on an `Err` value: NotSupported


failures:
    serde::deserializer::tests::test_str
    serde::deserializer::tests::test_struct
```

It is unclear to me whether this is a "bug in the tests"; in that I am unsure if they are testing for behviors which conflict with serde behavior.

jeswr: Created https://github.com/risc0/risc0/pull/2906 to try and answer this question, would be good if CI can be run on that PR.

flaub:
We'd like to deprecate our serde codec implementation. I'd recommend using `postcard` or some other more complete serde codec implementation.

PR #2906, opened 2025-02-28
chore(test): test functionality against serde's deserializeowned

jeswr:
Runs tests against Serdes `DeserializeOwned` as well as the custom one defined by the ZKVM; supporting the work being done on #2905


flaub:
We will be deprecating our custom serde code implementation. We recommend using another codec like `postcard` or `bincode`.

PR #2910, opened 2025-03-01
ZKVM-1171: Add kernel_id to ReceiptMetadata

flaub:


bobbobbio (approved):
we should probably just bump all of it to 2.0, we already have breaking changes in the other crates anyway

PR #2919, opened 2025-03-04
Minor fix for CI for params digest check

tzerrell:
Since https://github.com/risc0/risc0/pull/2619 changes the format of receipt errors, the tests need to be changed to expect the new errors. This PR does so, partially fixing CI for #2619.

Note that the actual changes in this PR are quite small, the large amount of code is due to merging `main`.


tzerrell:
[This was merged from git, not github, as we decided it made more sense to have these changes directly on #2619 than as a separate PR]

PR #2928, opened 2025-03-06
ZKVM-1186: Use wide types in poseidon2

bobbobbio:
This improves my poseidon2 benchmark from 1.77s to 1.43s. 

With this code if we turn on the vectorized instructions it goes down from 1.43s to 1.22s. We could do some runtime detection of the processor features in a future change maybe.


bobbobbio:
these use of wide types don't help the other poseidon2 implementation

PR #2930, opened 2025-03-07
ZKVM-1188: Report Ecall metrics

tzerrell:
Reports ecall metrics from the executor.


SchmErik (commented):
I think you need to increment the count for read in this line:
https://github.com/risc0/risc0/pull/2930/files#diff-eae5c13b15bbcd3ec4dfd8709517172725b75edf4cb81de553274d22334dc656L409
and write in this line: 
https://github.com/risc0/risc0/pull/2930/files#diff-eae5c13b15bbcd3ec4dfd8709517172725b75edf4cb81de553274d22334dc656L416

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -470,6 +486,7 @@ impl<S: Syscall> Risc0Context for Executor<'_, '_, S> {
 
     fn on_terminate(&mut self, a0: u32, a1: u32) -> Result<()> {
         self.user_cycles += 1;
+        // TODO: We don't want to count user_cycles towards the terminate ecall, right?
```
flaub, line 489: Why not?

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -417,13 +428,18 @@ impl<S: Syscall> Risc0Context for Executor<'_, '_, S> {
...
         _next: CycleState,
         _s0: u32,
         _s1: u32,
         _s2: u32,
+        kind: EcallKind,
     ) -> Result<()> {
         self.phys_cycles += 1;
+        self.ecall_metrics.0[kind].cycles += 1;
+        if cur == CycleState::MachineEcall {
+            self.ecall_metrics.0[kind].count += 1;
```
flaub, line 441: What's the rationale for this one?

risc0/circuit/rv32im/src/execute/executor.rs
tzerrell: I was thinking of ecall cycles as something separate from user or paging cycles -- that it was a different category and each cycle would be at most one of these. Sounds like this model isn't what we want, so I'll change this.

flaub (approved):

flaub:
Previous comments from Tim for posterity:

I am still examining whether this is fully correct, although it does seem to be largely correct (e.g. bigint cycle counts seem to line up correctly). The places that give me pause are:

* ~Reserved cycle counts are still too high even after subtracting off ecall cycles~ This is expected, because reserved cycle counts includes the padding cycles to get up to the next power of 2
* Not sure why syscall and ecall read/write numbers don't match.

I also chose not to adjust down the reserve cycle counts for ecalls, and instead just added a note in the log. However, I could also add an `ecall_cycles` field to `SessionStats` ([here](https://github.com/risc0/risc0/blob/c6a157b76933908b14886bd1d9f45feabacb703c/risc0/zkvm/src/host/prove_info.rs#L30)) if that's preferred.

~I'm happy to either land and iterate, or hold the PR open while I dig further.~ With the reserve cycles issue sorted out, I think this is good to go -- unless anyone sees a problem with the syscall vs ecall read/write counts not matching.

tzerrell:
Thanks for the cleanup & corrections!

PR #2935, opened 2025-03-07
ZKVM-1190: Cleanups in preparation for major version bump

flaub:


risc0/zkvm/src/host/client/prove/mod.rs
```diff
@@ -418,6 +418,11 @@ pub fn default_executor() -> Rc<dyn Executor> {
     Rc::new(ExternalProver::new("ipc", get_r0vm_path().unwrap()))
 }
 
+/// Return a local [Executor].
+pub fn local_executor() -> Rc<dyn Executor> {
+    Rc::new(self::local::LocalProver::new("local"))
+}
+
```
bobbobbio, line 425: nice!

bobbobbio (approved):
Love to see it.

I think there is still some v2 stuff remaining, I guess some of it we need to address first with zirgen changes right?

flaub:
> Love to see it.
> 
> I think there is still some v2 stuff remaining, I guess some of it we need to address first with zirgen changes right?

I could include the rename to `rv32im-v2` in this PR, or we could do it separately. I have a zirgen PR prepared.

bobbobbio:
> > Love to see it.
> > I think there is still some v2 stuff remaining, I guess some of it we need to address first with zirgen changes right?
> 
> I could include the rename to `rv32im-v2` in this PR, or we could do it separately. I have a zirgen PR prepared.
for the sake of reviewing, I think we should put the rename in its own PR

PR #2949, opened 2025-03-10
ZKVM-1203: add tagged tiny keccak patch to main

sashaaldrick:
changed:
`tiny-keccak = { git = "https://github.com/risc0/tiny-keccak", rev = "7ff2b8465891d1cbb55a69e553e05d369ce41d72" }`
to:
`tiny-keccak = { git = "https://github.com/risc0/tiny-keccak", tag = "tiny-keccak/v2.0.2-risczero.0" }`

to avoid confusion and errors (customer support, someone used the rev syntax and the error was fixed changing it to the above tagged version).


flaub:
Why is this preferred? I don't understand what this actually fixes.

pdg744:
I think the only reason this is preferred is in order to simplify the mental bandwidth for users & devrel -- i.e., when we answer the question of "how do i use the precompiles?" we typically point users to the precompiles.md page in the dev docs & the example in `risc0/examples`. I guess it's not strictly necessary, but it would be nice if the usage patterns in those assets align.

pdg744:
> and the error was fixed changing it to the above tagged version. 

I suspect this may be a misunderstanding -- I expect the actual fix [here](https://github.com/BrianSeong99/Agglayer_PessimisticProof/pull/7/) was based on the changes to where the patch was being applied within the directory structure, rather than the change to the patch statement.

flaub (approved):

PR #2950, opened 2025-03-10
ZKVM-1199: prepare risc0-build and rzup for toolchains >= 1.85.0

SchmErik:
Recent LLVM versions have changed `loweratomic` to `lower-atomic`. Change this for `rzup` and `risc0-build`.


risc0/build/src/lib.rs
```diff
@@ -426,8 +427,32 @@ pub(crate) fn cargo_command_internal(subcmd: &str, guest_info: &GuestInfo) -> Co
     cmd
 }
 
+fn get_rustc_version() -> Result<semver::Version> {
+    let output = Command::new("rustc")
+        .arg("--version")
+        .env("RUSTUP_TOOLCHAIN", "risc0")
+        .output()
+        .expect("failed to get risc0 toolchain's rustc version. Please ensure that your risc0 toolchain installed is correctly");
```
bobbobbio, line 435: If instead we could as `rzup` for the path to the Rust toolchain that would be better, maybe re-use the existing `rust_toolchain()` function in `risc0/build/src/lib.rs`

bobbobbio (changes requested):
The `rzup` changes look fine. Could maybe add a test, but might not be worth it, up to you.

I'm just asking that we use `rzup` to find the Rust toolchain in your new changes, I did changes previously to remove other places like the one you added.

flaub (approved):

flaub:
I'd love to somehow remove or reduce the duplication between `risc0-build` and `rzup` so that we have a single source of truth for building guests.

SchmErik:
@bobbobbio I addressed your feedback but also added names to the new option. --path is mutually exclusive with the tag/commit and the name helps differentiate which one is which. I think this could be a breaking change but this command is likely seldomly used so I'm hoping that this is ok. I'll follow up in a later PR with changes to rzup. This breaks the ability for new rzup to build older rust versions

rzup/src/cli/commands.rs
```diff
@@ -272,13 +272,23 @@ pub const BUILD_HELP: &str = "Discussion:
...
+        .args(&["tag_or_commit", "path"])
+        .required(true)
+))]
 pub(crate) struct BuildCommand {
-    /// Name of component to uninstall
+    /// Name of component to build
     #[arg(value_parser=parser("build"))]
     name: String,
 
     /// Tag or commit of the component to uninstall
```
bobbobbio, line 285: thanks for fixing the doc-comment, I think this one is wrong too

bobbobbio (approved):

PR #2952, opened 2025-03-10
Replace risc0-circuit-rv32im with risc0-circuit-rv32im-v2

flaub:


SchmErik (approved):

bobbobbio (approved):
GitHub makes it pretty hard to read these changes, but AFAICT LGTM

PR #2957, opened 2025-03-11
ZKVM-1208: Update crate version numbers with new scheme

bobbobbio:
In the new crate version numbering scheme each crate will have its own version.
The crates on main will either:
- have the version of the last release if there are no changes since that release
- have the version of the next release if there are changes since the last release

We are going to bump all the version numbers for the next release though (but not all major version numbers), to make things easier to start but also to avoid associating with the 1.3.0 release.

I am also working on changes to the CI semver checks so that these rules are always respected and the PR with changes to a crate will include bumping the version number. Before I can make those checks pass though, I need the main version numbers to be right.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -866,7 +866,7 @@ dependencies = [
 
 [[package]]
 name = "bonsai-sdk"
-version = "1.4.0-alpha.1"
+version = "1.4.0"
 dependencies = [
  "anyhow",
  "bincode",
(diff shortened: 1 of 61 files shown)


flaub:
Maybe we should wait for the CI checks to be in place before this one lands

bobbobbio:
> Maybe we should wait for the CI checks to be in place before this one lands
The CI checks won't pass unless we do this (because these version numbers aren't correct in a few ways), so it seems like this comes first to me. Unless you are thinking we have an interstitial state where we change the version numbers to something else before we do this.

flaub:
OK, timing wise I suppose we need to do this first. But do we have the CI actions developed yet?

flaub (approved):

rlukata (approved):

PR #2958, opened 2025-03-12
ZKVM-1210: Reduce bigint executor work

flaub:
The `BytePolyProgram` only needs to be run during preflight.


mars-risc0 (commented):
this makes a lot of sense

bobbobbio (approved):

PR #2960, opened 2025-03-12
ZKVM-1213: Add EC precompiles for 384 bits

tzerrell:
Adds precompile support for 384-bit elliptic curve operations (add and double, and thereby also mul). This supports elliptic curves over any prime field whose prime fits into 384 bits.

This does slightly slow down performance for 256-bit EC operations (by about 1%) to handle selecting the appropriately-sized precompile. We might be able to avoid this by getting rid of const generics for the bitwidth and making multiple structs & traits for the different bitwidths, but I'm not sure this is worth the re-architecting.


iddo-bentov (approved):
It looks ok to me (I see that it relies on https://github.com/risc0/zirgen/pull/202), but I'm not so familiar yet with EC bigint2 programs.

PR #2965, opened 2025-03-12
backport 2.0: prepare risc0-build and rzup for toolchains >= 1.85.0

SchmErik:
Recent LLVM versions have changed `loweratomic` to `lower-atomic`. Change this for `rzup` and `risc0-build` and add ability to build rust toolchain from a local directory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4481,6 +4481,7 @@ dependencies = [
  "risc0-zkp",
  "risc0-zkvm-platform",
  "rzup",
+ "semver",
  "serde",
  "serde_json",
  "stability",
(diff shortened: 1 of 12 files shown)


bobbobbio:
For 2.0 frank I discussed just taking all the commits to main when we are ready to release (or cut another RC), so I'm not sure we should bother backporting this one thing right now. If we need a new RC I think we can just merge all of main into this branch.

SchmErik:
> For 2.0 frank I discussed just taking all the commits to main when we are ready to release (or cut another RC), so I'm not sure we should bother backporting this one thing right now. If we need a new RC I think we can just merge all of main into this branch.

got it! Thanks. I'll close this then

PR #2966, opened 2025-03-12
backport 1.2: risc0-build: add support for rust toolchain versions >= 1.85

SchmErik:
This is a backport of https://github.com/risc0/risc0/pull/2950


risc0/build/src/lib.rs
```diff
@@ -457,15 +458,39 @@ pub fn cargo_command(subcmd: &str, rust_flags: &[&str]) -> Command {
     cmd
 }
 
+fn get_rustc_version() -> Result<semver::Version> {
+    let output = Command::new("rustc")
+    .arg("--version")
+    .env("RUSTUP_TOOLCHAIN", "risc0")
+    .output()
+    .expect("failed to get risc0 toolchain's rustc version. Please ensure that your risc0 toolchain installed is correctly");
```
bobbobbio, line 466: indentation is weird

bobbobbio (approved):

risc0/build/src/lib.rs
flaub: Is our check job not catching this with `cargo fmt`?

SchmErik: looks like it ran but didn't catch this. I also ran it locally and it also didn't catch this either..

PR #2967, opened 2025-03-13
ZKVM-1218: Skip git submodules in `cargo_local_patch.py`

capossele:
When applying the `cargo_local_patch.py` during the `risc0-ethereum` CI workflow, we also modify the git submodules if a match is found. That by itself is not an issue; however since we added the `--locked` when running the `risc0-forge-ffi` binary, we can produce an error like

```console
ERROR cheatcodes: non-empty stderr input=["cargo", "run", "--locked", "--manifest-path", "lib/risc0-ethereum/crates/ffi/Cargo.toml", "--bin", "risc0-forge-ffi", "-q", "prove", "/opt/actions-runner/_work/risc0/risc0/risc0-foundry-template/target/riscv-guest/methods/guests/riscv32im-risc0-zkvm-elf/release/is-even.bin", "0x0000000000000000000000000000000000000000000000000000000000bc614e"] stderr="\u{1b}[1m\u{1b}[31merror\u{1b}[0m\u{1b}[1m:\u{1b}[0m the lock file /opt/actions-runner/_work/risc0/risc0/risc0-foundry-template/lib/risc0-ethereum/Cargo.lock needs to be updated but --locked was passed to prevent this\nIf you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.\n"
```


bobbobbio:
This change seem pretty specific to `risc0-etherium` what if we did this by adding an `--ignore` flag to the script instead? If you want I could give it go myself.

capossele:
I mean this script is only used for risc0-ethereum and the foundry-template test, but sure adding an --ignore flag works as well

capossele:
@bobbobbio done in https://github.com/risc0/risc0/pull/2967/commits/ea4cc2c6ba9fd93399d72fc5f1213a1492d6b89a

.github/cargo_local_patch.py
```diff
@@ -71,14 +79,16 @@ def process_cargo_file(cargo_toml_path, dep_path_mapping, start_directory):
...
+def main(directory, dep_path_mapping, ignore_dirs):
+    cargo_toml_files = find_cargo_toml_files(directory, ignore_dirs)
     for cargo_toml_path in cargo_toml_files:
         process_cargo_file(cargo_toml_path, dep_path_mapping, directory)
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Modify dependencies in Cargo.toml files within a directory.")
     parser.add_argument("directory", help="Directory to search for Cargo.toml files")
+    parser.add_argument("--ignore", nargs="+", default=[],
+                        help="List of directories to ignore for patching")
```
bobbobbio, line 91: You should remove the `nargs="+"` I think, without it you are able to use the command like this:
`cargo_local_patch.py --ignore lib --ignore lib2 .` which I think makes more sense

.github/cargo_local_patch.py
```diff
@@ -3,9 +3,17 @@
...
+        ignore_dirs = []
+
     matches = []
-    for root, _, filenames in os.walk(start_path):
+    for root, dirs, filenames in os.walk(start_path):
+        # Skip directories specified in ignore_dirs
+        for ignore_dir in ignore_dirs:
+            if ignore_dir in dirs:
+                dirs.remove(ignore_dir)
+
```
bobbobbio, line 16: I don't think this works if one of `ignore_dirs` contains a `/`, maybe we fix that by doing this:
```python
root_abs = os.path.abspath(root)                                                                                                                                                                                                              
if any(root_abs.startswith(os.path.abspath(i)) for i in ignore):                                                                                                                                                                              
    dirs.clear()                                                                                                                                                                                                                              
    continue      
```

bobbobbio (changes requested):

.github/cargo_local_patch.py
bobbobbio: sorry and also add `action="append"`

bobbobbio:
> I mean this script is only used for risc0-ethereum and the foundry-template test, but sure adding an --ignore flag works as well
As it exists it seems useful to use in other contexts as a general tool when developing.

PR #2968, opened 2025-03-13
ZKVM-1222: Implement new SemVer checks

bobbobbio:
This patch implements the new SemVer checks we have discussed as a new `xtask` command

It first runs `cargo-semver-checks` via its Rust API normally, then it adds our check for patch releases by doing a new check.
This new check basically runs `cargo publish` on crates where the current version number matches the published version number and compares the version that would be published with the version that is published and if there are any changes, we require a bump of the version number.

This PR also enables the new checks in CI replacing the existing checks. It doesn't move to make it be required yet, I thought we would try it out for a bit before blocking commits behind it.


Cargo.toml
```diff
@@ -47,7 +47,7 @@ metal = "0.29"
 risc0-bigint2 = { version = "1.4.0", default-features = false, path = "risc0/bigint2" }
 risc0-binfmt = { version = "2.0.0", default-features = false, path = "risc0/binfmt" }
 risc0-build = { version = "3.0.0", default-features = false, path = "risc0/build" }
-risc0-build-kernel = { version = "1.4.0", default-features = false, path = "risc0/build_kernel" }
+risc0-build-kernel = { version = "2.0.0", default-features = false, path = "risc0/build_kernel" }
```
flaub, line 50: Curious about this bump

bobbobbio: Sorry I forgot to mention in the PR message, this was due to my change which added `non_exhaustive` to some type in this crate.

flaub (approved):

PR #2971, opened 2025-03-13
ZKVM-1224: Add ABI version number to ProgramBinary; Check version in zkVM

bobbobbio:
This adds a new header to the ProgramBinary. It stores it on disk as a list of enum values encoded using `postcard`. I wanted to use `bincode` at first, but it doesn't support `no_std` usage. It ignores values that it doesn't know about, potentially allowing for future versions of the header to still be decoded even if they contain new kinds of header values.

The value of this version is checked in the zkVM and an error is thrown if it doesn't match the one value we currently know about. This will prevent version 2.0.0 of the zkVM from attempting to execute guest programs built by newer version of the code. Also, this will allow future versions of zkVM to detect older versions of guest programs and handle that appropriately.

### DIFF
--- a/risc0/binfmt/src/elf.rs
+++ b/risc0/binfmt/src/elf.rs
@@ -20,6 +20,7 @@ use anyhow::{anyhow, bail, ensure, Context, Result};
 use elf::{endian::LittleEndian, file::Class, ElfBytes};
 use risc0_zkp::core::{digest::Digest, hash::sha::Impl};
 use risc0_zkvm_platform::WORD_SIZE;
+use serde::{Deserialize, Serialize};
 
 use crate::{Digestible as _, MemoryImage, SystemState, KERNEL_START_ADDR};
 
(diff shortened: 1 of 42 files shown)


risc0/binfmt/src/elf.rs
```diff
@@ -128,15 +129,162 @@ impl Program {
...
 
 const MAGIC: &[u8] = b"R0BF"; // RISC Zero Binary Format
-const VERSION: u32 = 1; // current version
+const BINARY_FORMAT_VERSION: u32 = 1; // RISC Zero Binary Format Version Number
 
-const VERSION_OFFSET: usize = MAGIC.len();
-const USER_LEN_OFFSET: usize = VERSION_OFFSET + WORD_SIZE;
-const USER_ELF_OFFSET: usize = USER_LEN_OFFSET + WORD_SIZE;
+#[derive(Serialize, Deserialize)]
+enum ProgramBinaryHeaderValueOnDisk {
```
SchmErik, line 135: do we want something like non-exhaustive here?

SchmErik: oh wait, this is private so it probably doesn't matter

SchmErik (approved):

flaub (approved):

flaub (approved):

PR #2972, opened 2025-03-14
refactor: Replace buffer array with named fields in WeierstrassCurve

crStiv:
Addresses TODO comment in WeierstrassCurve struct by:
- Replacing buffer array with named fields (prime, a, b)
- Updating constructor to use named fields
- Implementing as_u32s() with std::mem::transmute for compatibility
- Updating all buffer references throughout the codebase

This improves code readability and reduces potential for misuse while maintaining API compatibility.


bobbobbio:
Using `transmute` here is not desirable, and in my opinion incorrect. There is nothing that says the compiler can't reorder these fields.

PR #2974, opened 2025-03-14
Backport #2967

capossele:
When applying the `cargo_local_patch.py` during the `risc0-ethereum` CI workflow, we also modify the git submodules if a match is found. That by itself is not an issue; however since we added the `--locked` when running the `risc0-forge-ffi` binary, we can produce an error like

```console
ERROR cheatcodes: non-empty stderr input=["cargo", "run", "--locked", "--manifest-path", "lib/risc0-ethereum/crates/ffi/Cargo.toml", "--bin", "risc0-forge-ffi", "-q", "prove", "/opt/actions-runner/_work/risc0/risc0/risc0-foundry-template/target/riscv-guest/methods/guests/riscv32im-risc0-zkvm-elf/release/is-even.bin", "0x0000000000000000000000000000000000000000000000000000000000bc614e"] stderr="\u{1b}[1m\u{1b}[31merror\u{1b}[0m\u{1b}[1m:\u{1b}[0m the lock file /opt/actions-runner/_work/risc0/risc0/risc0-foundry-template/lib/risc0-ethereum/Cargo.lock needs to be updated but --locked was passed to prevent this\nIf you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.\n"
```

### DIFF
--- a/.github/cargo_local_patch.py
+++ b/.github/cargo_local_patch.py
@@ -3,9 +3,18 @@ import toml
 import os
 import fnmatch
 
-def find_cargo_toml_files(start_path):
+def find_cargo_toml_files(start_path, ignore_dirs=None):
+    if ignore_dirs is None:
+        ignore_dirs = []
+
     matches = []
-    for root, _, filenames in os.walk(start_path):
+    for root, dirs, filenames in os.walk(start_path):
+        # Skip directories specified in ignore_dirs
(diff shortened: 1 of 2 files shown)


bobbobbio:
thanks for making this but I'll do it myself by merging main into the release branch

PR #2975, opened 2025-03-14
ZKVM-1227: extern C for rust bigint2 blob fns

iddo-bentov:
Tested with external Rust and C projects (it doesn't seem to affect the cycle count of Rust programs)
For C code, this requires `use risc0_bigint2::field::*` in the guest program.


risc0/bigint2/src/field/mod.rs
```diff
@@ -182,7 +182,9 @@ pub fn modsub_384_unchecked(
 
 // These "checked" versions verify that `result < modulus`
 
-pub fn modadd_256(
+#[stability::unstable]
+#[no_mangle]
+pub extern "C" fn modadd_256(
```
flaub, line 187: I suggest we prefix extern "C" calls to make them globally unique.

```suggestion
pub extern "C" fn risc0_bigint_modadd_256(
```

tzerrell (commented):
Great, and thank you for checking this works on Rust programs too.

I think we should probably expose the "unchecked" versions externally as well, they are sometimes necessary for good performance and we may end up needing them if we end up patching precompiles into more C code.

risc0/bigint2/src/field/mod.rs
iddo-bentov: Ok, but then it won't work with patched crates (that are pure Rust) that are already using these functions, such as bn254
An alternative is to duplicate these functions (that is, have one version that's extern "C", and another version that's just for rust, where the body of these two versions is the same). @SchmErik also looked at the current approach and so far it seemed ok to him.

iddo-bentov:
> I think we should probably expose the "unchecked" versions externally as well, they are sometimes necessary for good performance and we may end up needing them if we end up patching precompiles into more C code.

No problem with me:)  For now I didn't even expose the generic deg2 mul functions, because the custom deg2 xxone functions are likely the only ones that'd be needed.

risc0/bigint2/src/field/mod.rs
flaub: This will break users downstream if the symbol names are not unique enough. 

Extern "C" functions need to be globally unique. You can have the extern "C" functions call the rust functions, that would be fine.

flaub (changes requested):
Need to make extern "C" symbols globally unique. Prefixing them is a common practice to help ensure this. Otherwise, downstream users get confusing and conflicting linking issues.

risc0/bigint2/src/field/mod.rs
iddo-bentov: Ok, I changed the extern "C" functions to use risc0_bigint_ prefix.
@tzerrell I did it in the risc0-bigint2 crate as you recommended (and not in zkvm-platform), though as mentioned the setup for 3rd-party crates will be more involved.

risc0/bigint2/src/field/syscall_field.rs
```diff
@@ -0,0 +1,375 @@
...
+    modulus: &[u32; FIELD_384_WIDTH_WORDS],
+    result: &mut [[u32; FIELD_384_WIDTH_WORDS]; EXT_DEGREE_2],
+) {
+    extfield_deg2_add_384_unchecked(lhs, rhs, modulus, result);
+}
+
+/*
+#[stability::unstable]
+#[no_mangle]
+pub extern "C" fn risc0_bigint_extfield_deg2_mul_256_unchecked(
```
tzerrell, line 128: Why are the (non-specialized) extension field multiplies commented out? I think we should either add them or drop them (and replace with a comment about why they're not supported)

risc0/bigint2/src/field/syscall_field.rs
tzerrell, line 1: To me `syscall_field` is an unintuitive name for this file, because the plain `field` file also just provides useful wrappers for syscalls. Perhaps `extern_field` or `field_externs` would be better?

risc0/bigint2/src/field/syscall_field.rs
iddo-bentov: We did the deg4 function to demontrate the generic extfield mul, it's not actually needed for common use-cases.
The generic deg2 extfield mul is commented out because the custom xxone variant is more efficient and used in common use-cases.
So I just removed the extern "C" versions of these functions because they aren't interesting enough (and to avoid confusion), the rust versions should still be kept.
So I guess that for now I will just remove the commented-out code completely.

risc0/bigint2/src/field/syscall_field.rs
iddo-bentov: Ok I will rename it to `extern_field`

risc0/bigint2/src/field/mod.rs
```diff
@@ -15,6 +15,8 @@
 #[cfg(test)]
 mod tests;
 
+pub mod syscall_field;
```
flaub, line 18: Could call it `ffi` (Foreign-function interface)


```suggestion
pub mod ffi;
```

flaub (approved):

risc0/bigint2/src/field/syscall_field.rs
flaub: Maybe `ffi`?

risc0/bigint2/src/field/mod.rs
iddo-bentov: ok, for now I changed it to `ffi_field`, I'm not sure what convention is considered best because later we could also have `ffi_ec` in the other directory of this bigint2 crate.

tzerrell (approved):

risc0/bigint2/src/field/mod.rs
tzerrell: If we wanted this for EC we could put it in the same file; there's no concern about name collisions as we have to prefix everything anyway. So I think ffi by itself would be fine

flaub: Since this is already nested in a subdir, I think `ffi` is sufficient.

PR #2977, opened 2025-03-14
ZKVM-1229: Add unchecked/checked versions of ext field ops

tzerrell:
Split the extension field precompile operations into checked and unchecked versions that do or don't (respectively) check whether any component of the result is >= the base field prime.

The unchecked version is still constrained to give a result equivalent to the checked result modulo the prime modulus; the check is just to ensure that the result is a representation of the field element "normalized" to its unique value `[0, p)`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -436,25 +436,10 @@ jobs:
       - uses: ./.github/actions/sccache
         with:
           key: ${{ matrix.os }}-default
-      - name: Install cargo-semver-checks
-        uses: risc0/cargo-install@9f6037ed331dcf7da101461a20656273fa72abf0
-        with:
-          crate: cargo-semver-checks
-          version: "0.36.0"
-      - name: run cargo semverchecks on ${{ github.event.pull_request.base.sha }}
-        run: |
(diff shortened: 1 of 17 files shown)


iddo-bentov (approved):
Looks good.
I get approximately 3% more user cycles with bls12_381 test vector (4.5% more total cycles, though that's more noisy).
Also I tried to unroll the is_less loop because the loop length is part of the function name and it seems to have a noticable effect on the cycle count (it's curious that the compiler doesn't seem to optimize that).

tzerrell:
The `is_less` unrolling also gave a minor perf improvement to the bn254 case. I'm also surprised the compiler didn't do this for us, but since it didn't, I'm glad you tried unrolling

bobbobbio:
> @bobbobbio I'm still figuring out how the new SemVer CI test works -- is the right thing to do when we have a SemVer-affecting change to bump the version of the affected crate in the PR? (In this case, we added new `unchecked` functions so I bumped the minor version)

That's correct but right now the check is broken and I'm fixing it. You shouldn't need a version bump yet since we haven't release v2

PR #2981, opened 2025-03-16
refactor: Improve CPP version parsing code structure

GarmashAlex:
This PR improves the code structure for handling special version formats
in the InstallCommand. It specifically addresses the TODO about moving away
from date-based version tags to semver for CPP components by:

1. Using a clearer pattern matching approach for component name checking
2. Adding a more descriptive comment about the future plan
3. Structuring the code to make it easier to add more special version
   format handlers for other components

The functionality remains unchanged, but the code is now more maintainable
and better documents the intention to move to semver in the future.

### DIFF
--- a/rzup/src/cli/commands.rs
+++ b/rzup/src/cli/commands.rs
@@ -74,10 +74,13 @@ impl InstallCommand {
             return Ok(None);
         };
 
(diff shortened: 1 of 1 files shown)


bobbobbio:
This is not an improvement in my view.

PR #2987, opened 2025-03-17
ZKVM-1238: update to circom 2.2.2 and disable x86 asm witness generation

mars-risc0:


mars-risc0:
obsoleted by 'proving-on-aarch64' branch

PR #2988, opened 2025-03-18
ZKVM-1239: [risc0-build] Improve error messages; Make C++ toolchain optional for building guests

bobbobbio:


risc0/build/src/lib.rs
```diff
@@ -410,8 +411,21 @@ pub(crate) fn cargo_command_internal(subcmd: &str, guest_info: &GuestInfo) -> Co
...
+            cmd.env("CC", toolchain_path.join("bin/riscv32-unknown-elf-gcc"));
+        } else {
+            // If you aren't compiling any C/C++ code, it might be just fine to not have a C++
+            // toolchain installed, but if you are then your compilation will surely fail. To avoid
+            // a potentially confusing error message, set the CC path to a bogus path that will
+            // hopefully make the issue obvious.
+            cmd.env(
+                "CC",
+                "/no_risc0_cpp_toolchain_installed_run_rzup_install_cpp",
+            );
```
flaub, line 424: Was going to suggest something like this, LGTM.

PR #2989, opened 2025-03-18
ZKVM-1240: Add missing constraint in rv32im v2

jacobdweightman:
companion of https://github.com/risc0/zirgen/pull/209


flaub (changes requested):
How did you run bootstrap? There shouldn't be any keccak changes and the rv32im changes don't make sense (there are new files, there should only be changes to existing files).

jacobdweightman:
Oh, I see. I needed to run `cargo bootstrap rv32im-v2 --output ~/code/risc0/` 🤦

flaub (approved):

PR #2990, opened 2025-03-18
ZKVM-1241: Support for dumping and replaying segments

flaub:
* During execution, if `RISC0_DUMP_PATH` is set and an error occurs, serialize the current segment and write it to the specified path.
* `r0vm --segment` can be used to replay this segment later, possibly with extra debug tracing.


risc0/circuit/rv32im/src/execute/segment.rs
```diff
@@ -46,3 +54,48 @@ pub struct Segment {
...
+            |_| Ok(()),
+        )?;
+        Ok(())
+    }
+}
+
+struct SegmentSyscallHandler<'a> {
+    segment: &'a Segment,
+    read_pos: AtomicUsize,
+    write_pos: AtomicUsize,
```
bobbobbio, line 86: Guess this could have been `Cell<usize>` I think, but this is fine

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -309,6 +322,50 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
...
+    fn dump_segment(
+        &mut self,
+        po2: usize,
+        segment_threshold: u32,
+        index: u64,
+    ) -> anyhow::Result<()> {
+        if let Some(dump_path) = std::env::var_os("RISC0_DUMP_PATH") {
+            tracing::error!(
+                "Execution failure, saving segment to {}:",
+                dump_path.to_string_lossy()
```
bobbobbio, line 334: its better to use `.display()` in this situation

bobbobbio (approved):

risc0/circuit/rv32im/src/execute/segment.rs
flaub: I couldn't figure out how to do `borrow_mut` without `self` also being `mut`.

risc0/circuit/rv32im/src/execute/executor.rs
flaub: Actually, this is an `OsString` and that doesn't have `display()`

bobbobbio: Whoops my mistake

bobbobbio (approved):
LGTM

PR #2996, opened 2025-03-19
Remove bytemuck_derive dependency

bobbobbio:
It is breaking the build due to new dependency on Rust 1.85.0 and providing only marginal benefit to the code anyway.

We can always revert it in the future if / when we support Rust 1.85.0


bobbobbio:
@jacobdweightman pinned the bytemuck_derive version in his commit

MarcusWentz:
Seems related to this issue with RISC Zero templates generated from cargo crate `cargo-risczero` here:
https://github.com/risc0/risc0/issues/3000

PR #2997, opened 2025-03-19
ZKVM-1245: Add a `cargo xtask` command to update repository lock files

bobbobbio:
I found myself running this sort of thing via a bash script all the time, just thought it would be easier to make our own tool so others can run it themselves.


flaub:
I usually just do:
```
$(cd examples; cargo c)
$(cd benchmarks; cargo c)
```

bobbobbio:
The advantage of doing it this way is that you don't have to build anything, which can save a lot of time depending on the situation.

flaub (approved):

PR #3001, opened 2025-03-20
ZKVM-1249: Constrain bigint memory accesses based on mode

flaub:


risc0/bigint2/methods/guest/src/bin/raw_test.rs
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use include_bytes_aligned::include_bytes_aligned;
+
+#[allow(unused)]
```
bobbobbio, line 17: this import is used it seems, do we need the attribute?

bobbobbio (commented):
Is there a mechanism that prevents a user program from doing an `ecall` directly (and setting machine mode themselves)?

Also, I guess we decided to not worry about the bigint program being in kernel memory? I guess this impact of this is a lot lower.

bobbobbio:
> Is there a mechanism that prevents a user program from doing an ecall directly (and setting machine mode themselves)?

I see, this is covered by machine mode vs user mode

bobbobbio (approved):

PR #3006, opened 2025-03-20
Bump guest rust toolchain to 1.85.0

SchmErik:
This PR updates the guest rust toolchain to use 1.85.0. A new guest builder with tag r0.1.85.0 has been uploaded to dockerhub. I've updated the guest builder to base off of ubuntu 22.04 because 1.85.0 requires a later glibc version that's not supported by 20.04.

Another item to note about 1.85 guest toolchain: crates with `cfg(test)` get compiled on the guest, these warnings are emitted:

```
risc0-zkvm-methods-guest: warning: unexpected `cfg` condition name: `test`
risc0-zkvm-methods-guest:    --> /Users/erik/risc0/rzup/src/lib.rs:125:11
risc0-zkvm-methods-guest:     |
risc0-zkvm-methods-guest: 125 |     #[cfg(test)]
risc0-zkvm-methods-guest:     |           ^^^^
risc0-zkvm-methods-guest:     |
risc0-zkvm-methods-guest:     = help: consider using a Cargo feature instead
risc0-zkvm-methods-guest:     = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
risc0-zkvm-methods-guest:              [lints.rust]
risc0-zkvm-methods-guest:              unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
risc0-zkvm-methods-guest:     = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(test)");` to the top of the `build.rs`
risc0-zkvm-methods-guest:     = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
```

According to https://doc.rust-lang.org/nightly/rustc/check-cfg.html#well-known-names-and-values,

> Starting with 1.85.0, the test cfg is consider to be a "userspace" config
> despite being also set by rustc and should be managed by the build system
> itself.

Therefore, I've added this lint to any crate that ends up being built on
the guest.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,3 +74,6 @@ opt-level = 3
 
 [profile.release]
 lto = true
+
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(test)'] }
--- a/risc0/bigint2/Cargo.toml
+++ b/risc0/bigint2/Cargo.toml
@@ -38,3 +38,6 @@ default = []
 num-bigint-dig = ["dep:num-bigint-dig"]
 num-bigint = ["dep:num-bigint"]
 unstable = []
+
+[lints]
+workspace = true
--- a/risc0/binfmt/Cargo.toml
+++ b/risc0/binfmt/Cargo.toml
@@ -47,3 +47,6 @@ std = [
   "tracing/log",
   "tracing/std",
 ]
+
+[lints]
+workspace = true
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -143,7 +143,7 @@ fn create_dockerfile(manifest_path: &Path, temp_dir: &Path, guest_info: &GuestIn
     .join(" ");
 
     let mut build = DockerFile::new()
-        .from_alias("build", "risczero/risc0-guest-builder:r0.1.81.0")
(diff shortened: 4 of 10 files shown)


bobbobbio:
I'm wondering if we should also update `cargo risczero` to add the ignore for that warning to the guest Cargo.toml? Or should we explore making `risc0-build` ignore that warning via CLI arguments to `cargo`?

SchmErik:
> I'm wondering if we should also update `cargo risczero new` to add the ignore for that warning to the guest `Cargo.toml`?
> Or should we explore making `risc0-build` ignore that warning via CLI arguments to `cargo`?

The risc0-build solution might be better but it kind of feels like an overreach in that users cannot unset this...

flaub:
We should be building rustc with static crt, so I'm not sure why you require a newer glibc.

bobbobbio:
> > I'm wondering if we should also update `cargo risczero new` to add the ignore for that warning to the guest `Cargo.toml`?
> > Or should we explore making `risc0-build` ignore that warning via CLI arguments to `cargo`?
> 
> The risc0-build solution might be better but it kind of feels like an overreach in that users cannot unset this...

Is there a reason why users would want to unset it?

PR #3008, opened 2025-03-20
ZKVM-1257: Update bigint2 version to 2.0.0

tzerrell:
I think semver requires us to update the major version given the renames in https://github.com/risc0/risc0/pull/3003. Sorry for the churn


tzerrell:
Nevermind -- the removed functions were in the `unstable` feature and don't require this version update

PR #3009, opened 2025-03-20
ZKVM-1258: Disable CI docker login on Mac

bobbobbio:
We don't want to support docker on Mac

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -244,6 +244,7 @@ jobs:
         with:
           key: ${{ matrix.os }}-${{ matrix.feature }}
       - name: Login to Docker Hub
+        if: matrix.os == 'Linux'
         uses: docker/login-action@v3
         with:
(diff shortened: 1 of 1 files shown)


mars-risc0:
FYI, https://github.com/risc0/risc0/pull/2987 will make stark_to_snark work on aarch64

bobbobbio:
> FYI, #2987 will make stark_to_snark work on aarch64. Perhaps it'd be better to get docker installed on all the Mac workers?

Is the plan to continue to use `docker` for `stark_to_snark`? If it is then I guess it makes sense to install `docker` like you say.

bobbobbio:
I'm going to see if we can install docker on the Mac workers missing it instead since it does seem like at least for now we are going to still use docker for `stark_to_snark`

bobbobbio:
We instead decided that we don't want to support docker on Mac

flaub (approved):

PR #3018, opened 2025-03-26
ZKVM-1263: Add Poseidon254 CUDA implementation

flaub:


risc0/sys/kernels/zkp/cuda/supra/api.cu
```diff
@@ -25,8 +27,6 @@ sppark_poseidon2_fold(poseidon_out_t* d_out, const poseidon_in_t* d_in, size_t n
   } catch (const cuda_error& e) {
     gpu.sync();
     return RustError{e.code(), e.what()};
-  } catch (...) {
-    return RustError(cudaErrorUnknown, "Generic exception");
```
dot-asm, line 29: Just in case for reference. The reason for removing this catch-all is that it's dead code, the above try block generates no non-cuda_error exceptions.

bobbobbio: thanks for providing the insight, good to know in case it comes up later

PR #3036, opened 2025-03-28
ZKVM-1281: rzup add CI recommendations

SchmErik:
I've added this to rzup's readme but this might be better placed in the developer documentation. @pdg744 or anyone else: let me know if you have any thoughts on this...

### DIFF
--- a/rzup/README.md
+++ b/rzup/README.md
@@ -149,3 +149,24 @@ rzup stores its installations in:
 When communicating with GitHub, it tries using authentication. This can be useful to get around
 rate-limiting. It attempts to get a token from the `GITHUB_TOKEN` environment variable, then from
 ~/.config/gh/hosts.yml.
+
+## Using `rzup` in CI
+
+When using `rzup` in CI environments, it's important to pin each component to
+the desired version. Invoking `rzup install` will download the latest software
+released by RISC Zero and may cause CI environments to suddenly fail when new
+versions of components are published.
+
+In order to avoid this, `rzup` should be run in CI using the following commands:
+
+```
+rzup install rust 1.81.0
+rzup install cpp 2024.1.5 # only required if the guest uses any C++ code or rust crates that bind to C++ code.
+rzup install r0vm 2.0.0 # should be the same version as cargo-risczero and risc0-zkvm crate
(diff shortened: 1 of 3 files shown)


bobbobbio:
The verbiage looks fine to me.

I'm slightly worried about how we are going to remember to update these version numbers in the future

bobbobbio:
Maybe good to mention the same thing on this page:
https://dev.risczero.com/api/zkvm/install#installation-for-x86-64-linux-and-arm64-macos

bobbobbio (approved):

rzup/README.md
```diff
@@ -149,3 +149,24 @@ rzup stores its installations in:
...
+the desired version. Invoking `rzup install` will download the latest software
+released by RISC Zero and may cause CI environments to suddenly fail when new
+versions of components are published.
+
+In order to avoid this, `rzup` should be run in CI using the following commands:
+
+```
+rzup install rust 1.81.0
+rzup install cpp 2024.1.5 # only required if the guest uses any C++ code or rust crates that bind to C++ code.
+rzup install r0vm 2.0.0 # should be the same version as cargo-risczero and risc0-zkvm crate
```
hmrtn, line 165: Just pointing out: `rzup install r0vm <VERSION>` and `rzup install cargo-risczero <VERSION>` are actually redundant, as the releases are packaged together, right? No harm in being pedantic, though.

bobbobbio: My thinking around this is that if we tell users to have both lines, it frees us up in the future to possibly split the packages apart.

SchmErik:
I'd like to hold off on landing this becuase I'd like input from @pdg744 on if this is the right place or not...

flaub:
I think a page on the dev docs would be better. I doubt many users actually read this README. 

A link from https://dev.risczero.com/api/zkvm/install would be good, or we could just add it to that page.

bobbobbio (approved):

SchmErik:
> The verbiage looks fine to me.
> 
> I'm slightly worried about how we are going to remember to update these version numbers in the future

oooh, this is a great point - these version numbers were meant to be an example but I can see how someone would take a look at this and copy paste verbatim..

PR #3037, opened 2025-03-28
ZKVM-1282: risc0-build: make docker container tag configurable

SchmErik:
This change allows users to set the docker container used for reproducible programmatically or by setting the `RISC0_DOCKER_CONTAINER_TAG` environment variable.

### DIFF
--- a/risc0/build/src/config.rs
+++ b/risc0/build/src/config.rs
@@ -36,6 +36,7 @@ pub struct DockerOptions {
     pub env: Vec<(String, String)>,
 
     /// Docker container tag to use
+    #[builder(setter(into, strip_option))]
     pub docker_container_tag: Option<String>,
 }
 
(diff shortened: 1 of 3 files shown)


risc0/build/src/config.rs
```diff
@@ -32,6 +34,9 @@ pub struct DockerOptions {
 
     /// Additional environment variables for the build container.
     pub env: Vec<(String, String)>,
+
+    /// Docker container tag to use
```
bobbobbio, line 38: do you need to add an attribute for the `Builder` derive?

bobbobbio (approved):
Since this is the addition of new public stuff maybe we should bump the minor version of the crate. `semver-checks` doesn't require a minor version bump in this case, but maybe we would prefer it.

Since environment variables are global names, I'm wondering if we should prefix it like `RISC0_BUILD_DOCKER_CONTAINER_TAG`, or if just `RISC0_DOCKER_CONTAINER_TAR` is specific enough

risc0/build/src/config.rs
SchmErik: thanks for catching this

PR #3040, opened 2025-03-30
Update lib.rs

leopardracer:


bobbobbio:
this change doesn't make any sense, there is not `Executor2` on main anymore.

PR #3041, opened 2025-03-30
Rename ReceiptMetadata to ReceiptClaim

leopardracer:


bobbobbio:
No explanation provided about why we should do this

bobbobbio:
Thank you for explaining, but I don't think it makes sense. There is already a `ReceiptClaim` struct and this type is distinct type, so calling this the same thing it confusing. Also my understanding is this information isn't part of the "claim". To rename this struct is also a breaking change we wouldn't do lightly.

flaub:
Claims and metadata are two distinct things. Metadata is not committed to, but the claim is.

PR #3044, opened 2025-03-31
Update README.md and CONTRIBUTING.md to use xfast instead of xtask

maximevtush:
**Description:**
This PR updates the documentation to replace `cargo xtask` commands with `cargo xfast` commands in both CONTRIBUTING.md and the browser-verify example README. This change reflects the updated tooling nomenclature for the project's build system.


bobbobbio:
This is not a change we want.

PR #3045, opened 2025-03-31
Move zip download tmpdir to $HOME/.risc0

zeevick10:


bobbobbio:
This change seems nonsensical to me.

flaub:
This can prevent issues with moving a temporary directory over into its final destination.

PR #3055, opened 2025-04-01
ZKVM-1290: Statically link in `liblzma`

bobbobbio:
No need to depend on the system provided version of this library on Linux, and on Mac it creates an even greater issue of somehow providing this library to users.

See https://github.com/risc0/risc0/issues/3054

Issue #3054, opened 2025-04-01
r0vm-2.0.0 depends on homebrew-installed liblzma.5.dylib dynamic library on macOS


kubkon:
### Is there an existing issue?

- [x] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [x] This is not a support question.

### Description of bug

Examining dynamic dependencies of released `r0vm` reveals:

```
$ otool -L ./r0vm
r0vm:
        /opt/homebrew/opt/xz/lib/liblzma.5.dylib (compatibility version 12.0.0, current version 12.2.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1700.255.0)
        /System/Library/Frameworks/Metal.framework/Versions/A/Metal (compatibility version 1.0.0, current version 343.14.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1345.100.2)
        /System/Library/Frameworks/CoreGraphics.framework/Versions/A/CoreGraphics (compatibility version 64.0.0, current version 1774.4.3)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 2420.0.0)
        /usr/lib/libobjc.A.dylib (compatibility version 1.0.0, current version 228.0.0)
        /usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
```

Firstly, I don't think you specified in your docs or release readme/changelog that the new cut of `r0vm` depends on `xz`, so the user who doesn't have it installed *and* in the homebrew-installation path will be greeted with a cryptic error of:

```
$ ./r0vm --version
dyld[14104]: Library not loaded: /opt/homebrew/opt/xz/lib/liblzma.5.dylib
  Referenced from: <BBFD7461-C4F8-32AA-8B86-656CAC29323C> ./r0vm
  Reason: tried: '/opt/homebrew/opt/xz/lib/liblzma.5.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/xz/lib/liblzma.5.dylib' (no such file), '/opt/homebrew/opt/xz/lib/liblzma.5.dylib' (no such file)
fish: Job 1, './r0vm --version' terminated by signal SIGABRT (Abort)
```

[...]

PR #3056, opened 2025-04-01
backport release-2.0: ZKVM-1290: Statically link in `liblzma` (#3055)

bobbobbio:
No need to depend on the system provided version of this library on Linux, and on Mac it creates an even greater issue of somehow providing this library to users.

See https://github.com/risc0/risc0/issues/3054


bobbobbio:
I'm not fixing this SemVer checks in this PR and instead am going to fix them in https://github.com/risc0/risc0/pull/3053

PR #3069, opened 2025-04-03
chore: removed redundant `&` in function calls

barajeel:
`lhs`, `rhs`, and `modulus` in several `unchecked::*` function calls for 384-bit operations were unnecessarily referenced. since they are already `&[u32; FIELD_384_WIDTH_WORDS]`, the extra `&` is redundant.

### DIFF
--- a/risc0/bigint2/src/field/mod.rs
+++ b/risc0/bigint2/src/field/mod.rs
@@ -52,7 +52,7 @@ pub fn modadd_384(
     modulus: &[u32; FIELD_384_WIDTH_WORDS],
     result: &mut [u32; FIELD_384_WIDTH_WORDS],
 ) {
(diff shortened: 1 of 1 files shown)


flaub:
@tzerrell @iddo-bentov clippy should be catching this, do we have clippy enabled for this crate?

tzerrell:
I think what's happening here is that clippy is running on this crate, but not with `-F unstable` and so is missing most things. A run of `cargo clippy -p risc0-bigint2` says everything passes, but `cargo clippy -F unstable -p risc0-bigint2` turns up a number of changes to make.

I can put together another PR incorporating this and making the other suggested clippy changes, and look into whether we can get an `-F unstable` clippy test into CI as well.

tzerrell:
I've included this commit in https://github.com/risc0/risc0/pull/3070 with the other clippy changes, so closing this PR.

PR #3070, opened 2025-04-03
ZKVM-1308: Fix & run Clippy for risc0-bigint2

tzerrell:
This adds a run with `-F unstable` to the clippy CI job. It also adds the corresponding fixes to risc0-bigint2. This caught a number of overlooked issues that only affected `unstable` functionality.

This includes the commit from https://github.com/risc0/risc0/pull/3069, which is a change clippy would have caught.


.github/workflows/main.yml
```diff
@@ -206,6 +206,8 @@ jobs:
           key: ${{ matrix.os }}-${{ matrix.feature }}
       - name: "root: cargo clippy"
         run: cargo clippy -F $FEATURE --all-targets
+      - name: "root: cargo clippy -F unstable"
+        run: cargo clippy -F $FEATURE -F unstable --all-targets
```
bobbobbio, line 210: I don't think we should have to run it with and without the "unstable" feature, probably just with it is good. Features are _supposed_ to be only additive (as in they shouldn't remove any code).
Making the CI job longer is also a non-goal.

bobbobbio (changes requested):

.github/workflows/main.yml
bobbobbio: although this is now making me question why we run clippy both with and without the CUDA feature flag

flaub: We have different code paths depending on feature flags. While it's true that they are additive from the workspace perspective, they still can cause exclusive sections of code to be enabled or disabled.

flaub: I personally think this is a reasonable change for this specific issue.

flaub (approved):

.github/workflows/main.yml
flaub: I also think we can just add it onto this existing clippy

bobbobbio: okay, I guess I'm wrong. You can totally have stuff compiled in only if the feature isn't enabled, and maybe that still fits within best practices, but I don't know how worried we are about that exactly

bobbobbio (approved):

flaub:
I would combine `-F unstable` to the existing clippy action

PR #3072, opened 2025-04-04
ZKVM-1309: use semver::VersionReq for comparisons in rzup build

### LINKED ISSUE #3071: rzup: building rust-toolchain <= 1.81.0 fails with invalid LLVM pass

### Description of bug

When attempting to build a Rust toolchain from source using `rzup build` fails with error `error: failed to run LLVM passes: unknown pass name 'lower-atomic'`. This LLVM pass was not introduced until after Rust version 1.81.0 (or LLVM version 19), and previously was known as `loweratomic` which is handled in `rzup` but we still try and use the new naming instead.

### Steps to reproduce

```
$ rzup build rust -t r0.1.81.0
...
   Compiling libc v0.2.155
   Compiling memchr v2.5.0
   Compiling std v0.0.0 (/home/kubkon/.risc0/tmp/build-rust-toolchain/library/std)
   Compiling rustc-std-workspace-core v1.99.0 (/home/kubkon/.risc0/tmp/build-rust-toolchain/library/rustc-std-workspace-core)
error: failed to run LLVM passes: unknown pass name 'lower-atomic'

error: could not compile `rustc-std-workspace-core` (lib) due to 1 previous error
error: could not compile `libc` (lib) due to 1 previous error
error: could not compile `core` (lib) due to 1 previous error
error: could not compile `compiler_builtins` (lib) due to 1 previous error
Build completed unsuccessfully in 0:29:34
```

kubkon:
Fixes #3071 

According to [docs.rs/semver](https://docs.rs/semver/latest/semver/), `semver::VersionReq` should be used for semantic version requirements (hence comparisons) since using standard Rust comparison operators directly on `semver::Version` leads to incorrect results if `semver::BuildMetadata` is present.

Consider this example

```
let mut v1 = semver::Version::new(1,81,0);
v1.build = semver::BuildMetadata::new("72a7a1c4").unwrap();

let min = semver::Version::new(1,81,0);
println!("{v1} > {min} := {}", v1 > min);

let req = semver::VersionReq::parse(">1.81.0").unwrap();
println!("{v1}{req} := {}", req.matches(&v1));
```

which yields

```
1.81.0+72a7a1c4 > 1.81.0 := true
1.81.0+72a7a1c4>1.81.0 := false
```

### DIFF
--- a/rzup/src/build.rs
+++ b/rzup/src/build.rs
@@ -219,7 +219,7 @@ pub fn build_rust_toolchain(
         message: "./x build".into(),
     });
 
-    let req = semver::VersionReq::parse(">1.81.0")?;
+    let req = semver::VersionReq::parse(">=1.82.0")?;
     let lower_atomic = if req.matches(&version) {
         "-Cpasses=lower-atomic"
     } else {
--- a/rzup/src/build.rs
+++ b/rzup/src/build.rs
@@ -219,7 +219,8 @@ pub fn build_rust_toolchain(
         message: "./x build".into(),
     });
 
-    let lower_atomic = if version > semver::Version::new(1, 81, 0) {
+    let req = semver::VersionReq::parse(">1.81.0")?;
+    let lower_atomic = if req.matches(&version) {
         "-Cpasses=lower-atomic"
     } else {
         "-Cpasses=loweratomic"
--- a/rzup/src/lib.rs
+++ b/rzup/src/lib.rs
@@ -324,6 +324,7 @@ impl Rzup {
 mod tests {
     use super::*;
     use crate::distribution::Os;
+    use std::collections::HashMap;
     use std::convert::Infallible;
     use std::io::Write as _;
     use std::net::SocketAddr;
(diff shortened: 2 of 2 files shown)


bobbobbio:
Thank you for the fix. I agree using `VersionReq` seems like the right way to do this. I'm going to push a few tests for this to the PR

kubkon:
> I change the logic slightly, it now applies the logic for Rust version 1.81.0 for 1.81.X, I don't think there is (or ever will be) a version that is `>1.81.0 <1.82.0` but if there ever is it should follow the `1.81.0` rules
> 
> It would be good if @SchmErik could take a look

If you need me to test it locally, happy to do so. For more context, the reason I go to such lengths as to build the toolchain from source is that I am running NixOS so standard ld.so paths do not apply and hence it's generally easier to build things from source when standard `patchelf` dance is difficult/does not apply.

SchmErik (approved):
Looks great. Thanks for catching this. `>=1.82.0` is the right call

PR #3073, opened 2025-04-04
fix: removed redundant & in function calls

zeevick10:


bobbobbio:
@tzerrell Is this another place clippy is missing? first place seems like it might be under the `num-bigint` feature

tzerrell:
This PR doesn't compile; these borrows are necessary.

> @tzerrell Is this another place clippy is missing? first place seems like it might be under the `num-bigint` feature

That said, we are not actually running clippy on these files in CI, because these are guest methods and hence are in their own workspace and not part of `--all-targets` from the root directory. Perhaps it would be worth getting clippy into CI for guests as a separate issue?

PR #3075, opened 2025-04-04
ZKVM-1311: Split RSA precompile into multiple modmuls

tzerrell:
Instead of running a single monolithic RSA precompile, split it into its constituent modmuls.


tzerrell:
Corresponding zirgen changes here: https://github.com/risc0/zirgen/pull/217

bobbobbio (approved):

PR #3076, opened 2025-04-05
ZKVM-1312: Add a program that tries to do worst-cost analysis on BigInt program

bobbobbio:
It tries to figure out the worst-case in terms of paging cycles.


flaub:
I think this could probably live in `risc0-circuit-rv32im` next to the executor.

bobbobbio:
> I think this could probably live in `risc0-circuit-rv32im` next to the executor.



> I agree that this doesn't feel like it has a great place -- the `risc0-circuit-rv32im` sounded reasonable enough to me.
> 
> Otherwise LGTM

I've moved the code into the `risc0-circuit-rv32im` crate and thus allows me to code-share. I moved the binary to that package also.

risc0/circuit/rv32im/src/execute/bigint/analyze.rs
```diff
@@ -0,0 +1,173 @@
...
+            page_in = true;
+            *num_pages += 1;
+        }
+
+        println!("load_arena({arena}, {offset}, {count}), page_in = {page_in}");
+    }
+
+    fn track_store_arena(&mut self, arena: u32, offset: u32, count: u32) {
+        self.cycles += PAGE_CYCLES;
+        self.cycles += 22 * NODE_CYCLES * 2;
```
SchmErik, line 92: I'm not sure about this line. Could you describe what's going on here? Why 22 instead of 21? Is store arena in kernel memory?

risc0/circuit/rv32im/src/execute/bigint/analyze.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn track_load_arena(&mut self, arena: u32, offset: u32, count: u32) {
+        let mut page_in = false;
+
+        let num_pages = self.loaded_arenas.entry(arena).or_insert(0);
+
+        while *num_pages < max_pages_for_io(offset + count) {
+            // load could go all the way up the tree
+            self.cycles += PAGE_CYCLES;
+            self.cycles += 21 * NODE_CYCLES;
```
SchmErik, line 82: nit: might be more readable if you declare the `21 * NODE_CYCLES` as a constant.

risc0/circuit/rv32im/src/execute/bigint/analyze.rs
bobbobbio: This being `22` instead of `21` was just a mistake I think. The `21` think is the height of the tree - 1, and I originally had `22` for all of these but missed fixing this one.
It isn't in kernel memory.

bobbobbio:
> I think it looks good. I had a question about the 22 vs 21 multiplication but other than that it looks good. This might be rooted in my lack of knowledge on where programs are stored... Approving - but it would be nice if you could answer the question (if anything for my own curiosity).

The `21` thing was just a mistake, and in fact those extra lines were inflating the estimate by a good amount

PR #3083, opened 2025-04-07
ZKVM-1313: Add 2.0 verifier contract deployment to docs

nategraf:
This PR adds documentation for deployed verifier addresses, generated from the `deployment.toml` file in https://github.com/risc0/risc0-ethereum/pull/508


nategraf:
Missed that this PR exists https://github.com/risc0/risc0/pull/3067/files

PR #3089, opened 2025-04-09
ZKVM-1327: Return an error if "sha-256" is used for the hashfn

bobbobbio:
Update `ProverOpts::fast()` to use poseidon2 for the hashfn, otherwise it will be unusable.

We didn't properly implement "sha-256" support for segment proving in v2. Instead of ignoring the hashfn value, we should check that it is "poseidon2", the only hashfn that we currently support.

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -268,7 +268,7 @@ mod test {
         compare_image_id(
             &guest_list,
             "hello_commit",
-            "6ebb7949869b15585da87873fd6c21b37b7fcb56d95a88031a04a382aba1bb10",
+            "1db1ae8a6e3ac8a2add00a80dde19c01ac6f6ffba1ff1c2331142f76a63e1d51",
         );
     }
 }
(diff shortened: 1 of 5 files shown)


SchmErik (approved):
looks good to me. My only thought is would this necessitate a version bump of risc0-zkvm to 2.1.0? These functions are lower-level so a bump to 2.0.2 might be sufficient... it is a "possibly breaking change" so it's kind of a grey area and I'm fine with the current patch bump but I'm also open to minor version bump as well

bobbobbio:
> looks good to me. My only thought is would this necessitate a version bump of risc0-zkvm to 2.1.0? These functions are lower-level so a bump to 2.0.2 might be sufficient... it is a "possibly breaking change" so it's kind of a grey area and I'm fine with the current patch bump but I'm also open to minor version bump as well

My thinking is we don't need to treat it like a breaking change (major version bump) because the previous behavior was "broken". We don't want users providing a `hashfn` and then ignore it, so we want to auto-update any users. Between it being a "patch" vs "minor" update, its more grey. I think "minor" is mostly for "new" things, and this feels more like a bugfix.

PR #3090, opened 2025-04-10
ZKVM-1328: Add xtask extract-elf subcommand

bobbobbio:
This command simply extracts the user and kernel elfs from a risc0 binfmt file.

This has been useful to have so things like `objdump` /   `gdb` / `llvm-dwarf-dump` can be run on the user elf.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8209,6 +8209,7 @@ dependencies = [
  "prost-build",
  "protobuf-src",
  "regex",
+ "risc0-binfmt",
  "risc0-circuit-keccak",
  "risc0-circuit-recursion",
  "risc0-core",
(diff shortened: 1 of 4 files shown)


SchmErik (commented):
This is super useful. I think this is better as a `cargo-risczero` command instead. I think the xtask is intended as commands that are useful/necessary for working with the risc0 repo but this command is much more general in that it could be ran by users as well so having it be a part of cargo-risczero makes this more accessible for those that need to debug their ELFs

flaub:
`risc0-build` does still produce an ELF along with the new `.bin`. I've been using this ELF directly for this purpose.

bobbobbio:
> Member

yes, it does. there was two instances where a user sent us a standalone `.bin` file

flaub (approved):

PR #3091, opened 2025-04-11
chore: improve error handling in `to_decimal` function

famouswizard:
I updated the `to_decimal` function to improve error handling. Previously, the function returned an `Option<String>`, which caused it to silently fail when the string format wasn't as expected.
Now, it returns a `Result<String>`, allowing us to catch parsing errors and provide more meaningful error messages.


flaub:
This breaks some tests:

```     
Running tests/external_prover.rs (target/debug/deps/external_prover-8a0c68031491cbdf)

running 2 tests
test basic_proof ... ok
test compressed_proof ... FAILED

failures:

---- compressed_proof stdout ----

thread 'compressed_proof' panicked at risc0/r0vm/tests/external_prover.rs:42:57:
called `Result::unwrap()` on an `Err` value: digest_to_decimal failed
```

PR #3092, opened 2025-04-11
fix: incorrect handling of result in `run` method

hexcow:
I addressed an issue in the `run` method where the line `Ok(result?)` was incorrectly used. Since `result` is already the outcome of `verify()`, using `result?` didn't make sense and caused issues. I replaced it with `Ok(())`, as the value of `result` isn't needed after the error handling.


bobbobbio:
> using result? didn't make sense and caused issues.

what issues?

Issue #3094, opened 2025-04-14
chore: bump getrandom to 0.3


kubkon:
### Is there an existing issue?

- [x] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [x] This is not a support question.

### Motivation

In our project, we rely on [tlsn](https://github.com/tlsnotary/tlsn) which recently bumped `getrandom` usage from 0.2 to 0.3, and as you may be aware, there have been significant breaking changes between the two versions. In order to now compile tlsn as risc0 guest, the zkvm needs to bump it to a compatible version as well, which is 0.3. This is required since `getrandom` changed how custom entropy backends are defined at compile and link time - namely you now need to export a prespecified global symbol named `__getrandom_v03_custom` (verbatim, no name mangling), and then pass a config flag to cargo `--cfg getrandom_backend="custom"`.

### Request

In order to unblock us from upgrading to latest tlsn, I would love if we could bump `getrandom` to 0.3. In fact, I am more than happy to do all the necessary leg work to get this resolved swiftly.

### Are you willing to help with this request?

Yes!

PR #3095, opened 2025-04-14
PER-37: chore(zkvm): bump getrandom to 0.3

kubkon:
Fixes https://github.com/risc0/risc0/issues/3094


bobbobbio (approved):

SchmErik (changes requested):
This solution essentially deprecates our support for getrandom 0.2... not sure if we want to do this... Maybe this is better hidden under a new feature flag specifically for 0.3? I'm going to request changes for now and follow up tomorrow with more info

kubkon:
> This solution essentially deprecates our support for getrandom 0.2... not sure if we want to do this... Maybe this is better hidden under a new feature flag specifically for 0.3? I'm going to request changes for now and follow up tomorrow with more info

Yeah, I thought about this. I am happy to tweak it so that this is under `unstable` or some other feature flag. Lemme know what the desired direction is and I'll make the necessary changes.

bobbobbio:
> This solution essentially deprecates our support for getrandom 0.2... not sure if we want to do this... Maybe this is better hidden under a new feature flag specifically for 0.3? I'm going to request changes for now and follow up tomorrow with more info

I wonder if there is a way we can support both without a flag, maybe even inspect the guest dependencies

kubkon:
> > This solution essentially deprecates our support for getrandom 0.2... not sure if we want to do this... Maybe this is better hidden under a new feature flag specifically for 0.3? I'm going to request changes for now and follow up tomorrow with more info
> 
> 
> 
> I wonder if there is a way we can support both without a flag, maybe even inspect the guest dependencies

While you could inspect the symbol table of the guest binary and assuming the symbol table was not stripped, risc0-zkvm still has to expose it at compile time though right? Otherwise you will get build errors that "this target is not supported, use custom backend" error. Or do you mean something else?

bobbobbio:
> > > This solution essentially deprecates our support for getrandom 0.2... not sure if we want to do this... Maybe this is better hidden under a new feature flag specifically for 0.3? I'm going to request changes for now and follow up tomorrow with more info
> > 
> > 
> > I wonder if there is a way we can support both without a flag, maybe even inspect the guest dependencies
> 
> While you could inspect the symbol table of the guest binary and assuming the symbol table was not stripped, risc0-zkvm still has to expose it at compile time though right? Otherwise you will get build errors that "this target is not supported, use custom backend" error. Or do you mean something else?

it seems to me the way to implement it for 0.2 and 0.3 could be done at the same time. I am trying it out on the PR branch, but ran into some other issues, particularly the `--cfg` flag that was added is causing warnings to pop up about the `test` config flag that I don't understand

bobbobbio:
> While you could inspect the symbol table of the guest binary and assuming the symbol table was not stripped, risc0-zkvm still has to expose it at compile time though right? Otherwise you will get build errors that "this target is not supported, use custom backend" error. Or do you mean something else?

since this is happening at build time when we run `cargo` anyway, you could somewhat easily run `cargo metadata` to find the guest dependencies, although if the two different ways of implementing it can be done simultaneously like I'm hoping we can avoid this

SchmErik:
I took your commits and ported them here: https://github.com/risc0/risc0/pull/3106 Seems like you gave Remi access but I'm unable to make any changes so I'm closing in favor of this new PR

Issue #2707, opened 2025-01-06
[User experience] Improve rzup install instructions and error messages


austinabell:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [X] This is not a support question.

### Motivation

When trying to use rzup without understanding the versioning semantics, it's unclear from the help info what type of versions are expected. For example with:
```sh
rzup install --help
Install toolchains or extensions

Usage: rzup install [OPTIONS] [NAME] [VERSION]

Arguments:
  [NAME]     Name of the toolchain or extension to install
  [VERSION]  Version tag of the toolchain or extension to install

Options:
  -f, --force  Force the installation, ignoring existing installations and downloads
  -h, --help   Print help

Discussion:
    With no toolchain or extension specified, the `install` command
    installs the default RISC Zero installation from the release channels.

    If given a toolchain or extension argument then `install` installs the
    specified toolchain or extension.
```

it's unclear what options there are for `[NAME]` (it seems to be {cargo-risczero,rust,cpp}) and for the versions of each, there isn't any `--help` commands to let any user know which version semantics are expected. 

In this case, the versioning for `cargo-risczero` expects `v{VERSION}` semantics, where if you just provide the version without the `v` prefix it just returns `Error: error decoding response body`, which is unhelpful. (using these git tags afaik https://github.com/risc0/risc0/releases)

For `rzup install rust` it expects the version to be `r0.{TOOLCHAIN VERSION}` which someone wouldn't stumble upon naturally as this isn't in any high-level docs either. (seems to be these git tags https://github.com/risc0/rust/releases)

For `rzup install cpp` it expects the git release tag in https://github.com/risc0/toolchain/releases where again if anything is incorrect, it just gives the error `Error: error decoding response body`

also, somewhat related to https://github.com/risc0/risc0/issues/2316, which might help with listing toolchain versions, but would also be helpful to know what toolchain/extensions are available and the versions associated with each.

### Request

Add information for the semantics of each so I can just link users trying to do any of these to that, and perhaps include a link to those docs inside of `rzup install --help`, or just improve `rzup install` commands and subcommands and hints to not require someone to read high-level docs.

Also, a nice to have would be to have a better error than `Error: error decoding response body` whenever a user has the incorrect format.

### Are you willing to help with this request?

Yes!

sashaaldrick:
@austinabell @SchmErik 

[...]

PR #3096, opened 2025-04-14
docs(cli): clarify version formats in rzup install help output

Himess:
## Description

This pull request resolves the core usability issue raised in [#2707](https://github.com/risc0/risc0/issues/2707) by updating the version parsing logic used in `rzup install`.

Previously, version parsing silently failed or produced unclear `error: error decoding response body` messages when the expected component-specific version prefix (e.g. `r0.`, `v`, or date format) was not provided in the exact semantic format. This PR introduces robust normalization logic and improved error reporting.

- **Component-specific prefix handling:**
  - `rust` accepts versions prefixed with `r0.` and parses them into `X.Y.Z` format.
  - `cargo-risczero` accepts `vX.Y.Z` and parses it correctly.
  - `cpp` accepts versions in `YYYY.MM.DD` format as before.

- **Semver normalization**:
  - Versions like `r0.1.85.0` or `v2.0.1` are normalized to valid `semver` `X.Y.Z`.

- **Invalid input handling**:
  - Clear error messages are returned if the prefix is missing or the semantic version is malformed.
  - Helpful `tip:` messages guide users toward correct usage.

- **Help text improved**:
  - Updated `--help` output now clearly documents expected version formats per component.

---

## Screenshots

image

Pasted Graphic 14

## Related

Fixes: [#2707](https://github.com/risc0/risc0/issues/2707)

---

## Test Plan

All changes tested locally:

```bash
cargo build
cargo fmt --all --check
cargo test

### DIFF
--- a/rzup/src/cli/commands.rs
+++ b/rzup/src/cli/commands.rs
@@ -45,7 +45,20 @@ pub const INSTALL_HELP: &str = "Discussion:
     `GITHUB_TOKEN` environment variable, then from ~/.config/gh/hosts.yml.
 
     The default version of the component is updated to the version that was
-    just installed.";
+    just installed.
+
+    ──────────────
+    Component List:
+      - rust            → version format: r0.1.85.0
+      - cargo-risczero  → version format: v2.0.1
+      - cpp             → version format: YYYY.MM.DD
+
+    You can run 'rzup show' to list installed versions.
+
+    For more information on available versions:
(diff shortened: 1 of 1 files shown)


bobbobbio:
Thanks for the PR, but I don't think we want to accept these version formats anymore. We made a conscious choice to try to switch to SemVer here.

PR #3101, opened 2025-04-16
ZKVM-1333: Improve Executor Segment Creation Performance

ec2:
Before, every time the executor creates a Segment, it cloned the whole MemoryImage which is expensive and then it rehashes everything as well.  

Before:
```
┌─────────┬────────┬────────────┬──────────┬─────────┬─────┬──────┐
│ name    │ hashfn │ throughput │ duration │ cycles  │ ram │ seal │
├─────────┼────────┼────────────┼──────────┼─────────┼─────┼──────┤
│ execute │ N/A    │ 40.8MHz    │ 24.3ms   │ 1014.1K │ N/A │ N/A  │
└─────────┴────────┴────────────┴──────────┴─────────┴─────┴──────┘
```
After:
```
┌─────────┬────────┬────────────┬──────────┬─────────┬─────┬──────┐
│ name    │ hashfn │ throughput │ duration │ cycles  │ ram │ seal │
├─────────┼────────┼────────────┼──────────┼─────────┼─────┼──────┤
│ execute │ N/A    │ 44.5MHz    │ 22.3ms   │ 1014.1K │ N/A │ N/A  │
└─────────┴────────┴────────────┴──────────┴─────────┴─────┴──────┘
```

I've ran this against zeth as well ` RUST_LOG="info" RISC0_INFO=1 target/release/zeth-ethereum run --cache=cache --block-number=22176200 --block-count=1 `

Before:
```
[2025-04-16T21:17:07Z INFO  risc0_zkvm::host::server::exec::executor] execution time: 27.017512174s
[2025-04-16T21:17:07Z INFO  risc0_zkvm::host::server::session] number of segments: 575
[2025-04-16T21:17:07Z INFO  risc0_zkvm::host::server::session] 602931200 total cycles
[2025-04-16T21:17:07Z INFO  risc0_zkvm::host::server::session] 502496192 user cycles (83.34%)
[2025-04-16T21:17:07Z INFO  risc0_zkvm::host::server::session] 96775344 paging cycles (16.05%)
[2025-04-16T21:17:07Z INFO  risc0_zkvm::host::server::session] 3659664 reserved cycles (0.61%)
```

After:
```
[2025-04-16T22:00:01Z INFO  risc0_zkvm::host::server::exec::executor] execution time: 23.589958846s
[2025-04-16T22:00:01Z INFO  risc0_zkvm::host::server::session] number of segments: 575
[2025-04-16T22:00:01Z INFO  risc0_zkvm::host::server::session] 602931200 total cycles
[2025-04-16T22:00:01Z INFO  risc0_zkvm::host::server::session] 502555065 user cycles (83.35%)
[2025-04-16T22:00:01Z INFO  risc0_zkvm::host::server::session] 96868690 paging cycles (16.07%)
[2025-04-16T22:00:01Z INFO  risc0_zkvm::host::server::session] 3507445 reserved cycles (0.58%)
```

### DIFF
--- a/risc0/circuit/rv32im/src/execute/pager.rs
+++ b/risc0/circuit/rv32im/src/execute/pager.rs
@@ -452,6 +452,39 @@ impl PagedMemory {
         (pre_image, pre_state, post_state)
     }
 
+    /// Similar to commit, but does not clone and return the pre-image.
+    pub(crate) fn flush(&mut self) -> (Digest, Digest) {
+        // tracing::trace!("commit: {self:#?}");
+
+        self.write_registers();
+
+        let pre_state = self.image.image_id();
+
+        let mut sorted_keys: Vec<_> = self.page_states.keys().collect();
+        sorted_keys.sort();
+
+        for node_idx in sorted_keys {
+            if node_idx < MEMORY_PAGES as u32 {
+                continue;
+            }
+
+            let page_state = self.page_states.get(node_idx);
+            let page_idx = page_idx(node_idx);
+            tracing::trace!("commit: {page_idx:#08x}, state: {page_state:?}");
+
+            // Update dirty pages into the image that accumulates over a session.
+            if page_state == PageState::Dirty {
+                let cache_idx = self.page_table.get(page_idx).unwrap();
+                let page = &self.page_cache[cache_idx];
+                self.image.set_page(page_idx, page.clone());
+            }
+        }
+        self.image.update_digests();
+
+        let post_state = self.image.image_id();
+        (pre_state, post_state)
+    }
+
     fn load_page(&mut self, page_idx: u32) -> Result<()> {
         tracing::trace!("load_page: {page_idx:#08x}");
         let page = self.image.get_page(page_idx)?;
@@ -519,6 +552,16 @@ pub(crate) fn compute_partial_image(
 
     for node_idx in &indexes {
         if *node_idx < MEMORY_PAGES as u32 {
+            let lhs_idx = *node_idx * 2;
+            let rhs_idx = *node_idx * 2 + 1;
+
+            // Otherwise, add whichever child digest (if any) is not loaded
+            if !indexes.contains(&lhs_idx) {
+                image.set_digest(lhs_idx, *input_image.get_existing_digest(lhs_idx));
+            }
+            if !indexes.contains(&rhs_idx) {
+                image.set_digest(rhs_idx, *input_image.get_existing_digest(rhs_idx));
+            }
             continue;
         }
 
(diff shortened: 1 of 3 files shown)


ec2:
oops meant to open as draft

risc0/circuit/rv32im/src/execute/pager.rs
```diff
@@ -419,12 +419,11 @@ impl PagedMemory {
         }
     }
 
-    pub(crate) fn commit(&mut self) -> (MemoryImage, Digest, Digest) {
+    pub(crate) fn commit(&mut self) -> (Digest, Digest) {
         // tracing::trace!("commit: {self:#?}");
 
         self.write_registers();
 
-        let pre_image = self.image.clone();
```
flaub, line 427: I don't see how this can work without this.

bobbobbio (changes requested):
The tests are failing with OOM and timeouts which is exactly the same failure mode I got before when I messed this up, its still confusing to me why it breaks that way

If we want to do this I think we would just have to undo this commit 2fd790272e28d12e718fbcde207553f6eb236e5e

But at the time doing the partial image creation on the background thread seemed to speed up the main thread, if its not that way now I'd like to understand why.

If the tests pass then the numbers are more interesting I think, its hard to reason about whats happening right now

ec2:
After making the page hashing done in parallel via rayon, I've managed to get `zeth` down to ~18s. 

```
[2025-04-21T21:19:54Z INFO  risc0_zkvm::host::server::exec::executor] execution time: 18.582866867s
[2025-04-21T21:19:54Z INFO  risc0_zkvm::host::server::session] number of segments: 588
[2025-04-21T21:19:54Z INFO  risc0_zkvm::host::server::session] 616562688 total cycles
[2025-04-21T21:19:54Z INFO  risc0_zkvm::host::server::session] 502406956 user cycles (81.49%)
[2025-04-21T21:19:54Z INFO  risc0_zkvm::host::server::session] 96793698 paging cycles (15.70%)
[2025-04-21T21:19:54Z INFO  risc0_zkvm::host::server::session] 17362034 reserved cycles (2.82%)```

ec2:
@bobbobbio Calculating partial images on a separate thread indeed was a speed up. However, we were cloning the whole `MemoryImage` and then sending it to another thread for calculation. As time goes on, each `MemoryImage` becomes larger and larger and thus more expensive... 

In addition, the way the partial image was being constructed in the separate thread was not very efficient because what we were doing was rehashing EVERYTHING to create this partial image even though the `MemoryImage` we are using already has the hashes calculated at that point.

PR #3104, opened 2025-04-17
ZKVM-1336: Executor adds soft cycle limit for per-segment execution

mars-risc0:
The `max_cycles` termination behavior on `Executor::run` is now configurable: the existing behavior is a "hard limit", which errors out if the limit is reached, but one can also request a "soft limit", which simply stops the VM and returns. This allows `Segment::execute` to provide the recorded suspend cycle as the cycle limit, which prevents execution from continuing to the next segment, for which there are no IO records.


flaub (approved):
Nice

PR #3107, opened 2025-04-17
ZKVM-1340: datasheet adds three segments from a po2=22 zeth run on block "shapella"

mars-risc0:


flaub (approved):

risc0/zkvm/examples/datasheet.rs
```diff
@@ -547,6 +553,44 @@ impl Datasheet {
         self.bigint2_prove_segment(&session, &segment);
     }
 
+    fn shapella_segment(&mut self, name: &str, bytes: &[u8]) {
+        // "shapella" is the name of ethereum block 17034870
+        // this test runs one segment of a zeth run on that block
+        println!("{}", name);
```
bobbobbio, line 559: FYI this can be `println!("{name}");`

bobbobbio (approved):

PR #3108, opened 2025-04-18
fix(build): escape special chars in rustc flags if requested

### LINKED ISSUE #3103: risc0-build: building with docker improperly escapes strings in metadata

### Description of bug

Passing strings with embedded strings in risc0 package metadata strips all quotes from inner strings in Docker environment leading to build errors. For example, consider the following metadata:

```toml
[package.metadata.risc0]
rustc-flags = ["--cfg", "getrandom_backend=\"custom\""]
```

When building *without* docker, `CARGO_ENCODED_RUSTFLAGS` will correctly include `--cfg getrandom_backend="custom"`. However, when run in Docker, we end up with `--cfg getrandom_backend=custom`.

### Steps to reproduce

I have prepared a simple reproduction:

```
$ git clone git@github.com:kubkon/risc0-docker-repro
$ cd risc0-docker-repro
$ cargo build # no docker, builds fine
$ USE_DOCKER=1 cargo build # fails
```

kubkon:
Fixes #3103 

cc @bobbobbio


bobbobbio (approved):
Thanks for this, its a much better version that what we were trying to get into the getrandom patch. @SchmErik maybe we can just rebase onto this?

flaub (approved):

PR #3109, opened 2025-04-18
ZKVM-1342: Fix an issue with sys_read where it would read past the end of the

bobbobbio:
supplied buffer, or fail in other ways.

The branch that had the issue was in the v2 kernel and only taken when the read is greater than 1024. By beefing up the test to do larger reads we can see the bug trigger.

The sys_read syscall would pass arguments in registers a0-a4. The first 3 registers (a0, a1, a2) are common to all the ecall_read syscalls. These are `syscall_ptr`, `buf`, and 'len`. The last two arguments (a3, ar) are `fd` and `nwords`.

On v2, we implement ecall_read differently where it accepts `fd`, `buf`, `len` as the first three common arguments. We end up adapting the API in the v2 kernel. The kernel calls the host ecall_read from kernel code. We stash the `syscall_ptr` in `fd` (the new ecall_read's `fd` argument not to be confused with sys_read's `fd` argument) and `buf` and `len` in their respective arguments, these are passed via kernel registers.

In the host ecall_read handler we unpack and adapt these arguments (the kernel registers a1 and a2) before calling into the syscall_read code (which we know to do thanks to a0) passing the two needed values (`buf` and `len` directly through).

Once in the syscall code we read the two other arguments, the actual `fd` and `nbytes` from the user register directly.

The return values are also different and there is code to adapt that including another ecall_read call to ferry the result, but its not relevant to the issue.

This all works fine, until we deal with reads greater than 1024. The new ecall API has the limitation of only doing reads <= 1024. So if we do reads >= 1024 the v2 kernel has to break up the read into smaller reads. This happens in the Rust implemented part of the kernel, and this is only taken for sys_read, the other syscalls aren't able to deal with chunking at all.

When we do this chunking though, we forget about the other sys_read arguments, in particular the value stored in a4. This value also contains the length and needs to be updated. By not updating it, the sys_read call does unexpected stuff, including reading not enough or too much.


risc0/zkos/v1compat/src/main.rs
```diff
@@ -198,11 +199,17 @@ mod zkvm {
     }
 
     #[no_mangle]
-    unsafe extern "C" fn ecall_software(fd: u32, mut buf: *const u8, mut len: u32) -> ! {
+    unsafe extern "C" fn ecall_software(fd: u32, mut buf: *const u8, mut len: u32) {
         // use no_std_strings::{str256, str_format};
         // let msg = str_format!(str256, "ecall_software_slow({fd:#010x}, {buf:?}, {len})");
         // print(msg.to_str());
 
+        if read_syscall_ptr(fd) != b"risc0_zkvm_platform::syscall::nr::SYS_READ" {
```
flaub, line 207: This is a good defensive move, but I wonder if this will cause perf regressions now that every software ecall has to go thru the string compare.

bobbobbio: thats a good point. I'm wondering what alternatives we have, its sorta unfortunate that we used a string here, I wonder if there is some way we can find the pointer to this string at build time or something.

I think this fixes another bug here which is basically if you call one of the other syscalls with a buffer larger than 1024 it will break in an unexpected way, maybe even corrupt memory like we saw with this one.

bobbobbio: actually maybe i have an idea, we could pass some other flag through to let the host know we're doing this chunking and then check in the host that we're in sys_read.

bobbobbio: okay, interestingly the only syscall where this is a real problem I think is getenv

bobbobbio:
Okay, here is my attempt to fix the situation. I went through all the other syscalls and made them somehow compatible with the chunking algorithm in the kernel. This was done by making them work with the chunking or making sure that they error if they go down the chunking path.

Either the syscall
- expects a zero buffer, and now errors if it doesn't get it
- expects a buffer of size < 1024 and errors if given a larger one (the keccak ones for example)
- has been changed to work with the chunking algorithm (sys_random, sys_slice_io)
- errors if all the data to write doesn't fit in the supplied buffer and the supplied buffer is MAX_IO_BYTES (sys_getenv, sys_argv)

risc0/zkvm/src/host/server/exec/syscall/args.rs
```diff
@@ -50,6 +55,11 @@ impl Syscall for SysArgs {
             let nbytes = min(to_guest.len() * WORD_SIZE, arg_val.len());
             let to_guest_u8s: &mut [u8] = bytemuck::cast_slice_mut(to_guest);
             to_guest_u8s[0..nbytes].clone_from_slice(&arg_val.as_bytes()[0..nbytes]);
+
+            if arg_val.len() == MAX_IO_BYTES as usize {
```
SchmErik, line 59:
```suggestion
            if arg_val.len() >= MAX_IO_BYTES as usize {
```
I think you meant this. There's a few spots where you have `==` instead of `>=`

SchmErik (approved):
Thanks! I followed the conversation and the changes that you've made based on feedback. All of this looks good to me but please address the `>=` vs `==`.

flaub (approved):

ercecan (approved):

PR #3111, opened 2025-04-21
feat: complete TODO by adding static entropy to memory image

Himess:
## Description

This PR completes the following TODO in `examples/c-guest/main.c`:

Zero-Knowledge (ZK) systems require non-deterministic execution traces to maintain privacy and soundness. To fulfill this requirement, a static 128-bit entropy value (four 32-bit words) is introduced and committed to the SHA256 hasher at the start of the guest program.

This ensures the memory image includes the minimum required entropy and aligns with protocol expectations.

## Changes

- Introduced a 128-bit fixed entropy array in `main.c`
- Committed it to the hasher before any host input is processed
- Removed the obsolete TODO comment


flaub:
Entropy would mean random data for each execution. This is just an example, so I don't think we need to worry about it too much.

PR #3115, opened 2025-04-21
ZKVM-1345: Load word-aligned regions by words, not bytes

mars-risc0:
The base addr and size parameters of a `load_region` call are very often word-aligned, so we can load more efficiently word-wise than with a byte-by-byte loop. This improves performance on some bigint-heavy workloads: datasheet exec time for `zeth_shapella_50` goes from 757ms to 413ms, with smaller improvements seen on the other segments.

### DIFF
--- a/risc0/circuit/rv32im/src/execute/r0vm.rs
+++ b/risc0/circuit/rv32im/src/execute/r0vm.rs
@@ -90,8 +90,16 @@ pub(crate) trait Risc0Context {
 
     fn load_region(&mut self, op: LoadOp, addr: ByteAddr, size: usize) -> Result<Vec<u8>> {
         let mut region = Vec::with_capacity(size);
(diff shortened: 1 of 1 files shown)


bobbobbio:
Good catch.

I think an even better algorithm would be to deal with the unaligned section until things are aligned then do the faster loading until you can't and then deal with the potentially unaligned bit at the end. That should speed up even unaligned loads that are large enough.

mars-risc0:
Yeah, that could be an improvement. In the case of the bigint-heavy zeth segments, I didn't actually see any unaligned loads, since the bigint blobs are all imported with `include_bytes_aligned!`. It could be interesting to run the whole block and see how many unaligned loads we get, and whether they are larger than word-size.

flaub (approved):

PR #3116, opened 2025-04-22
ZKVM-1346: Add DevModeContext to allow for programatic enabling of dev mode

nategraf:
Following up on https://github.com/risc0/risc0/pull/3102, this PR adds a programtic way to enable dev mode by creating a `DevModeContext` value that enables dev mode for the lifetime of the value. This mechanism allows for use of dev mode in the zkVM in special circumstances where is is desirable, such as testing of guests that recursively verify RISC Zero receipts. Using the [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization) pattern, simmilar to [MutexGuard](https://doc.rust-lang.org/std/sync/struct.MutexGuard.html), this implementation makes it possible to enable dev mode within a scope, and according to application logic with assurance that it is disabled outside that scope.


risc0/zkvm/src/dev_mode.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+impl Drop for DevModeContext {
+    /// Exit the dev mode context on drop.
+    fn drop(&mut self) {
+        DEV_MODE_CONTEXT_DEPTH.with_borrow_mut(|x| {
+            *x = x
+                .checked_sub(1)
+                .expect("underflow of DEV_MODE_CONTEXT_DEPTH occurred")
```
flaub, line 77: It doesn't seem good to have panics in a drop

risc0/zkvm/src/dev_mode.rs
```diff
@@ -0,0 +1,184 @@
...
+//! Module containing functions for using dev mode. See [self::is_dev_mode].
+
+use core::{cell::RefCell, marker::PhantomData};
+
+thread_local! {
+    /// Thread-local counter that is determine when the code is in a dev mode context.
+    ///
+    /// NOTE: This is thread-local because it entering a DevModeContext in one thread should not
+    /// effect other threads.
+    static DEV_MODE_CONTEXT_DEPTH: RefCell<u32> = RefCell::new(0);
```
flaub, line 24: Is this really a semaphore? Aren't there better primitives for this?

risc0/zkvm/src/dev_mode.rs
```diff
@@ -0,0 +1,184 @@
...
+///
+/// When dev mode is enabled, [FakeReceipt][crate::FakeReceipt] can be used to test applications.
+/// In particular, the verifier will accept fake receipts as a valid proof of
+/// execution, even though it provides no cryptographic guarantees. Additionally, the prover will
+/// produce a fake receipt instead of spending computation resources to produce a full receipt.
+///
+/// Dev mode can be enabled in one of two ways:
+/// 1. By setting the `RISC0_DEV_MODE` environment variable to "1", "true", or "yes". This does not
+///    work when building for the guest (i.e. `target_os = zkvm`).
+/// 2. By calling [DevModeContext::enter()] and holding the returned value.
```
flaub, line 92: What are some use cases for `DevModeContext::enter`, especially where `RISC0_DEV_MODE` falls short?

risc0/zkvm/src/host/api/mod.rs
```diff
@@ -222,11 +222,17 @@ fn get_server_version<P: AsRef<Path>>(server_path: P) -> Result<Version> {
...
-            .spawn()
-            .with_context(|| self.spawn_fail())?;
+        let mut cmd = Command::new(&self.server_path);
+        cmd.arg("--port").arg(addr.port().to_string());
+
+        // If dev mode is enabled, enable on the the child process with RISC0_DEV_MODE set. If this
+        // RISC0_DEV_MODE was set on this process, it will be inheritted anyway. If dev mode was
+        // set via DevModeContext, then it would not otherwise be set.
+        if crate::is_dev_mode() {
+            cmd.env("RISC0_DEV_MODE", "true");
```
flaub, line 232: This makes me a little nervous...

bobbobbio:
This seems to me a bit of a wild solution for something where we could plumb a value through normally (via function arguments) indicating the desire for dev-mode or not, unless I'm missing some reason why we would do that (other than SemVer).

This doesn't seem like the kind of problem that should require per-thread global counters...

PR #3120, opened 2025-04-23
Rename get_cycle_count to cycle_count

maximevtush:


rlukata:
Version 0.18 still uses [get_cycle_count](https://docs.rs/risc0-zkvm/0.18.0/risc0_zkvm/guest/env/fn.get_cycle_count.html)

PR #3122, opened 2025-04-24
groth16 implemented by using substate-bn directly (without arkworks)

iddo-bentov:
risc0/examples/groth16-verifier used to be ~161M cycles with arkworks, now it's ~46M cycles (and ~241M cycles when vanilla substrate-bn that doesn't include the bigint2 acceleration).

This PR still has many TODOs, there's another branch (iddo/groth16-tryfrom) that addresses some of the TODOs though in terms of cycles it's a regression. After resolving some minor TODOs, there are still other TODOs that are non-trivial/unclear for now, so it seems reasonable to merge this PR as it is because it's already working well.


flaub (changes requested):
I can help get the error forwarding resolved

PR #3124, opened 2025-04-25
ZKVM-1360: Drop legacy code and bump versions

flaub:


bobbobbio (commented):
We should probably update `workspace.package.version` to `3.0.0`, although I'm now not sure anything is using it anymore.

The semver checks are indicating some other bumps are needed it seems, but this looks good so far

bobbobbio (approved):

PR #3099, opened 2025-04-15
docs: add documentation for Poseidon2 hash constants

Dimitrolito:
This PR resolves a `TODO` by adding proper documentation for the Poseidon2 hash function constants.

The following constants are now properly documented:

- `CELLS`: Number of elements in the internal state.
- `ROUNDS_HALF_FULL`: Number of full S-box rounds (beginning and end).
- `ROUNDS_PARTIAL`: Number of partial S-box rounds (middle rounds).
- `ROUND_CONSTANTS`: Round constants array used during each permutation step.

PR #3093, opened 2025-04-12
Fix variable naming consistency in composite receipt handling

zeevick10:
Description:
This PR addresses naming consistency in the composite receipt handling code:

1. Standardizes variable naming from 'composite_recipt' to 'composite_receipt'
2. Updates all references to maintain consistent terminology
3. Improves code readability and reduces potential confusion

These changes help maintain our coding standards and make the codebase more maintainable.

PR #3126, opened 2025-04-28
ZKVM-1361: Integrate community contributions

flaub:
Source:

* https://github.com/risc0/risc0/pull/3093
* https://github.com/risc0/risc0/pull/3099


bobbobbio (approved):
I don't know enough about Poseidon2 to say those comments are right, so I'm assuming you have validated those yoursef.

PR #3135, opened 2025-04-30
ZKVM-1364: Replace decoding table with rust matches

flaub:


risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -429,15 +441,20 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
     }
 
     #[cold]
```
bobbobbio: the idea of this `cold` is to mark some branch as not taken very often, but now it is taken all the time.

we should remove this `#[cold]`  and maybe consider adding it to the tracing branch in a different way

bobbobbio (approved):
We can probably use the `likely` / `unlikely` to clean up the `trace_pager` calls too

PR #3139, opened 2025-05-01
Add way to use cuda to compute control id for bootstrap

bobbobbio:
This lets us compute the control id for bootstrap using the GPU.
It significantly speeds up bootstrap when enabled since doing these hashes on the CPU can take a long time.

### DIFF
--- a/risc0/zkp/src/hal/cuda.rs
+++ b/risc0/zkp/src/hal/cuda.rs
@@ -61,7 +61,9 @@ unsafe impl DeviceCopy for DeviceExtElem {}
 
 pub trait CudaHash {
     /// Create a hash implementation
-    fn new() -> Self;
+    fn new() -> Self
+    where
+        Self: Sized;
 
     /// Run the hash_fold function
     fn hash_fold(&self, io: &BufferImpl<Digest>, output_size: usize);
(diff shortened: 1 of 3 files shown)


risc0/zkp/src/hal/cuda.rs
```diff
@@ -438,7 +446,19 @@ impl<CH: CudaHash> CudaHal<CH> {
...
 }
 
-impl<CH: CudaHash> Hal for CudaHal<CH> {
+impl CudaHal<dyn CudaHash> {
+    pub fn new_from_hash_suite(hash_suite: HashSuite<BabyBear>) -> Self {
+        let hash_suite_box = match &hash_suite.name[..] {
+            "poseidon2" => Box::new(CudaHashPoseidon2::new()) as Box<dyn CudaHash>,
+            "poseidon254" => Box::new(CudaHashPoseidon254::new()) as Box<dyn CudaHash>,
+            "sha-256" => Box::new(CudaHashSha256::new()) as Box<dyn CudaHash>,
+            other => unimplemented!("unsupported hash_fn {other}"),
```
flaub, line 455: I think this needs to return an error rather than panic. I'm thinking about whether the hash function might be part of some user-controlled proof request.

flaub (approved):

PR #3142, opened 2025-05-02
Update cd path in Hello World tutorial

PixelPil0t1:
The old path cd hello-world assumes the project is created in the current directory. However, in the context of using the cloned risc0 monorepo, the actual path is risc0/examples/hello-world. This change corrects the path and prevents user confusion or errors like No such file or directory.


flaub:
If you use `cargo risczero new`, then won't it create a directory underneath the current directory?

PixelPil0t1:
> If you use `cargo risczero new`, then won't it create a directory underneath the current directory?

Good call — I’ve removed the cargo risczero new line. Since we’re cd’ing into an existing example at risc0/examples/hello-world, there’s no need to scaffold a new project. Should be cleaner now. Let me know if you spot anything else.

flaub:
The intent of this tutorial is to make use of `cargo risczero new`.

PR #3143, opened 2025-05-02
fix(risc0-build): canonicalize manifest dir

rkrasiuk:
## Description

I have a project that has guest methods located in the parent directory. e.g. 
```toml
[package.metadata.risc0]
methods = ["../../programs/risc0"]
```

Calling `risc0_build::embed_methods()` in `build.rs` results in
```
 ERROR: No package found in "<PATH>/../../programs/risc0"
```
because vanilla `Path` comparison here
https://github.com/risc0/risc0/blob/3df2b55e611bbd7c65303bc7ef35be69f5a93b4d/risc0/build/src/lib.rs#L258
only normalizes `.` occurrences.


rkrasiuk:
upd: swapped `risc0-build` dependency for the branch on top of `2.1.1` version, worked like a charm

flaub (approved):

PR #3145, opened 2025-05-02
ZKVM-1371: executor perf: Move all memory hashing to background thread

bobbobbio:
Thanks to the conversation that came from @ec2 's investigation we realized that none of the memory hashing has to happen on the main executor thread. This moves all of it to the background thread.

On my coreweave machine this took zeth execute down from 21s to 17s.

### DIFF
--- a/risc0/circuit/rv32im/src/execute/executor.rs
+++ b/risc0/circuit/rv32im/src/execute/executor.rs
@@ -31,7 +31,7 @@ use crate::{
 
 use super::{
     bigint,
-    pager::{compute_partial_image, PageTraceEvent, PagedMemory},
+    pager::{compute_partial_image, PageTraceEvent, PagedMemory, WorkingImage},
     platform::*,
     poseidon2::Poseidon2State,
     r0vm::{EcallKind, LoadOp, Risc0Context, Risc0Machine},
@@ -57,6 +57,7 @@ pub struct Executor<'a, 'b, S: Syscall> {
     user_pc: ByteAddr,
     machine_mode: u32,
     user_cycles: u32,
+    initial_image: MemoryImage,
     pager: PagedMemory,
     terminate_state: Option<TerminateState>,
     read_record: Vec<Vec<u8>>,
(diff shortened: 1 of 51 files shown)


risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -143,17 +150,29 @@ fn unlikely(b: bool) -> bool {
...
+    let initial_digest = existing_image.image_id();
+
     while let Ok(req) = recv.recv() {
-        let partial_image = compute_partial_image(req.image, req.page_indexes);
-        callback(Segment {
+        let pre_digest = existing_image.image_id();
+        let partial_image = compute_partial_image(&mut existing_image, req.page_indexes);
+
+        for (idx, page) in req.partial_image.pages {
+            existing_image.set_page(idx, page);
```
ec2, line 166: If you want this to be even faster, what you could do here is hash all the leaves (i.e.  Pages) in parallel using Rayon. I don't know if this makes everything faster overall, but if you make the segment creation faster, then you can start proving the first segment earlier and thus making the whole Proving pipeline faster.

bobbobbio: I think we could speed up the hashing maybe even more just by using the GPU to do it. At this moment like you say it probably won't speed up the overall time very much, but eventually it might be a problem.

risc0/circuit/rv32im/src/execute/pager.rs
```diff
@@ -52,7 +52,8 @@ pub(crate) const RESERVED_PAGING_CYCLES: u32 = LOAD_ROOT_CYCLES
     + POSEIDON_PAGING
     + STORE_ROOT_CYCLES;
 
-const NUM_PAGES: usize = 4 * 1024 * 1024;
+const PAGE_SIZE: usize = 1024;
+const NUM_PAGES: usize = 4 * PAGE_SIZE * PAGE_SIZE;
```
bobbobbio, line 56: this was a mistake, I'm going to revert this before merging

flaub (approved):

PR #3148, opened 2025-05-07
correct groth16 path in circom_stable.sh

zeevick10:
Updates the path in `circom_stable.sh` script from `compact_proof/groth16/stark_verify.circom` to `groth16_proof/groth16/stark_verify.circom` to match the directory structure changes.


.github/scripts/circom_stable.sh
```diff
@@ -5,7 +5,7 @@ if [ $RISC0_LIB_HASH != $RISC0_LIB_EXPECTED_HASH ]; then
     exit 1
 fi
     
-STARK_VERIFY_HASH=$(git hash-object -- compact_proof/groth16/stark_verify.circom | awk '{ print $1 }')
+STARK_VERIFY_HASH=$(git hash-object -- groth16_proof/groth16/stark_verify.circom | awk '{ print $1 }')
```
flaub, line 8: Isn't the path on line 1 also wrong?

flaub (approved):

PR #3149, opened 2025-05-07
fix: Add skip-cuda-components feature to reduce CUDA build times

### LINKED ISSUE #3023: Compiling with CUDA support takes a long time / is unable to complete

### Description of bug

On a [8 core Ryzen](https://www.techpowerup.com/cpu-specs/ryzen-7-1800x.c1879) I am experimenting with a sample project that takes on the order of ~5 minutes to build from scratch without the cuda feature enabled.

With cuda, it takes an indeterminate amount of time... I waited almost 2 hours and it's still running. Is that expected?

`risc0-circuit-recursion-sys(build), risc0-circuit-rv32im-sys(build)` took on the order of 10s of minutes, but `risc0-circuit-keccak-sys(build)` seems to be the longest running that I cannot get to complete.

**Is there a way to opt out of building any parts of the prover?** Specifically I don't believe I have any reliance on keccak at least for now in [my demo](https://github.com/nuke-web3/zkvm-chacha/tree/3ad7f11c14267ca32f389d533a0b0275697c9230). It would be great to be able to automatically or manually decide what things are needed to reduce compilation times. 

Possibly related to other CUDA compilation issues, notably other compilation issues like #3016 
https://github.com/risc0/risc0/issues?q=is%3Aissue%20state%3Aopen%20cuda

Note: I am using the `risc0-* v2.0` toolchain & deps. 

### Steps to reproduce

To duplicate, build the [demo project](https://github.com/nuke-web3/zkvm-chacha/tree/3ad7f11c14267ca32f389d533a0b0275697c9230) with and without cuda:
```sh
cargo b # order of a couple minutes
cargo b -F cuda # perhaps impossible to compile? Hours.
```

DeVikingMark:
Resolves #3023

- Added new `skip-cuda-keccak` feature to risc0-circuit-keccak-sys
- Added `skip-cuda-components` feature to zkvm that enables skip-cuda-keccak
- Created stub library during build when using skip features to satisfy dependencies
- Added documentation in READMEs explaining feature usage and benefits
- Dramatically reduces CUDA build times when Keccak functionality isn't needed

### DIFF
--- a/risc0/circuit/keccak-sys/Cargo.toml
+++ b/risc0/circuit/keccak-sys/Cargo.toml
@@ -21,5 +21,6 @@ glob = "0.3"
 risc0-build-kernel = { workspace = true }
 
 [features]
-default = []
+default = ["keccak"]
+keccak = []
 cuda = ["dep:cust", "dep:sppark", "risc0-sys/cuda"]
--- a/risc0/circuit/keccak-sys/README.md
+++ b/risc0/circuit/keccak-sys/README.md
@@ -4,30 +4,34 @@ This package contains the generated HAL code for the keccak circuit.
 
 ## CUDA Compilation
 
-The CUDA compilation for this circuit can be time-consuming and may take hours on certain systems. If you don't need the keccak circuit when using CUDA, you can use the `skip-cuda-keccak` feature to skip compiling the CUDA kernels for this circuit.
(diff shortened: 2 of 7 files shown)


flaub:
Generally, since cargo feature flags are additive, it's better to have a `keccak` feature flag that is on by default and that users can turn off default flags if they want to avoid building keccak.

PR #2945, opened 2025-03-10
ZKVM-1191: Refactor Clap attribute macros from #[clap(...)] to #[arg(...)] and #[command(...)] in v4.x

Himess:
This PR updates deprecated `#[clap(...)]` attributes to their modern equivalents in `clap` 4.x.
The current codebase still uses outdated syntax that has been deprecated since version 4.0.
By making this update, we ensure compatibility with future versions and maintain code quality.

Fixes #2936

License Check Completed Successfully
Pasted Graphic 23

This PR replaces #2937 , which was closed to ensure a cleaner code structure and a well-organized commit history.

PR #3150, opened 2025-05-07
fix: replace deprecated #[clap]

### LINKED ISSUE #2936: Migrate #[clap(...)] to #[arg(...)] / #[command(...)] in clap 4.x
In **clap** version 4.x and above, the use of `#[clap(...)]` is no longer valid.  
Many instances of `#[clap(...)]` are still present in the project.  
These should be replaced with `#[arg(...)]` or `#[command(...)]`.

Himess:
Closes #2936 

Suggestion : .
Let’s enable a feature like clap/deprecated in the project so that everyone automatically gets warnings.
This way, if someone in the team accidentally uses old #[clap(...)] attributes or name, they will immediately get a warning.
Otherwise, people might unknowingly keep adding deprecated code again and again.

I Can take it. If team is ok.

I noticed that the #[clap changes I made in #2945 have been reintroduced. They were added again 3 weeks ago. We can fix this as well if that's okay.

This PR updates deprecated #[clap(...)] attributes to their modern equivalents in clap 4.x.
The current codebase still uses outdated syntax that has been deprecated since version 4.0.
By making this update, we ensure compatibility with future versions and maintain code quality.

### DIFF
--- a/bento/crates/api/src/lib.rs
+++ b/bento/crates/api/src/lib.rs
@@ -161,50 +161,50 @@ impl IntoResponse for AppError {
 }
 
 #[derive(Parser, Debug)]
-#[clap(author, version, about, long_about = None)]
+#[arg(author, version, about, long_about = None)]
 pub struct Args {
     /// Bind address for REST api
-    #[clap(long, default_value = "0.0.0.0:8080")]
+    #[arg(long, default_value = "0.0.0.0:8080")]
     bind_addr: String,
 
     /// SQL DB Connection pool connections
-    #[clap(long, default_value_t = 10)]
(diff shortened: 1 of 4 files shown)


Himess:
Let’s enable a feature like clap/deprecated in the project so that everyone automatically gets warnings.
This way, if someone in the team accidentally uses old #[clap(...)] attributes or name, they will immediately get a warning.
Otherwise, people might unknowingly keep adding deprecated code again and again.

I can take it, if team is ok.

flaub:
I'm not sure what problem we're trying to solve with this. I'll note that big changes to `bento` are likely coming so I'm not too keen on making more changes at the moment.

PR #3158, opened 2025-05-09
chore: Update bls12_381 dependency from commit hash to stable tag

DeVikingMark:


flaub:
This needs the following command to run and then commit the results:
```
cargo xtask update-lock-files
```

flaub:
Superseded by #3369

PR #3163, opened 2025-05-12
ZKVM-1378: Set `sm_code` flags to lowest supported default

morgangibbs87:
- `ARG NVCC_APPEND_FLAGS` defined with `sm` values used to successfully build bento agent image POC pushed to Dockerhub


flaub:
@bobbobbio this needs to be included in `release-2.1`

flaub (approved):

PR #3169, opened 2025-05-15
Return correct error message for `Inv`

Wollac:


Wollac:
I ran into this: the blst crate tried to invert zero and I just got a cryptic panic.
Malachite does this check in an assertion anyway, so I doubt the performance impact will be much.
We should at least change the existing error message, as it is not correct.

PR #3170, opened 2025-05-15
fix: Correct function pointer type for nvtxNameCudaStreamA_impl_fnptr in NVTX implementation

avorylli:
Fix type mismatch in nvtxImpl.h by changing the function pointer type declaration from nvtxNameCudaStreamA_fakeimpl_fntype to the correct nvtxNameCudaStreamA_impl_fntype. 

It should solve a potential runtime issue when calling CUDA-related NVTX functions for profiling and tracing.


flaub:
This is vendor code which we copied from NVidia. If there are issues there, we should fix it upstream. I'm not aware of any issues with the current version we have checked in. Closing for now.

PR #3189, opened 2025-05-27
Executor Bigint Smallvec Spillage

ec2:
In the executor where bigints are handled, when loading, a Smallvec is allocated to read limbs into. Currently the inlined smallvec value is optimized for 8 u32 limbs. Although this code isn't related to BLS, BLS is used often by many people in the ZKVM and limbs sizes (at least in blst are 12 u32s), so through some profiling, I observed that we are constantly spilling over and heap allocating here anyways.

### DIFF
--- a/risc0/circuit/rv32im/src/execute/bigint.rs
+++ b/risc0/circuit/rv32im/src/execute/bigint.rs
@@ -70,7 +70,7 @@ impl<Risc0ContextT: Risc0Context> BigIntIO for BigIntIOImpl<'_, Risc0ContextT> {
         check_bigint_addr(start_addr, self.mode)?;
 
         let word_count = (count + 3) / 4;
-        let mut limbs = SmallVec::<[u32; 8]>::with_capacity(word_count as usize);
+        let mut limbs = SmallVec::<[u32; 12]>::with_capacity(word_count as usize);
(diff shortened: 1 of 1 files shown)


risc0/circuit/rv32im/src/execute/bigint.rs
```diff
@@ -70,7 +70,7 @@ impl<Risc0ContextT: Risc0Context> BigIntIO for BigIntIOImpl<'_, Risc0ContextT> {
         check_bigint_addr(start_addr, self.mode)?;
 
         let word_count = (count + 3) / 4;
-        let mut limbs = SmallVec::<[u32; 8]>::with_capacity(word_count as usize);
+        let mut limbs = SmallVec::<[u32; 12]>::with_capacity(word_count as usize);
```
bobbobbio, line 73: I would leave a small comment here about where 12 came from, now that you have some actual data, less we accidentally change it later thinking it was arbitrary.

bobbobbio (approved):
LGTM assuming you do as I suggest

risc0/circuit/rv32im/src/execute/bigint.rs
ec2: Done!

PR #3193, opened 2025-05-28
lower noisy bento log levels

austinabell:
https://github.com/boundless-xyz/boundless/pull/491 seemed to be dropped when migrating to risc0, so applying roughly those changes here as the log levels are very noisy and spam logs by default. Not urgent, but would be nice for the next builds.


flaub (approved):

bobbobbio:
> @bobbobbio this also needs to go into `release-2.1`

release-2.1 doesn't have bento at all

PR #3198, opened 2025-05-29
ZKVM-1410: Add metrics/logging to telemetry

bobbobbio:
This is a rebased and merged with RPC system version of https://github.com/risc0/risc0/pull/3191

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -464,6 +464,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert_matches"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


bobbobbio:
The failure is here:
```
thread 'tokio-runtime-worker' panicked at risc0/r0vm/src/actors/job.rs:253:14:
called `Option::unwrap()` on a `None` value
```
which is this new code added in this PR
```
  247     fn span_event<T>(&mut self, header: TaskHeader, name: T)
  248     where
  249         T: Into<Cow<'static, str>>,
  250     {
  251         self.pending_spans
  252             .get_mut(&header.global_id.task_id)
A>253             .unwrap()
  254             .add_event(name, vec![]);
  255     }
```

flaub (approved):

PR #3203, opened 2025-06-01
Upgrade default_prover to use new actor system

flaub:


risc0/r0vm/src/actors/mod.rs
```diff
@@ -114,6 +143,145 @@ pub(crate) async fn async_main(args: &Cli) -> Result<(), Box<dyn StdError>> {
...
+            .arg(secondary_config.file.path())
+            .arg("--addr")
+            .arg(local_addr.to_string())
+            .spawn()
+            .with_context(|| spawn_fail(&r0vm_path))?;
+        children.push(child);
+    }
+
+    let stdin_fd = stdin().as_raw_fd();
+    let socket = unsafe { StdUnixStream::from_raw_fd(stdin_fd) };
```
bobbobbio, line 222: I think technically you're not suppose to do this. Now two different objects think they own stdin and it will call `close` on it twice. This can have unintended consequences if you open another file-descriptor between the two `close` calls, although perhaps unlikely in this case since it closes `stdin` (if at all) after main exits.

perhaps the right thing to do here is call `dup` on this before handing it to the stream code

risc0/zkvm/src/host/client/prove/default.rs
```diff
@@ -0,0 +1,163 @@
...
+/// An implementation of a [Prover] that runs proof workloads via local `r0vm` cluster.
+impl DefaultProver {
+    /// Construct a [DefaultProver].
+    pub fn new<P: AsRef<Path>>(r0vm_path: P) -> Result<Self> {
+        let r0vm_path = r0vm_path.as_ref();
+
+        let (parent_socket, child_socket) = UnixStream::pair()?;
+        let mut cmd = Command::new(r0vm_path);
+        let child_fd = child_socket.into_raw_fd();
+        let child_file = unsafe { File::from_raw_fd(child_fd) };
```
bobbobbio, line 52: i think you can just do `Stdin::from_raw_fd` here without going to a `File`, the `stdin` call seems to convert it to that type anyway.

risc0/zkvm/src/host/client/prove/default.rs
```diff
@@ -0,0 +1,163 @@
...
+        let _ = self.stop().inspect_err(|errors| {
+            for err in errors {
+                tracing::warn!("{err}");
+            }
+        });
+    }
+}
+
+fn spawn_fail(path: &Path) -> String {
+    format!("Could not launch \"{}\".", path.to_string_lossy())
```
bobbobbio, line 86: `path.display()` is preferred

risc0/zkvm/src/host/client/prove/default.rs
```diff
@@ -0,0 +1,163 @@
...
+            .read_exact(&mut buf)
+            .context("error receiving RPC header")?;
+        let body_len: u32 = bincode::deserialize(&buf).context("error deserializing RPC header")?;
+        let mut buf = vec![0u8; body_len as usize];
+        self.socket
+            .borrow_mut()
+            .read_exact(&mut buf)
+            .context("error receiving RPC body")?;
+        let job_info: JobInfo =
+            bincode::deserialize(&buf).context("error deserializing RPC body")?;
```
bobbobbio, line 130: we could probably extract this into a function that we could re-use in the `r0vm` `rpc_main`

bobbobbio (approved):

risc0/zkvm/src/host/client/prove/default.rs
flaub: There's a conversion to a `Stdio` and only from `File` (and other types) are supported, but I couldn't get going direction from `UnixStream` to work.

risc0/zkvm/src/host/client/prove/default.rs
flaub: For sure, this isn't really ready to land yet.

risc0/zkvm/src/host/client/prove/default.rs
bobbobbio: I had I typo with what I was saying, I guess I meant you could use this function https://doc.rust-lang.org/std/process/struct.Stdio.html#impl-FromRawFd-for-Stdio

flaub: OK, I fixed this.

PR #3217, opened 2025-06-18
Fix potential panic in sys_alloc_aligned functions

reject-i:
I found a potential issue in the syscall.rs file where the sys_alloc_aligned functions could panic when called with invalid parameters. The functions use unwrap() on the result of Layout::from_size_align(), which will panic if the alignment is not a power of two or if it's too large.

Since these are low-level system calls in the zkVM, a panic could lead to unpredictable behavior. I've added parameter validation to prevent panics by:

1. Checking if alignment is a power of two
2. Ensuring alignment is not too large (< 1 << 29)
3. Verifying bytes don't exceed isize::MAX
4. Properly handling errors from Layout::from_size_align()


bobbobbio:
The checks you added seem covered by the `Layout` constructor and thus redundant.

> a panic could lead to unpredictable behavior.

what do you mean by this? It seems more like the opposite is true, returning `NULL` from the memory allocation function can result in undefined behavior easily if you're not careful, whereas a panic is very well defined.

PR #3220, opened 2025-06-19
Implement proof of verifiable work (PoVW)

nategraf:
This PR implements the PoVW prover, core data types, Merkle tree for nonces, and the Log Builder guest.

Related: https://github.com/risc0/zirgen/pull/250
Related: https://github.com/boundless-xyz/boundless/pull/847


risc0/binfmt/src/elf.rs
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-extern crate alloc;
-
 use alloc::{collections::BTreeMap, vec, vec::Vec};
```
flaub: I have a feeling this was done this way to allow `no_std` to work with different feature flags on this crate.

nategraf: Ah, thank you. I will revert this change

risc0/circuit/recursion/src/prove/program.rs
```diff
@@ -54,7 +54,7 @@ impl Program {
             po2,
         };
         assert_eq!(prog.code.len() % RECURSION_CODE_SIZE, 0);
-        assert!(prog.code.len() <= (RECURSION_CODE_SIZE * (1 << po2) - ZK_CYCLES));
+        assert!(prog.code.len() <= (RECURSION_CODE_SIZE * ((1 << po2) - ZK_CYCLES)));
```
flaub, line 57: oh snap!

risc0/circuit/rv32im/src/prove/witgen/mod.rs
```diff
@@ -306,6 +307,27 @@ fn build_global_vec(segment: &Segment, trace: &PreflightTrace) -> Vec<Val> {
     // shutdown_cycle
     global[LAYOUT_GLOBAL.shutdown_cycle._super.offset] = segment.segment_threshold.into();
 
+    // povw nonce
+    // Split the U256 nonce into shorts LE shorts and assign to the globals.
```
flaub:
```suggestion
    // Split the U256 nonce into LE shorts and assign to the globals.
```

risc0/povw/guests/log-builder/src/main.rs
```diff
@@ -0,0 +1,90 @@
...
+    guest::{Input, Journal, State},
+    Job, WorkLog,
+};
+use risc0_zkvm::{guest::env, sha::Digestible, Digest};
+use ruint::Uint;
+
+type U96 = Uint<96, 2>;
+
+fn main() {
+    let input: Input = borsh::from_slice(&env::read_frame()).unwrap();
```
flaub, line 25: Why borsh?

risc0/povw/src/consts.rs
```diff
@@ -0,0 +1,581 @@
...
+    digest!("a2728ecfea9ff6d878ecf5e34d8fcd75b223f2d558704eabce4705b6abb89b81"),
+    digest!("d017503dbfbd79541e5c3cff98d49d634635428b8a0ff90dcc2f90381fb2b5de"),
+    digest!("2adceb7571aac8b5d1cdacb00cc6f3e4ced0c5f390380e54326e4020bc734593"),
+    digest!("9efa59a9bc1af73953e3c616542bec53b78ad16dab8638079a8f3c08b6cecd41"),
+    digest!("7a026caea53e29e8401b9754bcff860d7045d2e7d03eb45bf4f1dbc8567898ca"),
+    digest!("f6de01586aab597043763280bd07342f7c6783a900bf74a02142c6d90d40acc8"),
+];
+
+#[cfg(test)]
+use crate::{join, Bitmap};
```
flaub:
```suggestion
#[cfg(test)]
mod tests {
use crate::{join, Bitmap};
...
}
```

risc0/povw/src/lib.rs
```diff
@@ -0,0 +1,996 @@
...
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct WorkLog {
+    pub jobs: BTreeMap<u64, Job>,
+}
+
+impl WorkLog {
+    pub const TREE_HEIGHT: usize = Job::TREE_HEIGHT + 64;
```
flaub: Why 64?

risc0/circuit/recursion/src/prove/program.rs
nategraf: Debugging would have been a lot faster if this check was working 😅

risc0/povw/src/consts.rs
nategraf: 25ffcf2298ee7087caa2a38fec91cfa51009b6a1

risc0/povw/src/lib.rs
nategraf: I've done some work to try to make this more clear in 25c5694a2daae335b09f65a1a98b2a684c324352

risc0/povw/guests/log-builder/src/main.rs
nategraf: It's not a strong preference. I considered `bincode`, `postcard` and `borsh` and ended up deciding to use `borsh` because it guarantees the property of its encodings being canonical, which prevents issues with `env::verify` [below](https://github.com/risc0/risc0/blob/25c5694a2daae335b09f65a1a98b2a684c324352/risc0/povw/guests/log-builder/src/main.rs#L31) in that if when re-serialized the journal does not match what was committed, the verification will fail.

risc0/circuit/recursion/build.rs
```diff
@@ -50,15 +52,25 @@ fn download_zkr() {
...
+    // Check the ZKR archive in the out dir and the src dir. Use the out file if it matches the
+    // hash above, but also warn the developer if there exists a copy in the src dir and it does
+    // not match.
+    let zkr_out_valid = out_path.exists() && check_sha2(&out_path);
+
+    if src_path.exists() {
+        if check_sha2(&src_path) {
+            if !zkr_out_valid {
+                fs::copy(&src_path, &out_path).unwrap();
+            }
```
flaub: I think we need to be careful about changes here. The last set of changes were specifically allow nix users to be able to override this behavior so that they can ensure no downloads occur during a build. What specifically was the goal with this refactor?

risc0/circuit/recursion/src/prove/mod.rs
```diff
@@ -66,6 +66,13 @@ impl RecursionReceipt {
     pub fn seal_size(&self) -> usize {
         core::mem::size_of_val(self.seal.as_slice())
     }
+
+    /// Allocated a [VecDeque] and copies the output stream into it for decoding.
```
flaub:
```suggestion
    /// Allocates a [VecDeque] and copies the output stream into it for decoding.
```

risc0/circuit/recursion/src/prove/witgen.rs
```diff
@@ -98,6 +98,23 @@ where
...
+                &noise,
+                CIRCUIT.data_size(),      // from_rows
+                ZK_CYCLES,                // from_cols
+                0,                        // from_offset
+                ZK_CYCLES,                // from_stride
+                total_cycles - ZK_CYCLES, // into_offset
+                total_cycles,             // into_stride
+            );
+        });
+
```
flaub, line 117: Does this imply that noise was not being added? I'm a bit confused by this.

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -279,14 +284,19 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
...
                         dump_path: None,
+                        povw_nonce: self.povw_nonce(segment_counter),
                     };
                     if commit_sender.send(req).is_err() {
                         return Err(segment_callback_thread.join().unwrap().unwrap_err());
                     }
 
-                    segment_counter += 1;
+                    // NOTE: There is no reasonable scenario where a session will have more than 4B
+                    // segments, but its possible.
```
flaub, line 296: Wasn't `segment_counter` a `u64` before, so we wouldn't need to worry about the overflow. For my understanding, what requires the `segment_counter` to be a `u32`?

risc0/core/Cargo.toml
```diff
@@ -9,7 +9,7 @@ repository = { workspace = true }
 
 [dependencies]
 bytemuck = { version = "1.12", features = ["derive"] }
-bytemuck_derive = "1.9.3"
+bytemuck_derive = "1.8.1" # DO NOT MERGE
```
flaub: Curious about this one.

risc0/povw/guests/log-builder/src/main.rs
flaub: 👍

risc0/zkvm/src/host/protos/core.proto
```diff
@@ -41,6 +42,14 @@ message InnerReceipt {
   }
 }
 
+message GenericReceipt {
+  oneof kind {
+    SuccinctReceipt succinct = 2;
+    FakeReceipt fake = 3;
+    Groth16Receipt groth16 = 4;
+  }
+}
```
flaub: We already have an `InnerReceipt` type, could we just re-use that?

risc0/zkvm/src/host/server/prove/prover_impl.rs
```diff
@@ -147,58 +152,76 @@ impl ProverServer for ProverImpl {
...
+            );
+            return Ok(ProveInfo {
+                receipt,
+                work_receipt: work_receipt.map(Into::into),
+                stats: session.stats(),
+            });
+        }
+
+        // As long as the checks above are exhaustive, this code is unreachable. If this statement
+        // is reached, this is an implementation error.
```
flaub, line 220: Is there a reason we can't use a `match` as before so that the compiler can tell us when we're missing an arm?

risc0/zkvm/src/receipt/succinct.rs
```diff
@@ -35,16 +35,17 @@ use risc0_zkp::{
         hash::{hash_suite_from_name, sha::Sha256},
     },
     verify::VerificationError,
+    MAX_CYCLES_PO2,
```
flaub: This will need to be selective based on the unstable feature, which is why it avoided being brought in with a `using`.

risc0/povw/src/lib.rs
```diff
@@ -0,0 +1,1107 @@
...
+            .unwrap_or(&Job::EMPTY)
+            .bitmap_at(index.wrapping_to::<u32>())
+    }
+}
+
+#[derive(Clone, Debug)]
+/// Representation of a range of used nonces in a job (i.e. a single continuation).
+///
+/// Stores the maximum used nonce index as a shorthand for the range [0, index_max].
+/// When set to None, represents an empty job with no used nonces.
```
flaub: Generally I think we put comments first and then attributes. Just a nit.

risc0/povw/src/prover.rs
```diff
@@ -0,0 +1,210 @@
...
+                self.log_id == work.nonce_max.log,
+                "work claim nonce min log id is not equal to the expected log id: {} != {}",
+                work.nonce_max.log,
+                self.log_id
+            );
+            let job_number = work.nonce_min.job;
+            ensure!(
+            job_number == work.nonce_max.job,
+            "work claim nonce min job number is not equal to nonce max job number: {} != {job_number}",
+            work.nonce_min.job
```
flaub: seems like cargo fmt didn't work on `ensure!` likely since it's a macro

risc0/zkvm/Cargo.toml
```diff
@@ -68,10 +68,11 @@ tracing = { version = "0.1", default-features = false, features = [
 
 # Host dependencies
 [target.'cfg(not(target_os = "zkvm"))'.dependencies]
-addr2line = { version = "0.24.2", optional = true, features = [
+addr2line = { version = "0.24.2", features = [
   "rustc-demangle",
   "std",
-] }
+], optional = true }
```
flaub: Could you revert this part? We're trying to have the `features = ` be the last part of the dependency line since it's almost always dynamic.

risc0/zkvm/src/receipt/succinct.rs
nategraf: 2d86aae43fcdad759cf427f6fa6c5e2de32d9bf7

risc0/circuit/recursion/build.rs
nategraf: My goal here is to get a warning if I forget to update the hash in the `build.rs` file above. It's something I had forgotten needed to be done, and took me a minute to remember we I first started, and is also a step that's easy to forget when making multiple iterations. It's not supposed to have functional changes though, so I can revert this if you think it's an issue.

risc0/circuit/recursion/src/prove/witgen.rs
nategraf: That's right. When Jeremy and I took a look at this, in debugging the issue he noticed that this was missing.

risc0/circuit/rv32im/src/execute/executor.rs
nategraf: It's a u32 for the purposes of the segment index in the PoVW nonce.

risc0/core/Cargo.toml
nategraf: I've found that Cargo will not let me have `risc0-zkvm 2.x` and use `main` in the same workspace, which I do in the Boundless repo to develop with the new PoVW features on my PoVW PR there.

I imagine this may be problem we run into as well if we do implement backwards compatibility for verification. I left this note here to avoid this getting into `main`. And I've been meaning to ask: why did we upgrade this requirement 1.9.3? If 1.9.3 is required, then maybe there is still a way around this that we can find. It also may not be the end of the world, but a bit friction-full if we absolutely cannot have those two versions in the same workspace.

risc0/zkvm/src/host/protos/core.proto
nategraf: Yes, that's an option. It simply will be a deserialization error if the `kind` is `composite`. That seems reasonable though. Do you think that's better? If so, I'll do that

risc0/zkvm/src/host/server/prove/prover_impl.rs
nategraf: I tried factoring it that way, and it ended up being pretty ugly in my opinion. To have this be checked properly, this has to happen in a single `match`, which means repeating an shared parts of each flow. I can do that though if you think it would be better.

risc0/povw/src/lib.rs
nategraf: 7e4468e83e6f0ecc076032c69b8f77f2a1ca785f

risc0/zkvm/Cargo.toml
nategraf: Ok, yes. I accidentally changed that when resolving a merge conflict ae0f0e446449f90cb1c795a15536924f8acc741a

risc0/povw/src/prover.rs
nategraf: 89d62d8a35d77a0e42538f42d1f62dfb82c6f036

risc0/zkvm/src/host/protos/core.proto
flaub: I think it'd be better for now since it will make this PR smaller.

risc0/circuit/recursion/build.rs
flaub: I'm trying to remember why we had a `remove_file` that is now dropped, so it does feel like something functional did change, but I don't fully understand this diff.

nategraf: I've reverted this change in 9df10c28d647e1b7ce480d048601e7aea2bb691b

nategraf:
I worked on updating `PovwLogId` to be opaque bytes instead of an alias for `U160`. In my opinion, it's more pain than its worth to implement all the required traits and functions.

nategraf:
I've pushed a commit, fa285f62cdcf81db69321a245bc9fd5cd9773f5c, to avoid the addition of an extra comparison in the common case in `load_u32` in preflight. I don't have a ready benchmark to say the previous version was an issue, or that this version is better.

risc0/zkvm/src/host/protos/core.proto
nategraf: c175ea5612c7ff42a1a26d50d49bb41aae3fca60

risc0/core/Cargo.toml
bobbobbio: My suggestion of what to try here is to change this to be just `bytemuck_derive = "1"` , I believe this should allow it to compile in your other repo, and as long as we don't actually care about which `1.x` version in the code we are using (cargo should already prefer latest when possible for us) it seems fine.

In my view the real sin was to have `bytemuch_derive = "=XX"` in the 2.X crates, and this just works around it. Going forward I think we should avoid requiring exact versions for crates since it causes issues like this. 

Another solution would be to backport a change to the 2.x crates, but I think making this be just `= "1"` on main doesn't seem bad to me at all.

bonsai/sdk/src/lib.rs
```diff
@@ -299,7 +299,7 @@ pub mod responses {
     /// Snark proof request object
     #[derive(Deserialize, Serialize)]
     pub struct SnarkReq {
-        /// Existing Session ID from [super::SessionId]
+        /// Existing Session ID from `SessionId`
```
flaub, line 302: Why this change?

nategraf: I noticed a few broken links and ended up running `cargo doc` across the repo. This was one of the broken links I couldn't actually resolve because whether the target should be `blocking` or `non_blocking` depends on feature flags. So I removed the link.

risc0/core/Cargo.toml
flaub: I'm fine with this on paper. It will usually work out, but in the past we've run into situations where a specific version won't compile for one reason or another (usually conflicts with other dependencies), and so to workaround we've had to pin to a specific version. This usually isn't permanent and a newer version usually does come out so we don't need to do a pin. 

In this specific case I think using `bytemuck_derive = "1"` is appropriate.

flaub (approved):

PR #3225, opened 2025-06-24
Prepare for guest code upgrade to Rust 1.88.0

bobbobbio:
This updates rzup to compile toolchains of a later Rust version.

This also fixes a warning seen from compiling guest code with Rust 1.88.0


bobbobbio:
I made sure that Rust 1.81.0 also still builds after my changes to `rzup`

PR #3230, opened 2025-06-27
Fix cargo risczero build for guests using bigint2

nategraf:
Reported by bobafetdaor on Discord, `cargo risczero build` installed via `rzup` currently does not work to build guests that use `bigint2`. The issue is that the `unstable` flag is required on `risc0-build`, since this feature is still marked as `unstable` and a build-time env var needs to be set to enable it (see https://github.com/risc0/risc0/pull/2567).


flaub:
I'm ok with making this all stable

PR #3232, opened 2025-06-29
Example and docs for the ApiClient module

rezahsnz:
This PR adds usage examples and documentation to the ApiClient module. Here is the list of affected methods:
  - `from_env`
  - `prove`
  - `execute`
  - `prove_segment`
  - `lift`
  - `join`
  - `verify`

The usage examples are written in a way to be concise, step by step, and all-inclusive with the focus being on a pipeline-like proving scheme: proof = execute -> prove -> lift -> join. Besides examples, a bit of context have been added to some methods while preserving the original documentation.

### DIFF
--- a/risc0/zkvm/src/host/api/client.rs
+++ b/risc0/zkvm/src/host/api/client.rs
@@ -82,6 +82,12 @@ impl Client {
     }
 
     /// Construct a [Client] based on environment variables.
+    ///
+    /// Example:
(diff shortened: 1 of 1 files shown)


rezahsnz:
`reports-and-benchmarks` check(s) fail. Not sure what is going on here and I think I cannot do anything about it. Let me know if something is wrong on my end.

PR #3247, opened 2025-07-08
Unify byte slice handling in run_guest_framed calls in serde benchmark

Galoretka:
Replaced unnecessary usage of bytemuck::cast_slice(&encoded) with a direct reference to &encoded in the run_guest_framed calls within the serde benchmark function. This change improves code clarity and consistency, as the encoded data is already a Vec<u8>, making the cast redundant. All framed benchmark calls now use a unified approach for passing byte slices.


flaub (approved):

flaub:
Could you run `cargo fmt` please?

PR #3256, opened 2025-07-11
fix duplicate in Update ceremony.Dockerfile

moshderte:
Repeating WORKDIR /src/ multiple times in a Dockerfile is not technically an error — Docker allows multiple WORKDIR instructions and will simply change the working directory each time.

However:

Repeating the same WORKDIR instruction consecutively is unnecessary and redundant.

It clutters the Dockerfile and may confuse readers, making it seem like something is being overridden or forgotten.

For clarity and maintainability, it’s better to set WORKDIR /src/ once and avoid repeating it.

So, while having it three times won’t break your build, it’s cleaner and more efficient to keep only one WORKDIR /src/ and remove the duplicates.


moshderte:
@flaub if you have time check pls

PR #3262, opened 2025-07-15
Backport bento upload limit removal

austinabell:
Given we need this to be on 2.1 specifically


bobbobbio:
seemed like the conclusion was to forgo backporting this to 2.1, we can reopen if that changes.

PR #3263, opened 2025-07-15
Add risc0-groth16 component to rzup

flaub:


flaub:
I think it's because this got initially based on my other PR. Let me start a new branch to see if we can't get this to go.

PR #3269, opened 2025-07-17
specify Circom version for trusted setup ceremony

### LINKED ISSUE #3224: Error when independently verifying Risc0 trusted setup

### Motivation

Hi!
As a part of L2BEAT effort of verifying rollup's zk setups, I tried to repeat the steps from [your docs](https://dev.risczero.com/api/trusted-setup-ceremony#the-transcript-matches-the-circuit) to verify the trusted setup ceremony. I got the error `[ERROR] snarkJS: INVALID:  Different circuit parameters` after literally performing the 6 steps you mention.

I think the issue happens on step 4 of generating r1cs circuit from `stark_verify.circom` file, the hash that I get differs from the one provided on the docs:
```
$ shasum -a 256 stark_verify.r1cs 
515e28822790f51e22f3ac33815fbd9a9532e2b3937c39aa7ed4a6fdbb6d2702  stark_verify.r1cs
```
The hash value in the docs is `84d3c34b7c0eb55ad1b16b24f75e0b9de307f7b74089ea4a20a998390ee24178`.

When I directly download the r1cs circuit file from the server, the trusted setup is verified correctly. `snarkJS: ZKey Ok!`

### Request

Please update the Risc0 docs to allow users independently verify the trusted setup.

### Solution

My guess is that the current version of circom differs from the one used when writing the docs and thus the `stark_verify.circom` circuit is compiled differently. IMO the best solution would be to pinpoint the exact version of circom that is needed to be used for compiling `stark_verify.circom`, this should be easy enough since circom installation guide builds it from the GH repo source. So the docs could say: "check out this particular circom version before building and installing it".

### Are you willing to help with this request?

Maybe... (I have elaborated on how above), Yes!

Wollac:
fixes #3224


flaub:
We probably need to backport this update to all the previous versions that document this.

Wollac:
I backported those documentation changes for all v2.x versions.
 It seems strange to do it for earlier versions since they are older than the mentioned Circom v2.2.2 version. For those, we would have to reference a different Circom version that does not have non-x86 support, which also seems undesirable. Overall, I think changing it to v2 is a reasonable compromise.

PR #3271, opened 2025-07-17
Add s3 region to Bento

ec2:
Bento works with minio, but fails when using S3 in any other region besides us-west-2. This allows the region to be set.

### DIFF
--- a/bento/crates/api/src/lib.rs
+++ b/bento/crates/api/src/lib.rs
@@ -191,6 +191,10 @@ pub struct Args {
     #[clap(env)]
     s3_url: String,
 
+    /// S3 region, can be anything is using minio
+    #[clap(env, default_value = "us-west-2")]
+    s3_region: String,
+
     /// Executor timeout in seconds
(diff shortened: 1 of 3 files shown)


bento/crates/api/src/lib.rs
```diff
@@ -191,6 +191,10 @@ pub struct Args {
     #[clap(env)]
     s3_url: String,
 
+    /// S3 region, can be anything is using minio
```
bobbobbio: is -> if

bento/crates/workflow/src/lib.rs
```diff
@@ -90,6 +90,10 @@ pub struct Args {
     #[clap(env)]
     pub s3_url: String,
 
+    /// S3 region, can be anything is using minio
```
bobbobbio: is -> if

ec2:
@bobbobbio I think we should leave this as is for now. We can open up a follow up for clean up later :)

bobbobbio (approved):

PR #3276, opened 2025-07-19
rzup: Add publish command to upload to S3

bobbobbio:
This PR adds two new commands to the `rzup` CLI.
- `publish upload` uploads a `.tar.xz` file to S3 under `rzup/components/<component>/sha265` and updates the `rzup/components/<component>/distribution_manifest.json` to mention the artifact
- `publish set-latest` which updates the `latest_version` field fo the `distribution_manifest.json`

`publish` help text
```
Publish a component on S3

Usage: rzup publish [OPTIONS] <COMMAND>

Commands:
  upload
  set-latest
  help        Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose  Enable verbose output
  -q, --quiet    Suppress output
  -h, --help     Print help

Discussion:
    Publishes a component to S3 at a particular version, or updates the current latest version.

    Requires that AWS credentials be available in the environment.

    Each version of a component is either tagged with a target, or is target agnostic. A particular
    version can only have one target-agnostic artifact, or many target specific artifacts, but each
    target may only have one artifact each.
```
`publish upload` help text
```
Usage: rzup publish upload [OPTIONS] <--target-agnostic|--target-triple <TARGET_TRIPLE>> <NAME> <VERSION> <PAYLOAD>

Arguments:
  <NAME>     Name of component to publish (e.g. rust) [possible values: cargo-risczero, cpp, gdb, r0vm, rust, risc0-groth16, self]
  <VERSION>  Version of the component to publish (e.g. 1.0.0)
  <PAYLOAD>  The path to the archive

Options:
      --target-agnostic                Is the component the same for all targets
      --target-triple <TARGET_TRIPLE>  The target triple of the component
      --force                          If the component at the given target / version already exists, replace it
  -v, --verbose                        Enable verbose output
  -q, --quiet                          Suppress output
  -h, --help                           Print help
```

`publish set-latest` help text
```
Usage: rzup publish set-latest [OPTIONS] <NAME> <VERSION>

Arguments:
  <NAME>     Name of component to update the latest version of (e.g. rust) [possible values: cargo-risczero, cpp, gdb, r0vm, rust, risc0-groth16, self]
  <VERSION>  Version to set (e.g. 1.0.0)

Options:
  -v, --verbose  Enable verbose output
  -q, --quiet    Suppress output
  -h, --help     Print help
```

### DIFF
--- a/rzup/Cargo.toml
+++ b/rzup/Cargo.toml
@@ -12,15 +12,29 @@ name = "rzup"
 path = "src/bin/rzup.rs"
 
 [features]
-default = ["cli", "install"]
+default = ["cli", "install", "publish"]
 cli = ["dep:clap", "dep:indicatif", "dep:colored", "dep:is-terminal"]
 install = ["dep:reqwest", "dep:tar", "dep:xz", "dep:flate2", "dep:fs2"]
+publish = [
+    "dep:aws-config",
+    "dep:aws-credential-types",
+    "dep:aws-sigv4",
+    "dep:http",
+    "dep:reqwest",
+    "dep:serde_json",
+    "dep:sha2",
+    "dep:tokio",
+]
 
 [dependencies]
+aws-config = { version = "1.8", optional = true, features = ["behavior-version-latest"] }
+aws-credential-types = { version = "1.2", optional = true }
+aws-sigv4 = { version = "1.3", optional = true }
 clap = { version = "4.5.23", features = ["derive"], optional = true }
 colored = { version = "3.0.0", optional = true }
 flate2 = { version = "1.0.35", optional = true }
 fs2 = { version = "0.4.3", optional = true }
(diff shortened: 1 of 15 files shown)


rzup/Cargo.toml
```diff
@@ -12,15 +12,29 @@ name = "rzup"
 path = "src/bin/rzup.rs"
 
 [features]
-default = ["cli", "install"]
+default = ["cli", "install", "publish"]
```
flaub, line 15: I'm thinking maybe this shouldn't be the default so that it's easier for folks to test out an internal version.

bobbobbio: We could easily do this, I'm not sure I understand the reasoning though. Just compiling in publishing support shouldn't really affect much of anything. It should be a small amount of extra compilation time and code.

PR #3306, opened 2025-08-09
fix(povw/prover): correct error messages in WorkLogUpdateProver

radik878:
- Use "nonce max log id" in the second ensure! message (was "nonce min log id")
- Print both nonce_min.job and nonce_max.job in the job-number mismatch message

Improves clarity of diagnostics when validating work claims.


flaub:
@nategraf Need your help on this one

risc0/povw/src/prover.rs
```diff
@@ -155,15 +155,16 @@ impl<P: Prover> WorkLogUpdateProver<P> {
...
                 self.log_id
             );
             let job_number = work.nonce_min.job;
             ensure!(
                 job_number == work.nonce_max.job,
-                "work claim nonce min job number is not equal to nonce max job number: {} != {job_number}",
-                work.nonce_min.job
+                "work claim nonce min job number is not equal to nonce max job number: {} != {}",
+                work.nonce_min.job,
+                work.nonce_max.job
```
nategraf: This change is incorrect

nategraf (changes requested):

risc0/povw/src/prover.rs
radik878: > This change is incorrect

sorry, my bad, fixed

nategraf (approved):

PR #3309, opened 2025-08-11
docs(field): correct polynomial reduction comment in baby_bear.rs

Bashmunta:
Correct comment to “reduce modulo x^4 + BETA (i.e., x^4 = -BETA)” in risc0/core/src/field/baby_bear.rs.
Remove unused import alias Elem as FieldElem.


Bashmunta:
oops, messed up with import

PR #3317, opened 2025-08-12
risc0-linux toolchain proof of concept

flaub:


risc0/build/src/config.rs
```diff
@@ -100,15 +102,35 @@ pub struct GuestOptions {
...
+    #[builder(setter(into, strip_option))]
     pub kernel: Option<Vec<u8>>,
+
+    /// Use specified toolchain.
+    #[builder(setter(into, strip_option))]
+    pub toolchain: Option<String>,
+
+    /// Use specified target.
+    #[builder(setter(into, strip_option))]
+    pub target: Option<String>,
```
bobbobbio, line 114: I wonder if we should instead use two `enum`s here, and also we should think about how we want to specify toolchain version.

risc0/build/src/lib.rs
```diff
@@ -392,11 +379,20 @@ pub fn cargo_command(subcmd: &str, rustc_flags: &[String]) -> Command {
...
+    let target = guest_info.options.target();
+    let rustc = sanitized_cmd("rustup")
+        .arg(format!("+{toolchain}"))
+        .args(["which", "rustc"])
+        .output()
+        .expect("rustup failed to find risc0 toolchain")
+        .stdout;
+    let rustc = String::from_utf8(rustc).unwrap();
+    let rustc = rustc.trim();
+    println!("Using rustc: {rustc}");
```
bobbobbio, line 392: I think ultimately we want to use `rzup` to find the toolchains, its the only way that will support multiple versions when we allow guests to specify the version of the toolchain it wants

Also it doesn't feel like an improvement to go from using the toolchain specified in the `rzup` settings to now relying on this rustup symlink again. I'd rather not rely on the symlink

bobbobbio (commented):
If this is just a proof of concept, I wonder if it would be better to just make separate code paths for this that can be enabled via an unstable flag or something we don't plan to keep.

I think it would be easier to just get this committed if we avoided changing the behavior of the normal code-path and avoided making changes to the public interface.

risc0/build/src/lib.rs
flaub: I don't think rzup should be trying to replace what rustup does, basically. Could we maybe layer to have both? I think eventually we won't need rzup to even manage the rust toolchain at all.

risc0/build/src/config.rs
flaub: I think `rust-toolchain.toml` can be used to specify the version. I think we need a string here to just point to the named toolchain that rustup will manage.

PR #3318, opened 2025-08-12
Add support for Groth16 shrink-wrapping to the actor system

bobbobbio:
Adds new `ShrinkWrapRequest` to actor system which is fulfilled by `shrink_wrap::JobActor`. The `ShrinkWrapRequest` has a `kind` field which is `ShrinkWrapKind`, the only variant in it right now is `Groth16`.

There is a new `ShrinkWrap` task type that workers can ask for. We could split this out more in the future so that workers could subscribe to only certain kinds of shrink wrapping. The `ShrinkWrap` task always runs as a GPU task. (the workers don't have support for CPU proving right now)

Implements the `compress` method of `DefaultProver` when asking for a groth16 receipt, also add groth16 shrink-wrapping when calling `prove_with_ctx` and the `receipt_kind` is `Groth16`

Implements the snark API calls in the actor system API.

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -262,7 +262,7 @@ mod test {
         compare_image_id(
             &guest_list,
             "hello_commit",
-            "0b25154f31ff970988c49d81539a4daaccee785b48e10775b6ee37b58112b0c6",
+            "4a038d5f7c0332ced32f114ac8c2ed1af693e8b9a638e5f9d28b75e00cef7469",
         );
     }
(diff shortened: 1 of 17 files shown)


bobbobbio:
I think these files need to be updated to include a proper delay for groth16 but I'm not sure I want to do that as part of this PR
```
./risc0/r0vm/testdata/rtx_4000_ada-20.toml
./risc0/r0vm/testdata/rtx_4000_ada-21.toml
```

PR #3323, opened 2025-08-15
Fix sys_read reported length for short unaligned reads

nategraf:
I stumbled upon this with the following guest code:

```rust
let mut buf = Vec::new();
env::stdin().read_to_end(&mut buf).unwrap()
```

When the host supplies `b"hello"`, the result is `b"hello\0\0\0"`.

With some work, I found two issues in the `sys_read` implementation:
1. On a short read with unaligned start less than `unaligned_at_start`, the function will continue as if it had read enough bytes to become aligned.
2. The host is allowed to have the `nread` value exceed the requests bytes for the read, since the checks we made are in `debug_assert` and guests are essentially always compiled in release mode.

### DIFF
--- a/risc0/zkvm/platform/src/syscall.rs
+++ b/risc0/zkvm/platform/src/syscall.rs
@@ -562,13 +562,8 @@ pub unsafe extern "C" fn sys_read(fd: u32, recv_ptr: *mut u8, nread: usize) -> u
         assert!(nread_first as usize <= unaligned_at_start);
 
         // Copy the firstword provided by the host into the unaligned start bytes.
-        // NOTE: When nread_first < unaligned_at_start, we still copy unaligned_at_start bytes.
+        // NOTE: When nread_first < unaligned_at_start, we still copy unaligned_at_start. See #1557
         let main_ptr = fill_from_word(recv_ptr, firstword, unaligned_at_start);
-        // If the host returned less data than is required to align up to the word boundary, return
-        // early, as the main read cannot proceed.
-        if (nread_first as usize) < unaligned_at_start {
-            return nread_first as usize;
-        }
         // If the amount read is the full amount requested, return early.
         if nread == unaligned_at_start {
             return nread;
--- a/risc0/zkvm/platform/src/syscall.rs
+++ b/risc0/zkvm/platform/src/syscall.rs
@@ -562,6 +562,7 @@ pub unsafe extern "C" fn sys_read(fd: u32, recv_ptr: *mut u8, nread: usize) -> u
         assert!(nread_first as usize <= unaligned_at_start);
 
         // Copy the firstword provided by the host into the unaligned start bytes.
+        // NOTE: When nread_first < unaligned_at_start, we still copy unaligned_at_start bytes.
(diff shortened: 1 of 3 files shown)


risc0/zkvm/platform/src/syscall.rs
```diff
@@ -559,12 +559,18 @@ pub unsafe extern "C" fn sys_read(fd: u32, recv_ptr: *mut u8, nread: usize) -> u
         // Read unaligned bytes into "firstword".
         let Return(nread_first, firstword) =
             syscall_2(nr::SYS_READ, null_mut(), 0, fd, unaligned_at_start as u32);
-        debug_assert_eq!(nread_first as usize, unaligned_at_start);
+        assert!(nread_first as usize <= unaligned_at_start);
 
-        // Align up to a word boundary to do the main copy.
+        // Copy the firstword provided by the host into the unaligned start bytes.
+        // NOTE: When nread_first < unaligned_at_start, we still copy unaligned_at_start bytes.
         let main_ptr = fill_from_word(recv_ptr, firstword, unaligned_at_start);
```
bobbobbio, line 566: I think this function will still write zeros into the calling buffer when the new `if` is taken, I think that is (albeit subtle) unexpected behavior, (if the return-value of this function indicates that we only read x bytes into the input buffer but wrote x+1 it violates normal constraints placed by similar read like functions)

I suggest we instead modify this function to also take in `nread_first` and only copy over the `min(nread_first, unaligned_at_start)` and somehow indicated to the caller that it needs to return, (maybe make it return the number of bytes copied have the caller calculate `main_ptr`)


Also inside `fill_from_word` there is another `debug_assert` that we should probably change to `assert`

bobbobbio (commented):
seems like there are still test failures, I will review more in-depth the tests later

risc0/zkvm/platform/src/syscall.rs
nategraf: We should add this as context, but check out https://github.com/risc0/risc0/pull/1557. There is some discussion in the PR, and the behavior we landed on what that we would always overwrite the amount requested instead of just the amount the host indicated that they provided.

Part of the context here is that the "copy-in" ecall we use to actually implement I/O always presents the host the option to overwrite the full length of the aligned ("main") part of the buffer. It does not stop short if the host indicates fewer bytes are read (in fact, the circuit does not even try to interpret the return registers). When there is less than the requested amount of data, our implemented host behavior is to overwrite with zeroes.

With the unaligned tail, we chose in #1557 to essentially continue this behavior forwards, implemented in rv32im. This can certainly all be seen as a bit odd, but its a situation where I don't know is there is a solution that would feel fully natural.

nategraf: > Also inside fill_from_word there is another debug_assert that we should probably change to assert

Its not as clear to me that that should be an assert. In this case, the `nfill` value is not controlled by the host. However, it is true that its not providing value in practice.

nategraf:
Looking at this again, my updated implementation does fix the issue I set our to fix, but breaks the target behavior set out in https://github.com/risc0/risc0/pull/1557.

PR #3326, opened 2025-08-15
Upgrade to Rust 2024 edition

bobbobbio:


risc0/zkp/src/verify/fri.rs
```diff
@@ -126,7 +126,7 @@ where
         let final_digest = hashfn.hash_elem_slice(final_coeffs);
         self.iop().commit(&final_digest);
         // Get the generator for the final polynomial evaluations
-        let gen = <F::Elem as RootsOfUnity>::ROU_FWD[log2_ceil(domain)];
+        let gen_ = <F::Elem as RootsOfUnity>::ROU_FWD[log2_ceil(domain)];
```
flaub, line 129: What's this about?

bobbobbio: https://doc.rust-lang.org/edition-guide/rust-2024/gen-keyword.html

PR #3327, opened 2025-08-18
Add support for RISC-V fence instruction

codygunton:
Companion of https://github.com/risc0/zirgen/pull/277. This implements handling of `fence` instructions as nops that only advance the pc. See the zirgen PR for more details.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -277,6 +277,8 @@ jobs:
         if: matrix.os == 'macOS'
       - name: test workspace
         run: ${{ matrix.test_command }} --release -F $FEATURE -F prove -F redis -F unstable --workspace --timings --exclude doc-test
+      - name: test risc0-r0vm (disable-dev-mode)
+        run: cargo test --release -p risc0-r0vm -F $FEATURE -F disable-dev-mode -- dev_mode
       - uses: actions/upload-artifact@v4
         with:
           name: cargo-timings-${{ matrix.os }}-${{ matrix.device }}-${{ matrix.partition }}
@@ -329,10 +331,6 @@ jobs:
       - run: cargo run --bin rzup -- --verbose install rust $RISC0_RUST_TOOLCHAIN_VERSION
       - run: cargo run --bin rzup -- --verbose install cpp $RISC0_CPP_TOOLCHAIN_VERSION
       - run: cargo run --bin rzup -- --verbose install risc0-groth16 $RISC0_GROTH16_VERSION
-      - name: test risc0-r0vm
-        run: ${{ matrix.test_command }} --release -p risc0-r0vm -F $FEATURE
-      - name: test risc0-r0vm (disable-dev-mode)
-        run: ${{ matrix.test_command }} --release -p risc0-r0vm -F $FEATURE -F disable-dev-mode
       - run: ${{ matrix.test_command }} -p cargo-risczero -F experimental
       - name: run datasheet generator (smoke test)
         run: cargo run --locked -F $FEATURE -F prove --example datasheet -- --json tmp/datasheet.json lift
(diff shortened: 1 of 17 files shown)


fence-test/add_fence_test.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
codygunton: Script to show how the fence test is defined and how it was added to the archives. Intend to remove before merging ofc

risc0/circuit/rv32im/src/execute/testutil.rs
```diff
@@ -302,6 +302,10 @@ impl Assembler {
     }
 
     pub fn die(&mut self) {
+        self.text.push(0x00000000);
```
codygunton, line 305: https://github.com/riscv/riscv-isa-manual/blob/5aaf918ae0264a64c70fb9079316761a5e70b6fd/src/intro.adoc?plain=1#L504 says that this is guaranteed to be an illegal instruction

risc0/circuit/rv32im/src/prove/tests.rs
```diff
@@ -83,7 +85,9 @@ mod riscv {
 
             let program = Program::load_elf(&elf, u32::MAX).unwrap();
             run_program(program);
+            return;
```
codygunton, line 88: Without this return and the panic, the test would pass even if the archive does not contain a test matching the supplied test_name.

risc0/zkvm/src/host/server/prove/tests.rs
```diff
@@ -387,7 +389,9 @@ mod riscv {
 
             let env = ExecutorEnv::default();
             prove_elf(env, &elf).unwrap();
+            return;
```
codygunton, line 395: Ditto above

codygunton:
> I suggest we update the existing riscv-test suite so that includes `fence`, see:
> 
> https://github.com/risc0/risc0/tree/main/risc0/zkvm/src/host/server/testdata https://github.com/risc0/risc0/tree/main/risc0/circuit/rv32im/src/prove/testdata

Thanks, yes, I had done one but missed the other. But suites updated.

codygunton:
> We'll need to replace all the current uses of `fence` before being able to land this.
> 
> For example: https://github.com/risc0/risc0/blob/main/risc0/zkos/v1compat/src/kernel.s#L102

Done. I replaced the use of `fence` by `0x00000000`, rebuilt kernel.s. Does that work? Wasn't sure how to test

codygunton:
Alright, we're ready for another CI run. I had to some pinned hash values: recursion zkr, 'stability' test values and also some control ids.

codygunton:
```

Caused by:
  process didn't exit successfully: `/home/ghrun-4/actions-runner/_work/risc0/risc0/target/release/build/risc0-circuit-recursion-f346369dd302e81f/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=RECURSION_SRC_PATH

  --- stderr
  Downloading https://risc0-artifacts.s3.us-west-2.amazonaws.com/zkr/81b25917a407f04006502cf8b4884ff90821a6247e71792cff58612efee6860c.zip

  thread 'main' panicked at risc0/circuit/recursion/build.rs:80:47:
  called `Result::unwrap()` on an `Err` value: Download(/home/ghrun-4/actions-runner/_work/risc0/risc0/target/release/build/risc0-circuit-recursion-2e2532e79cb52f7f/out/recursion_zkr.zip: (verification: unverified):
  ```
Should the upload of this new artifact have happened automatically via me using git lfs? I recall seeing git lfs uploads when pushing in recent days.

codygunton:
Hoping that the problem with the license check is that there was a blank line? There is clearly a license at the top of the files and all I did was generate them anyway...

flaub:
Check is now passing. I wonder if we need to modify the codegen in zirgen.

PR #3329, opened 2025-08-19
Add cargo risczero build --archive flag to save guest binaries to source tree

nategraf:
- **add support for archiving guests and flag for Docker in cargo-risczero build**
- **switch from --docker to --no-docker**
- **comment out setting of the root dir**
- **add ci job for cargo risczero build --archive**

PR #3330, opened 2025-08-19
Build PoVW log builder and check into git

nategraf:
In order to allow the Boundless PoVW code to use a stable Log Builder image ID and binary, this PR includes pre-built binaries. These binaries are produced using the `cargo risczero build --archive` which is implemented in https://github.com/risc0/risc0/pull/3329

- Re-export povw types from binfmt
- Build and include guest programs for povw


jbruestle (approved):

flaub:
Is it possible to use the `cargo risczero bake` command? It seems there's missing metadata to allow that to work.

risc0/povw/src/guest.rs
```diff
@@ -21,6 +21,15 @@ use serde::{Deserialize, Serialize};
 
 use crate::{Job, SubtreeOpening, WorkLog};
 
+/// Guest program for the Log Builder.
+#[cfg(feature = "prover")]
+pub const RISC0_POVW_LOG_BUILDER_ELF: &[u8] =
+    include_bytes!("../guests/log-builder/elfs/risc0-povw-log-builder.bin");
```
flaub, line 27: I would have expected this path to be:
```
../elfs/risc0-povw-log-builder.bin
```

flaub (approved):

risc0/povw/src/guest.rs
nategraf: It's in its own workspace. So that's where it is right now. I would like to explore merging the workspaces, but I'm still not sure how to do it well.

nategraf:
Looks like this got kicked out of the merge queue due to failing the the `check-semver` and `check-template` jobs.
https://github.com/risc0/risc0/actions/runs/17079677090/job/48430086856

PR #3331, opened 2025-08-19
3.0: Update docs & groth16 component error message

bobbobbio:
This updates docs concerning the following this
- README for `rzup publish`
- README for groth16 around publishing the component
- developer docs for groth16 shrink-wrapping
- developer docs for gdb support
- error message when groth16 component is missing


risc0/groth16/README.md
```diff
@@ -39,4 +39,23 @@ crate. With `ProverOpts::groth16()` it will produce a Groth16 proof.
 
 [risc0-zkvm]: https://docs.rs/risc0-zkvm/latest/risc0_zkvm/
 
+# Publishing the rzup Component
+The groth16 proving on GPU requires a special rzup component be installed. Publishing this
+omponent can be done with the following commands
```
pdg744:
```suggestion
component can be done with the following commands:
```

pdg744: [feel free to drop the colon at the end if that doesn't match typical dev style; opened the comment for the typo at the start of the line and kinda threw in the nit about the colon on a whim]

website/api/blockchain-integration/shrink-wrapping.md
```diff
@@ -0,0 +1,16 @@
+# Shrink-Wrapping Proofs
+
+The default proving options will result in the final proof being a composite receipt. This receipt
+can be quite large and grows linearly with respect to execution length. So instead a smaller receipt
```
pdg744:
```suggestion
can be quite large. So instead a smaller receipt
```

The detail about "growing linearly with execution length" is true of CompositeReceipts but not of SuccinctReceipts. (CompositeReceipts grow linearly with execution length; SuccintReceipts don't grow linearly but are still too large for posting on-chain.) 

In theory, users may want to use any of these three proof types: CompositeReceipts are fastest & largest, SuccinctReceipts are under 250kB but take a bit longer than CompositeReceipts, and Groth16Receipts are the smallest and have the highest latency. 

Suggesting a simple revision to sweep the detail under the rug, though I still kinda prefer the organizational choice to discuss these three receipt formats together rather than decoupling the ShrinkWrap from the other size-vs-latency optionality. Happy to defer to your judgment, @bobbobbio

bobbobbio: Removing this line I think is fine. I think it definitely could use more verbiage to describe the different receipt types. I'm going to try to keep it simple for the purposes of the release, but we should consider updating it.

website/api/generating-proofs/proving-options.md
```diff
@@ -1,13 +1,5 @@
 # Options for Proof Generation
```
pdg744, line 1: IMHO it feels like a mistake to me to have a page called "proving options" that omits a path-to-information about optionality around proof size, but I'm not opinionated enough to consider it blocking or require further discussion.

pdg744 (approved):
Approved with some notes -- feel free to resolve & land as you see fit (happy to discuss more if desired)

website/api/generating-proofs/proving-options.md
bobbobbio: I've added in a link to the new shrink-wrapping page in the note on this page

flaub (approved):
LGTM

PR #3337, opened 2025-08-21
Drop bento

flaub:
This has been migrated over to https://github.com/boundless-xyz/boundless/tree/main/bento


bobbobbio (approved):
A little unsavory to me to update `CODEOWNERS` in this way at the same time, maybe just mention it in the commit message

flaub:
OK, I'll split that out

PR #3339, opened 2025-08-21
Remove all the file I/O and JSON stuff from the groth16 kernels

bobbobbio:
This helps minimize the groth16 C++ code. Also ferrying stuff through files and JSON needlessly just wastes time.


xtask/src/setup_groth16.rs
```diff
@@ -77,14 +77,53 @@ impl SetupGroth16 {
         }
 
         // stark_verify.circom -> stark_verify_graph.bin
-        let graph_path = &witness_params.graph_path;
-        if !sh.path_exists(graph_path) {
+        let graph_path = setup_dir.join("stark_verify_final.zkey");
```
ec2: Is this not supposed to be `stark_verify_graph.bin`?

bobbobbio: yes it is. good catch

flaub (approved):

PR #3342, opened 2025-08-22
Improve CI performance

flaub:


Cargo.toml
```diff
@@ -25,7 +25,6 @@ members = [
   "risc0/zkos/v1compat",
   "risc0/zkp",
   "risc0/zkvm",
-  "risc0/zkvm/compat",
```
bobbobbio, line 28: Did you mean to disable this?

flaub: Yes, it should speed up tests

PR #3351, opened 2025-08-25
Refactor guest I/O with host

flaub:
* Heavy lifting now occurs in the kernel
* A syscall number is passed to the kernel so that it can dispatch, avoiding the need for an expensive strcmp
* Replace RISC-V instruction fence with unimp, which represents an intentional illegal instruction
* This implementation drops support for the v1 host/guest protocol.


bobbobbio (changes requested):
Seems like this undoes the recent `rzup` commit

bobbobbio (approved):

PR #3357, opened 2025-08-26
V3 compatibility prototype

jacobdweightman:


risc0/zkp/src/prove/merkle.rs
```diff
@@ -394,4 +409,91 @@ mod tests {
...
+            TAPSET
+        }
+    }
+    impl CircuitCoreDef<BabyBear> for HelloCircuit {}
+
+    // pub trait CircuitCoreDef<F: Field>: CircuitInfo + PolyExt<F> + TapsProvider {}
+
+    #[test]
+    fn verify_v3_stark_proof() {
+        // TODO: is there a better way to include this?
```
flaub: How about using `include_bytes!`?

flaub (approved):

PR #3362, opened 2025-08-28
Move guest binaries in risc0-povw to allow publishing and add defaults to log-builder prover

nategraf:
In https://github.com/risc0/risc0/pull/3330, I added built binaries for the Log Builder guest. Because I added them to the inner guest workspace, and then imported them in the library, this fails to build when pulling from `crates.io`.

This PR moves the pre-built guest binaries. It also sets a default on the Log Builder prover to use the included bin and ID for easier instantiation.


nategraf:
testing that this indeed fixes the publishing issue is not as straightforward as I would hope. It should be possible to run the following command, however it currently fails with an error that may be a bug.

```
cargo +nightly-2025-08-06 package --workspace --exclude risc0-povw-guests --exclude doc-test --exclude risc0-tools --exclude xtask
```

PR #3370, opened 2025-08-29
groth16: Expose seal_to_json

flaub:


flaub:
@ec2 Does this work for you?

risc0/groth16/src/prove/mod.rs
```diff
@@ -25,6 +25,8 @@ use anyhow::Result;
 
 use crate::Seal;
```
bobbobbio, line 27: lets put a comment on this re-export to ourselves in the future to make sure we don't remove this line.

bobbobbio (approved):

PR #3398, opened 2025-09-02
Minor nits in the soundness calculators

asn-d6:
- Improve consistency along the python calculator
- Fix the command in the rust calculator (perhaps there are better commands, but this is the one I found)

### DIFF
--- a/risc0/zkp/src/prove/soundness.rs
+++ b/risc0/zkp/src/prove/soundness.rs
@@ -29,7 +29,7 @@
 //! Increasing SEGMENT_SIZE to the maximum of 2^24 cycles reduces the result to 95 bits of conjectured soundness.
 //!
 //! Use the following command to run the calculator:
-//! RUST_LOG=risc0_zkp=info cargo run --release
+//! RUST_LOG=risc0_zkp=debug cargo run -p risc0-zkvm --example fib --features=prove --release -- --iterations 1
 //!
 //! Running the calculator results in a terminal printout for scenarios (1) and (3) in the list above.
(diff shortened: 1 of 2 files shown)


risc0/zkp/src/prove/soundness.rs
```diff
@@ -29,7 +29,7 @@
 //! Increasing SEGMENT_SIZE to the maximum of 2^24 cycles reduces the result to 95 bits of conjectured soundness.
 //!
 //! Use the following command to run the calculator:
-//! RUST_LOG=risc0_zkp=info cargo run --release
+//! RUST_LOG=risc0_zkp=debug cargo run -p risc0-zkvm --example fib --features=prove --release -- --iterations 1
```
pdg744, line 32: I *think* all that's necessary here is to change `=info` to `=debug`. @bobbobbio @flaub can one of you confirm that? (I tried myself but running into some version management issues)

asn-d6: Hmm, which directory am I supposed to run this from? Perhaps I'm not running the command from the right directory.

Because if I run it from the root directory, I get the following (regardless of info or debug):
```
❯ RUST_LOG=risc0_zkp=info cargo run --release                                                                                                                                17:45:15
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
```

asn-d6:
Closing this one. The  soundness calculator needs more love. I might open a new PR in the future that also fixes the UDR.

PR #3401, opened 2025-09-02
rzup: Update install help text. It doesn't install all components.

bobbobbio:
The rzup install help text mentions that a bare `install` without any arguments will install all the components, this is no longer correct in the latest version. It installs only the "default" components. This PR updates the text to reflect this.


bobbobbio:
The CI is failing due to needing this fix: https://github.com/risc0/risc0/pull/3402

PR #3406, opened 2025-09-02
Add --kernel-id mode to r0vm

nategraf:
This PR adds a --kernel-id mode to `r0vm`, which may be useful in debugging.

This PR is marked as draft as I am not sure if we want to expose the API in this way, so I wanted to get feedback.

### DIFF
--- a/risc0/binfmt/src/elf.rs
+++ b/risc0/binfmt/src/elf.rs
@@ -383,19 +383,31 @@ impl<'a> ProgramBinary<'a> {
         ret
     }
 
+    fn user_program(&self) -> Result<Program> {
+        Program::load_elf(self.user_elf, KERNEL_START_ADDR.0).context("Loading user ELF")
+    }
+
+    fn kernel_program(&self) -> Result<Program> {
+        Program::load_elf(self.kernel_elf, u32::MAX).context("Loading kernel ELF")
+    }
+
     /// Convert this binary into a `MemoryImage`.
     pub fn to_image(&self) -> Result<MemoryImage> {
-        let user_program =
-            Program::load_elf(self.user_elf, KERNEL_START_ADDR.0).context("Loading user ELF")?;
-        let kernel_program =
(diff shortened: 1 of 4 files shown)


risc0/binfmt/src/lib.rs
```diff
@@ -49,3 +49,8 @@ pub(crate) const PAGE_WORDS: usize = PAGE_BYTES / WORD_SIZE;
 pub fn compute_image_id(blob: &[u8]) -> Result<Digest> {
     ProgramBinary::decode(blob)?.compute_image_id()
 }
+
+/// Compute and return the kernel image ID of the specified combined user ELF + kernel ELF binary.
```
flaub: This is only the kernel_id, not combined with user.

risc0/binfmt/src/elf.rs
```diff
@@ -383,20 +383,32 @@ impl<'a> ProgramBinary<'a> {
...
     }
 
     /// Compute and return the ImageID of this binary.
     pub fn compute_image_id(&self) -> Result<Digest> {
         let merkle_root = self.to_image()?.image_id();
         Ok(SystemState { pc: 0, merkle_root }.digest::<Impl>())
     }
+
+    /// Compute the memory merkle root of the kernel.
+    pub fn kernel_image_id(&self) -> Result<Digest> {
```
flaub:
```suggestion
    pub fn kernel_id(&self) -> Result<Digest> {
```

risc0/binfmt/src/lib.rs
nategraf: Right, I wasn't sure how to explain it. The output is a hash of the kernel. The input is the combined bin file

nategraf: I tried to clear it up with https://github.com/risc0/risc0/pull/3406/commits/b16332e02dba6b34883a3c6292f5d9fe0bab3034

PR #3409, opened 2025-09-03
First pass at config file for actor system

flaub:


Cargo.toml
```diff
@@ -77,7 +77,7 @@ lto = true
 opt-level = 3
 
 [profile.test]
-opt-level = 0
+opt-level = 3
```
bobbobbio, line 80: what's going on here? maybe split this out.

risc0/r0vm/testdata/dev.toml
```diff
@@ -0,0 +1,34 @@
+# r0vm --manager --addr 0.0.0.0:9000 --worker executor
```
bobbobbio, line 1: is this file unused? if its intended for developer usage it being in `testdata` seems confusing

risc0/r0vm/src/actors/mod.rs
```diff
@@ -67,69 +68,18 @@ use self::{
...
 #[tokio::main]
-pub(crate) async fn async_main(args: &Cli) -> Result<(), Box<dyn StdError>> {
-    let is_manager = args.mode.manager;
-
-    let config = if let Some(ref path) = args.mode.config {
+pub(crate) async fn async_main(config_path: Option<PathBuf>) -> Result<(), Box<dyn StdError>> {
+    let config: AppConfig = if let Some(ref path) = config_path {
         let str = tokio::fs::read_to_string(path).await?;
-        Some(toml::from_str(&str)?)
+        toml::from_str(&str)?
```
bobbobbio, line 80: We should probably check the `version` field in the TOML before we fully deserialize it so we fail gracefully when encountering a configuration file from the future, and it would be great to test that.

bobbobbio (approved):
I like the `.toml` file format so far.

risc0/r0vm/testdata/dev.toml
flaub: I recently added, I could drop it. But it would be nice to have a stash of config files someplace. Maybe `testdata` is not a good place for this.

Cargo.toml
flaub: Just trying to speed up local development. I'm just trying to tune build/runtimes.

risc0/r0vm/src/actors/mod.rs
bobbobbio, line 80: something like this maybe? I didn't test it.
```suggestion
        let table: toml::Table = toml::from_str(&str)?;
        let version_entry = table.get("version").ok_or_else(|| anyhow!("missing version"))?;
        let version = version_entry.as_integer().ok_or_else(|| anyhow!("version not integer"))?;
        ensure!(version == 1, "version {version} not supported");
        toml::from_str(&str)?
```

risc0/r0vm/src/actors/mod.rs
```diff
@@ -72,6 +70,11 @@ use self::{
 pub(crate) async fn async_main(config_path: Option<PathBuf>) -> Result<(), Box<dyn StdError>> {
     let config: AppConfig = if let Some(ref path) = config_path {
         let str = tokio::fs::read_to_string(path).await?;
+        let version: VersionConfig = toml::from_str(&str)?;
```
bobbobbio, line 75: nice, using this `VersionConfig` is smart

flaub: I actually have a test that already uses it, I just wasn't quite ready to deal with versioning, but this is fine for now.

PR #3412, opened 2025-09-05
Create revm example

aviggiano:
# REVM Example

This example demonstrates how to execute EVM bytecode inside the RISC Zero zkVM using the [revm](https://github.com/bluealloy/revm) library.

# Motivation

The motivation for this example is a Zero Knowledge Bug Bounty Program, in which a whitehat can prove they know how to construct an attack to extract funds from a contract without revealing the transaction information, minimizing trust issues with protocols and revealing the information only after a payout.

### Roadmap

* **Version 0**
  Proof of concept where REVM runs inside RISC Zero and public bytecode with private calldata yields a public predicate result, committing only code/calldata hashes and the boolean outcome—establishing the verifiable basis for a ZK bug bounty.

* **Version 1**
  Execution anchored to real chain state using a witnessed snapshot at a specific block; initially a trusted witness, then upgraded to in-guest Merkle Patricia Tries verification against the public state root with a fully explicit, deterministic TxEnv.

* **Version 2**
  On-chain verification and payouts via a RISC Zero verifier, a predicate registry, and a vault that escrows bounties; claims are submitted and accepted, and payment releases through a pay-to-reveal hash timelocked contract keyed to the calldata commitment.


flaub:
Perhaps a better place for this to live is over in https://github.com/risc0/demos

PR #3415, opened 2025-09-06
fix: correct spelling errors in scripts/workflows

phrwlk:
- Remove trailing whitespace in prover.sh
- correct 'typo' "D0 NOT MERGE" -> "DO NOT MERGE"


.github/workflows/main.yml
```diff
@@ -166,7 +166,7 @@ jobs:
         with:
           python-version: "3.10"
       - run: python license-check.py
-      - name: check for "D0 NOT MERGE" comments
+      - name: check for "DO NOT MERGE" comments
```
flaub, line 169: This will make it so that we can never make changes again....

PR #3430, opened 2025-09-24
Initial commit for rv32im-m3 circuit

flaub:


tools/gpu-guard/macros/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+    //     #vis fn #name() {
+    //         let _guard = gpu_guard::acquire_gpu_semaphore()
+    //             .expect("Failed to acquire GPU semaphore");
+    //         #block
+    //     }
+    // };
+
+    let injected = quote! {
+        let _guard = gpu_guard::acquire_gpu_semaphore()
+            .expect("Failed to acquire GPU semaphore");
```
bobbobbio:
```suggestion
        let _gpu_guard_function_guard = ::std::result::Result::<_, _>::expect(
            ::gpu_guard::acquire_gpu_semaphore(), "Failed to acquire GPU semaphore");
```
better to use absolute paths in macro output since you don't know whats in scope

also I wondered if you want a more unique name for `_guard` since I think it will be accessible in scope for the function.

Another option is to instead rename the decorated function can declare a function with the original name that calls the original function thus creating a new scope.

tools/gpu-guard/src/semaphore.rs
```diff
@@ -0,0 +1,61 @@
...
+                .read(true)
+                .write(true)
+                .create(true)
+                .truncate(true)
+                .open(&lock_path)?;
+
+            if file.try_lock_exclusive().is_ok() {
+                println!("[gpu-guard] Acquired slot {i} (attempt {attempt})");
+                return Ok(file);
+            }
```
bobbobbio:
```suggestion
            match file.try_lock_exclusive() {
                Ok(_) => {
                    println!("[gpu-guard] Acquired slot {i} (attempt {attempt})");
                    return Ok(file);
                },
                Err(error) => {
                    if error.kind() == fs2::lock_contended_error().kind() {
                        continue;
                    } else {
                        return Err(error);
                    }
                }
            }
```
might be better to not catch all errors here, if there is some other issue with locking the file due to the OS or environment and we just ignore it might be hard to diasnose.

tools/gpu-guard/src/semaphore.rs
```diff
@@ -0,0 +1,61 @@
...
+    path::Path,
+    thread::sleep,
+    time::Duration,
+};
+
+use fs2::FileExt;
+use once_cell::sync::Lazy;
+
+static SEMAPHORE_DIR: Lazy<String> = Lazy::new(|| {
+    std::env::var("GPU_SEMAPHORE_PATH").unwrap_or_else(|_| "/tmp/gpu_semaphore".to_string())
```
bobbobbio, line 27: without including GPU identifying information in this path, the CI jobs that use two different GPUs will conflict.  Perhaps you want to set `GPU_SEMAPHORE_PATH` in the CI job or something like that?

tools/gpu-guard/src/semaphore.rs
```diff
@@ -0,0 +1,61 @@
...
+                .truncate(true)
+                .open(&lock_path)?;
+
+            if file.try_lock_exclusive().is_ok() {
+                println!("[gpu-guard] Acquired slot {i} (attempt {attempt})");
+                return Ok(file);
+            }
+        }
+
+        sleep(Duration::from_millis(50));
```
bobbobbio, line 66: sleeping 50ms helps but it is a bit unfortunate that this is basically a slower busy loop, the waiters will burn some amount of CPU

bobbobbio:
I ran into this error on my PR too
```
thread 'main' panicked at xtask/src/semver_checks.rs:777:13:
failed to retrieve index of crate versions from registry
```

it was due to enabling some feature in `reqwest` that the `cargo-semver-checks` crate didn't like (which ultimately is a bug with the `cargo-semver-checks` crate IMO). But looking at your changes it to the lockfile it seems very unclear to me what might be causing this. So it could be something else entirely like some network issue.

PR #3437, opened 2025-09-26
Force executor to have a minimum segment po2 of 14

bobbobbio:
This ensures whatever segment it produces is able to be lifted later (since 14 is the minimum size of the lift programs)

I made new constants in `risc0-circuit-rv32im` for min/max po2 and used them in places that seemed right to. These are now in contrast to the ones in `risc0-zkp`


risc0/zkvm/src/host/recursion/prove/zkr.rs
```diff
@@ -45,7 +45,7 @@ pub fn test_recursion_circuit(hashfn: &str) -> Result<(Program, Digest)> {
 }
 
 pub fn lift(po2: usize, hashfn: &str) -> Result<(Program, Digest)> {
-    if (MIN_CYCLES_PO2..MAX_CYCLES_PO2).contains(&po2) {
+    if (MIN_CYCLES_PO2..=MAX_CYCLES_PO2).contains(&po2) {
```
bobbobbio, line 49: FYI this seems like a sorta unrelated bug, it would reject 24 for no good reason

flaub: I don't think we have po2=24 lift predicates generated

risc0/zkp/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub mod verify;
 
 pub use risc0_core::field;
 
-pub const MIN_CYCLES_PO2: usize = 13;
+pub const MIN_CYCLES_PO2: usize = 14;
```
flaub: This is the zkp crate, and so this isn't specific to a particular circuit. Other circuits may indeed be smaller, so this value represents the limit of the proof system.

risc0/zkvm/src/host/recursion/prove/zkr.rs
bobbobbio: we do
```
      (
          "lift_rv32im_v2_24.zkr",
          digest!("06987236d7c48128ab8c3505b88269637cbbd540636ccb4b4552d100e7f41742"),
      ),
      ...
            (
          "lift_rv32im_v2_povw_24.zkr",
          digest!("d805245016e3d277cc4040064517f53112a6b719383ed2195395196ac1d46d42"),
      ),
```

risc0/zkp/src/lib.rs
bobbobbio: I've created new constants in `risc0-circuit-rv32im`

PR #3446, opened 2025-09-30
Finish work on Hello-M3 circuit

jacobdweightman:
This PR does a few things:
* Adds code for generating ZKR for Hello-M3
* Adds full test verifying Hello-M3 proof using the ZKR inside the recursion circuit
* Consolidates Rust verifier test to live alongside the other Hello-M3 code

With this, Hello-M3 is complete!


risc0/circuit/hello-m3/src/hello_lift_12.zkr
flaub, line 1: I don't think it's good to commit such a large binary file directly to git. We either need to use LFS or put this into a `.zip` like we do in other places.

risc0/circuit/hello-m3/src/proof.bin
flaub, line 1: Can we not generate this on the fly now?

flaub (approved):

PR #3448, opened 2025-09-30
Add a hint for Docker permissioned denied with Groth16

nategraf:
This is a small edit I made while debugging why `bootstrap-groth16` was not working. One of the issues I ran into was that the process is trying to run `docker` without sufficient permissions. The error currently just gives an exit code of `126`, which does indicate a permissions issue but it took me a while to realize the particular permissions issue. This PR adds a hint.

I also included two other minor edits:
* Remove `bento/Cargo.toml` from linked projects for the VS Code config. This file no longer exists.
* Add a `cuda` feature to bootstrap. I decided to do this instead of using Docker to sidestep the permissions issue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -454,8 +454,8 @@ jobs:
     env:
       RUSTDOCFLAGS: "--cfg docsrs -D warnings"
       # Building with docs.rs config requires the nightly toolchain.
-      # nightly-2025-05-09 corresponds to 1.88
-      NIGHTLY_RUST_TOOLCHAIN: nightly-2025-05-09
+      # nightly-2025-06-20 corresponds to 1.89
+      NIGHTLY_RUST_TOOLCHAIN: nightly-2025-06-20
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/rustup
--- a/risc0/circuit/rv32im-m3-sys/cxx/hal/cuda/hal.cpp
+++ b/risc0/circuit/rv32im-m3-sys/cxx/hal/cuda/hal.cpp
@@ -46,7 +46,7 @@ extern "C" bool cuda_zero_dev(void* buf, size_t size);
 // api.cu
 extern "C" SparkError sppark_poseidon2_fold(void* d_out, const void* d_in, size_t num_hashes);
 extern "C" SparkError sppark_poseidon2_rows(void* d_out, const void* d_in, uint32_t count, uint32_t col_size);
-extern "C" SparkError sppark_prefix_sum(void* d_inout, uint32_t count);
+extern "C" void prefix_sum(Fp* d_inout, uint32_t count);
 extern "C" SparkError supra_poly_divide(void* d_inout, size_t len, void* remainder, FpExt pow);
 
 // query.cu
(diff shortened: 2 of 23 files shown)


xtask/Cargo.toml
```diff
@@ -49,6 +49,7 @@ release = false
 
 [features]
 default = ["zkvm"]
+cuda = ["risc0-zkvm/cuda"]
```
bobbobbio: Adding this feature back seems potentially okay, but I think there is something weird here with the dependencies. Also I'm confused how it related with what you were doing.

Earlier in the file we depend on the `cuda` feature in `risc0-groth16-sys` unconditionally. I would think that we should enable that cuda feature conditionally also. Also I think the `risc0-circuit-recursion` `risc0-zkp` crates should have their cuda feature enabled here too since depend on them directly (it might happen indirectly due to `risc0-zkvm` but this more of a side-effect).

IIRC though, there was issues with making it compile when the cuda feature is not enabled for `risc0-groth16-sys` so maybe we need to make `setup-groth16` instead depend on the cuda feature itself.

risc0/groth16/src/prove/docker.rs
```diff
@@ -60,8 +60,12 @@ pub(crate) fn shrink_wrap(identity_p254_seal_bytes: &[u8]) -> Result<Seal> {
         .stderr(Stdio::piped())
         .output()?;
     if !output.status.success() {
+        let hint = match output.status.code() {
+            Some(126) => "\nThis process may not have permission to run containers.",
+            _ => "",
+        };
```
bobbobbio: interesting, the internet seems to say this is coming from docker copying chroot exit codes https://tldp.org/LDP/abs/html/exitcodes.html

maybe throw this chunk of code into its own function, we could also add some other messages and maybe mention where this information comes from.

bobbobbio (changes requested):
I think the features in `xtask` aren't quite right. Either punt on adding the cuda flag for now or somehow make it more legit.

xtask/Cargo.toml
```diff
@@ -49,6 +49,7 @@ release = false
...
 setup-groth16 = ["dep:risc0-groth16-sys", "dep:memmap2"]
```
bobbobbio: Here is what I'm suggesting

```suggestion
cuda = [
  "risc0-circuit-recursion/cuda",
  "risc0-groth16-sys/cuda",
  "risc0-zkp/cuda",
  "risc0-zkvm/cuda",
]
setup-groth16 = ["dep:risc0-groth16-sys", "dep:memmap2", "cuda"]
```

And also remove the `cuda` feature from `risc0-groth16-sys`

xtask/Cargo.toml
nategraf: Interesting. I added this because without it the prover would try to use Docker and fail. I didn't use the `setup-groth16` flag as I though it was for generating Groth16 verifier key parameters.

bobbobbio: I see. Yes `bootstrap-groth16` does use this I think. Yes `setup-groth16` does not use this and instead uses the `setup-groth16` feature.

bobbobbio (approved):

risc0/groth16/src/prove/docker.rs
nategraf: 126 is the one I know `docker` uses. Running this command in my shell, I get the following more complete stderr output.

```
docker: permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Head "http://%2Fvar%2Frun%2Fdocker.sock/_ping": dial unix /var/run/docker.sock: connect: permission denied

Run 'docker run --help' for more information
```

I thought about putting the full stderr in the error context, but I don't think that actually makes sense.

bobbobbio: > I thought about putting the full stderr in the error context, but I don't think that actually makes sense.

I'm not sure why you say that

PR #3449, opened 2025-10-01
actor system: Improve error handling

bobbobbio:
Every actor now deals with errors in some way rather than panic'ing.

Most non-test code `unwrap`s are removed and the error is handled, or an `expect` with a message is added.

### DIFF
--- a/risc0/r0vm/Cargo.toml
+++ b/risc0/r0vm/Cargo.toml
@@ -35,7 +35,6 @@ opentelemetry-otlp = "0.30.0"
 opentelemetry_sdk = "0.30.0"
 reqwest = { version = "0.12", default-features = false }
 risc0-circuit-rv32im = { workspace = true }
-risc0-zkp = { workspace = true }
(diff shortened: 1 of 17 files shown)


risc0/r0vm/Cargo.toml
```diff
@@ -35,6 +35,7 @@ opentelemetry-otlp = "0.30.0"
 opentelemetry_sdk = "0.30.0"
 reqwest = { version = "0.12", default-features = false }
 risc0-circuit-rv32im = { workspace = true }
+risc0-zkp = { workspace = true }
```
flaub: Why is this needed? Can we forward whatever we need thru `risc0-zkvm` instead?

bobbobbio: at one point I use the type`risc0_zkp::verify::VerificationError`. We could re-export this through `risc0-zkvm` which seems appropriate perhaps.

bobbobbio: I've done just that and re-exported `VerificationError` from `risc0-zkvm`

PR #3451, opened 2025-10-01
Fix a non-determinism bug in CUDA

jbruestle:


jbruestle:
Also, added some tests that run in bazel to verify the ffi.

bobbobbio (approved):

PR #3459, opened 2025-10-04
actor system: Factory deals with workers disconnecting 

bobbobbio:
Have the factory deal with workers disconnecting, and reschedule jobs that were running on disconnected workers.

Also backfill some other factory tests..


risc0/r0vm/src/actors/actor.rs
```diff
@@ -95,11 +95,28 @@ impl<ActorT: Actor> ActorRunner<ActorT> {
...
 
+    /// Receive and potentially reply to a single message without blocking.
+    #[allow(dead_code)]
+    pub async fn try_handle_one(&mut self) -> Result<(), tokio::sync::mpsc::error::TryRecvError> {
+        assert!(self.actor_ref.is_none(), "start not called");
+
+        let handle_msg = self.recv.try_recv()?;
+        handle_msg(&mut self.actor).await;
+        Ok(())
+    }
```
flaub, line 106: Is this still in use?

flaub (approved):

risc0/r0vm/src/actors/actor.rs
bobbobbio: It is used, but currently only by tests. I could put `cfg(test)` but I was worried that might cause other compilation weirdness. Maybe the right thing is something like `cfg_attr(not(test), allow(dead_code))`.

PR #3462, opened 2025-10-06
Get verification working in rust on V3

jbruestle:


risc0/circuit/rv32im-m3/src/prove.rs
```diff
@@ -14,7 +14,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 #[cfg(test)]
-#[cfg(feature = "cuda")]
+#[cfg(feature = "prove")]
```
flaub: This will cause CI to run very slowly.

flaub (approved):

PR #3463, opened 2025-10-06
semver-checks: Show full error information

bobbobbio:


flaub:
I've included this inside the upcoming PR: #3446

PR #3470, opened 2025-10-08
actor system: Improved telemetry and worker parallel jobs

bobbobbio:
This does two different things

- Improve the tracing
    - the worker adds tracepoints for the various stages of running a task parented under the remote job task.
    - have a task job include the time it spends waiting to run
- Execute CPU jobs in the worker in parallel
    - This will let preflights run in parallel.

### DIFF
--- a/risc0/r0vm/src/actors/worker.rs
+++ b/risc0/r0vm/src/actors/worker.rs
@@ -70,6 +70,7 @@ struct GpuTaskMsg {
     to_reserve: Vec<HardwareReservation>,
     reserved: Vec<HardwareReservation>,
     tracing: job::tracer::SavedContext,
+    allocate_tracer: TaskTracer,
 }
 
 enum CpuTask {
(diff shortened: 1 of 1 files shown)


risc0/r0vm/src/actors/worker.rs
```diff
@@ -478,26 +524,49 @@ impl GpuProcessor {
...
-            GpuTask::ShrinkWrap(task) => self.shrink_wrap(msg.header, task).await,
-        };
+        tokio::task::spawn(async move {
+            let res = processor.run_task(header, msg.task, msg.reserved).await;
+            if let Err(error) = res {
+                tracing::error!("GPU task runner failed: {error}");
+            }
 
-        self.send_done(header, result, msg.reserved).await?;
+            drop(tracer);
```
flaub: Curious why we need to drop the tracer explicitly. Is it to force a flush?

bobbobbio: this was just a trick to get it to move into the closure

flaub: I noticed other drops that seemed to be about the tracer too and thought maybe there was some pattern.

bobbobbio: This tracer object (`TaskTracer`) which I implemented in this file ends the span when it gets dropped. I move into the closure so that it ends the span when the task finishes.

I did this on `drop` so that the span would finish even if we get an error, although maybe we don't care about that.

PR #3486, opened 2025-10-15
Replace ExternalProver with actor-based DefaultProver

bobbobbio:
- `default_prover` and `default_executor` return `DefaultProver` by default
- `ExternalProver` is removed
- zkvm client / server interface is removed
- prover example is removed
- dev-mode support is added to actor system / `DefaultProver`
- guest stdout forwarded in `DefaultProver` (no streaming)


risc0/r0vm/src/actors/worker.rs
```diff
@@ -821,32 +849,35 @@ impl CpuProcessor {
...
+                    env.add_assumption(assumption.clone());
+                }
+                if let Some(po2) = task.request.segment_limit_po2 {
+                    env.segment_limit_po2(po2);
+                }
+                let env = env
+                    .stdout(&mut stdout)
+                    .write_slice(&task.request.input)
+                    .coprocessor_callback(coproc)
+                    // .session_limit(limit) // TODO
```
flaub, line 866: Might need to hook this up at some point

flaub (approved):

PR #3496, opened 2025-10-20
atomics support for v1compat

stskeeps:
This adds atomics support (emulated) for v1compat and introduces a zkos/common for future sharing.

This works only with rv32im-m3 circuit.

Known issue: ma_data (not supported)

To get this to work:
- Align r0vm execution with traps
- Remove fence from preflight/executor as circuit doesn't understand it

### DIFF
--- a/benchmarks/methods/build.rs
+++ b/benchmarks/methods/build.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RISC Zero, Inc.
+// Copyright 2026 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
--- a/benchmarks/methods/guest/src/bin/big_blake2b.rs
+++ b/benchmarks/methods/guest/src/bin/big_blake2b.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RISC Zero, Inc.
+// Copyright 2026 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
--- a/benchmarks/methods/guest/src/bin/big_blake3.rs
+++ b/benchmarks/methods/guest/src/bin/big_blake3.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RISC Zero, Inc.
+// Copyright 2026 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
--- a/benchmarks/methods/guest/src/bin/big_keccak.rs
+++ b/benchmarks/methods/guest/src/bin/big_keccak.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RISC Zero, Inc.
+// Copyright 2026 RISC Zero, Inc.
 //
 // Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
--- a/benchmarks/methods/guest/src/bin/big_sha2.rs
+++ b/benchmarks/methods/guest/src/bin/big_sha2.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 RISC Zero, Inc.
+// Copyright 2026 RISC Zero, Inc.
 //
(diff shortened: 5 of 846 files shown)


Cargo.toml
```diff
@@ -1,43 +1,45 @@
...
-  "risc0/zkvm/platform",
-  "risc0/zkvm/receipts",
-  "rzup",
-  "tools/gpu-guard",
-  "tools/gpu-guard/macros",
-  "tools/hotbench",
-  "vendor/sppark",
-  "website/doc-test/main",
-  "xtask",
+    "bonsai/sdk",
```
flaub, line 4: Would prefer to remove all these whitespace changes.

bonsai/sdk/Cargo.toml
```diff
@@ -7,6 +7,14 @@ license = { workspace = true }
 homepage = { workspace = true }
 repository = { workspace = true }
 
+[package.metadata.docs.rs]
```
flaub, line 10: Are you using some editor that automatically formats Cargo.toml files? I'm ok with them being consistently formatted, but I'd like for there to be a CI action to do a check and also a local command line to use to autoformat.

risc0/zkos/common/src/atomics.rs
```diff
@@ -0,0 +1,433 @@
...
+pub fn emulate_atomic_instruction(rd: u32, rs1: u32, rs2: u32, funct3: u32, funct7: u32) -> ! {
+    let funct5 = (funct7 & 0b1111100) >> 2;
+    let _aq = (funct7 & 0b0000010) >> 1;
+    // acquire access
+    let _rl = funct7 & 0b0000001;
+    // release access
+
+    match (funct3, funct5) {
+        (0x2, 0x00) => {
+            // amoadd.w - atomic memory operation: add word
```
flaub: Can we split this into multiple functions? They can be marked inline to avoid any overhead.

stskeeps: Yup, coderabbit actually suggested a macro cos they’re so similar, will get on that

rebuild-and-test.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
flaub: I think the way to handle this is with the bazel rule in the cxx codebase. What is the overall intent with having to make all these patches to the test suite?

risc0/binfmt/src/elf.rs
```diff
@@ -25,7 +25,8 @@ use serde::{Deserialize, Serialize};
 
 use crate::{ByteAddr, Digestible as _, KERNEL_START_ADDR, MemoryImage, SystemState};
 
-const USER_START_ADDR: ByteAddr = ByteAddr(0x0001_0000);
+/// The starting address for user programs in the RISC Zero zkVM.
+pub const USER_START_ADDR: ByteAddr = ByteAddr(0x0001_0000);
```
flaub: I'm a little concerned about making these public. It makes it much harder to relocate this address in the future.

risc0/binfmt/src/elf.rs
```diff
@@ -146,6 +147,11 @@ impl Program {
         self.image.insert(SUSPEND_PC_ADDR.0, self.entry);
         self.image.insert(SUSPEND_MODE_ADDR.0, 1);
     }
+
+    /// Write a word to the image
+    pub fn write_u32(&mut self, address: &u32, value: u32) {
+        self.image.insert(*address, value);
+    }
```
flaub: It's unclear why we need this since the only thing that should be inserting the start word is `prepare_user` or `prepare_kernel` above.

risc0/zkos/common/src/atomics.rs
stskeeps: Changed to macros and multiple functions, should be clearer to read now

risc0/circuit/rv32im-sys/cxx/rv32im/rvtest-guest/riscv_test.h
```diff
@@ -0,0 +1,296 @@
...
+// Pass/Fail Macro
+//-----------------------------------------------------------------------
+
+#define RVTEST_PASS                                                     \
+        fence;                                                          \
+        li TESTNUM, 1;                                                  \
+        li t0, 0;                                                       \
+        li a0, 0;                                                       \
+        la a1, final_digest;                                            \
+        ecall
```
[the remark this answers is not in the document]
stskeeps: @coderabbitai for future reference, we are actually emulating the fence instruction in v1compat which these tests run under, and these tests are for the emulated environment

risc0/zkvm/src/host/server/exec/tests.rs
```diff
@@ -219,7 +219,7 @@ fn bigint_accel() {
 }
 
 #[test_log::test]
-#[should_panic(expected = "IllegalInstruction")]
+#[should_panic(expected = "Illegal trap in machine mode")]
```
flaub, line 222: It seems like the kernel could still panic or cause the host to panic to retain these specific errors.

PR #3512, opened 2025-10-28
Collection of optimizations for profiling work the week from October 22

nategraf:
This PR contains a collection of optimizations from a week of profiling work starting on October 22. Each is relatively minor.

The changes with at least some perf improvement are:

- **use array instead of Vec for Page to remove one pointer indirection**
- **extract split_segment into function from Executor::run hot loop**
- **stop early in MemoryImage::mark_dirty and add extra checks to error paths**

### DIFF
--- a/risc0/circuit/rv32im/src/execute/executor.rs
+++ b/risc0/circuit/rv32im/src/execute/executor.rs
@@ -512,7 +512,6 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
 
     #[inline(always)]
     fn trace_instruction(&mut self, cycle: u64, kind: InsnKind, decoded: &DecodedInstruction) {
-        // TODO(victor/perf): This tracing enabled check is showing up in the trace.
         if unlikely(tracing::enabled!(tracing::Level::TRACE)) {
             tracing::trace!(
                 "[{}:{}:{cycle}] {:?}> {:#010x}  {}",


risc0/binfmt/src/image.rs
```diff
@@ -178,8 +181,11 @@ impl MemoryImage {
         self.pages.keys().copied().collect()
     }
 
-    /// Sorted iterator over page digests (page_idx -> Digest)
+    /// Sorted iterator over page digests (digest_idx -> Digest)
     pub fn digests(&self) -> impl Iterator<Item = (&'_ u32, &'_ Digest)> + '_ {
+        if !self.dirty.is_empty() {
+            panic!("attempted to get digests on a dirty memory image")
+        }
```
bobbobbio: seems like an assert: `assert!(self.dirty.is_empty(), "...")`

risc0/binfmt/src/image.rs
```diff
@@ -225,12 +237,18 @@ impl MemoryImage {
     pub fn get_digest(&mut self, digest_idx: u32) -> Result<&Digest> {
         // Expand if needed
         self.expand_if_zero(digest_idx);
+        if self.dirty.contains(&digest_idx) {
+            bail!("digest marked as dirty: {digest_idx}");
+        }
```
bobbobbio: this seems like it could be an ensure: `ensure!(!self.dirty.contains(&digest_idx), "....")`

risc0/binfmt/src/image.rs
```diff
@@ -345,17 +367,16 @@ impl MemoryImage {
 
 impl Default for Page {
     fn default() -> Self {
-        Self::from_vec(vec![0; PAGE_BYTES])
+        Self::from_arr([0; PAGE_BYTES])
```
bobbobbio, line 372: I think this is actually not great, IIRC this will end up allocating 1k of zeros on the stack and then copy it to the heap, there are ways around this I think maybe using some `unsafe`, but then again if its not showing up on your trace maybe it doesn't matter 🤷

risc0/circuit/rv32im/Cargo.toml
```diff
@@ -104,3 +104,4 @@ std = [
   "serde/std",
 ]
 witgen_debug = []
+exec_debug = []
```
bobbobbio: are features snake_case or kebab-case? I don't have the answer myself

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -326,37 +296,20 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
...
+            )?;
 
             drop(commit_sender);
 
-            segment_callback_thread.join().unwrap()
+            segment_callback_thread
+                .take()
+                .expect("segment_callback_thread is None")
+                .join()
+                .expect("segment_callback_thread panicked")
```
bobbobbio: the message passed into `expect` is suppose to be an expectation, not what went wrong. I like to think about "should" statements like `"segment_callback_thread should be Some(..)"`

https://doc.rust-lang.org/stable/std/error/index.html#common-message-styles

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -16,6 +16,7 @@
 use std::{
     cell::RefCell,
     collections::BTreeSet,
+    fmt::Debug,
```
bobbobbio, line 19: this is needed for something?

risc0/circuit/rv32im/Cargo.toml
nategraf: Idiomatically, they are snake-case. I just matched the other debug feature flag in this crate. I could go ahead and change both though.

risc0/circuit/rv32im/src/execute/executor.rs
nategraf: Indeed, but the signature of the new `split_segment` function
https://github.com/risc0/risc0/blob/d4c7e34fa472ede21423266ec315fc7fdb3a88e1/risc0/circuit/rv32im/src/execute/executor.rs#L343

risc0/binfmt/src/image.rs
nategraf: 50e9a5ec8af600762b6e515baff26ccdd05577e9

risc0/binfmt/src/image.rs
nategraf: 50e9a5ec8af600762b6e515baff26ccdd05577e9

risc0/circuit/rv32im/src/execute/executor.rs
nategraf: Ya, this is one I don't always follow. I have a feeling when I read these messages, espcially when something actually goes wrong, of "Ok, sure I know what _should have_ happened. What actually happened?". I adjusted this though. The second one should simply be an unwrap I think, since we are just propagating the `panic`

50e9a5ec8af600762b6e515baff26ccdd05577e9

risc0/circuit/rv32im/Cargo.toml
flaub: There isn't a standard in rust, and we haven't been super consistent about it. But I've always used `kebab-case`, so maybe we should try and do that everywhere.

risc0/circuit/rv32im/src/execute/mod.rs
```diff
@@ -43,6 +43,29 @@ pub(crate) fn node_idx(page_idx: u32) -> u32 {
...
+    ($($arg:tt)*) => {
+        #[cfg(feature = "exec_debug")]
+        {
+            ::tracing::trace!($($arg)*);
+        }
+    };
+}
+use exec_trace;
+
+macro_rules! exec_debug {
```
flaub: I'd prefer to avoid macros if possible. Can't we use a function instead?

risc0/circuit/rv32im/src/execute/rv32im.rs
```diff
@@ -303,11 +303,6 @@ impl Emulator {
         }
 
         let word = ctx.load_memory(pc.waddr())?;
-        if word & 0x03 != 0x03 {
-            ctx.trap(Exception::IllegalInstruction(word, 0))?;
-            return Ok(());
-        }
-
```
flaub, line 314: I don't think we can do this...

flaub:
Does `exec-debug` actually buy us any perf? I'd like to avoid it since it's painful to do development to turn it on/off all the time.

risc0/circuit/rv32im/src/execute/executor.rs
```diff
@@ -208,6 +215,7 @@ impl<'a, 'b, S: Syscall> Executor<'a, 'b, S> {
             ring: AllocRingBuffer::new(10),
             povw_job_id,
             circuit_version,
+            segment_counter: 0,
```
flaub, line 218: BTW, this is unneeded and undefined in m3.

risc0/binfmt/src/image.rs
nategraf: Ok, I spent possibly too long on this 😅 .

First, I was able to confirm that the page loads when switching to using array elide the bounds checks and pointer indirection, which is nice. This gives the ~100ms improvement on 8s. (Nothing to write home about, but worth the minor change).

I was also able to confirm that the `Default` implementation here produces the sane code you would hope: it allocates, calls `memset` to zero it out, then returns. So its not putting 1kB on the stack 👍

risc0/circuit/rv32im/src/execute/rv32im.rs
nategraf: I looked at all the code paths fairly carefully and could not find any observable difference in results from checking here versus in the `match`. I can back this change out if you think this serves a purpose though.

risc0/circuit/rv32im/src/execute/mod.rs
nategraf: It needs to be a macro if things like this are to be supported
https://github.com/risc0/risc0/blob/11def0b95d43c5e1a2e41dc4c6f7ea3cef4511bd/risc0/circuit/rv32im/src/execute/bigint.rs#L184-L186

risc0/circuit/rv32im/src/execute/executor.rs
nategraf: Good to know. That does make me realize we'll need a new way of assigning nonces for PoVW in M3. I'm going to leave this in this PR, since its still required while v2 is supported and the change on this line is to do with extracting it onto the `Executor` struct to allow for the `split_segment` function to be factored out.

risc0/circuit/rv32im/src/execute/rv32im.rs
nategraf: Reverted in 45de2d02c56dd826152902041e3efd818fd563e9

risc0/circuit/rv32im/src/execute/pager.rs
```diff
@@ -236,15 +236,16 @@ pub(crate) struct WorkingImage {
...
     #[inline(always)]
-    fn get_page(&mut self, page_idx: u32) -> Result<Page> {
-        // If page exists, return it
-        if let Some(page) = self.pages.get(&page_idx) {
-            return Ok(page.clone());
-        }
-        self.pages.insert(page_idx, zero_page().clone());
-
-        Ok(zero_page().clone())
+    fn get_page(&mut self, page_idx: u32) -> Result<&Page> {
```
nategraf, line 244: This change, of returning `&Page` instead of `Page` is less about performance, and instead is to resolve a lifetimes build error when I switched the `peek_page` and `load_page` signatures on `SyscallContext` to return `&[u8; PAGE_BYTES]`. Its also based on the observation that only one call site of this function actually needs an owned `Page`

risc0/circuit/rv32im/src/execute/sha2.rs
```diff
@@ -99,7 +99,6 @@ pub fn ecall(ctx: &mut impl Risc0Context) -> Result<()> {
         ctx.store_u32(sha2.state_out_addr + 7u32 - i, e)?;
     }
 
-    // HERE!
```
flaub, line 102: Could we keep these?

nategraf: Sure, I'll add them back

nategraf: Done in 505312e41de604da6c8713f71c967494e3dfc94e

PR #3514, opened 2025-10-29
Refactor executor APIs to allow for deferred hashing memory images

nategraf:
In the current API, inclusion of `Segment` and `ReceiptClaim` force hashing of memory images when it is not required. In order to allow for deferred hashing of memory images (e.g. to preflight) and for execution entirely without hashing (e.g. in dev mode), this PR refactors the executor APIs.

In addition to moving the construction of `Segment` and `ReceiptClaim` types, it makes the `risc0-zkvm` and `risc0-circuit-rv32im` executor implementations more compatible with iterator-like execution.

This PR includes changes the `MemoryImage` and `Segment` to allow the executor to avoid fully hashing the memory image on each segment split, instead only computing the hashes required to provide the partial memory image to the preflight (the Merkle tree uncle nodes).

Additionally, it moves the code that handles creating a dedicated hashing thread into `risc0_zkvm` and remove hashing entirely from the execution-only code paths.

This is a follow-up to https://github.com/risc0/risc0/pull/3512

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -267,7 +267,7 @@ mod test {
         compare_image_id(
             &guest_list,
             "hello_commit",
-            "c4a025f218c023fa78d1b32431c05405f63214fbef29d1e3bc6a9780fcaad44b",
+            "6ee5c73bdabf2959d26e232d9ca7cb1a8b9e21a4b5cf04c2ca678341c3db4f7c",
         );
     }
 }


risc0/zkvm/src/host/server/exec/executor.rs
```diff
@@ -90,6 +95,88 @@ pub(crate) fn circuit_version() -> u32 {
...
+                let segment_ref =
+                    segment_callback(segment).context("Segment callback returned an error")?;
+
+                self.segment_ref_channel
+                    .send(segment_ref)
+                    .map_err(|err| anyhow!("Failed to send segment ref: {err:?}"))?;
+
+                // Update the digests in the working image. This ensures that all digests are
+                // populated in the produced Segment and the receiver does not need to hash.
+                image.update_digests();
```
nategraf, line 173: Note that due to this line, this PR does not currently turn on any form of lazy hashing by default. With this, every `Segment` will contain a memory image with all digests populated.

risc0/circuit/rv32im/src/execute/executor.rs
nategraf, line 1: One reason the lines in this files change so much is that I extracted the logic to run the sidecar hashing thread into the `risc0_zkvm` crate. This results in a level of unnesting in the `run` function.

nategraf:
Recent proving test failures were indeed do to a change I made here. I realized that a change I made to take the `PagedMemory::loaded pages` from the `page_state` tracker instead of `WorkingImage` ended up creating an issue where the `loaded_pages` function is called before any pages have been loaded into the tracker. As a result, the `PagedMemory::loaded_pages` overestimates the loaded pages, since it includes every page that is provided in the `WorkingImage` (e.g. every non-zero page in an ELF if this were called in the `Executor`) but as it is used, it only overestimates for the case of the bigint2 non-det program or memory which is peeked but never loaded.

I reverted by change in https://github.com/risc0/risc0/pull/3514/commits/ab0326fbc5c0e3bc10c935bdb755e7f508453327 and added some comments about the odd behavior in https://github.com/risc0/risc0/pull/3514/commits/79548ff6bd729accdb47346cb85f2af7a566aa6a

PR #3515, opened 2025-10-29
Adjust ProverServer trait to support rv32im-m3

flaub:


risc0/zkvm/benches/fib.rs
```diff
@@ -51,8 +51,9 @@ fn warmup(_group: &mut BenchGroup) {
         let prover = get_prover_server(&opts).unwrap();
         let session = setup_exec(1).run().unwrap();
         let segment = session.segments[0].resolve().unwrap();
-        let receipt = prover.prove_segment(&ctx, &segment).unwrap();
-        prover.lift(&receipt).unwrap();
+        let receipts = prover.prove_segment(&ctx, &segment).unwrap();
+        let receipt = receipts.first().unwrap();
```
nategraf, line 55: It would be better if there was a way to assert that there is exactly one receipt, rather than taking the first and dropping any others. In the current factoring, this will fail in hard to debug ways if `prove_segment` happens to produce more than one receipt.

One way to do this while keeping the return type of `prove_segment` as `Vec<_>` is

```suggestion
        let [receipt] = receipts.try_into().unwrap();
```

risc0/zkvm/src/host/server/prove/dev_mode.rs
```diff
@@ -411,3 +407,12 @@ where
     let secs = f64::deserialize(deserializer)?;
     Ok(Duration::from_secs_f64(secs))
 }
+
+/// Utility macro to compress repeated checks that dev mode is not disabled.
```
nategraf:
```suggestion
/// Utility function to compress repeated checks that dev mode is not disabled.
```

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -50,6 +50,8 @@ mod private {
     impl Sealed for DevModeProver {}
 }
 
+pub type PreflightIter = Box<dyn Iterator<Item = Result<PreflightResults>>>;
```
nategraf, line 53: Making it a struct instead of a type alias would give more flexibility to change this in the future, and to add traits (e.g. `FusedIterator` or `Drop`) to `PreflightIter`
```suggestion
pub struct PreflightIter(Box<dyn Iterator<Item = Result<PreflightResults>>>);

impl Iterator for PreflightIter {
  type Item = Result<PreflightResults>;

  fn next(&mut self) -> Option<Self::Item> {
    self.0.next()
  }
}
```

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -70,17 +72,26 @@ pub trait ProverServer: private::Sealed {
...
+            let receipt = self.prove_preflight(ctx, chunk?)?;
+            receipts.push(receipt);
+        }
+        Ok(receipts)
     }
 
-    /// Run preflight on the specified [Segment].
-    fn segment_preflight(&self, segment: &Segment) -> Result<PreflightResults>;
+    /// Perform preflight on a [Segment] to produce a sequence of [PreflightResults].
+    fn segment_preflight(&self, segment: &Segment) -> Result<PreflightIter>;
```
nategraf, line 91: Do we want to use the same `Segment` and `Session` types for m3 as v2? If not, this is an opportunity to define them as associated types and define them for each system. This would allow use to do something like

```rust
trait ProverServer {
  type Segment;
  type SegmentIter: Iterator<Item = Result<Self::Segment>>;
  type Preflight;
  type PreflightIter: Iterator<Item = Result<Self::Preflight>>;
  type Receipt;

  fn execute(&self, env: ExecutorEnv, program: &[u8]) -> SegmentIter;

  fn preflight(&self, segment: &Segment) -> PreflightIter;

  fn prove_preflight(&self, preflight: &Preflight) -> Result<Receipt>;
}
```

The idea being that this is a pipeline that can be used on an opaque `ProverServer` without needing to unpack the specific `Segment` or `Preflight` or have them agree between v2 and m3 (or e.g. dev mode).

risc0/zkvm/src/host/server/prove/mod.rs
nategraf: Note that this would not be required if using associated types.

risc0/zkvm/src/host/server/session.rs
```diff
@@ -147,8 +147,10 @@ impl Segment {
 
 /// The results of running preflight on a [Segment].
 pub struct PreflightResults {
+    #[cfg(not(feature = "rv32im-m3"))]
     pub(crate) inner: risc0_circuit_rv32im::prove::PreflightResults,
-
+    #[cfg(feature = "rv32im-m3")]
+    pub(crate) prover: Option<std::sync::Arc<dyn risc0_circuit_rv32im_m3::prove::SegmentProver>>,
```
nategraf: If someone calls `prove` more than once on the prover provided here, what happens?

risc0/circuit/rv32im-m3/src/prove.rs
```diff
@@ -24,48 +27,24 @@ use crate::verify::verify;
...
 pub type Seal = Vec<u32>;
 
-pub trait SegmentProver {
-    fn prove(&self, segment: &Segment) -> Result<Seal>;
+pub trait SegmentProver: Send + Sync {
+    fn load_segment(&self, segment: &Segment) -> Result<()>;
+
+    fn preflight(&self) -> Result<bool>;
+
+    fn prove(&self) -> Result<Seal>;
```
nategraf: As factored, this `trait` only makes sense when using interior mutability. Should the `&mut self` references be `&mut self` to communicate the fact that these operation mutate the internal state?

risc0/circuit/rv32im-m3/src/prove.rs
```diff
@@ -24,48 +27,24 @@ use crate::verify::verify;
...
-    }
-
-    #[cfg(feature = "cuda")]
-    fn new_cuda(po2: usize) -> Self {
-        Self {
-            ctx: unsafe { risc0_circuit_rv32im_m3_prover_new_cuda(po2) },
-        }
-    }
+unsafe impl Send for SegmentProverImpl {}
+unsafe impl Sync for SegmentProverImpl {}
```
bobbobbio: is this correct? It would mean I think that either nothing is mutated in the `ProverContext` (or anything it has pointers to) or else it has internal locking.

nategraf: Does `ProverContext` have its own synchronization for its interior mutability? If not, it is not really safe to implement `Send` or `Sync`. If the `trait` above is updated to require `&mut self`, this this could potentially be `Send` and therefore `Mutex<SegmentProverImpl>` would be `Sync`.

risc0/zkvm/src/host/server/prove/mod.rs
flaub: Default types are not allowed for associated types (at least it's not yet stable).

risc0/zkvm/src/host/server/prove/mod.rs
flaub: I don't see a need to do this, it seems to just overly complexify everything.

flaub: Being able to support v2 and m3 is a temporary situation.

risc0/circuit/rv32im-m3/src/prove.rs
flaub: Yes, it is intended that the C++ will become thread-safe later on.

risc0/circuit/rv32im-m3/src/prove.rs
flaub: It calls down into C++ code, so there's no real mutability going on from Rust's perspective.

risc0/zkvm/benches/fib.rs
flaub: This isn't the final factoring of this. This is just a stepping stone. All of these issues will be addressed in the future.

risc0/circuit/rv32im-m3/src/prove.rs
flaub: Yes, there will be internal locking.

risc0/zkvm/src/host/server/session.rs
flaub: This whole thing will be refactored later

risc0/zkvm/src/host/server/prove/mod.rs
nategraf: What is the connection of defaults for associated types and this suggestion?

risc0/circuit/rv32im-m3/src/prove.rs
```diff
@@ -114,47 +94,91 @@ impl SegmentProverImpl {
...
-impl Drop for SegmentProverImpl {
+impl Drop for ProverContext {
     fn drop(&mut self) {
-        unsafe { risc0_circuit_rv32im_m3_prover_free(self.ctx) };
+        unsafe { risc0_circuit_rv32im_m3_prover_free(self.ctx.as_ptr()) };
     }
 }
 
-pub fn segment_prover(po2: usize) -> Result<Box<dyn SegmentProver>> {
+pub fn segment_prover(po2: usize) -> Result<Rc<ProverContext>> {
```
bobbobbio: Should we really turn a `Rc` here? It seems like the caller could easily create their own `Rc` if they need it

risc0/zkvm/src/host/server/prove/prover_impl.rs
```diff
@@ -437,3 +463,61 @@ fn check_claims(
...
+                Ok(preflight) => {
+                    self.is_done = preflight.is_final();
+                    let results = PreflightResults {
+                        inner: preflight,
+                        terminate_state: self.terminate_state,
+                        output: if self.is_done {
+                            self.output.clone()
+                        } else {
+                            None
+                        },
```
bobbobbio: this can be `self.is_done.then(|| self.output.clone())`

bobbobbio (approved):
I find the thread-local error a bit regrettable but I understand why you did it, hopefully we can improve that later.

I know you're not worrying about this yet, but there are going to have to be further changes to the actor system to make this work. Right now it will try to finish the task after proving a single preflight_results, and any further segments will cause it to send a duplicate task completion response (both to the allocator and the job)

risc0/circuit/rv32im-m3/src/prove.rs
flaub: I think this was leftover from `dyn SegmentProver`, now with a concrete sized type, we don't need to box it so I agree.

risc0/zkvm/src/host/server/prove/prover_impl.rs
flaub: Yes, but we need a `flatten()` at the end.

Issue #2461, opened 2024-10-30
Multi-thread testing results in error


SkymanOne:
### Is there an existing issue?

- [X] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [X] This is not a support question.

### Description of bug

When running multiple tests (in production mode) involving building a prover with `default_prover()` for each test and executing the same guest program on the same or different data inputs, a local prover can either panic with:
```
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
  left: [1109155563, 76776968, 726152720, 1606742656]
 right: [0, 0, 0, 0]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
  left: [1937466798, 422672686, 84982592, 642582437]
 right: [0, 0, 0, 0]
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
  left: [916425584, 1323635903, 560063291, 61845575]
 right: [0, 0, 0, 0]
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
  left: [283844281, 1796260247, 1438206498, 1271669024]
 right: [0, 0, 0, 0]
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
  left: [1889568656, 1020291148, 1642680298, 421080635]
 right: [0, 0, 0, 0]
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
  left: [414445105, 1970834655, 939333111, 991974128]
 right: [0, 0, 0, 0]
thread '<unnamed>' panicked at /Users/risc0/actions-runner/_work/risc0/risc0/risc0/zkp/src/prove/prover.rs:338:33:
assertion `left == right` failed
[...]

PR #3522, opened 2025-11-03
Add `cudaDeviceSynchronize` to the start of kernels missing it

bobbobbio:
There is a race between copying data to the GPU using `cust` and a GPU kernel reading the data.

It exists in a demonstrable way with the `poly_divide` kernel, and I've added a test that demonstrates it in this branch with basically 100% hit rate. The failure this causes can be seen in https://github.com/risc0/risc0/issues/2461 as well as my own testing and automation runs with `compute-sanitizer`

Other kernels avoid this problem right now by adding `cudaDeviceSynchronize` to the start of them which seems to wait for the memory copy to fully complete. I've fixed the issue with `poly_divide` by doing the same.

I've also added `cudaDeviceSynchronize` to other kernels which I believe should have theoretically the same issue, although they may or may not have it in practice due to how they are called.


bobbobbio:
rather than do this, I'd like to better fix the host / device synchronization.

It might still be good to come back and save the poly_divide test though

PR #3528, opened 2025-11-04
Fix risc0_groth16::prove::cuda::CalcWitness::as_bytes slice length

bobbobbio:
The element size is 32 bytes not 4 bytes. Just use mem::size_of to hopefully get the right answer.

This length I think is unused, so the impact of this is pretty low.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -114,7 +114,7 @@ jobs:
       - check-semver
       - check-template
       - clippy
-      - compute-sanitizer
+      # - compute-sanitizer
       - doc
       - docker
       - docs-rs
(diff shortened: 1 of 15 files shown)


risc0/groth16/src/prove/cuda.rs
```diff
@@ -193,7 +193,7 @@ struct CalcWitness {
 impl CalcWitness {
     fn as_bytes(&self) -> &[u8] {
         let r: &[wtns_file::FieldElement<32>] = &self.witness[..];
-        unsafe { std::slice::from_raw_parts(r.as_ptr() as *const u8, r.len() * 4) }
+        unsafe { std::slice::from_raw_parts(r.as_ptr() as *const u8, std::mem::size_of_val(r)) }
```
nategraf, line 196: `size_of_val` appears to only return the size of a single element
```suggestion
        unsafe { std::slice::from_raw_parts(r.as_ptr() as *const u8, std::mem::size_of_val(r) * r.len()) }
```

https://doc.rust-lang.org/std/mem/fn.size_of_val.html

bobbobbio: I think you are mistaken, look at what it says about `[T]`.

also here is a playground for proof: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a90fae17ac7ed0dfbbb66b2b7b4c3197

nategraf: Indeed I am! Ok, thank you, I misread the docs

nategraf (approved):

PR #3529, opened 2025-11-04
actor-system: Run multiple GPU jobs at once and tune the token amounts

bobbobbio:
This does the following:
- Schedule work in the allocator based on token amount instead of task count
    - change `worker_task_limit` to `worker_queing_factor`
- Show task token counts and gpu token limits in status API
- Run GPU jobs in parallel on the worker

Now that most of the concurrency issues have been resolved I am able to complete proofs on the cluster and see improved times.


flaub (approved):

bobbobbio:
This is failing the examples I think because I only tuned the token amounts for po2=21 and the default prover uses some other po2, and po2=21 doesn't fit on the CI cards probably

PR #3532, opened 2025-11-05
Introduce Blake3 Groth16 compression

ec2:
Introduces a new Groth16 compression using Blake3. 

There is no rzup component yet, so to run this, you need to export the env var BLAKE3_GROTH16_SETUP_DIR which is where you ran `cargo xtask-blake3-groth16`. 

TODO:
[] Create rzup component


risc0/zkvm/src/host/rpc.rs
```diff
@@ -50,6 +50,10 @@ pub struct ProofRequest {
 pub enum ShrinkWrapKind {
     /// TODO
     Groth16,
+
+    #[cfg(feature = "blake3")]
```
bobbobbio, line 54: since this is part of the on-the-wire encoding, I wouldn't conditionally compile this in, we should add it unconditionally and if we request it we can return an error

this is better than getting a can't decode request error

risc0/zkvm/src/host/server/prove/mod.rs
```diff
@@ -205,6 +207,37 @@ pub trait ProverServer: private::Sealed {
         })
     }
 
+    #[cfg(feature = "blake3")]
+    /// Compress a [SuccinctReceipt] into a [Groth16Receipt].
```
bobbobbio, line 211: maybe update this doc-comment to differentiate it from the other groth16 function

risc0/zkvm/src/host/client/prove/opts.rs
```diff
@@ -75,6 +75,15 @@ pub enum ReceiptKind {
...
     Groth16,
+
+    #[cfg(feature = "blake3")]
+    /// Request that a Blake3 [Groth16Receipt][crate::Groth16Receipt] be generated.
+    ///
+    /// Groth16 receipts are proven using Groth16, are constant in size, and are the smallest
+    /// available receipt format. A Groth16 receipt can be serialized to a few hundred bytes.
+    /// This variant uses the BLAKE3 hash function and is only compatible with guests that
+    /// have journal size of 32 bytes.
+    Blake3Groth16,
```
bobbobbio, line 86: this too because it is serialized / deserialized I think it better to always compile it in and just error if encountered.

xtask/src/setup_blake3_groth16.rs
```diff
@@ -0,0 +1,197 @@
...
+            };
+            fuzzed_results_file
+                .set_len(fuzzed_results_size as u64)
+                .unwrap();
+            preprocessed_coeffs_file
+                .set_len(preprocessed_coeffs_size as u64)
+                .unwrap();
+        }
+    }
+}
```
bobbobbio, line 197: a lot of this chunk of code seems similar to the other groth16 setup, it would be nice to not have to maintain two different versions of this code, I think we can collapse it somehow.

xtask/src/bootstrap_groth16.rs
```diff
@@ -60,7 +60,7 @@ const SOLIDITY_VERIFIER_SOURCE: &str = "groth16_proof/groth16/verifier.sol";
 const SOLIDITY_VERIFIER_TARGET: &str = "contracts/src/groth16/Groth16Verifier.sol";
 const SOLIDITY_CONTROL_ID_PATH: &str = "contracts/src/groth16/ControlID.sol";
 const SOLIDITY_TEST_RECEIPT_PATH: &str = "contracts/test/TestReceipt.sol";
-const RUST_VERIFIER_PATH: &str = "risc0/groth16/src/verifier.rs";
+const RUST_VERIFIER_PATH: &str = "risc0/groth16/src/verifier/risc0.rs";
```
bobbobbio, line 63: whats going on here?

xtask/src/bootstrap_blake3_groth16.rs
```diff
@@ -0,0 +1,219 @@
...
+                "const {}: &str =[\\r\\n\\s]*\"\\d+\";",
+                rust_constants[i]
+            ))
+            .unwrap();
+            rust_code = rust_re
+                .replace(
+                    &rust_code,
+                    &format!("const {}: &str = \"{}\";", rust_constants[i], &caps[1]),
+                )
+                .to_string();
```
bobbobbio, line 113: oof, I guess this works, if it was me I would use `syn` which contains a Rust parser.

PR #3538, opened 2025-11-11
rv32im-m3: Fix for out of bounds access in bigint

flaub:


jbruestle (approved):
A significant improvement in clarity as well as a bug fix.

PR #3545, opened 2025-11-13
Avoid unbounded host buffer allocation during execution

nategraf:
In various ecall and syscalls, the guest passes a pointer and length to a region in memory that should be read by the host. Currently on `main`, many of these allocate a `Vec<u8>` to copy the region pointed to by the guest. This opens up an issue for provers that need to execute arbitrary guest programs (e.g. on Boundless) in that the guest can supply a very large (up to 4GB) size to read, and the host will allocate a large buffer. This can be used to cause an OOM condition in the executor. 

When running unknown guest programs, limits should be placed on the amount of memory the process can use (e.g. via `cgroups`) and in practice this is done. This mitigates the most significant possible effects, but this kind of issue can still cause friction.

This PR addresses this issue by providing the ability to read from guest memory without copying, through the `std::io::Read` trait. This is then used in all relevant callsites to avoid allocating a large buffer unless required (e.g. the journal needs to be read into a buffer that will be returned with the session). At those callsites, limits are added on how much can be read.

This PR is built on top of https://github.com/risc0/risc0/pull/3512

### DIFF
--- a/risc0/sys/build.rs
+++ b/risc0/sys/build.rs
@@ -50,11 +50,12 @@ fn build_cuda_kernels(cxx_root: &Path) {
             "kernels/zkp/cuda/combos.cu",
             "kernels/zkp/cuda/eltwise.cu",
             "kernels/zkp/cuda/ffi.cu",
+            "kernels/zkp/cuda/fri_prove.cu",
             "kernels/zkp/cuda/kernels.cu",
             "kernels/zkp/cuda/sha.cu",
             "kernels/zkp/cuda/supra/api.cu",
-            "kernels/zkp/cuda/supra/poseidon2.cu",
             "kernels/zkp/cuda/supra/ntt.cu",
+            "kernels/zkp/cuda/supra/poseidon2.cu",
         ])
         .deps(["kernels/zkp/cuda", "kernels/zkp/cuda/supra"])
         .flag("-DFEATURE_BABY_BEAR")
--- a/risc0/sys/kernels/zkp/cuda/ffi.cu
+++ b/risc0/sys/kernels/zkp/cuda/ffi.cu
@@ -152,4 +152,31 @@ const char* risc0_zkp_cuda_combos_prepare(cudaStream_t stream,
   return nullptr;
 }
 
+const char* risc0_zkp_cuda_fri_prove(cudaStream_t stream,
+                                     uint32_t* out_values,
+                                     const size_t values_column_width,
+                                     uint32_t* out_digests,
+                                     const size_t digests_column_width,
+                                     uint32_t* positions,
+                                     const size_t positions_len,
+                                     CudaMerkleTreeProver* trees,
+                                     const size_t trees_len,
+                                     uint32_t* groups) {
+  try {
+    LaunchConfig cfg1 = getSimpleConfig(trees_len * positions_len * values_column_width);
+    fri_prove_values<<<cfg1.grid, cfg1.block, 0, stream>>>(
+        out_values, values_column_width, positions, positions_len, trees, trees_len, groups);
+
+    LaunchConfig cfg2 = getSimpleConfig(trees_len * positions_len * digests_column_width);
+    fri_prove_digests<<<cfg2.grid, cfg2.block, 0, stream>>>(
+        out_digests, digests_column_width, positions, positions_len, trees, trees_len, groups);
+
+    CUDA_OK(cudaStreamSynchronize(stream));
+  } catch (const std::exception& err) {
+    CUDA_OK(cudaStreamSynchronize(stream));
+    return strdup(err.what());
+  }
+  return nullptr;
+}
+
 } // extern "C"
--- a/risc0/sys/kernels/zkp/cuda/fri_prove.cu
+++ b/risc0/sys/kernels/zkp/cuda/fri_prove.cu
@@ -1,0 +1,88 @@
+// Copyright 2025 RISC Zero, Inc.
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+#include <assert.h>
+
+#include "kernels.h"
+#include "fp.h"
+
+
+__global__ void fri_prove_values(uint32_t* out_values,
+                                 const size_t values_column_width,
+                                 uint32_t* positions,
+                                 const size_t positions_len,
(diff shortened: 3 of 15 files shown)


risc0/circuit/rv32im/src/execute/pager.rs
```diff
@@ -360,16 +461,25 @@ impl PagedMemory {
         }
     }
 
-    pub(crate) fn peek_page(&mut self, page_idx: u32) -> Result<&[u8; PAGE_BYTES]> {
+    pub(crate) fn peek_page(&mut self, page_idx: u32) -> Result<&Page> {
+        // Registers are not stored in the main RAM during execution. If the page containing the
+        // memory mapped registers is requested, then they must first be written to RAM.
+        if page_idx == USER_REGS_ADDR.page_idx() || page_idx == MACHINE_REGS_ADDR.page_idx() {
+            self.write_registers()
+        }
```
flaub: This doesn't seem right to me

nategraf: Can you elaborate? If I understand the code correctly, it is true that the registers are kept in a separate buffer on the `PagedMemory` struct and the actually `Page` is not updated on each change. If the page is requested, the values at the register memory map location will not be accurate to the current machine state. And so if that page ends up being used, then it may lead to bugs.

As far as I can tell, registers and this page more generally are always accessed through the load paths that only fetch a single word, and so this writing is not needed, as the caller is not requesting a single buffer with both the registers and the rest of RAM.

So, unless I am missing something, this check is required to faithfully implemented the semantics of loading this page.

flaub: Users should not be doing peek page on the system page.

risc0/zkvm/src/host/server/exec/syscall/panic.rs
```diff
@@ -13,12 +13,18 @@
...
+
 use anyhow::{Result, bail};
 use risc0_binfmt::ByteAddr;
 use risc0_zkvm_platform::syscall::reg_abi::{REG_A3, REG_A4};
 
 use super::{Syscall, SyscallContext};
 
+// Maximum size, in bytes, for the panic message provided by the guest. Messages above this size
+// are truncated.
+const MAX_PANIC_MSG_SIZE: u64 = 1 << 20; // 1 MB
```
flaub: I think 1KB is more than enough for this.

risc0/circuit/rv32im/src/execute/pager.rs
nategraf: I did figure out this does interact poorly with the cycle counting, and breaks segment splitting for small segments. The issue is that the registers are currently only written, and the paging cycles only counted, _after_ the executor has decided to split the segment. So in the case that no other words in the system page are written, the paging cycles essentially eat into the part that's reserved at the end of the segment. Calling `write_registers` outside of `commit` moves this accounting to before the split, and throws off the counting logic. It's not clear to me this is all intentional, and I'm curious if this could give rise to a completeness issue.

As for this PR, I am going to remove the check and add a note. The check is not needed since, as noted, the registers are only ever loaded through the other code paths.

EDIT:

This turned out not to be the issue. Although this seemed likely, there is at least one write in `Risc0Machine::resume` to the system page, and this prevents the issue I was thinking was happening.

The actual issue was an interaction in the new code that allowed the page state to go from `Dirty` back to `Loaded`, which is fixed in 05da30e72fe10d879cc9d8282140061d6198a427.

nategraf: With the `Region` code, there is at least one place where this happens, which is `on_terminate`
https://github.com/risc0/risc0/blob/05da30e72fe10d879cc9d8282140061d6198a427/risc0/circuit/rv32im/src/execute/executor.rs#L671-L674

Do you want me to refactor this to avoid this ever happening? Or do you think we should land this PR?

risc0/zkvm/src/host/server/exec/syscall/panic.rs
nategraf: As discussed, this might also include a stack trace

risc0/circuit/rv32im/src/execute/bigint.rs
```diff
@@ -162,6 +170,13 @@ pub fn ecall_execute(ctx: &mut impl Risc0Context) -> Result<usize> {
     Ok(cycles)
 }
 
+// A thin function to drain a reader. Optimizes well with inlining.
+fn drain(mut reader: impl Read) -> std::io::Result<()> {
```
bobbobbio, line 174: I guess it optimizes better than `std::io::copy(reader, std::io::sink())`?

risc0/circuit/rv32im/src/execute/r0vm.rs
```diff
@@ -90,19 +90,60 @@ pub(crate) trait Risc0Context {
...
+        impl<C: Risc0Context + ?Sized> Read for Reader<'_, C> {
+            #[inline(always)]
+            fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+                let end = ByteAddr::min(self.pos.saturating_add(buf.len() as u32), self.end);
+                let size = (end.0 - self.pos.0) as usize;
+                if self.pos.is_aligned()
+                    && let Ok(wbuf) = bytemuck::try_cast_slice_mut(&mut buf[..size])
+                {
+                    let waddr = self.pos.waddr();
+                    for (pos, word) in wbuf.iter_mut().enumerate() {
```
bobbobbio, line 117: maybe `(waddr..).zip(wbuf.iter_mut())`?

risc0/circuit/rv32im/src/execute/r0vm.rs
```diff
@@ -90,19 +90,60 @@ pub(crate) trait Risc0Context {
...
+                        *word = self
+                            .ctx
+                            .load_u32(self.op, waddr + pos)
+                            .map_err(std::io::Error::other)?
+                            .to_le();
+                    }
+                } else {
+                    for (pos, byte) in buf[..size].iter_mut().enumerate() {
```
bobbobbio, line 125: similar

risc0/circuit/rv32im/src/execute/bigint.rs
nategraf: That's right. `std::io::copy(reader, std::io::sink())` ends up having some non-inlined calls internally that prevent the compiler from realizing that it can just drop the copy to the buffer.

risc0/circuit/rv32im/src/execute/r0vm.rs
nategraf: That would be slightly nicer. It doesn't compile as is because `WordAddr` does not implement the required step to allow it to form ranges like that.

risc0/zkvm/src/host/server/exec/syscall/verify2.rs
```diff
@@ -37,9 +37,10 @@ impl Syscall for SysVerify2 {
         if !to_guest.is_empty() {
             bail!("invalid sys_verify2 call");
         }
+
         let from_guest_ptr = ByteAddr(ctx.load_register(REG_A3));
-        let from_guest_len = ctx.load_register(REG_A4);
-        let from_guest: Vec<u8> = ctx.load_region(from_guest_ptr, from_guest_len)?;
+        let from_guest_len = DIGEST_BYTES as u32 * 2;
+        let from_guest = ctx.load_region(from_guest_ptr, from_guest_len)?;
```
flaub, line 43: 👍

flaub:
Thanks, lgtm

flaub (approved):

PR #3552, opened 2025-11-20
m3 executor splitting

bobbobbio:
This PR implements a new segment splitting logic for m3.

It counts the number of "row points" that is portions of a row that each block takes and uses this to keep track of when to split.

This creates a one-to-one relationship between the executor segments and the m3 preflight segments again. To make this work I had to reverse some multi-segment code in m3. Some multi-segment code here remains still like the preflight iterator, we can remove this in a future PR.

The one-to-one relationship fixes the actor-system to work with m3 now.

po2 > 20 doesn't seem to work for m3 yet for unknown reasons.


risc0/circuit/rv32im/src/execute/block_tracker.rs
```diff
@@ -0,0 +1,237 @@
...
+        blocks.add_blocks(BlockType::MakeTable, (256 + 65536) / 16);
+
+        Self {
+            pc_cache: Default::default(),
+            blocks,
+        }
+    }
+}
+
+impl BlockTracker {
```
flaub, line 152: I like that this is a distinct thing that can be used standalone.

risc0/circuit/rv32im/src/execute/block_tracker.rs
```diff
@@ -0,0 +1,237 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use risc0_circuit_rv32im_m3_sys::BlockType;
+
+use super::pager::NUM_PAGES;
+use super::platform::{PAGE_BYTES, WORD_SIZE};
+use super::rv32im::InsnKind;
+
+pub const POINTS_PER_ROW: u64 = 5040;
```
flaub, line 23: A comment on where this number comes from would be useful.

PR #3554, opened 2025-11-21
Use risc0_zkp impl of Poseidon2 mix in executor

nategraf:
Poseidon2 hashing, most commonly used by the keccack acceleration, is a significant part of execution time. This PR optimizes Poseidon2 hashing during execution by using the `risc0_zkp` implementation of the mix function, rather than the version that tracks all intermediate states for preflight.


risc0/circuit/rv32im/src/prove/witgen/poseidon2.rs
```diff
@@ -191,13 +191,13 @@ impl Poseidon2 {
     pub fn read_node(ctx: &mut impl Risc0Context, node_idx: u32) -> Result<()> {
         // tracing::trace!("read_node: {node_idx:#010x}");
         let mut p2 = Poseidon2State::new_node(node_idx, true);
-        p2.rest(ctx, CycleState::PoseidonPaging)
+        p2.run(ctx, CycleState::PoseidonPaging)
```
flaub: The name `rest` comes from the preflight implementation. There's an advantage to keeping the names similar so it's easier to ensure they remain consistent.

risc0/circuit/rv32im/src/execute/poseidon2.rs
```diff
@@ -79,89 +79,141 @@ impl Poseidon2State {
...
         &mut self,
-        ctx: &mut impl Risc0Context,
+        ctx: &mut C,
         cur_state: &mut CycleState,
         next_state: CycleState,
         sub_state: u32,
-    ) {
+    ) where
+        C: Risc0Context + ?Sized,
+    {
```
flaub: I'm wondering why this is needed?

flaub (approved):

risc0/circuit/rv32im/src/prove/witgen/poseidon2.rs
nategraf: ca572c0a45aa513db23159afdb1b3d2047867d60

risc0/circuit/rv32im/src/execute/poseidon2.rs
nategraf: This must have been an artifact of different revisions. I've changed it back to an `impl`
bdc75e2802cf3f7a3043c1b84e69ae84eb03f85e

PR #3561, opened 2025-11-24
Dropping Bonsai from docs

pdg744:
As we move forward with shutting down Bonsai, this PR removes Bonsai from the docs in most locations. In most cases, we simply update the link to point to [Boundless](https://boundless.network).

### Future Work 
There are a few remaining Bonsai references, which can be addressed later: 
- A reference to "Bonsai Pay"
- A reference to "Bonsai Governance"
- A reference to the "litepaper" 

I think we should remove the litepaper, because it's showing up prominently in the search tool in the dev docs. This may cause some problems with broken links in old versions, because the litepaper is stored in `pages/` which lives outside the api versioning system. I figure we'll just disable those links, but there's already a decently sized diff on this PR so will save that for a future PR. Not immediately sure what we want to do about the other two items in this list, so I'm just leaving them for now.


pdg744:
I figured I'll port (i.e., duplicate) the changes to the current api_versioned_docs/ after the initial review.

PR #3564, opened 2025-11-25
Fix operator precedence issue in check_bigint_addr

nategraf:
In the executor, there is a bug in the implementation of the check for whether a bigint2 memory op is withing the allowed address range. This PR fixes this issue.

The issue was reported to us by ChainSentry on Hackenproof

### DIFF
--- a/risc0/circuit/rv32im/src/execute/bigint.rs
+++ b/risc0/circuit/rv32im/src/execute/bigint.rs
@@ -317,8 +317,7 @@ mod tests {
 
         test_context.store_region(base_addr, &input)?;
 
-        let mode = if machine_mode { 1 } else { 0 };
-        let mut io = BigIntIOImpl::new(&mut test_context, mode);
+        let mut io = BigIntIOImpl::new(&mut test_context, machine_mode as u32);
         let natural_out = io.load(ARENA_REG as u32, 0, count).unwrap();
         assert_eq!(natural_out.to_limbs_asc(), expected);
         Ok(())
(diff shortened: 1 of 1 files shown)


risc0/circuit/rv32im/src/execute/bigint.rs
```diff
@@ -273,22 +284,65 @@ mod tests {
...
+    fn bigint_write_oob(base_addr: ByteAddr, count: u32, machine_mode: bool) -> Result<()> {
         let mut test_context = TestRisc0Context::default();
-        let arena = REG_A0;
-        let base_addr = ByteAddr(0x1234);
-        test_context
-            .store_register(MACHINE_REGS_ADDR.waddr(), arena, base_addr.0)
-            .unwrap();
+        test_context.store_register(MACHINE_REGS_ADDR.waddr(), ARENA_REG, base_addr.0)?;
+
+        let mode = if machine_mode { 1 } else { 0 };
```
bobbobbio, line 295: you can use `as` to convert a bool to an int like this

risc0/circuit/rv32im/src/execute/bigint.rs
```diff
@@ -273,22 +284,65 @@ mod tests {
...
+        expected: Vec<u32>,
+    ) -> Result<()> {
+        let mut test_context = TestRisc0Context::default();
+        test_context.store_register(MACHINE_REGS_ADDR.waddr(), ARENA_REG, base_addr.0)?;
 
         test_context.store_region(base_addr, &input)?;
 
-        let mut io = BigIntIOImpl::new(&mut test_context, /*mod=*/ 0);
-        let natural_out = io.load(arena as u32, 0, count).unwrap();
+        let mode = if machine_mode { 1 } else { 0 };
```
bobbobbio: same

bobbobbio (approved):

risc0/circuit/rv32im/src/execute/bigint.rs
nategraf: 8cf655c6e81239733a1501b56c1eb68224dd64a0

risc0/circuit/rv32im/src/execute/bigint.rs
nategraf: 8cf655c6e81239733a1501b56c1eb68224dd64a0

PR #3570, opened 2025-12-04
perf(binfmt): avoid alloc in PovwJobId::to_bytes

ANtutov:
Replaced concat+try_into with direct writes into a fixed-size array in PovwJobId::to_bytes. This removes an unnecessary heap allocation and copy for a statically sized 28-byte output while preserving layout and endianness. The API and semantics are unchanged, and existing round-trip tests continue to validate correctness.

### DIFF
--- a/risc0/binfmt/src/povw.rs
+++ b/risc0/binfmt/src/povw.rs
@@ -76,8 +76,7 @@ impl PovwJobId {
     pub fn to_bytes(self) -> [u8; U160::BYTES + U64::BYTES] {
         let mut out = [0u8; U160::BYTES + U64::BYTES];
         out[..U64::BYTES].copy_from_slice(&self.job.to_le_bytes());
-        out[U64::BYTES..]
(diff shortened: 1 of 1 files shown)


risc0/binfmt/src/povw.rs
```diff
@@ -74,13 +74,11 @@ impl PovwJobId {
...
-        [
-            self.job.to_le_bytes().as_slice(),
-            self.log.to_le_bytes::<{ U160::BYTES }>().as_slice(),
-        ]
-        .concat()
-        .try_into()
-        .unwrap()
+        let mut out = [0u8; U160::BYTES + U64::BYTES];
+        out[..U64::BYTES].copy_from_slice(&self.job.to_le_bytes());
+        out[U64::BYTES..U64::BYTES + U160::BYTES]
```
nategraf:
```suggestion
        out[U64::BYTES..]
```

nategraf (approved):

PR #3582, opened 2025-12-09
Fix issues with m3 poseidon2 ecall and exit code handling

nategraf:
In https://github.com/risc0/risc0/pull/3554, I accidentally removed some logic for M3 handling in a bad merge.

This PR fixes that issue. It also fixes a couple of issues in claim decoding that I noticed when running the proving tests.
* First, the seal parsing swapped the high and low parts of the exit code and did not decode them from Montgomery form.
* Second, the parsing was using `Elem::new_raw` which has undefined (arithmetic) behavior if the value is greater than `p`. A checked cast must be used when decoding the seal.

With the PR, `cargo test -p risc0-zkvm -Fcuda,rv32im-m3 -- host::server::prove` now passes on my machine (Some of the execution tests still fail).

- **fix build issues in risc0-zkvm when -F rv32im-m3 is enabled**
- **add back writing of values to the trash address for M3**
- **fix decoding to termination reg a0 and a1 (exit code) from seal**
- **eliminate use of Elem::new_raw and check read length instead of panicking**
- **add back the setting of the output on the final segment in ExecutorImpl**

### DIFF
--- a/risc0/circuit/rv32im/src/execute/segment.rs
+++ b/risc0/circuit/rv32im/src/execute/segment.rs
@@ -25,8 +25,7 @@ use crate::{
     TerminateState,
     execute::{
         ExecutionLimit, Executor, ExecutorResult, RV32IM_M3_CIRCUIT_VERSION,
-        RV32IM_V2_CIRCUIT_VERSION,
-        executor::ExecutionError,
+        RV32IM_V2_CIRCUIT_VERSION, executor::ExecutionError,
     },
 };
 


risc0/circuit/rv32im-m3/src/lib.rs
```diff
@@ -63,49 +63,51 @@ pub struct Claim {
...
         let (slice, remain) = self.seal.split_at(len);
         self.seal = remain;
         Ok(slice)
     }
 
+    /// Read a u32 from the seal, decoding from Montgomery form.
     fn read_u32(&mut self) -> Result<u32> {
-        let slice = self.read(1)?;
-        Ok(slice[0])
+        Ok(self.read(1)?[0].as_u32())
```
flaub: I would revert this, since the intent of `read_u32` is to read a `u32` directly from the stream. There are headers that are injected into the seal that are not related to elements at all, and so aren't associated with a Montgomery form at all.

risc0/circuit/rv32im-m3/src/lib.rs
```diff
@@ -118,14 +120,14 @@ impl Claim {
...
+        let version = decoder.read_u32_raw()?;
         ensure!(version == RV32IM_SEAL_VERSION, "seal version mismatch");
 
-        let po2 = decoder.read_u32()?;
+        let po2 = decoder.read_u32_raw()?;
 
         let pre_state = decoder.read_digest_from_words()?;
         let post_state = decoder.read_digest_from_words()?;
-        let is_terminate = decoder.read_u32_from_elem()?;
+        let is_terminate = decoder.read_u32()?;
```
flaub: The intent here is to interpret the `u32` in the seal as an element. This seems to imply we're reading a `u32` when actually we're reading an element.

risc0/circuit/rv32im-m3/src/lib.rs
```diff
@@ -63,49 +63,51 @@ pub struct Claim {
 }
 
 struct Decoder<'a> {
-    seal: &'a [u32],
+    seal: &'a [Elem],
```
flaub: This isn't accurate. The seal is made up of `u32` and sometimes we want to interact directly with the `u32` instead of dealing with elements. The elements are interpretations of the underlying `u32` stream.

risc0/circuit/rv32im-m3/src/lib.rs
```diff
@@ -118,14 +120,14 @@ impl Claim {
     pub fn decode(seal: &[u32]) -> Result<Self> {
         let mut decoder = Decoder::new(seal);
 
-        let version = decoder.read_u32()?;
+        let version = decoder.read_u32_raw()?;
         ensure!(version == RV32IM_SEAL_VERSION, "seal version mismatch");
 
-        let po2 = decoder.read_u32()?;
+        let po2 = decoder.read_u32_raw()?;
```
flaub: These values aren't elements, and it's perfectly fine for them to fall outside the range of the baby-bear field.

risc0/circuit/rv32im/src/execute/poseidon2.rs
```diff
@@ -126,6 +126,10 @@ impl Poseidon2State {
             self.load_buf_in(ctx, &mut cur_state)?;
             mix(self, &mut cur_state, ctx)?;
             self.count -= 1;
+
+            if self.count != 0 {
+                self.trash_digest(ctx)?;
+            }
```
flaub, line 132: I don't think it's only when `self.count != 0`, I don't recall this being behind a conditional

risc0/circuit/rv32im/src/execute/poseidon2.rs
```diff
@@ -134,6 +138,8 @@ impl Poseidon2State {
 
         if self.has_state == 1 {
             self.store_p2_state(ctx, &mut cur_state)?;
+        } else {
+            self.trash_state(ctx)?;
```
flaub, line 142: I'm not sure this is correct

risc0/zkvm/src/host/server/prove/prover_impl.rs
```diff
@@ -243,7 +243,7 @@ impl ProverServer for ProverImpl {
     fn segment_preflight(&self, segment: &Segment) -> Result<PreflightIter> {
         tracing::debug!("segment_preflight");
         Ok(Box::new(rv32im_m3::PreflightIter::new(
-            segment,
+            segment.clone(),
```
flaub: This can be expensive, why are we requiring a clone here now?

risc0/zkvm/src/host/server/prove/prover_impl.rs
```diff
@@ -446,28 +446,30 @@ mod rv32im_m3 {
...
+        output: MaybePruned<Option<Output>>,
         terminate_state: Option<TerminateState>,
         is_done: bool,
         segment_index: u32,
     }
 
     impl PreflightIter {
         pub(crate) fn new(
-            segment: &Segment,
+            segment: Segment,
```
flaub: What's the reason for this?

risc0/circuit/rv32im/src/execute/poseidon2.rs
nategraf: Here is what is replaces, from `1bc05a1635e1ae7e03510a280cf41d93c197e73c~1`
https://github.com/risc0/risc0/blob/dcba2183bd8a377c62a6315f0819b58a41315c1f/risc0/circuit/rv32im/src/execute/poseidon2.rs#L186-L198

risc0/circuit/rv32im/src/execute/poseidon2.rs
nategraf: This is what it replaces from `1bc05a1635e1ae7e03510a280cf41d93c197e73c~1`
https://github.com/risc0/risc0/blob/dcba2183bd8a377c62a6315f0819b58a41315c1f/risc0/circuit/rv32im/src/execute/poseidon2.rs#L153-L157

risc0/circuit/rv32im/src/execute/poseidon2.rs
flaub: Ok. The reason the code looks the way it used to was so that we can line it up with preflight to ensure it's doing the same thing. With the refactor, this makes it a bit harder.

risc0/circuit/rv32im/src/execute/poseidon2.rs
flaub: ok

flaub: Shouldn't it be `self.count != 1`?

risc0/zkvm/src/host/server/prove/prover_impl.rs
nategraf: This is to fix an issue I introduced, but didn't realize, in https://github.com/risc0/risc0/pull/3514

In that PR, I thought changing `SegmentContext::new` to take an owned `Segment` would avoid a clone and I did not need any callers that it caused issues for. I had neglected to run the tests locally with `-Frv32im-m3` though.

This is one way to fix it, but looking at it again I think https://github.com/risc0/risc0/pull/3582/commits/9dcff510803288d3aec262c548e700782e0e6a4a is a better solution

flaub (approved):

PR #3585, opened 2025-12-10
Enable m3 by default

bobbobbio:
This fixes lingering issues with m3 and enables the feature-flag by default

In order to make this work, I had to do two regrettable things:
- disable creation of po2=13 final segments in executor, it reveals issues with block tracker that need a deeper fix.
- disable the POVW test since m3 segments don't have support for it yet

I think it is worthwhile to move forward with this PR ahead of fixing those issues to swap our CI coverage over to m3 so we can avoid breaking m3 with further commits.

### DIFF
--- a/risc0/circuit/rv32im/src/execute/tests.rs
+++ b/risc0/circuit/rv32im/src/execute/tests.rs
@@ -95,3 +95,26 @@ fn system_split() {
     assert!(segments[0].read_record.is_empty());
     assert!(segments[0].write_record.is_empty());
 }
+
+#[test]
+fn insufficient_segment_limit() {
+    const ITERATIONS: u32 = 10_000;
+
+    let program = testutil::kernel::simple_loop(ITERATIONS);
+    let image = MemoryImage::new_kernel(program);
+
+    let error = testutil::execute(
+        image,
+        ExecutionLimit::default()
+            .with_segment_po2(13)
+            .with_max_insn_cycles(0),
+        testutil::NullSyscall,
+        None,
+    )
+    .map(|_| ())
+    .unwrap_err();
+    assert!(
+        error.to_string().contains("too small"),
+        "too small not found in {error}"
+    );
+}
--- a/risc0/zkvm/src/host/server/exec/tests.rs
+++ b/risc0/zkvm/src/host/server/exec/tests.rs
@@ -76,19 +76,6 @@ fn cpp_test() {
     assert_eq!(message.as_str(), "blst is such a blast");
 }
 
-#[rstest]
-#[ignore = "in m3 this way of testing this no longer works"]
-#[should_panic(expected = "too small")]
-fn insufficient_segment_limit() {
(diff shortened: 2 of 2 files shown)


risc0/circuit/rv32im-m3/src/lib.rs
```diff
@@ -169,3 +173,15 @@ impl Claim {
...
 }
+
+/// Decodes a PoVW nonce from a segment seal.
+pub fn decode_povw_nonce(segment_seal: &[u32]) -> Result<PovwNonce> {
+    let seal_version = segment_seal[0];
+    ensure!(
+        seal_version == RV32IM_SEAL_VERSION,
+        "seal version mismatch: actual={seal_version} expected={RV32IM_SEAL_VERSION}"
+    );
+    let _segment_seal = &segment_seal[1..];
```
flaub, line 184: This seems odd?

risc0/circuit/rv32im/src/execute/poseidon2.rs
```diff
@@ -15,22 +15,23 @@
...
 
 use crate::execute::{
     platform::*,
     r0vm::{LoadOp, Risc0Context},
 };
 
+#[cfg(feature = "prove")]
 const BABY_BEAR_P_U32: u32 = baby_bear::P;
+
+#[cfg(feature = "prove")]
```
nategraf, line 34: It would be nice to extract all the witgen parts into a separate `impl Poserdon2Stat` into a `mod witgen`, or into `crate::prove::witgen`.

risc0/circuit/rv32im-m3/src/lib.rs
bobbobbio: the unused variable? yeah its not needed for anything, this function is just unimplemented at the moment, as I mentioned in the description povw doesn't work right now, so nothing is calling this yet.

risc0/zkvm/src/claim/receipt.rs
```diff
@@ -346,6 +346,7 @@ impl MaybePruned<WorkClaim<ReceiptClaim>> {
     }
 }
 
+#[cfg_attr(feature = "rv32im-m3", allow(dead_code))]
```
nategraf, line 349: Why `allow(dead_code)` instead of conditionally compiling it out?

nategraf:
Oh, I see this is in the merge queue. My comments are not really worth unqueuing it.

risc0/zkvm/src/claim/receipt.rs
bobbobbio: In general I try to avoid compiling things out when possible because I find it to be a better strategy to ensure it continue to compile. Unused code will get deleted by the compiler later on anyway.

risc0/circuit/rv32im/src/execute/poseidon2.rs
bobbobbio: I definitely agree, not going to do in this PR

PR #3601, opened 2025-12-16
Parallelize the M3 CPU backend

bobbobbio:
This addresses previously identified places in the m3 CPU HAL where the work could be parallelized.


jbruestle (approved):
LGTM.

PR #3640, opened 2026-01-07
Resolve future-incompatibility warning by bumping num-bigint-dig to 0.8.6 in Cargo.lock

nategraf:


nategraf:
On any invocation of a `cargo` command that builds the code (e.g. `build`, `check`, `test`) I see the following at the end of the log just after the `Finished` line

```
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.4
```

I had assumed everyone would see this message. Do you?

PR #3697, opened 2026-01-30
Add publish=false to Cargo.toml files to signal cargo publish to ignore them

nategraf:
This is a change I made locally in trying to get `cargo +nightly publish --workspace --dry-run` to work. It still doesn't work, but this is a reasonable change to make by itself and gets it _closer_ to working.

### DIFF
--- a/examples/keccak/methods/Cargo.toml
+++ b/examples/keccak/methods/Cargo.toml
@@ -2,6 +2,7 @@
 name = "keccak-methods"
 version = "0.1.0"
 edition = "2024"
+publish = false
 
 [build-dependencies]
 risc0-build = { path = "../../../risc0/build" }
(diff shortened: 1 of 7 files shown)


examples/keccak/methods/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "keccak-methods"
 version = "0.1.0"
 edition = "2024"
+publish = false
```
bobbobbio, line 5: this is in the examples workspace, do we need this?

bobbobbio (commented):
The release tool we use `cargo release` doesn't use this, thats why we have the other annotations. `cargo release` has its own dry-run (which also has its issues), I'm not sure why we would use `cargo publish` over this.

This change I guess is fine, but I worry it will rot quickly with nothing checking it. If we want to keep this correct we could easily lint it.

examples/keccak/methods/Cargo.toml
nategraf: I ran into it when trying to use `cargo publish --workspace`

PR #3705, opened 2026-02-05
Tiny thread count change

jbruestle:


risc0/circuit/rv32im-sys/cxx/hal/cuda/kernels/eval_check_impl.h
```diff
@@ -16,6 +16,22 @@
 #include "hal/cuda/kernels/base.h"
 #include "rv32im/circuit/eval_check.h"
 
+template <typename Kernel> int max_legal_1d_block_threads(Kernel k, int device = -1) {
+  if (device < 0)
+    cudaGetDevice(&device);
```
bobbobbio: I think it would better to just call `cudaStreamGetDevice` instead

jbruestle: Done, also checked returns, etc.  Also moved call to static initialization, c++ statics are threadsafe during an initialization.

PR #3637, opened 2026-01-06
Remove asserts and potentially panicking index operations based on seal data

nategraf:
This PR addresses locations in the verification code that would panic on input of a seal that is malformed, e.g. by being too short. Code paths within the verification logic now return an error when the seal is too short or too long, and asserts on the po2 are replaced with error returns.

PR #3712, opened 2026-02-11
Avoid panicking index operations in the v2 seal decode and verify functions

nategraf:
Related to the issues fixed in https://github.com/risc0/risc0/pull/3637, this PR resolves a panic in code paths that did not exist on `main` at the time of that fix. Note that this PR is applied to the `release-3.0` branch, and not `main` because these code paths no longer exist there.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
       - uses: baptiste0928/cargo-install@v3
         with:
           crate: cargo-binstall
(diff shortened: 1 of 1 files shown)


nategraf:
Ok, I've added a test. I also noticed that the PoVW path had the same issue and so fixed it there
https://github.com/risc0/risc0/pull/3712/commits/aea6a0ac1d26752a6587f7b4df038991f1188306

PR #3719, opened 2026-02-17
feat(build): add BuildKit cache mounts to Docker guest builds

jimmy-claw:
## Summary

Add BuildKit cache mounts for the cargo registry and git index to the generated Dockerfile used for reproducible guest builds. This dramatically speeds up rebuilds by avoiding re-downloading all dependencies when the source tree changes.

## Problem

`cargo risczero build` uses Docker for reproducible guest builds. The generated Dockerfile does:

```dockerfile
COPY . .
RUN cargo +risc0 fetch ...
RUN cargo +risc0 build --release ...
```

Any source file change invalidates the `COPY . .` layer, which means `cargo fetch` re-downloads **all** dependencies from scratch on every build. On slower machines (Raspberry Pi, CI runners with cold caches), this adds significant time to every iteration.

## Solution

Add `--mount=type=cache` for `/usr/local/cargo/registry` and `/usr/local/cargo/git/db` to both the fetch and build `RUN` commands:

```dockerfile
RUN --mount=type=cache,target=/usr/local/cargo/registry \
    --mount=type=cache,target=/usr/local/cargo/git/db \
    cargo +risc0 fetch ...
RUN --mount=type=cache,target=/usr/local/cargo/registry \
    --mount=type=cache,target=/usr/local/cargo/git/db \
    cargo +risc0 build --release ...
```

Also:
- Prepend `# syntax=docker/dockerfile:1` for BuildKit syntax support
- Set `DOCKER_BUILDKIT=1` on the `docker build` command

## Why this is safe for reproducibility

1. **Cache mounts are not part of any image layer** — they exist only during the `RUN` step and are not included in the final output
2. **`--locked` pins exact dependency versions** from `Cargo.lock`, so the same crates are used regardless of cache state
3. **The output ELF depends only on source code + compiler + flags**, not on how dependencies were downloaded
4. **The `scratch` export stage** only copies the final binary — no cache artifacts leak

## Performance impact

| Scenario | Before | After |
|----------|--------|-------|
| First build | Same | Same |
| Rebuild (same deps) | Re-downloads all crates | Instant (cached) |
| Rebuild (changed deps) | Re-downloads all crates | Downloads only changed crates |

## Requirements

- Docker 18.09+ (BuildKit support) — this has been the default since Docker 23.0
- No changes required for users already using `cargo risczero build`

### DIFF
--- a/risc0/build/src/docker.rs
+++ b/risc0/build/src/docker.rs
@@ -174,11 +174,18 @@ fn create_dockerfile(manifest_path: &Path, temp_dir: &Path, guest_info: &GuestIn
         build = build.env(&docker_env);
     }
 
+    // Use BuildKit cache mounts to persist the cargo registry and git index between
+    // builds. This avoids re-downloading all dependencies on every build when the
+    // source tree changes (which invalidates the COPY layer cache).
+    //
+    // Cache mounts are not part of any image layer and do not affect the output binary.
+    // Reproducibility is preserved because `--locked` pins exact dependency versions.
+    let cache_mounts = "--mount=type=cache,target=/usr/local/cargo/registry \
+                        --mount=type=cache,target=/usr/local/cargo/git/db";
+
     build = build
-        // Fetching separately allows docker to cache the downloads, assuming the Cargo.lock
-        // doesn't change.
-        .run(&fetch_cmd)
-        .run(&build_cmd);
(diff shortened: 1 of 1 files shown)


3esmit:
@jimmy-claw Take a look in this https://github.com/logos-blockchain/lez-programs/pull/212

PR #3725, opened 2026-02-25
Experiment with cubecl

flaub:


cubecl.toml
```diff
@@ -0,0 +1,4 @@
+[compilation]
+[compilation.logger]
+level = "full"
+file = "/home/flaub/src/risc0/tmp/cubecl-codegen.log"
```
bobbobbio, line 4: this tripped me up at first, trying to run the cubecl code crashes since this path doesn't exist on my machine

risc0/zkp/src/hal/cubecl/ntt.rs
```diff
@@ -0,0 +1,1251 @@
...
+    }
+    if comptime!(iterations > 4u32) {
+        ct_warp_butterfly(&mut r0, &mut r1, radix_twiddles, unit_pos, 4u32);
+    }
+    if comptime!(iterations > 5u32) {
+        ct_warp_butterfly(&mut r0, &mut r1, radix_twiddles, unit_pos, 5u32);
+    }
+
+    // Shared memory stages: CT processes s from 6 up to iterations-1.
+    if comptime!(iterations > 6u32) {
```
bobbobbio, line 468: seems similar to `vendor/sppark/ntt/kernels/ct_mixed_radix_wide.cuh` although our usage of sppark for NTT actually uses `vendor/sppark/ntt/kernels/ct_mixed_radix_narrow.cuh` which is much simpler

risc0/zkp/src/hal/cubecl/ntt.rs
```diff
@@ -0,0 +1,1251 @@
...
+    }
+    if comptime!(iterations > 3u32) {
+        gs_warp_butterfly(&mut r0, &mut r1, radix_twiddles, unit_pos, 3u32);
+    }
+    if comptime!(iterations > 2u32) {
+        gs_warp_butterfly(&mut r0, &mut r1, radix_twiddles, unit_pos, 2u32);
+    }
+    if comptime!(iterations > 1u32) {
+        gs_warp_butterfly(&mut r0, &mut r1, radix_twiddles, unit_pos, 1u32);
+    }
```
bobbobbio, line 341:
```suggestion
         // Shared memory stages: GS processes s from iterations-1 down to 6 (when s >= 6).
      #[unroll]
      for i in 1..=comptime!(iterations.saturating_sub(6)) {
          gs_shared_butterfly(
              &mut r0,
              &mut r1,
              &mut xchg,
              radix_twiddles,
              unit_pos,
              comptime!(iterations - i),
              radix,
          );
      }

      // Warp shuffle stages: GS processes s from min(5, iterations-1) down to 1.
      #[unroll]
      for i in 0..comptime!(std::cmp::min(5, iterations - 1)) {
          gs_warp_butterfly(
              &mut r0,
              &mut r1,
              radix_twiddles,
              unit_pos,
              comptime!(std::cmp::min(5, iterations - 1) - i),
          );
      }
```

Here and other places doing this manual loop unrolling is unnecessary

PR #3736, opened 2026-03-25
Verify proper cycle order for BigInt operations

bobbobbio:
Adds verification in the circuit for the order of bigint operations in a bigint program.


jbruestle (approved):
LGTM!

flaub (approved):

PR #3751, opened 2026-05-11
Removes unused build-deps: downloader, hex, liblzma

citizen-stig:
Do reduce number of warnings from `cargo audit` in downstream crates


citizen-stig:
Hey @bobbobbio would it be possible to review this PR? It is minor clean up, but it silences false-positive security reports.

PR #3761, opened 2026-05-25
perf(cuda): enable po2=22 proving on RTX4090 GPUs (+40% throughput)

hugo-blue:
## Motivation

The work originated as a ZKPool team's optimization on a downstream fork;
this PR is the upstreamable subset: the two features (`low_vram` and
`pinned_witgen`) that are directly responsible for the 4090 throughput gain.

The target is to increase the performance of the RiscZero ZKVM and also
support the Boundless prover community to get better device performance
and a higher profit.

The current `risc0-zkvm` cuda path cannot run `po2 = 22` segments on
24 GB consumer GPUs (RTX 4090): the LDE codeword + auxiliary buffers
exceed 24 GB VRAM. Users are forced to `po2 = 21`, doubling segment
count and leaving significant throughput on the table.

This PR enables `po2 = 22` on RTX 4090 and improves `fib_succinct`
throughput by **~49%** vs v3.0.4 baseline.

## Changes

| Layer | Change |
|---|---|
| **Two opt-in features** (default-on under `cuda`) | `low_vram`: early release / reuse of codeword GPU memory + 1 GB host-pinned arena for code Merkle Tree Node. `pinned_witgen`: `cudaMallocHost`-backed witgen H2D buffers (PCIe gen4 full bandwidth) |
| **`eval_check` flatten** | Restructures the nested call chain in `eval_check_*.cu` into sequential sibling calls. **No computation logic changes** — same Fp/FpExt expressions, same `arg0[N]` slot writes, same constraints. Only the call structure changes. |
| **CUDA HAL trait** | 10 new methods + signature change on `zk_shift` (gains `beta, factor`) |
| **NTT kernels** | `ntt_kernels.cu` + `ntt_lde.cuh` under `risc0/sys/kernels/zkp/cuda/supra/`, forked from Supranational `sppark` (Apache-2.0) — back the new `zk_shift` + batch coeffs-bitrev path |

## Performance

**Hardware**:

| | |
|---|---|
| GPU       | NVIDIA RTX 4090 (24 GB, driver 570.133.07) |
| CPU       | AMD EPYC 7K62 (48 cores / 96 threads, 2.6 GHz) |
| RAM       | 125 GB |
| OS        | Ubuntu 22.04.3 LTS, kernel 6.8.0 |
| Toolchain | rustc 1.96.0 (host), rzup `r0.1.94.1` (rustc 1.94.1) for guest, CUDA 12.8 |

[...]

PR #3762, opened 2026-05-25
perf(cuda): enable po2=22 proving on RTX4090 GPUs (+40% throughput)

hugo-blue:
Adds two opt-in Cargo features that together unlock po2=22 segment proving on 24 GB consumer GPUs and improve fib_succinct throughput by ~40% on RTX 4090.

Features
--------
  low_vram      Reduces per-segment GPU memory footprint:
                  - early release / reuse of codeword GPU memory at
                    po2 >= LARGE_SEGMENT_PO2 (= 22)
                  - early release / reuse of rv32im data/accum codeword buffers
                  - host-pinned 1 GB arena for code Merkle Tree Node
  pinned_witgen cudaMallocHost-backed PinnedVec for witgen H2D buffers
                (cycles, txns, bigint_bytes); host->device DMA at full
                PCIe gen4 bandwidth instead of via the driver's
                pageable bounce buffer.

Both features are cuda-gated via `cfg(all(low_vram, cuda))` / `cfg(all(pinned_witgen, cuda))` and default-on in `risc0-zkvm` default features. The cuda code path matches v3.0.4 when the features are not active.

Performance
-----------
  RTX 4090, fib 12M cycles:
    v3.0.4 prebuilt baseline (po2=21):                724.7 KHz
    this PR --features cuda+low_vram+pinned_witgen
       po2=22:                                       ~1015  KHz
    throughput improvement vs v3.0.4 baseline:        +40.0%

Backend support
---------------
  cuda + low_vram + pinned_witgen     primary perf path
  cuda only                            compiles + runs; identical
                                       to v3.0.4 behavior
  cpu                                  compiles + runs
                                       (build_injector gated to v3.0.4
                                       behavior outside low_vram+cuda)
  metal                                compiles via default
                                       unimplemented!() trait impls for
                                       six new HAL methods; panics if
                                       called, but the metal prove path
                                       does not call them under
                                       po2 < LARGE_SEGMENT_PO2

Reproduce
---------
  cargo bench -F cuda --bench fib --no-run
  FIB_N=12000000 FIB_PO2=22 \
    target/release/deps/fib-<hash> --bench fib_succinct

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ compile_commands.json
 .vercel
 node_modules
 *.code-workspace
+logs/
--- a/risc0/circuit/keccak/Cargo.toml
+++ b/risc0/circuit/keccak/Cargo.toml
@@ -63,3 +63,5 @@ prove = [
   "std",
 ]
 std = ["risc0-zkp/std", "risc0-binfmt/std"]
+
+low_vram = []
--- a/risc0/circuit/keccak/src/prove/hal/cpu.rs
+++ b/risc0/circuit/keccak/src/prove/hal/cpu.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::rc::Rc;
-
 use anyhow::Result;
 use rayon::{iter::IntoParallelIterator, prelude::*};
 use risc0_circuit_keccak_sys::{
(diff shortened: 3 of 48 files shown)


hugo-blue:
Closing in favor of #3761, which is the version we're moving forward with. Apologies for the duplicate noise.

PR #3763, opened 2026-05-27
Return serde error for partial-word input

peter941221:
Motivation

`risc0_zkvm::serde::from_slice<T, P>` returns `Result<T>`, but partial-word byte input could panic while handling the bytemuck cast path instead of returning a deserialization error.

Semantic change

This patch rejects inputs whose total byte length is not a multiple of `WORD_SIZE` before attempting the cast to `[u32]`.

That preserves the existing unaligned-but-complete copy path, while making incomplete byte input return `Error::DeserializeUnexpectedEnd` instead of panicking.

Preserved behavior

Complete but unaligned byte slices still deserialize through the existing copy path.

Testing

I added two focused serde tests in `risc0/zkvm/src/serde/deserializer.rs`:
- `partial_word_byte_slice_returns_error`
- `unaligned_complete_byte_slice_deserializes`

Local verification on this Windows host used a focused smoke crate against the patched local `risc0-zkvm` checkout. That smoke test passed for both behaviors with a locally configured `CXX` override.

### DIFF
--- a/risc0/zkvm/src/serde/deserializer.rs
+++ b/risc0/zkvm/src/serde/deserializer.rs
@@ -14,6 +14,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use alloc::{string::String, vec};
+use core::mem::size_of_val;
 
 use bytemuck::Pod;
 use risc0_zkvm_platform::WORD_SIZE;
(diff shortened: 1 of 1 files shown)


peter941221:

PR #3764, opened 2026-05-27
docs: fix release-3.0 Metal proving notes

peter941221:
This updates the release-3.0 documentation to match the prover selection code on that branch.

Failure mechanism:
The release docs still say Metal proving is enabled by default on Apple Silicon, but both prover selection paths on release-3.0 fall back to CPU unless CUDA is enabled:
- risc0/circuit/rv32im/src/prove/mod.rs
- risc0/circuit/recursion/src/prove/mod.rs

Semantic change:
- replace the deprecated feature-table wording in the top-level README and zkVM docs
- update the local proving docs to state that release-3.0 does not enable Metal proving by default on Apple Silicon and falls back to CPU in this release line

Files changed:
- README.md
- risc0/zkvm/src/lib.rs
- risc0/zkvm/README.md
- website/api/generating-proofs/local-proving.md

Testing:
- inspected the release-3.0 diff for scope and wording only
- did not run a docs build


peter941221:

Issue #3708, opened 2026-02-09
Manual build: `build-essential` dependency undocumented


Arjentix:
### Is there an existing issue?

- [x] I have searched the existing issues

### Experiencing problems? Have you tried our Discord first?

- [x] This is not a support question.

### Motivation

I tried to install risc0 in Docker from my Mac Apple Silicon. Decided to go with native build as no risc0 binaries released for `linux/arm64`. I've followed [installation page](https://dev.risczero.com/api/zkvm/install#manual-installation-and-installation-for-all-other-systems-eg-x86-64-macos-arm64-linux). The problem comes from `rzup build` step.

It states that `ninja/ninja-build` and `cmake` dependencies are required and actually `rzup build` will even fail without them with a clear error message. But then in proceeds to run `cmake` and `cmake` fails without any clear error message:

```
1013.8 thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cmake-0.1.54/src/lib.rs:1119:5:
1013.8 
1013.8 command did not execute successfully, got: exit status: 1
1013.8 
1013.8 build script failed, must exit now
1013.8 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
1013.8 Build completed unsuccessfully in 0:00:45
``` 

So I had no idea what's wrong. I had to search for your release ci [`Dockerfile`](https://github.com/risc0/rust/blob/r0.1.91.1/.github/containers/release/Dockerfile) to find out that you also install `build-essential`. After I installed it, the error was gone. I understand that this is a common thing to install in Docker, but it wasn't needed before in my concrete use-case. 

### Request

1. A propper error message from build step with output from `cmake` invocation OR
2. A help-message in the build-script with `Is build-essential installed?` AND
3. Add to documentation on the website.

### Solution

Install `build-essential` package

### Are you willing to help with this request?

Maybe... (I have elaborated on how above)

PR #3766, opened 2026-05-27
docs: mention build-essential for rzup build

peter941221:
This documents an existing manual-build prerequisite for `rzup build rust`.

Failure mechanism:
The manual installation docs currently mention `cmake` and `ninja`, but they do not mention that a native C/C++ toolchain is also required. On Debian/Ubuntu systems, missing `build-essential` can leave users with an opaque `cmake` build failure during `rzup build rust`.

Evidence:
- issue #3708 reports the missing `build-essential` dependency in the manual installation flow
- `risc0/cargo-risczero/docker/Dockerfile.release` already installs `build-essential`

Semantic change:
- update the manual installation docs to mention the native build-tool requirement explicitly
- add the same prerequisite hint to `rzup build` help text and the `rzup` README so the CLI and docs stay aligned

Files changed:
- website/api/zkvm/install.md
- rzup/src/cli/commands.rs
- rzup/README.md

Testing:
- `cargo fmt --all --check`
- `cargo test -p rzup` still fails on Windows because of pre-existing Unix-only code paths in `rzup`, unrelated to this doc/help-text change

### DIFF
--- a/rzup/README.md
+++ b/rzup/README.md
@@ -123,6 +123,10 @@ the default version.
 
 The resulting version of the component will contain the commit hash.
 
+Building the Rust toolchain also requires native build dependencies such as `cmake`,
+`ninja`, and a C/C++ toolchain. On Debian/Ubuntu systems, install packages such as
+`build-essential`, `cmake`, and `ninja-build` first.
+
 ## Components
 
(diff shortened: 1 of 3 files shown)


peter941221:

PR #3776, opened 2026-06-20
Gracefully skip Metal kernels when tools are unavailable

leopoldjoy:
## Summary
- detect missing `metal` / `metallib` tools with `xcrun --sdk ... --find` before building Metal kernels
- warn and emit the same empty metallib artifact shape used by `RISC0_SKIP_BUILD_KERNELS` when the Metal toolchain is unavailable
- allow CPU proving builds on macOS systems that have Command Line Tools but not full Xcode installed

## Testing
- `cargo fmt --package risc0-build-kernel -- --check`
- `cargo clippy -p risc0-build-kernel --all-targets -- -D warnings`
- `cargo check -p risc0-sys` on macOS without `xcrun metal`
- `cargo check -p risc0-zkp --features prove` on macOS without `xcrun metal`

### DIFF
--- a/risc0/build_kernel/src/lib.rs
+++ b/risc0/build_kernel/src/lib.rs
@@ -352,6 +352,22 @@ impl KernelBuild {
             panic!("unsupported target: {target}")
         };
 
+        let missing_tool = if env::var("RISC0_SKIP_BUILD_KERNELS").is_err() {
+            missing_metal_tool(sdk_name)
+        } else {
+            None
+        };
+        if let Some(tool) = missing_tool {
+            println!(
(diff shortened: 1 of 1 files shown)


risc0/build_kernel/src/lib.rs
```diff
@@ -352,6 +352,22 @@ impl KernelBuild {
             panic!("unsupported target: {target}")
         };
 
+        let missing_tool = if env::var("RISC0_SKIP_BUILD_KERNELS").is_err() {
+            missing_metal_tool(sdk_name)
+        } else {
+            None
+        };
```
bobbobbio, line 359: could be `.is_er().then_some(|| ...)`

bobbobbio (approved):

PR #3777, opened 2026-06-23
fix: return error instead of panicking in from_slice on invalid byte slice

### LINKED ISSUE #3754: serde::from_slice panics on partial-word byte slices

### Description of bug

`risc0_zkvm::serde::from_slice<T, P>` returns `Result<T>`, but a byte slice whose length is not a multiple of a `u32` word can panic while handling the bytemuck cast instead of returning a deserialization error.

For example, `from_slice::<u32, u8>(&[0, 1, 2])` reaches the fallback panic for `PodCastError::OutputSliceWouldHaveSlop`.

The relevant code is in `risc0/zkvm/src/serde/deserializer.rs`:

```rust
pub fn from_slice<T: DeserializeOwned, P: Pod>(slice: &[P]) -> Result<T> {
    match bytemuck::try_cast_slice(slice) {
        Ok(slice) => {
            let mut deserializer = Deserializer::new(slice);
            T::deserialize(&mut deserializer)
        }
        // P is u8 or another value without word-alignment. Data must be copied.
        Err(bytemuck::PodCastError::TargetAlignmentGreaterAndInputNotAligned) => {
            let vec = bytemuck::allocation::pod_collect_to_vec::<P, u32>(slice);
            let mut deserializer = Deserializer::new(vec.as_slice());
            T::deserialize(&mut deserializer)
        }
        Err(ref e) => panic!("failed to cast or read slice as [u32]: {e}"),
    }
}
```

Expected behavior: `from_slice` should return an error, such as `Error::DeserializeUnexpectedEnd`, when the input does not contain a complete final word.

Actual behavior: partial-word byte inputs can panic despite the API returning `Result<T>`.

## Difference from #381

I found the older closed issue #381 while searching. This report is distinct:

- #381 was about serde's treatment of `&[u8]` versus `Vec<u8]` and whether deserializing `&[u8]` behaves as expected.
- This report is about `from_slice::<_, u8>` receiving a byte input whose total length is not a complete number of `u32` words. In that case, `bytemuck::try_cast_slice::<u8, u32>` returns `OutputSliceWouldHaveSlop`, and the current fallback panics instead of returning the `Result` error promised by the API.
- The repro does not deserialize into `&[u8]`; it deserializes a `u32` from a malformed raw byte input: `from_slice::<u32, u8>(&[0, 1, 2])`.

A small fix direction is to reject inputs whose byte length is not a multiple of `WORD_SIZE` before attempting the bytemuck cast, while preserving the existing unaligned-but-complete byte-slice copy path.

### Steps to reproduce

A minimal regression test is:

```rust
#[test]
fn partial_word_byte_slice_returns_error() {
    assert_eq!(
        Err(risc0_zkvm::serde::Error::DeserializeUnexpectedEnd),
        risc0_zkvm::serde::from_slice::<u32, u8>(&[0, 1, 2])
    );
}
```

It is also useful to verify that complete but unaligned byte input still works:

```rust
#[test]
fn unaligned_complete_byte_slice_deserializes() {
    let bytes = [0xff, 1, 0, 0, 0];
    assert_eq!(1u32, risc0_zkvm::serde::from_slice::<u32, u8>(&bytes[1..]).unwrap());
}
```

Locally, I verified the regression/fix with the focused serde deserializer tests:

```text
RISC0_SKIP_BUILD=1 cargo test -p risc0-zkvm --lib serde::deserializer::tests -- --nocapture
```

`RISC0_SKIP_BUILD=1` was needed in my local Termux/Android environment because guest build scripts otherwise fail with `UnsupportedPlatform("unknown OS: android")`.

Sertug17:
When `from_slice` receives a byte slice whose length is not a multiple of 4 (the u32 word size), `bytemuck::try_cast_slice` can fail with errors other than alignment mismatch. The catch-all arm was panicking instead of returning an `Err`, violating the function's `Result` contract.

Fixes #3754

### DIFF
--- a/risc0/zkvm/src/serde/deserializer.rs
+++ b/risc0/zkvm/src/serde/deserializer.rs
@@ -85,7 +85,7 @@ pub fn from_slice<T: DeserializeOwned, P: Pod>(slice: &[P]) -> Result<T> {
             let mut deserializer = Deserializer::new(vec.as_slice());
             T::deserialize(&mut deserializer)
         }
-        Err(ref e) => panic!("failed to cast or read slice as [u32]: {e}"),
+        Err(_) => return Err(Error::DeserializeUnexpectedEnd),
     }
 }
 


Sertug17:
Hi! Just wanted to check in on this PR it's been open for about a month. 
Happy to rebase or make any changes if needed. Let me know if there's anything 
blocking review. Thanks!

PR #3781, opened 2026-07-11
fix(zkp): widen poseidon2 row-commit CUDA offset to 64-bit

### LINKED ISSUE #3773: CUDA prover: u32 `col*domain` index overflow in poseidon2 row-commit for large traces (po2>=21, wide AIRs)
Edit sorry I accidentally submitted early. 

Anyway, please note the following issue is AI generated but I it's accurate to my understanding. I'm happy to answer questions or whatever if it's helpful. I wanted to submit an issue because if it is something you want to fix before v5, better to make the issue now. 

Everything following is AI:

---

## Summary
The CUDA prover silently produces an **incorrect** proof for sufficiently large traces due to a u32 integer overflow in column-major buffer indexing. We hit it with a custom ~694-column BabyBear AIR at po2=21 (evaluation domain = 2^21 * INV_RATE(4) ~= 8.39M rows). CPU is unaffected; the GPU seal silently diverges from the CPU seal (no error, just a wrong/invalid proof).

## Where
`risc0-sys` 5.0.0-rc.1, `kernels/zkp/cuda/supra/poseidon2.cuh`, `_poseidon2_rows`:
```cpp
cells[j] = i < dim_y ? matrix[i * dim_x] : fr_t{0};
```
`i` (column index, up to ~693) and `dim_x` (row count = domain ~= 8.39M) are both `uint32_t`, so `i * dim_x` overflows 2^32 once `columns * domain > 4.29e9` (here 693 * 8.39M ~= 5.8e9). This is the active commit-hash kernel, so it reads the wrong cells. Threshold: roughly `columns * domain > 2^32` (~columns >= 512 at po2-21). Narrow circuits / smaller po2 are unaffected.

## Fix
Widen the multiply to 64-bit, e.g. `matrix[(size_t)i * dim_x]`. (We also found the same u32 `col*size` overflow class in our own generated eval_check kernel; the risc0-zkp NTT/FRI/sha paths we checked already use size_t.)

## Repro
Prove a ~694-column AIR at po2=21 on CUDA and compare the seal to the CPU prover -- they differ; with the size_t cast they are bit-identical.

## Env
risc0 5.0.0-rc.1, CUDA 12.4, sm_86 (A40) and sm_89 (RTX 5880 Ada).

bit2swaz:
Fixes #3773

## What's wrong

The SupraSeal poseidon2 row-commit kernel `_poseidon2_rows` computes the column-major matrix index with a 32-bit multiply:

```cpp
cells[j] = i < dim_y ? matrix[i * dim_x] : fr_t{0};
```

`i` and `dim_x` are both `uint32_t`, so `i * dim_x` runs in 32 bits and wraps mod 2^32. Following the Rust caller in `risc0/zkp/src/hal/cuda.rs` (`hash_rows`), `dim_x` is the domain size (the `output.size()` stride, ~8.39M at po2-21) and `i` walks the columns, so `i * dim_x` is `column * domain`. Once that product crosses 2^32 (roughly columns ≥ 512 at po2-21) the kernel reads the wrong cell of `matrix` and commits a hash over the wrong data. There's no panic and no error: the GPU just returns a valid-looking seal that differs from the CPU seal.

This is what @XertroV hit with the ~694-column AIR at po2-21 (`column * domain` ≈ 5.8e9). Narrow circuits and smaller po2 stay under the threshold, which is why the default rv32im/recursion paths never trip it.

## The fix

Widen the multiply to 64-bit at both sites:

```cpp
cells[j] = i < dim_y ? matrix[(size_t)i * dim_x] : fr_t{0};
```

`(size_t)i` promotes the expression to 64-bit before the multiply, so the index is correct for any `column * domain`. Below the threshold nothing changes. This matches the `size_t` indexing the NTT/FRI/sha kernels already use.

The kernel is vendored in two byte-identical copies; the issue points at the first, both are patched:

- `risc0/sys/kernels/zkp/cuda/supra/poseidon2.cuh:151`
- `risc0/circuit/rv32im-sys/cxx/hal/cuda/kernels/poseidon2.cuh:165`

## Test

`risc0/circuit/rv32im-sys/cxx/hal/test/poseidon2_offset_test.cpp` covers the arithmetic directly: it asserts the widened multiply gives 2^32 exactly where the bare `uint32_t` one wraps to 0. It fails against the pre-fix expression and passes after.

There's no kernel-level regression test, and I want to be upfront about why. Tripping the overflow needs `column * domain > 2^32`, which is a matrix on the order of 8.39M rows × ~693 cols × 4 bytes ≈ 23 GB of device memory. That's not runnable in CI or on a normal dev GPU, so a small-input test would pass before and after and prove nothing. The unit test above is the honest substitute: it checks the exact expression that was wrong.

## Verification

Run locally:

- `python3 license-check.py`: passes
- test arithmetic confirmed red-green (wraps to 0 pre-fix, gives 2^32 post-fix)

Not run locally (no NVIDIA toolchain or >20 GB GPU here), left for CI / a maintainer with hardware:

- the gtest under Bazel (rides the existing `hal/test:test` glob target, no BUILD change)
- the full po2-21 repro: prove the ~694-column AIR on CUDA, compare the seal to the CPU path. Should differ before this change and be bit-identical after.

## One thing left out

`risc0/sys/kernels/zkp/cuda/combos.cu:47` has the same `uint32_t * uint32_t` pattern (`cycles * comboCount`). It's a different, much smaller buffer gated to 2 threads, so the product is nowhere near 2^32 and I've left it out to keep this PR scoped to the two poseidon2 sites. Happy to track it here or separately if you'd rather.


bit2swaz:
@bobbobbio could you take a look when you get a chance?

bit2swaz:
@leopoldjoy mind taking a look at this when you get a chance? its a one line fix for a u32 offset overflow in the CUDA poseidon2 row-commit path, with a test

PR #3800, opened 2026-08-19
r0vm: cancel pending RPC callbacks when sending fails

zloglevel:
## Summary

`RpcSender::ask` registers the response callback before sending the RPC message. When `send()` fails, the callback was left in `pending_requests` indefinitely.

Repeated write failures could therefore retain captured callback state and grow the pending request registry. In the theoretical case of message ID wraparound, this could also trigger the existing `RPC message_id reuse` assertion.

## Changes

- Add `RpcRegistry::cancel_request`.
- Remove the pending callback whenever `RpcSender::ask` returns a send error.
- Add a regression test using an `AsyncWrite` implementation that always returns `BrokenPipe`.
- Verify that repeated failed requests do not accumulate pending callbacks.

## Testing

- `cargo +1.96.0 fmt --all -- --check`
- `RISC0_SKIP_BUILD=1 RISC0_SKIP_BUILD_KERNELS=1 cargo +1.96.0 check -p risc0-r0vm --tests`
- `RISC0_SKIP_BUILD=1 RISC0_SKIP_BUILD_KERNELS=1 cargo +1.96.0 clippy -p risc0-r0vm --tests --no-deps -- -D warnings`


zloglevel:
@bobbobbio Could you please take a look when you have a chance? Thanks!

zloglevel:
@flaub Hi — this is a small r0vm RPC lifecycle fix. If sending an RPC fails after its response callback has been registered, the pending callback is now removed instead of being retained indefinitely. The PR includes a BrokenPipe regression test and verifies repeated failures don't accumulate callbacks.

Would you mind taking a look, or pointing me to the right r0vm reviewer? Thanks!

